PR #1, opened 2019-07-08
add initial project structure

moshloop:


timothysc:
@moshloop always 
/assign @timothysc

scripts/release.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -x
```
timothysc: I typically prefer `make release`

PR #8, opened 2019-07-23
konfigadm based image builder

moshloop:


images/konfigadm/Makefile
```diff
@@ -0,0 +1,27 @@
+K8S_VERSION=1.15
+KONFIGADM_VERSION=0.4.0
+CMD=konfigadm images build --resize +2G --output-dir images -vv "k8s-$(K8S_VERSION).yml"
+$(shell mkdir -p images)
+.PHONY: setup
+setup:
+	$(shell which konfigadm > /dev/null || \
+		wget https://github.com/moshloop/konfigadm/releases/download/$(KONFIGADM_VERSION)/konfigadm.deb && \
```
timothysc: ? shouldn't this be self-contained in this repo and not pull from your private?

moshloop: So the same could be said for packer and ansible...

Extracting the image building out of konfigadm isn't straight-forward: The mechanism that konfigadm uses to build an image is to boot a cloud OS image with a cloud-init script (generated by konfigadm) and then cleanup and shutdown at the end. 

Before I do the work of moving the code across I wanted a bit of discussion/bike-shedding in terms of how it is moved across e.g:

1. Copy and paste the image building bits and import konfigadm as a go module
2. Copy and paste both image building + configuration code.
3. Other ?

timothysc: It's fine to reference the canonical source of truth from an OSS tool, but it gets weird to reference your specific account.

moshloop: So at the moment this is the canonical source for konfigadm, still working out the practicalities of which org this gets moved to - `kubernetes-sigs` is a potential (still needs a bit of work in terms of contributor onboarding UX before moving to any org)

timothysc: Can you open an issue and put a comment with a TODO that cross references the issue and I'll unblock this and keep moving.

PR #14, opened 2019-08-12
update the ubuntu iso url

figo:
the 18.04.2 url been moved to old-releases domain, which is permanent now.

cc @akutz

### DIFF
--- a/images/capi/config/ubuntu-1804.json
+++ b/images/capi/config/ubuntu-1804.json
@@ -3,7 +3,7 @@
   "distro_name": "ubuntu",
   "os_display_name": "Ubuntu 18.04",
   "guest_os_type": "ubuntu-64",
-  "iso_url": "http://cdimage.ubuntu.com/ubuntu/releases/bionic/release/ubuntu-18.04.2-server-amd64.iso",
+  "iso_url": "http://old-releases.ubuntu.com/releases/18.04.2/ubuntu-18.04.2-server-amd64.iso",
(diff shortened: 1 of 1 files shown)


figo:
/hold

lets find out whether ubuntu has permanent iso url

moshloop:
I know that the cloud-images are archived and idempotent at https://cloud-images.ubuntu.com/releases/18.04 

But it looks like the cdimages are more aggressively archived, this might be a better option: http://cdimage.ubuntu.com/ubuntu/releases/18.04.3/release/  

Is this something we should raise with Ubuntu/Canonical?

akutz:
@moshloop Thanks for looking at this. To answer your question, "perhaps." We use the ISOs to control exactly what is added to the image via a pre-seed file for Debian/Ubuntu and Kickstart file for RHEL/CentOS. The cloud images are prescriptive, and thus are not used.

Additionally, the boot commands differ based on the ISO type, so it wouldn't be trivial to switch to the cloud images even if they were an option (which, in my mind, they aren't). I know you weren't recommending this -- I just wanted to note it here.

I may know someone at Canonical I can ask. Thanks!

figo:
@moshloop @akutz 

18.04.2 iso been moved to old-releases domain as new patch release released,
let's relying on this permanent URL: http://old-releases.ubuntu.com/releases/18.04.2/ubuntu-18.04.2-server-amd64.iso.

thanks for explaining the detail, @akutz

PR #16, opened 2019-08-13
Allow customize kubernetes, cni source

figo:
the image-builder tool getting kubernetes, cni from two source types:
1) rpm, deb repos, container registry
2) binary download site

the default behavior does not change, still getting them from upstream rpm, deb repo and k8s. gcr.io. this change enable us to configure rpm,deb source and binary source for kuberentes, cni

note: one more change needed to allow customize container registry for k8s images.

cc @codenrhoden @akutz @moshloop @timothysc

### DIFF
--- a/images/capi/ansible/roles/kubernetes/defaults/main.yml
+++ b/images/capi/ansible/roles/kubernetes/defaults/main.yml
@@ -12,20 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-kubernetes_semver: "latest"
-kubernetes_source: "pkg"
-kubernetes_version: "latest"
-kubernetes_cni_version: "latest"
-
-kubernetes_rpm_repo: "https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64"
-kubernetes_rpm_gpg_key: "https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg"
-kubernetes_rpm_gpg_check: True
-
-# Currently Kubernetes does not publish a repo specifically for Bionic, instead hardcode xenial here for now
-# https://github.com/kubernetes/release/issues/557
-#kubernetes_deb_repo: "http://apt.kubernetes.io/ kubernetes-' ~ (ansible_distribution_release | lower)"
-kubernetes_deb_repo: "http://apt.kubernetes.io/ kubernetes-xenial"
-kubernetes_deb_gpg_key: "https://packages.cloud.google.com/apt/doc/apt-key.gpg"
+kubernetes_goarch: "amd64"
 
 kubernetes_bins:
 - kubeadm
@@ -37,7 +24,3 @@ kubernetes_imgs:
 - kube-controller-manager.tar
 - kube-scheduler.tar
 - kube-proxy.tar
-
-# TODO(akutz) Refactor this the same way Kubernetes versions are now handled.
-#             For now this is only used when kubernetes_source != "pkg".
-kubernetes_cni_source: "https://storage.googleapis.com/capv-images/extra/cni-plugins-amd64-v0.7.5.tgz"
--- a/images/capi/ansible/roles/kubernetes/tasks/main.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/main.yml
@@ -19,7 +19,7 @@
   when: kubernetes_source == "pkg" and ansible_os_family == "RedHat"
 
 - import_tasks: url.yml
-  when: kubernetes_source != "pkg"
+  when: kubernetes_source != "pkg" and kubernetes_cni_source != "pkg"
 
 - name: Ensure that the kubelet is running
   service:
--- a/images/capi/ansible/roles/kubernetes/tasks/url.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/url.yml
@@ -32,8 +32,8 @@
 
 - name: Download CNI tarball
   get_url:
-    url: "{{ kubernetes_cni_source }}"
-    checksum: "sha1:{{ kubernetes_cni_source }}.sha1"
+    url: "{{ kubernetes_cni_source }}/{{ kubernetes_cni_semver }}/cni-plugins-{{ kubernetes_goarch }}-{{ kubernetes_cni_semver }}.tgz"
+    checksum: "sha1:{{ kubernetes_cni_source }}/{{ kubernetes_cni_semver }}/cni-plugins-{{ kubernetes_goarch }}-{{ kubernetes_cni_semver }}.tgz.sha1"
     dest: /tmp/cni.tar.gz
     mode: 0755
     owner: root
@@ -52,7 +52,7 @@
 
 - name: Download Kubernetes binaries
   get_url:
-    url: "{{ kubernetes_source }}/bin/linux/amd64/{{ item }}"
+    url: "{{ kubernetes_source }}/bin/linux/{{ kubernetes_goarch }}/{{ item }}"
     # TODO(akutz) Write a script to separately download the checksum
     #             and verify the associated file using the correct
     #             checksum file format
@@ -65,7 +65,7 @@
 
 - name: Download Kubernetes images
   get_url:
-    url: "{{ kubernetes_source }}/bin/linux/amd64/{{ item }}"
+    url: "{{ kubernetes_source }}/bin/linux/{{ kubernetes_goarch }}/{{ item }}"
     # TODO(akutz) Write a script to separately download the checksum
     #             and verify the associated file using the correct
     #             checksum file format
@@ -87,6 +87,14 @@
   file:
     state: directory
     path: /etc/kubernetes/manifests
+    mode: 0755
+    owner: root
+    group: root
+
+- name: Create kubelet sysconfig directory
+  file:
+    state: directory
+    path: /etc/sysconfig
     mode: 0755
     owner: root
     group: root
--- a/images/capi/config/kubernetes.json
+++ b/images/capi/config/kubernetes.json
@@ -1,5 +1,10 @@
 {
   "kubernetes_semver": "v1.13.6",
   "kubernetes_version": "1.13.6-00",
-  "kubernetes_source": "pkg"
+  "kubernetes_source": "pkg",
+  "kubernetes_rpm_repo": "https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64",
+  "kubernetes_rpm_gpg_key": "\"https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg\"",
+  "kubernetes_rpm_gpg_check": "True",
+  "kubernetes_deb_repo": "\"https://apt.kubernetes.io/ kubernetes-xenial\"",
+  "kubernetes_deb_gpg_key": "https://packages.cloud.google.com/apt/doc/apt-key.gpg"
 }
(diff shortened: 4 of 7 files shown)


images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -19,7 +19,7 @@
   when: kubernetes_source == "pkg" and ansible_os_family == "RedHat"
 
 - import_tasks: url.yml
-  when: kubernetes_source != "pkg"
+  when: kubernetes_source != "pkg" and kubernetes_cni_source != "pkg"
```
detiber: Should there be an error if `kubernetes_source != kubernetes_cni_source`?

images/capi/ansible/roles/kubernetes/tasks/url.yml
```diff
@@ -32,8 +32,8 @@
 
 - name: Download CNI tarball
   get_url:
-    url: "{{ kubernetes_cni_source }}"
-    checksum: "sha1:{{ kubernetes_cni_source }}.sha1"
+    url: "{{ kubernetes_cni_source }}/{{ kubernetes_cni_semver }}/cni-plugins-amd64-{{ kubernetes_cni_semver }}.tgz"
+    checksum: "sha1:{{ kubernetes_cni_source }}/{{ kubernetes_cni_semver }}/cni-plugins-amd64-{{ kubernetes_cni_semver }}.tgz.sha1"
```
detiber: I'm not sure we should always assume amd64, but that is something that could be tracked and fixed later.

images/capi/packer/packer.json
```diff
@@ -6,10 +6,16 @@
...
+    "kubernetes_cni_version": "",
+    "kubernetes_cni_semver": "",
+    "kubernetes_cni_source": "",
+    "kubernetes_semver": "",
+    "kubernetes_source": "",
+    "kubernetes_version": "",
+    "kubernetes_rpm_repo": "",
+    "kubernetes_rpm_gpg_key": "",
+    "kubernetes_deb_repo": "",
+    "kubernetes_deb_gpg_key": "",
```
detiber: Will this cause issues if run without including the kubernetes.json and cni.json files? Would it be better to not define them here so that it will result in an error if not set? Or possibly default the values to `null` here for a nicer error?

akutz: Hi @detiber,

I agree/disagree. I think @figo is combining the CNI change with our discussion about converging all defaults into a single location. I do wonder:

1. Where should that location be?
2. What happens if a default isn't provided?

It seems to me that there will be cases where the defaults *should* be in an external file, especially if CAPA and CAPV end up using different `packer.json` entry points. Therefore does it even make sense to define the user variables with empty values at all in `packer.json`?

detiber: @akutz if we expect that the files that define these values will be shared across providers, then I think setting them in 1 or more variables files makes sense, I just want to make sure that we don't end up with some weird and hard to troubleshoot error in the case that someone forgets to include the variables file.

images/capi/ansible/roles/kubernetes/tasks/main.yml
figo: @detiber your question make me realize that cni never come with rpm/deb format, the cni gz file need be downloaded from url, i am making a fix on this.

iow, `kubernetes_cni_source` will always been `non-pkg`

images/capi/ansible/roles/kubernetes/tasks/url.yml
akutz: Hi @detiber,

I agree, but I also think the happy path is linux_amd64 for now, and as long as we're not burning PROM chips, I think we're okay to change it when necessary.

images/capi/ansible/roles/kubernetes/tasks/main.yml
akutz: Hi @detiber,

This is a good point. I initially wanted to separate the Kubernetes and CNI components, but since the Kubernetes packages depend on the CNI packages, maybe it **does** mean we have to treat them as an atomic group, even when installing from a non-package source?

images/capi/packer/packer.json
akutz: Hi @detiber,

I agree completely, but I'm also not sure what we can do about that. Are you saying if we set them to `null` in the `packer.json` entry-point that Packer will complain the values are `null`?

detiber:
Yes, a null value will indicate that it is required for a user to set it (as an individual variable or through a card file)

images/capi/packer/packer.json
detiber: Yes, a null value will indicate that it is required for a user to set it (as an individual variable or through a card file)

figo: setting them to `null` and raise error if not been set by user sounds reasonable to me, will give it a try, thx

images/capi/ansible/roles/kubernetes/tasks/main.yml
figo: @detiber  when ```kubernetes_source == kubernetes_cni_source == pkg```, it install k8s, cni from same rpm/deb repo. (which can be configured through `kubernetes_rpm_repo` and `kubernetes_deb_repo`)

but when ```kubernetes_source != pkg and kubernetes_cni_source != pkg```,  those two variables can point to different binary download site.

i guess we should throw error when only one of it `==pkg`, we don't support that flexibility, not sure we need to.

images/capi/packer/packer.json
figo: done, thx

images/capi/ansible/roles/kubernetes/tasks/url.yml
figo: @detiber @akutz added two vars `kubernetes_arch` `kubernetes_cni_arch` default to `amd64`

images/capi/ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -15,7 +15,7 @@
 kubernetes_semver: "latest"
 kubernetes_source: "pkg"
 kubernetes_version: "latest"
-kubernetes_cni_version: "latest"
+kubernetes_arch: "amd64"
```
detiber: I'd suggest only having a single arch variable, since I don't think there is a case where we would support mixing of arch values. It might also be good to indicate that this is a `goarch` value rather than the arch value returned from something such as `uname -m`.

Actually, I would probably echo @akutz here and say we should probably exclude this for now, since if we wanted to support multi-arch down the road we should probably do it in a more comprehensive manner (we already hardcode x68_64 in some other places as well) and leverage facts to determine which arch value to use.

images/capi/ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -38,6 +38,7 @@ kubernetes_imgs:
 - kube-scheduler.tar
 - kube-proxy.tar
 
-# TODO(akutz) Refactor this the same way Kubernetes versions are now handled.
-#             For now this is only used when kubernetes_source != "pkg".
-kubernetes_cni_source: "https://storage.googleapis.com/capv-images/extra/cni-plugins-amd64-v0.7.5.tgz"
+kubernetes_cni_version: "latest"
```
moshloop: I think latest is dangerous, it would probably be better to error if no version is specified.

images/capi/config/kubernetes.json
```diff
@@ -1,5 +1,9 @@
 {
   "kubernetes_semver": "v1.13.6",
   "kubernetes_version": "1.13.6-00",
-  "kubernetes_source": "pkg"
+  "kubernetes_source": "pkg",
+  "kubernetes_rpm_repo": "https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64",
+  "kubernetes_rpm_gpg_key": "\"https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg\"",
+  "kubernetes_deb_repo": "\"http://apt.kubernetes.io/ kubernetes-xenial\"",
```
moshloop: I know http vs https doesn't really matter with apt, but it should probably be https to be consistent

images/capi/packer/packer.json
```diff
@@ -60,7 +66,7 @@
       ],
       "extra_arguments": [
         "--extra-vars",
-        "kubernetes_cni_version={{user `kubernetes_cni_version`}} kubernetes_semver={{user `kubernetes_semver`}} kubernetes_source={{user `kubernetes_source`}} kubernetes_version={{user `kubernetes_version`}}"
+        "kubernetes_rpm_repo={{user `kubernetes_rpm_repo`}} kubernetes_rpm_gpg_key={{user `kubernetes_rpm_gpg_key`}} kubernetes_deb_repo={{user `kubernetes_deb_repo`}} kubernetes_deb_gpg_key={{user `kubernetes_deb_gpg_key`}} kubernetes_cni_version={{user `kubernetes_cni_version`}} kubernetes_cni_semver={{user `kubernetes_cni_semver`}} kubernetes_cni_source={{user `kubernetes_cni_source`}} kubernetes_semver={{user `kubernetes_semver`}} kubernetes_source={{user `kubernetes_source`}} kubernetes_version={{user `kubernetes_version`}}"
```
moshloop: Perhaps this should be exported to a file beforehand, especially as ansible variable inheritance rules are crazy

images/capi/ansible/roles/kubernetes/defaults/main.yml
detiber: I have no objection to removing the default value here, especially since we are already going to be throwing an error if it is not set in packer.

images/capi/packer/packer.json
detiber: While the precedence is not necessarily intuitive, I don't see an issue related to precedence, since extra_vars always take precedence, and it wouldn't matter if it was specified individually on the command line as it is here or through a file.

images/capi/config/kubernetes.json
figo: good point, will change it

images/capi/packer/packer.json
figo: agree with @detiber since extra_vars always take precedence

images/capi/ansible/roles/kubernetes/defaults/main.yml
figo: initially andrew and i discussed about preserving ansible independency (run without packer), it is not complete anyway today, also i don't see real need of run ansible without packer (tell me wrong here), so i have no problem of removing those default vars that already required by packer.json

images/capi/ansible/roles/kubernetes/defaults/main.yml
figo: i agree support multi-arch requires more comprehensive change, i will change it to one var now.

images/capi/packer/packer.json
moshloop: then why not just reference the config file directly? e.g. `--extra-vars @config/kubernetes.json`

images/capi/ansible/roles/kubernetes/defaults/main.yml
moshloop: This depends on the testing strategy - running ansible standalone offers a bit more flexibility on how/where you run integration tests

akutz: It was decided that all defaults would be defined in a single location, as visible as possible. The intent of this builder is not to invoke Ansible directly.

figo: code has been updated that all packer required vars has been removed from ansible default

images/capi/packer/packer.json
figo: @moshloop  https://www.packer.io/docs/templates/user-variables.html has `Even if you want a user variable to default to an empty string, it is best to explicitly define it. This explicitness helps reduce the time it takes for newcomers to understand what can be modified using variables in your template.`, and it does make sense to me.

user can edit config/*.json at their will, but define those variables at packer.json does bring benefits.

PR #21, opened 2019-08-15
Use a newer version of cloud-init for CentOS

### LINKED ISSUE #17: Build/deploy cloud-init 19.1/2 for CentOS 7
This issue tracks the work necessary to automate the build of cloud-init 19.1/2 as an RPM for CentOS 7. More information on why this is necessary may be found at https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/41.

Neither [pkgs.org](https://pkgs.org/download/cloud-init) nor rpmfind listed cloud-init 18.4+ for CentOS or RHEL. There are RPMs for other OSs, but they link against python3 and other dependencies unavailable to even the CentOS EPEL repository.

There's [a branch](https://github.com/akutz/cluster-api-provider-vsphere/tree/feature/cloud-init-centos-7/hack/tools/cloud-init/centos7) of the CAPV project that includes a Dockerfile for building the cloud-init RPM for CentOS 7. More information may be found at https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/41#issuecomment-512610753.

However, during a recent call with Tim St. Clair, he mentioned possibly using [Copr](http://copr.fedorainfracloud.org) to automate the build and deployment of this package. As @codenrhoden has experience with Copr, this issue is being assigned to him.

/assign @codenrhoden

codenrhoden:
This patch changes the ansible script to install cloud-init from a
different COPR repository, namely
https://copr.fedorainfracloud.org/coprs/trhoden/cloud-init/, which
contains a cloud-init package newer than what can be found on EPEL. This
version is currently 19.2.

Fixes #17 

I tested that I could build the CentOS image locally, and after booting it up, saw that cloud-init had run (I could SSH in with the injected SSH key) and could look at the cloud-init logs and package version. Currently installed version is `cloud-init-19.2-1.el7.noarch`.

### DIFF
--- a/images/capi/ansible/roles/common/defaults/main.yml
+++ b/images/capi/ansible/roles/common/defaults/main.yml
@@ -37,5 +37,5 @@ common_pinned_debs:
 - "cloud-init={{ cloud_init_version }}"
 common_extra_debs: []
 common_redhat_epel_rpm: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
-common_redhat_cloud_init_copr_rpm: "https://copr.fedorainfracloud.org/coprs/jdetiber/cloud-init/repo/epel-7/jdetiber-cloud-init-epel-7.repo"
+common_redhat_cloud_init_copr_rpm: "https://copr.fedorainfracloud.org/coprs/trhoden/cloud-init/repo/epel-7/trhoden-cloud-init-epel-7.repo"
 common_deb_cloud_init_ppr_repo: "ppa:detiber/cloud-init"


akutz:
/hold

This should be held until both CAPV and CAPA v1 support the non-kubeadm module approach to cloud-init.

- [ ] CAPV use `runcmd` to execute `kubeadm` instead of `kubeadm` module (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/499)
- [ ] CAPA use `runcmd` to execute `kubeadm` instead of `kubeadm` module

@detiber, are there any plans to support the non-kubeadm module approach in CAPA v1alpha1?

detiber:
@akutz CAPA v1alpha1 can continue to use the existing image build tooling we have, while v1alpha2 has already moved away from the kubeadm module.

akutz:
Hi @detiber,

How is CAPA handling versioning the images based on CAPA version? CAPV versions based on Kubernetes version, and so if we created new images, it would possibly impact v1alpha1 providers.

akutz:
Hi @codenrhoden,

I wonder if we should expand this PR by retitling it "Use upstream version of cloud-init" and switch Debian/Ubuntu to the distro-specific version as well, using your Copr build for CentOS.

detiber:
@akutz ah, indeed, that could definitely have an impact. That said, we could always update the lookup pattern we are using for v1alpha2 to exclude the v1alpha1-oriented images.

akutz:
> @akutz ah, indeed, that could definitely have an impact. That said, we could always update the lookup pattern we are using for v1alpha2 to exclude the v1alpha1-oriented images.

True. I also think perhaps the machine image names should include `v1alpha1`, etc. as well as the Kubernetes version?

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -37,5 +37,5 @@ common_pinned_debs:
 - "cloud-init={{ cloud_init_version }}"
 common_extra_debs: []
 common_redhat_epel_rpm: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
-common_redhat_cloud_init_copr_rpm: "https://copr.fedorainfracloud.org/coprs/jdetiber/cloud-init/repo/epel-7/jdetiber-cloud-init-epel-7.repo"
-common_deb_cloud_init_ppr_repo: "ppa:detiber/cloud-init"
\ No newline at end of file
+common_redhat_cloud_init_copr_rpm: "https://copr.fedorainfracloud.org/coprs/trhoden/cloud-init/repo/epel-7/trhoden-cloud-init-epel-7.repo"
```
figo: very dummy question, where is the code to re-create this cloud-init rpm?

figo: i guess we should check-in the Dockerfile that re-build the cloud-init rpm

detiber:
@akutz @randomvariable If we eliminated the need for jinja templating, and rather did the variable substitution through other means (such as a script locally that runs envsubst prior to invoking kubeadm), would that also eliminate the need for a custom version of cloud-init for CentOS and Amazon Linux2?

akutz:
Hi @detiber,

For CAPV the answer is yes. Two things though:

* The instance data and query language provides a very nice way for operators with multi-device networks to control which dynamic addresses that may wish to use. Whatever replaces that would need to be fairly feature-full with a similar query language IMO. I recommend a data structure into a Go template.
* I'd also **prefer** to see the _custom)_ version anyway as it's really the upstream version much more closely related to the one Ubuntu uses. It's not a fork -- just keeping up with the actual cloud-init project.

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: I used the [Dockerfile](https://github.com/akutz/cluster-api-provider-vsphere/blob/feature/cloud-init-centos-7/hack/tools/cloud-init/centos7/Dockerfile) that Andrew already created, copied out the SRPM, and uploaded that to COPR. COPR will build new packages and create the repo for you from an SRPM.

But yes, if we continue to carry this build, checking that Dockerfile in somewhere would be good.

From all the comments on this, it sounds like this is become a more contentious issue. I had thought the ask/task was simple. :)

akutz: Aforementioned [`Dockerfile`](https://github.com/akutz/cluster-api-provider-vsphere/tree/feature/cloud-init-centos-7/hack/tools/cloud-init/centos7/Dockerfile)

moshloop:
> @akutz @randomvariable If we eliminated the need for jinja templating, and rather did the variable substitution through other means (such as a script locally that runs envsubst prior to invoking kubeadm), would that also eliminate the need for a custom version of cloud-init for CentOS and Amazon Linux2?

I would much prefer this approach, there are still a couple of unsolved problems with this PR.

Perhaps we should have an image-builder go helper that can bring in env substitution, vmware instance metadata and maybe even log shipping for diagnostics. 


1) Provenance, The rpm should ideally be hosted by a company through a promotion process, e.g. `packages.cloud.google.com` or a vmware hosted repo

2) Security, Combined with no 1, when a cloud-init CVE is released as per the project goals we need to release a new version within 48 hours, that requires a commitment from multiple people to maintain the rpm, and the ability to test it to ensure a rapid release does not have regression.

akutz:
Env substitution is not a solution, and this PR is fine as Copr builds are an accepted solution for packages. They’re built from source via RedHat hosting.

moshloop:
> Copr builds are an accepted solution for packages. They’re built from source via RedHat hosting.

Copr builds are fine, but the source cannot be controlled by an individual, this needs to be controlled by an org e.g. https://copr.fedorainfracloud.org/coprs/kubernetes-sigs/cloud-init/ or https://copr.fedorainfracloud.org/coprs/vmware/cloud-init/ 

> Env substitution is not a solution

I am not recommending that, I am recommending building a go based app that can provide Jinja like templating of files in the same way cloud-init 19+ does - https://github.com/flosch/pongo2 already provides Jinja style templates in go. 
This go package could be built using goreleaser and the debs and rpms hosted on a kubernetes-sigs github packages repo. 

This app could then also incorporate https://github.com/vmware/cloud-init-vmware-guestinfo and potentially handle some of the /etc/hosts issues that have come up. 

This would allow the use of standard packages and potentially the use of non-baked images with the kubeadm / docker installation also performed in cloud-init

randomvariable:
@detiber 
> If we eliminated the need for jinja templating, and rather did the variable substitution through other means (such as a script locally that runs envsubst prior to invoking kubeadm)

Yes, that would work. I was also thinking for v1alpha1 that the kubeadm module could  be extended to merge the value for the CertSANs into the kubeadm config instead of doing it via jinja templating.

One thing we need to note about cloud-init is that Amazon have not upstreamed their cloud-init modifications that extend the distro detection. Am gathering these at the moment.

Also, should https://github.com/number5/cloud-init should be swapped out for https://github.com/cloud-init/cloud-init or https://git.launchpad.net/cloud-init ?

randomvariable:
I've gathered all of the patches from `cloud-init-18.2-72.amzn2.0.7` into this branch https://github.com/randomvariable/cloud-init/tree/amazon-linux-2 .

Not all of them are necessary, but anything around OS detection and EC2 metadata improvements I think should be merged.

detiber:
> Also, should https://github.com/number5/cloud-init be swapped out for https://github.com/cloud-init/cloud-init or https://git.launchpad.net/cloud-init ?

:100: let's make sure we are using an official mirror.

detiber:
> I've gathered all of the patches from cloud-init-18.2-72.amzn2.0.7 into this branch randomvariable/cloud-init#1 .
>
> Not all of them are necessary, but anything around OS detection and EC2 metadata improvements I think should be merged.

This gets us back into a similar situation as we were with the kubeadm module, which if that is the case, maybe we should just continue to use the kubeadm module and incorporate the templating changes we were talking about there?

randomvariable:
We could try and get these merged upstream with some help from AWS peeps at SIG Cloud Provider?

moshloop:
/hold

I think we are making light of very serious security concerns and taking shortcuts where we shouldn't be.  One of the stated goals of this project is to provide security assurance and auditability, and as it stands this PR is a step backwards, not forwards.

Creating a new cloud-init package without a governance process in place to ensure it is secure  is a recipe for a global 0-day.

Cloud-init runs before anything else, including security auditing and monitoring packages that may be installed on top by end-users, malware that finds itself into this package will, therefore, be able to hide its tracks before any potential security monitoring is in place.

It is all very good to make the development against these images easier by being bleeding edge, but it can't be at the expense of security.

System packages must be maintained by the owner of the OS distribution who has made commitments to release and test updates for CVE's.  I don't think that is a commitment this group can or should make.

If we want to make use of newer system packages, we need to do the work upstream.

I am more than happy to pitch in on a go app (that can be installed after security agents) or any other type of solution we can come up with.

/assign @timothysc @justinsb

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -37,5 +37,5 @@ common_pinned_debs:
...
+common_deb_cloud_init_ppr_repo: "ppa:detiber/cloud-init"
```
timothysc: @codenrhoden - shouldn't this be your repo?

akutz: Hi @timothysc,

The second line is the Debian PPA, and @codenrhoden only built the RPM. Please see my above comment that addresses this:

> I wonder if we should expand this PR by retitling it "Use upstream version of cloud-init" and switch Debian/Ubuntu to the distro-specific version as well, using your Copr build for CentOS.

detiber: As an aside, I'm more than happy to add maintainers to both the COPR and PPA repos that I created originally.

codenrhoden: yeah I was under the impression that this only needed to be done for rhel-like distros, not debian. If we want to use a PPA for debian I can do that as well.

akutz: > As an aside,

As a bside, it amazes me Queen didn't recognize the potential of "We will rock you".

akutz: Hi @codenrhoden,

Your impression was correct. Debian/Ubuntu was going to just switch to the upstream version of cloud-init, hence my earlier comment.

timothysc: I guess my question is: "Why is this PR adding @detiber 's repo for debian"
I thought ubuntu worked fine, or is this for debians core distribution.

akutz: It's not. It was there, and due to a line ending or some space the line was deemed modified and included in the commit.

codenrhoden:
> > Also, should https://github.com/number5/cloud-init be swapped out for https://github.com/cloud-init/cloud-init or https://git.launchpad.net/cloud-init ?
> 
> 💯 let's make sure we are using an official mirror.

@akutz what was the reason for basing your Dockerfile example off the `number5` repo? I agree that I should be using the official mirror, but want to make sure I'm not missing something important.

images/capi/ansible/roles/common/defaults/main.yml
akutz: image

akutz:
> > > Also, should https://github.com/number5/cloud-init be swapped out for https://github.com/cloud-init/cloud-init or https://git.launchpad.net/cloud-init ?
> > 
> > 
> > 💯 let's make sure we are using an official mirror.
> 
> @akutz what was the reason for basing your Dockerfile example off the `number5` repo? I agree that I should be using the official mirror, but want to make sure I'm not missing something important.

Because I didn't notice that? Does that work?

randomvariable:
To be honest, I don't think https://github.com/cloud-init/cloud-init existed until relatively recently

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: yeah, that was my editor adding a newline at the end of the file when there wasn't one before.

akutz:
> To be honest, I don't think https://github.com/cloud-init/cloud-init existed until relatively recently

It did. It was an accident on my part. I've been using the official mirror for months, and I just goofed.

codenrhoden:
Okay, I'd like to circle the wagons here. Before I do more work on this, I'd like to get an agreed on approach.

(1) should we rebuild and package an upstream cloud-init from the official git repo (either github or launchpad?). That bit in itself was contentious in this PR, but I think I'm hearing the answer is yes...

(2) should we do it for both centos and debian? Is it need for debian?

(3) should I do it in my own repo hosting, or be added as a maintainer for @detiber's ?

(4) should we check-in any files here (like a Dockerfile) that describe how we build the packages?

detiber:
> (1) should we rebuild and package an upstream cloud-init from the official git repo (either github or launchpad)? That bit in itself was contentious in this PR, but I think I'm hearing the answer is yes...

If we want to incorporate the changes that @randomvariable identified for better supporting Amazon Linux 2, then we would either need to carry a fork or a patchset on top of the upstream repo.

detiber:
> (3) should I do it in my own repo hosting, or be added as a maintainer for @detiber's ?

I'm open to either solution here, but whichever way we go we should likely make sure that we have multiple people that have access to build new versions as needed.

akutz:
> If we want to incorporate the changes that @randomvariable identified for better supporting Amazon Linux 2, then we would either need to carry a fork or a patchset on top of the upstream repo.

I must disagree here. @randomvariable's do not need to be part of the cloud-init package. Just like [the image builder installs](https://github.com/kubernetes-sigs/image-builder/blob/170d7635960419a6c3a71a8aa24f1a61c458a23c/images/capi/ansible/roles/providers/tasks/vmware.yml#L39-L50) the [Cloud-init Datasource for VMware Guest Info](https://github.com/vmware/cloud-init-vmware-guestinfo), the image builder should be updated to incorporate the Amazon Linux-specific patches/distro/datasource.

randomvariable:
1) Happy to start with the upstream. Reading through the patches, I *think* We'll have enough working modules for K8s to bootstrap, but customizations orgs might want to do might not work, and we don't yet support that anyway.

2) Possibly for consistency, have the same version of cloud-init across the distros

3) Keep it in yours for now, I think, just in case we need to continue building another image for v1alpha1 in the immediate future for some reason. Also we should request a group to be created in Fedora Authentication Service for COPR management from fedora-infra. I think that might also allow us to create a kubernetes-sigs/cloud-init style COPR (happy to file the request if we can agree a name).

4) This project should probably contain any of the build scripts used by the COPR external to the upstream repository.

detiber:
> We shouldn't maintain a package fork when we can simply install distro/datasource files as part of the image build anyway.

Maybe? This isn't something that is supported by cloud-init and in practice can lead to issues, which is why we moved to a fork of cloud-init for the AWS provider.

akutz:
> > We shouldn't maintain a package fork when we can simply install distro/datasource files as part of the image build anyway.
> 
> Maybe? This isn't something that is supported by cloud-init and in practice can lead to issues, which is why we moved to a fork of cloud-init for the AWS provider.

Fair.

detiber:
> > > We shouldn't maintain a package fork when we can simply install distro/datasource files as part of the image build anyway.

> > Maybe? This isn't something that is supported by cloud-init and in practice can lead to issues, which is why we moved to a fork of cloud-init for the AWS provider.

> Fair.

That said, I did the forked repo out of expediency, another alternative that we could do is to just maintain a patchfile and an RPM Spec file that applies the patch on top of cloud-init.

moshloop:
>  b/c it's really hard [to upstream]

Just b/c something is hard doesn't mean you don't do it 

> Creating a copr repo is by far the best industry best practice

My concern is not with the technology, but the underlying processes and governance required to keep it maintained and secure.

> We can add travis to the cloud-init package maintainers so he can get all the CVE spam

Will travis agree to provide a 48h turnaround time on CVE updates, even when he is on PTO - I don't think that is a reasonable or fair expectation, In my mind 3 should be the minimum number of maintainers.

> I think we should document and add details 

I couldn't agree more, this PR should include this documentation and at a minimum answer:

* Who are the current maintainers for the package
* What is the process for reporting security issues and the expected turnaround time
* What is the testing strategy for this package and which operating systems and cloud environments is it tested against / supports.

I also have yet to see why all the alternatives proposed are not feasible?
1. Bash/Envsubst - I can understand why this is not suitable for multiple interfaces
2. Go package that gets installed to handle instance metadata lookup and templating
3. Python script or similar

timothysc:
This is not worth holding up for two weeks, please add a comment and document in the readme and let's get this one done.

codenrhoden:
I did not look at this at all over the last week and a half as I was focused on getting things into k/k before 1.16 code freeze.

I also hadn't worked on this because in https://github.com/kubernetes-sigs/image-builder/pull/21#issuecomment-521778726, it sounded like it shouldn't go in yet. In https://github.com/kubernetes-sigs/image-builder/issues/17#issuecomment-523918613, it seemed like this may not be necessary at all should be closed.

/close

If this comes back up, I will re-open, and add the needed comments and and documentation.

figo:
/reopen 

@codenrhoden we do need this PR to get Centos working properly with CAPV.
i do agree with @timothysc 's comment, lets document the improvement we need and get this check-in.

akutz:
It shouldn’t be a problem unless we remove Jason’s fork ahead of this, which no one should have done. And CentOS 7.7 will include the newer version. Bottom line, don’t build RHEL images without Jason’s CI fork until RHEL 7.7 is released.

codenrhoden:
After a bit more discussion, we concluded that we indeed don't need this. The current fork from Jason will get us by until cloud-init 18.5 is available along with CentOS 7.7.
/close

PR #24, opened 2019-08-21
fix makefile

figo:
fix makefile: the if..else condition need be shell inside target in order to get target name

cc @detiber


images/capi/packer/ami/packer.json
```diff
@@ -1,7 +1,7 @@
 {
   "variables": {
-    "aws_access_key": "",
-    "aws_secret_key": "",
+    "aws_access_key": null,
```
detiber: These will generally default to using the sdk lookup (env variables, default config, etc), with this change that would no longer be the case.

figo: make sense, i will switch it back.

PR #25, opened 2019-08-23
[capi images] Expose containerd version variables through packer

detiber:


figo:
this looks good to me, thanks for doing this.
one suggestion, we can expose containerd download URL to json file as well.
it all about making source configurable and visible.

detiber:
> this looks good to me, thanks for doing this.
one suggestion, we can expose containerd download URL to json file as well.
it all about making source configurable and visible.

I worry about this a bit because we aren't using the official containerd releases for this, it is a repackaged version that is used within the k8s infra. If we reworked this to use the containerd releases, then it would be easier to expose this without it being a major footgun.

PR #29, opened 2019-09-09
Build CAPI Ubuntu image for GCE

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -1,0 +1,75 @@
+{
+  "variables": {
+    "project_id": "{{env `GCP_PROJECT_ID`}}",
+    "build_timestamp": "{{timestamp}}",
+    "source_image_family": "ubuntu-1804-lts",
+    "machine_type": "n1-standard-1",
+    "encrypted": "false",
+    "zone": null,
+    "containerd_version": null,
+    "containerd_sha256": null,
+    "kubernetes_cni_rpm_version": null,
+    "kubernetes_cni_deb_version": null,
+    "kubernetes_cni_semver": null,
+    "kubernetes_cni_source_type": null,
+    "kubernetes_cni_http_source": null,
+    "kubernetes_series": null,
+    "kubernetes_semver": null,
+    "kubernetes_source_type": null,
+    "kubernetes_http_source": null,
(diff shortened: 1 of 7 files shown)


images/capi/packer/gce/packer.json
```diff
@@ -0,0 +1,75 @@
+{
+  "variables": {
+    "project_id": "{{env `GCP_PROJECT_ID`}}",
+    "account_file": "{{env `GOOGLE_APPLICATION_CREDENTIALS`}}",
```
detiber: Shouldn't these be picked up by the environment anyway if not specified?

images/capi/packer/gce/packer.json
```diff
@@ -0,0 +1,75 @@
...
+    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}"
+  },
+  "builders": [
+    {
+      "name": "ubuntu-1804",
+      "type": "googlecompute",
+      "project_id": "{{ user `project_id` }}",
+      "account_file": "{{ user `account_file` }}",
+      "machine_type": "n1-standard-1",
+      "source_image_family": "ubuntu-1804-lts",
```
detiber: Should this be exposed as a variable to allow users to override it?

images/capi/packer/gce/packer.json
vincepri: They weren't, at least in testing :/

images/capi/packer/gce/packer.json
vincepri: Is there a way to default it to this, unless the user specifies it?

detiber: Yes, setting it as the default value in the `variables` definition.

images/capi/packer/gce/packer.json
detiber: It should according to the docs: https://www.packer.io/docs/builders/googlecompute.html#precedence-of-authentication-methods

vincepri: Nice, need more coffee today 😄

images/capi/packer/gce/packer.json
vincepri: Fixed!

PR #41, opened 2019-09-17
Add targets to build images on remote ESX host

jiatongw:
Add `build-esx-ova-ubuntu-1804` and `build-esx-ova-centos-7` two targets to enable build on remote ESX host.

Special note: replaced "+" to "-" in VM names because `ovftool` unescapes "+" to " " (whitespace) and this will cause error when transferring images back to local.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -22,6 +22,7 @@ all: build
 OVA_BUILD_NAMES ?= ova-centos-7 ova-ubuntu-1804
 AMI_BUILD_NAMES ?= ami-default
 GCE_BUILD_NAMES ?= gce-default
+OVA_BUILD_NAMES_ESX ?= esx-ova-centos-7 esx-ova-ubuntu-1804
 
 # The version of Kubernetes to install.
 KUBE_JSON ?= packer/config/kubernetes.json packer/config/cni.json packer/config/containerd.json
@@ -35,6 +36,8 @@ PACKER_FLAGS += $(OLD_PACKER_FLAGS)
 OVA_BUILD_TARGETS := $(addprefix build-,$(OVA_BUILD_NAMES))
 AMI_BUILD_TARGETS := $(addprefix build-,$(AMI_BUILD_NAMES))
 GCE_BUILD_TARGETS := $(addprefix build-,$(GCE_BUILD_NAMES))
+OVA_BUILD_TARGETS_ESX := $(addprefix build-,$(OVA_BUILD_NAMES_ESX))
+
 $(OVA_BUILD_TARGETS):
 	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" packer/ova/packer.json
 .PHONY: $(OVA_BUILD_TARGETS)
@@ -46,6 +49,10 @@ $(AMI_BUILD_TARGETS):
 $(GCE_BUILD_TARGETS):
 	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/gce/$(subst build-,,$@).json)" packer/gce/packer.json
 .PHONY: $(GCE_BUILD_TARGETS)
+
+$(OVA_BUILD_TARGETS_ESX):
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=shell-local packer/ova/packer.json
+.PHONY: $(OVA_BUILD_TARGETS_ESX)
 
 CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES)) $(addprefix clean-,$(AMI_BUILD_NAMES)) $(addprefix clean-,$(GCE_BUILD_NAMES))
 $(CLEAN_TARGETS):
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -3,13 +3,15 @@
     "boot_wait": "10s",
     "build_timestamp": "{{timestamp}}",
     "capi_version": "v1alpha1",
-    "disable_public_repos": "false",
-    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
-    "extra_repos": "",
-    "headless": "true",
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "disable_public_repos": "false",
+    "disk_type_id": "0",
+    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
+    "extra_repos": "",
+    "format": "",
+    "headless": "true",
     "kubernetes_cni_rpm_version": null,
     "kubernetes_cni_deb_version": null,
     "kubernetes_cni_semver": null,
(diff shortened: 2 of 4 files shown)


jiatongw:
@figo 
cc @akutz

images/capi/packer/ova/variables.json
```diff
@@ -0,0 +1,6 @@
+{
```
figo: suggest to name the file as esx.json

images/capi/packer/ova/ova-centos-7-esx.json
```diff
@@ -0,0 +1,15 @@
+{
```
figo: if those two json works for non-esx case, then we only need to have one version of it.

images/capi/packer/ova/ova-ubuntu-1804-esx.json
```diff
@@ -0,0 +1,15 @@
...
+  "build_name": "ubuntu-1804",
+  "distro_name": "ubuntu",
+  "os_display_name": "Ubuntu 18.04",
+  "guest_os_type": "ubuntu-64",
+  "iso_url": "http://old-releases.ubuntu.com/releases/18.04.2/ubuntu-18.04.2-server-amd64.iso",
+  "iso_checksum": "a2cb36dc010d98ad9253ea5ad5a07fd6b409e3412c48f1860536970b073c98f5",
+  "iso_checksum_type": "sha256",
+  "ssh_username": "ubuntu",
+  "ssh_password": "ubuntu",
+  "boot_command_prefix": "<esc><wait><esc><wait><enter><wait>/install/vmlinuz auto console-setup/ask_detect=false console-setup/layoutcode=us console-setup/modelcode=pc105 debconf/frontend=noninteractive debian-installer=en_US fb=false initrd=/install/initrd.gz kbd-chooser/method=us keyboard-configuration/layout=USA keyboard-configuration/variant=USA locale=en_US netcfg/get_domain=local netcfg/get_hostname=localhost grub-installer/bootdev=/dev/sda preseed/url=",
```
figo: if removing `noapic` works for non-esx case, then we don't need to create new json file

akutz:
Hi @jiatongw,

Did you look at [this plug-in](https://github.com/jetbrains-infra/packer-builder-vsphere)? It would be nice to make this work.

jiatongw:
> Hi @jiatongw,
> 
> Did you look at [this plug-in](https://github.com/jetbrains-infra/packer-builder-vsphere)? It would be nice to make this work.

Hi Andrew, yea you mentioned this to me last time. I know your concern. Although the method in this PR require access to ESX host, it is working stably, and requesting an ESX is faster than requesting a VC. So after consideration, I was thinking adding this first and then I'll dig into the plugin and add another method. So we will have two methods to let people choose.

images/capi/packer/ova/ova-ubuntu-1804-esx.json
jiatongw: OK. I'll do some change

images/capi/packer/ova/ova-centos-7-esx.json
codenrhoden: This file is identical to the non-esx version. It's not needed that I can see:

```sh
$ git diff master:images/capi/packer/ova/ova-centos-7.json jiatong/packer-esx:images/capi/packer/ova/ova-centos-7-esx.json
$
```

images/capi/packer/ova/packer-esx.json
```diff
@@ -0,0 +1,109 @@
...
+        "build_date": "{{isotime}}",
+        "capi_version": "{{user `capi_version`}}",
+        "kubernetes_cni_semver": "{{user `kubernetes_cni_semver`}}",
+        "kubernetes_semver": "{{user `kubernetes_semver`}}",
+        "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
+        "iso_checksum": "{{user `iso_checksum`}}",
+        "iso_checksum_type": "{{user `iso_checksum_type`}}",
+        "iso_url": "{{user `iso_url`}}",
+        "os_name": "{{user `os_display_name`}}"
+      }
```
codenrhoden: So when I look at the differences between this file and the non-ESX version, it's basically the `+` vs `-` change (which could be adopted locally), and a few different settings:

```
-    "vnc_bind_address": "127.0.0.1",
     "vnc_port_min": "5900",
     "vnc_port_max": "6000"
   },
   "builders": [
     {
       "name": "{{user `build_name`}}",
-      "vm_name": "{{build_name}}+kube-{{user `kubernetes_semver`}}",
+      "vm_name": "{{build_name}}-kube-{{user `kubernetes_semver`}}",
       "vmdk_name": "{{build_name}}",
-      "output_directory": "./output/{{build_name}}+kube-{{user `kubernetes_semver`}}",
+      "output_directory": "./output/{{build_name}}-kube-{{user `kubernetes_semver`}}",
       "type": "vmware-iso",
       "cpus": 1,
       "cores": 1,
       "memory": 2048,
       "disk_size": 20480,
-      "disk_type_id": 0,
+      "disk_type_id": "thin",
       "disk_adapter_type": "scsi",
@@ -57,10 +56,21 @@
       ],
       "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E {{user `shutdown_command`}}",
-      "skip_compaction": false,
+      "skip_compaction": true,
       "vnc_bind_address": "{{user `vnc_bind_address`}}",
       "vnc_port_min": "{{user `vnc_port_min`}}",
-      "vnc_port_max": "{{user `vnc_port_max`}}"
+      "vnc_port_max": "{{user `vnc_port_max`}}",
+      "remote_type": "esx5",
+      "remote_host": "{{user `esxi_host`}}",
+      "remote_host": "{{user `esxi_host`}}",
+      "remote_datastore": "{{user `esxi_datastore`}}",
+      "remote_username": "{{user `esxi_username`}}",
+      "remote_password": "{{user `esxi_password`}}",
+      "keep_registered": true,
+      "vnc_disable_password": true,
+      "format": "ova",
+      "vmx_data": {
+        "ethernet0.networkName": "VM Network"
+      }
```

All of that could be parameterized in `esx.json`, so that  you "turn" on remote ESX by adding that as a different var file (we could set the defaults up above in the `variables` section).

The **big** differences is the removal of the two post-processors. Their skipping [can be made conditional](https://www.packer.io/docs/templates/post-processors.html#run-on-specific-builds).

So I think it's possible to accomplish all this by adding only one new file (`esx.json`) and passing it in for the specific ESX build target.

images/capi/packer/ova/packer.json
```diff
@@ -30,20 +30,29 @@
...
-      "output_directory": "./output/{{build_name}}+kube-{{user `kubernetes_semver`}}",
+      "output_directory": "./output/{{build_name}}-kube-{{user `kubernetes_semver`}}",
       "type": "vmware-iso",
       "cpus": 1,
       "cores": 1,
       "memory": 2048,
       "disk_size": 20480,
       "disk_type_id": 0,
+      "disk_type_id": "{{user `disk_type_id`}}",
       "disk_adapter_type": "scsi",
```
jiatongw: need to have this duplicate key to pass user variables when remote build and keep local build the `id:0`

images/capi/packer/ova/packer.json
```diff
@@ -62,10 +71,20 @@
...
+      "remote_host": "{{user `remote_host`}}",
+      "remote_datastore": "{{user `remote_datastore`}}",
+      "remote_username": "{{user `remote_username`}}",
+      "remote_password": "{{user `remote_password`}}",
+      "vnc_disable_password": "{{user `vnc_disable_password`}}",
+      "format": "{{user `format`}}",
+      "vmx_data": {
+        "ethernet0.networkName": "VM Network"
+      }
     }
```
jiatongw: `vmx_data` is used for remote ESX build. But it doesn't interfere local build as I tested. Putting it there is because passing through a user variable will cause json `marshal error`

images/capi/Makefile
```diff
@@ -35,8 +36,10 @@ PACKER_FLAGS += $(OLD_PACKER_FLAGS)
 OVA_BUILD_TARGETS := $(addprefix build-,$(OVA_BUILD_NAMES))
 AMI_BUILD_TARGETS := $(addprefix build-,$(AMI_BUILD_NAMES))
 GCE_BUILD_TARGETS := $(addprefix build-,$(GCE_BUILD_NAMES))
+OVA_BUILD_TARGETS_ESX := $(addprefix build-,$(OVA_BUILD_NAMES_ESX))
+
 $(OVA_BUILD_TARGETS):
-	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" packer/ova/packer.json
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" -var 'disk_type_id=0' packer/ova/packer.json
 .PHONY: $(OVA_BUILD_TARGETS)
```
jiatongw: b/c disk_type_id will set to a string while build remotely, if set `disk_type_id: 0` as default and pass `disk_type_id: "thin"` in a user variable, it will cause json marshal error. So I deleted `disk_type_id` key in `packer.json`, but to set it there

codenrhoden: Thanks for the explanation... So, from what I understand, the issues is because we want to set a default of `0`, an integer, but then later override it with `"thin"`, a string...

I know the docs aren't entirely helpful on this, but is there any way we can figure out what the string version of `0` is? Or if we have to go the other way, what the int version of `"thin"` is? Then we could avoid this extra `-var`, which I think is going to trip people up...

codenrhoden: One more comment here... Since the [docs](https://www.packer.io/docs/builders/vmware-iso.html#disk_type_id) say this field is actually a string, I wonder, does setting the default to `"0"` instead of `0` work? If it does, then overriding it with `"thin"` should work as well. Maybe I'll try that this morning if I have time.

codenrhoden: I gave it a quick try, and I confirm that using `"0"` and `"1"` here worked fine. On the desktop (Fusion), 0 created a single disk, and 1 created several split disks, just as it was supposed to. So I think you can just set it using the string, and then override it with `"thin"` in `esx.json`.

jiatongw: Fixed @codenrhoden

codenrhoden:
I was able to test that the "non-ESX" case worked fine locally.

/approve

PR #42, opened 2019-09-18
Support for configuration of non-default VPC and Subnet

gsadhani:
Added two new user configurable parameters `vpc_id` and `subnet_id`.

This allows us to build AMIs by launching instances in a specific VPCs which may have routing configured as needed.

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -28,7 +28,9 @@
     "ami_users": "",
     "snapshot_groups": "all",
     "snapshot_users": "",
-    "ami_regions": "ap-south-1,eu-west-3,eu-west-2,eu-west-1,ap-northeast-2,ap-northeast-1,sa-east-1,ca-central-1,ap-southeast-1,ap-southeast-2,eu-central-1,us-east-1,us-east-2,us-west-1,us-west-2"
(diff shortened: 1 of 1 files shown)


images/capi/packer/ami/packer.json
```diff
@@ -28,7 +28,9 @@
     "ami_users": "",
     "snapshot_groups": "all",
     "snapshot_users": "",
-    "ami_regions": "ap-south-1,eu-west-3,eu-west-2,eu-west-1,ap-northeast-2,ap-northeast-1,sa-east-1,ca-central-1,ap-southeast-1,ap-southeast-2,eu-central-1,us-east-1,us-east-2,us-west-1,us-west-2"
+    "ami_regions": "ap-south-1,eu-west-3,eu-west-2,eu-west-1,ap-northeast-2,ap-northeast-1,sa-east-1,ca-central-1,ap-southeast-1,ap-southeast-2,eu-central-1,us-east-1,us-east-2,us-west-1,us-west-2",
+    "vpc_id": null,
+    "subnet_id": null
```
detiber: Instead of `null`, which makes these required variables can this use `""` instead? That way the current behavior is preserved if the `vpc_id` and `subnet_id` are not provided.

gsadhani: Fixed.

codenrhoden:
Thanks for the quick turn around, @gsadhani !
/lgtm

PR #43, opened 2019-09-18
Add ability to disable existing repos

codenrhoden:
This patch adds two new variables to Ansible, and Packer support for
    them when building.

    disable_public_repos: Setting this flag to true will disable all
    existing repo definitions, therefore requiring that new repos are added
    via `extra_repos`. The default behavior is false.

    reenable_public_repos: Setting this flag to false will cause the
    repos disabled by setting `disable_public_repos: true` to *not* be
    reenabled and leave them in a disabled state. The default behavior is
    true.

/cc @figo


codenrhoden:
I'll devise a cleaner way to disable existing repo definitions than just deleting them...  For now, though, the injection of arbitrary repo files is handled in #44.

codenrhoden:
Okay, I've come up with a cleaner and more flexible solution. This patch builds off of #44 (this should go in after).

With this and #44 in place, a consumer can disable all the existing repos, inject their own, remove those extra repos at the end, and choose whether or not to re-enable the existing repos when everything is done.

Note this does *not* change anything to do with the Kubernetes repo that is added if installing Kubernetes from a package source.

codenrhoden:
/hold cancel
/assign @detiber 

Okay, this is ready for review!

images/capi/ansible/roles/common/tasks/redhat.yml
```diff
@@ -30,6 +30,19 @@
...
+- name: Find existing repo files
+  find:
+    depth: 1
+    paths: /etc/yum.repos.d
+    patterns: '*.repo'
+  register: repo_files
+  when: disable_public_repos|bool
+
+- name: Disable repos
+  command: "mv {{ item.path }} {{ item.path }}.disabled"
```
detiber: Does this work for yum repo configs? I would have expected that we would need to modify the config for the defined repos instead.

codenrhoden: It does. It will only read files ending in `*.repo`

```sh
[root@localhost yum.repos.d]# ls
CentOS-Base.repo  CentOS-CR.repo  CentOS-Debuginfo.repo  CentOS-fasttrack.repo  CentOS-Media.repo  CentOS-Sources.repo  CentOS-Vault.repo  epel.repo  epel-testing.repo
[root@localhost yum.repos.d]# yum repolist
repo id                                                                          repo name                                                                                                      status
base/7/x86_64                                                                    CentOS-7 - Base                                                                                                10,019
epel/x86_64                                                                      Extra Packages for Enterprise Linux 7 - x86_64                                                                 13,343
extras/7/x86_64                                                                  CentOS-7 - Extras                                                                                                 435
updates/7/x86_64                                                                 CentOS-7 - Updates                                                                                              2,500
```
^ You can see EPEL is present. Then after moving it...

```sh
[root@localhost yum.repos.d]# mv epel.repo epel.repo.disabled
[root@localhost yum.repos.d]# yum repolist
repo id                                                                                        repo name                                                                                        status
base/7/x86_64                                                                                  CentOS-7 - Base                                                                                  10,019
extras/7/x86_64                                                                                CentOS-7 - Extras                                                                                   435
updates/7/x86_64                                                                               CentOS-7 - Updates                                                                                2,500
```

PR #44, opened 2019-09-18
Add ability to inject/remove extra repos

codenrhoden:
This patch adds to two new variables to Ansible, and Packer support for
them when building.

`extra_repos`: This is a comma delimited list of files to copy into the
build target as repo definitions. For CentOS, these should be `.repo`
files and will go to `/etc/yum.repos.d`, and for Ubuntu these should be
`.list` files and will to `/etc/apt/sources.list.d`.

The list of files can use relative or absolute paths. If using relative
paths, the paths must reference files within Ansible's search
directories.

`remove_extra_repos`: This is a boolean flag that controls whether the
repo files added via `extra_repos` will be deleted during the sysprep
task. The default is `false`.

The motivation behind this PR is for enterprise consumers to be able to add repo definitions pointing to internal resources. If used, it may be desirable to remove those internal repo definitions afterwards to avoid leaking internal details.

### DIFF
--- a/images/capi/ansible/roles/common/tasks/debian.yml
+++ b/images/capi/ansible/roles/common/tasks/debian.yml
@@ -16,6 +16,13 @@
   apt_repository:
     repo: "{{ common_deb_cloud_init_ppr_repo }}"
 
+- name: Install extra repos
+  copy:
+    src: "{{ item }}"
+    dest: "/etc/apt/sources.list.d/{{ item | basename }}"
+  loop: "{{ extra_repos.split(',') }}"
+  when: extra_repos != ""
+
 - name: update apt cache
   apt:
     force_apt_get: True
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -4,6 +4,7 @@
     "build_timestamp": "{{timestamp}}",
     "capi_version": "v1alpha1",
     "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
+    "extra_repos": "",
     "headless": "true",
     "containerd_version": null,
     "containerd_sha256": null,
@@ -24,6 +25,7 @@
     "kubernetes_deb_gpg_key": null,
     "kubernetes_rpm_gpg_check": null,
     "kubernetes_container_registry": null,
+    "remove_extra_repos": "false",
     "vnc_bind_address": "127.0.0.1",
     "vnc_port_min": "5900",
     "vnc_port_max": "6000"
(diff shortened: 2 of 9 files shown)


codenrhoden:
I think the one thing I would request feedback on is if we want to provide a list of repos to inject, do people prefer a comma separated list, or space delimited? Right now I'm doing space delimited, so you can do something like this:

`ansible-playbook --extra-vars "extra_repos='~/git/repo1.list ~/git/repo2.list' remove_extra_repos=true" -i ./hosts -l 192.168.43.164 images/capi/ansible/playbook.yml`

Or if within a Packer var file...

```
extra_repos: "~/git/repo1.list ~/git/repo2.list"
```

gsadhani:
> I think the one thing I would request feedback on is if we want to provide a list of repos to inject, do people prefer a comma separated list, or space delimited? Right now I'm doing space delimited, so you can do something like this:
> 
> `ansible-playbook --extra-vars "extra_repos='~/git/repo1.list ~/git/repo2.list' remove_extra_repos=true" -i ./hosts -l 192.168.43.164 images/capi/ansible/playbook.yml`
> 
> Or if within a Packer var file...
> 
> ```
> extra_repos: "~/git/repo1.list ~/git/repo2.list"
> ```

I personally prefer comma separated list.

images/capi/packer/ova/packer.json
```diff
@@ -4,6 +4,7 @@
     "build_timestamp": "{{timestamp}}",
     "capi_version": "v1alpha1",
     "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
+    "extra_repos": "",
```
detiber: Should this be plumbed through the other `packer.json` files as well?

codenrhoden: probably. :)

I'll fix that up, and switch to a comma separated list on Monday.

codenrhoden:
@gsadhani Updated code and commit message that `extra_repos` is now comma separated rather than space separated.
@detiber Add support for the new new variables into the AMI and GCE builders as well as OVA.

I believe this one is ready to go.

images/capi/ansible/roles/common/tasks/debian.yml
```diff
@@ -16,6 +16,13 @@
   apt_repository:
     repo: "{{ common_deb_cloud_init_ppr_repo }}"
 
+- name: Install extra repos
+  copy:
+    src: "{{ item }}"
+    dest: "/etc/apt/sources.list.d/{{ item | basename }}"
+  loop: "{{ extra_repos.split(',') }}"
+  when: extra_repos != ""
```
detiber: Have you tested to see if this `when` is actually necessary? I would expect `split()` and `loop` to do the right thing if extra_repos == "" here.

Not a blocker for this PR by any means, though.

codenrhoden: when I comment out that `when` line, I see this:
```sh
$ ansible-playbook -i ./hosts -l 192.168.43.176 images/capi/ansible/playbook.yml

PLAY [all] *******************************************************************************************************************************************************************************************

TASK [Gathering Facts] *******************************************************************************************************************************************************************************
ok: [192.168.43.176]

TASK [common : Add cloud-init ppa repo] **************************************************************************************************************************************************************
ok: [192.168.43.176]

TASK [common : Install extra repos] ******************************************************************************************************************************************************************
failed: [192.168.43.176] (item=) => {"ansible_loop_var": "item", "changed": false, "item": "", "msg": "src (or content) is required"}

PLAY RECAP *******************************************************************************************************************************************************************************************
192.168.43.176             : ok=2    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

I didn't pass any `extra_vars` in that call, so the default empty string is being passed. I was surprised.

PR #45, opened 2019-09-19
OVA: Make CentOS kickstart not rely on internet

codenrhoden:
This patch changes the ova centos kickstart to only install packages
from the cdrom/iso. We are already relying on Ansible as a follow-on
provisioner to install required packages after the OS is installed, so
these steps aren't really necessary. There was also a good bit of
overlap where the Ansible tasks were installing packages that were
listed in the kickstart -- this just leaves it to Ansible.

This makes building in enterprise environments w/o internet access
simpler as well.

### DIFF
--- a/images/capi/ansible/roles/common/defaults/main.yml
+++ b/images/capi/ansible/roles/common/defaults/main.yml
@@ -13,14 +13,18 @@
 # limitations under the License.
 ---
 common_rpms:
-- yum-utils
-- python2-pip
-- python-requests
+- ca-certificates
+- conntrack-tools
+- curl
 - ebtables
-- socat
-- ntp
-- jq
 - nfs-utils
+- ntp
+- open-vm-tools
+- python2-pip
+- python-netifaces
+- python-requests
+- socat
+- yum-utils
 common_extra_rpms: []
 common_debs:
 - openssh-client
--- a/images/capi/packer/ova/linux/centos/http/7/ks.cfg
+++ b/images/capi/packer/ova/linux/centos/http/7/ks.cfg
@@ -15,7 +15,6 @@
 # Perform a fresh install, not an upgrade
 install
 cdrom
-url --url=http://mirror.centos.org/centos/7/os/x86_64/
 
 # Perform a text installation
 text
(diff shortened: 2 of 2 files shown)


images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -13,14 +13,17 @@
...
-- socat
-- ntp
 - jq
 - nfs-utils
+- ntp
+- python2-pip
+- python-netifaces
+- python-requests
+- socat
+- vim
```
figo: we probably don't need vim

codenrhoden: it was getting installed before, so I kept it. I personally would want vim on any machine I touch, but I also realize these are supposed to be sealed images.

I'd like to make as few changes here as possibly, so I'd like to keep it for now.

codenrhoden:
> Are you able to test the image end to end?

I haven't yet. I should fire up my account on VMC and try it out I suppose.

codenrhoden:
/hold
^ to do E2E tests with CAPV

images/capi/ansible/roles/common/defaults/main.yml
timothysc: we should bring only what we need to survive and enable consumers to add any packages they need as a consumer of this repo.

codenrhoden:
Okay, this is finally tested at ready for re-review. To recap...

The kickstart no longer allows reaching out to the internet to install packages. Furthermore, the packages that do get installed for the base OS only came from the ISO, and are only the following:

```
openssh-server
sed
sudo
```

This ensures a minimal image and one that is ready for installing packages later with Ansible. All other packages are left to be installed with Ansible. There was feedback on the Ubuntu version of this PR that we want the smallest number of packages installed by default in this repo, and leave it to others to customize with other packages later. I'll put some in-line comments on the packages that were left around.

/hold cancel
@detiber can you take a look at this?

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -13,14 +13,18 @@
...
 ---
 common_rpms:
-- yum-utils
+- ca-certificates
+- conntrack-tools
+- curl
+- ebtables
+- nfs-utils
+- ntp
+- open-vm-tools
```
codenrhoden: This wasn't present before, but should be there for images running on vSphere infra. This is the "VM tools" package.

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -13,14 +13,18 @@
...
 - python2-pip
+- python-netifaces
 - python-requests
-- ebtables
 - socat
-- ntp
-- jq
-- nfs-utils
+- yum-utils
```
codenrhoden: `vim`, `jq`, and `nginx` were all removed. Any reason that `nginx` would be required?

images/capi/packer/ova/linux/centos/http/7/ks.cfg
```diff
@@ -47,35 +46,13 @@ eula --agreed
...
-ebtables
-jq
-libnetfilter_conntrack
-libnetfilter_cthelper
-libnetfilter_cttimeout
-libnetfilter_queue
-nfs-utils
-nginx
-ntp
-openssh-clients
```
codenrhoden: an SSH client does not seem strictly required.

images/capi/ansible/roles/common/defaults/main.yml
detiber: :grimacing: I hope not...

PR #46, opened 2019-09-20
Support for non-default containerd download URL

gsadhani:
Adding support to configure `containerd` download URL from non-default location from the internet.

This is needed to support installation of locally built `containerd`.

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -22,7 +22,7 @@
 # See https://github.com/containerd/containerd/issues/1508 for context.
 - name: download containerd
   get_url:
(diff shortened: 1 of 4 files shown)


gsadhani:
cc: @figo @codenrhoden

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -22,7 +22,7 @@
 # See https://github.com/containerd/containerd/issues/1508 for context.
 - name: download containerd
   get_url:
-    url: https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{ containerd_version }}.linux-amd64.tar.gz
+    url: "{{ containerd_url }}"
```
detiber: Should there be a followup here to add some documentation around which containerd tarballs satisfy the requirements to be used here? I worry that someone might attempt to use the tarballs provided by the containerd project rather than the ones that are built as part of the Kubernetes build, test, and release process.

PR #47, opened 2019-09-20
OVA: Make ubuntu preseed not rely on internet

codenrhoden:
This patch changes the ova ubuntu preseed to only install packages
from the cdrom/iso. We are already relying on Ansible as a follow-on
provisioner to install required packages after the OS is installed, so
these steps aren't really necessary. There was also a good bit of
overlap where the Ansible tasks were installing packages that were
listed in the kickstart -- this just leaves it to Ansible.

This makes building in enterprise environments w/o internet access
simpler as well.

This is the Ubuntu flavor of #45.

/assign @detiber 

I've again left nginx out. If that's needed I can add it back.

### DIFF
--- a/images/capi/ansible/roles/common/defaults/main.yml
+++ b/images/capi/ansible/roles/common/defaults/main.yml
@@ -25,15 +25,28 @@ common_rpms:
 - python-requests
 - socat
 - yum-utils
+
 common_debs:
-- openssh-client
-- openssh-server
 - apt-transport-https
+- conntrack
+- conntrackd
+- curl
 - ebtables
-- socat
-- ntp
 - jq
-- nfs-client
+- gnupg
+- libnetfilter-acct1
+- libnetfilter-cttimeout1
+- libnetfilter-log1
+- linux-cloud-tools-virtual
+- linux-tools-virtual
+- nfs-common
+- ntp
+- open-vm-tools
+- python3-distutils
+- python3-netifaces
+- python3-pip
+- socat
+
 common_photon_rpms:
 - conntrack-tools
 - distrib-compat
--- a/images/capi/ansible/roles/common/tasks/debian.yml
+++ b/images/capi/ansible/roles/common/tasks/debian.yml
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+- name: Put templated sources.list in place
+  template:
+    src: etc/apt/sources.list.j2
+    dest: /etc/apt/sources.list
+
 - name: Find existing repo files
   find:
     depth: 1
@@ -38,7 +43,6 @@
   apt:
     force_apt_get: True
     update_cache: True
-    cache_valid_time: 3600
 
 - name: perform a dist-upgrade
   apt:
--- a/images/capi/ansible/roles/common/templates/etc/apt/sources.list.j2
+++ b/images/capi/ansible/roles/common/templates/etc/apt/sources.list.j2
@@ -1,0 +1,4 @@
+deb http://us.archive.ubuntu.com/ubuntu {{ ansible_distribution_release }} main restricted universe
+deb http://us.archive.ubuntu.com/ubuntu {{ ansible_distribution_release }}-updates main restricted universe
+deb http://us.archive.ubuntu.com/ubuntu {{ ansible_distribution_release }}-backports main restricted universe
+deb http://security.ubuntu.com/ubuntu {{ ansible_distribution_release }}-security main restricted universe
--- a/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
+++ b/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
@@ -90,37 +90,15 @@ d-i grub-installer/with_other_os boolean true
 d-i finish-install/reboot_in_progress note
 d-i pkgsel/update-policy select none
 
-# Select the apt mirror.
-d-i mirror/country string manual
-d-i mirror/http/hostname string us.archive.ubuntu.com
-d-i mirror/http/directory string /ubuntu
-d-i mirror/http/proxy string
+# Disable use of the apt mirror during base install
+# This means only packages available in the ISO can be installed
+d-i apt-setup/use_mirror boolean false
+
+# Disable the security repo as well (it's on by default)
+d-i apt-setup/services-select multiselect none
 
 # Customize the list of packages installed.
-d-i pkgsel/include string apt-transport-https \
-                          curl \
-                          ebtables \
-                          gnupg gnupg1 gnupg2 \
-                          jq \
-                          linux-cloud-tools-virtual \
-                          linux-tools-virtual \
-                          open-vm-tools \
-                          openssh-client openssh-server \
-                          conntrack conntrackd \
-                          libnetfilter-acct1 \
-                          libnetfilter-conntrack3 \
-                          libnetfilter-cthelper0 \
-                          libnetfilter-cttimeout1 \
-                          libnetfilter-log1 \
-                          libnetfilter-queue1 \
-                          nfs-client \
-                          ntp \
-                          python3-distutils \
-                          python3-netifaces \
-                          python3-pip \
-                          sed \
-                          socat \
-                          vim
+d-i pkgsel/include string openssh-server
 
 
 # Ensure questions about these packages do not bother the installer.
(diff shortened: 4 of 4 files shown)


codenrhoden:
/hold
^ to do E2E tests with CAPV

codenrhoden:
Dang it -- I found some more issues with this. Have I mentioned that PRESEED files are so much harder than kickstart?  😠 

By default, the preseed will still install security repos (security.ubuntu.com). Which means that every `apt-get` reaches out to the internet to fetch package lists, even if nothing is installed from them. I'll have to explicitly disable those, then add them back in with the Ansible.

Furthermore, I missed another important detail. Even though I left the part about defining the mirror, but then tell it not to *use* the mirror, it actually does not drop those repo definitions in. So I will have to explicitly add those back via Ansible as well.

There appears to be know way to tell the system "define all your normal repos, but don't use them". I shouldn't be surprised since there is no `apt` equivalent of `yum`'s `--disablerepo`. Oh well... Back to the drawing board.

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -24,18 +24,29 @@ common_rpms:
...
-- socat
-- ntp
 - jq
-- nfs-client
+- gnupg
+- gnupg1
+- gnupg2
+- libnetfilter-acct1
+- libnetfilter-cttimeout1
+- libnetfilter-log1
```
detiber: Do these by chance get installed as dependencies with conntrack/conntrackd?

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -24,18 +24,29 @@ common_rpms:
...
+- linux-cloud-tools-virtual
+- linux-tools-virtual
+- nfs-common
```
detiber: Was the change from `nfs-client` to `nfs-common` intentional here?

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -24,18 +24,29 @@ common_rpms:
...
 - jq
-- nfs-client
+- gnupg
+- gnupg1
+- gnupg2
+- libnetfilter-acct1
+- libnetfilter-cttimeout1
+- libnetfilter-log1
+- linux-cloud-tools-virtual
+- linux-tools-virtual
```
detiber: I'm not familiar with `linux-tools-virtual` or `linux-cloud-tools-virtual`, should I be aware of any implications by installing these on the gcp/aws images as well?

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -24,18 +24,29 @@ common_rpms:
...
+- conntrack
+- conntrackd
 - ebtables
-- socat
-- ntp
 - jq
-- nfs-client
+- gnupg
+- gnupg1
+- gnupg2
```
detiber: I'm not overly familiar with the vsphere provider, is there a specific requirement there that requires the `gnupg` packages?

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: They don't. But  libnetfilter-conntrack3 and libnetfilter-cthelper0 do, which is why I did not move them over from the preseed file.

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: It was. when you run `apt-get install nfs-client`, it spits out the following:

```
Note, selecting 'nfs-common' instead of 'nfs-client'
```

There [is no package](https://packages.ubuntu.com/search?keywords=nfs-client&searchon=names&suite=bionic&section=all) named `nfs-client` in Bionic, but they appear to have a handy redirect for it.

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: I'm not either. This was just keeping what was already there. I can take it as a follow-on to investigate if these are actually needed.

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: Funny enough, I had to make a local change (that I haven't pushed yet) to fix something around this. It's not about CAPV -- when we do that PPA repo add for cloud-init, it imports the PPA GPG key. If these aren't installed, it fails.

However, I do think it can be reduced to just `gnupg`, as that is an alias for `gnupg2`, and that should be fine.

codenrhoden:
/remove-lifecycle stale
/hold cancel

Okay, I finally came back to this. Working with preseeds here was a bit more complicated that kickstart files...

The strategy here is to disable the apt mirror in the preseed, and only install whatever is needed to get ansible to work (and it comes from the ISO). In Ubuntu's case, that's now only one package: `openssh-server`. Everything else is done via Ansible, so it's a clearer to see in one single place what the list of packages to install are.

However, there is no way to say in the preseed "disable the package mirror during the OS install, but re-enable it when done". If it is disabled in the preseed, it is disabled in the OS. Thus, an Ansible template is added to put in the correct package list at the very begging of the playbook. This package list will still get overridden if you are specifying a custom mirror later on.

images/capi/ansible/roles/common/tasks/debian.yml
```diff
@@ -38,7 +43,6 @@
   apt:
     force_apt_get: True
     update_cache: True
-    cache_valid_time: 3600
```
codenrhoden: I'm surprised I didn't hit this in earlier testing when using alternate repo definitions, but this line is causing `apt-get update` to not run if the cache is recent. We pretty much always want it to run because we've made changes to repo definitions, so removing this line takes it back to it's default value of `0`.

images/capi/ansible/roles/common/templates/etc/apt/sources.list.j2
```diff
@@ -0,0 +1,4 @@
+deb http://us.archive.ubuntu.com/ubuntu {{ ansible_distribution_release }} main restricted universe
```
codenrhoden: FYI, this leaves out the `multiverse` repo as it wasn't needed. I tried to leave out `universe` as well, but it's needed for at least `conntrackd`. I haven't checked if `restricted` is required.

codenrhoden:
@detiber PTAL again when you are free

images/capi/ansible/roles/common/defaults/main.yml
akutz: Both packages provide parts of the kernel required to run on certain, popular virtualization platforms, like...VMware :) I added those packages [because](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/supported-ubuntu-virtual-machines-on-hyper-v):

> linux-cloud-tools packages to install tools and daemons for use with virtual machine

However, I later discovered that since Ubuntu 12, the virtual kernel bits are [part of the generic kernel image](https://help.ubuntu.com/community/ServerFaq#What_are_the_differences_between_the_server_and_virtual_kernels.3F).

In fact, if you look at the actual packages, you'll notice the only files they include are [a changeling and license](https://packages.ubuntu.com/bionic/amd64/linux-cloud-tools-virtual/filelist).

I think these are safe to remove, but they're just as likely safe to keep as well.

codenrhoden: Ah, thanks for digging that up @akutz. I'll just leave for now I guess, but good to know we can remove in future if desired.

I've rebased this to resolve conflicts.

codenrhoden:
@detiber Any chance you can take another look at this one?

images/capi/ansible/roles/common/templates/etc/apt/sources.list.j2
detiber: Would it make sense to use the apt_repository module for this instead of a templated file?

detiber: Not a blocker by any means, but the use of the apt_repository module, may allow for you to refresh the cache there rather than having to update the cache in the apt module call above.

codenrhoden: I did look at this...

If I use `apt_repository`, each repo definition is added as a new `.list` file in `/etc/apt/sources.list.d`. Which normally would be desired. But in this case, I specifically want to overwrite the default file at `/etc/apt/sources.list`, and I don't see a way to do that with `apt_repository` or `apt_repo`.

PR #49, opened 2019-09-23
Azure support

vincepri:


CecileRobertMichon:
To use:
1) install Packer and Ansible
2) define all env variables
3) create RG and storage account in correct sub/location
3) from capi dir, run `make build-azure-default`

vincepri:
@CecileRobertMichon is gonna be back next week, we'll sync up and draw a plan of action for this PR

justaugustus:
/assign
(will add to this week's capz agenda)

images/capi/packer/azure/packer.json
```diff
@@ -0,0 +1,113 @@
...
+    "kubernetes_rpm_gpg_key": null,
+    "kubernetes_deb_repo": null,
+    "kubernetes_deb_gpg_key": null,
+    "kubernetes_rpm_gpg_check": null,
+    "kubernetes_container_registry": null,
+    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}"
+  },
+  "builders": [
+    {
+      "name": "vhd-ubuntu-1804",
```
CecileRobertMichon: @vincepri is this what you meant?

CecileRobertMichon:
(this is still work in progress)

images/capi/packer/azure/Makefile
```diff
@@ -0,0 +1,39 @@
+# Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
CecileRobertMichon: yeah I think you're right, this was probably copy-pasted from the aws folder

justaugustus:
@CecileRobertMichon @vincepri -- Would suggest:
- Gallery name: `ClusterAPI` (not sure if this can be hyphenated)
- Image name: `ClusterAPIUbuntu`
- Version: `0.3.0-alpha.1-<some semi-unique value>` (like timestamp or `0..9`)

CecileRobertMichon:
@justaugustus 

> Gallery name: ClusterAPI (not sure if this can be hyphenated)
Image name: ClusterAPIUbuntu
Version: 0.3.0-alpha.1-<some semi-unique value> (like timestamp or 0..9)

Gallery name can't be hyphenated. For now I set it to `capzsig` by default (but it's user configurable), I can change the default to `ClusterAPI` if we prefer that.
Image name: I would prefer something like `capi-ubuntu-1804` that includes the ubuntu SKU so we can have multiple base ubuntu SKUs in the future. 
Version: agree with adding timestamp in the version so it can be unique. Why do we need `0.3.0-alpha.1`? Do we intend the images to be tightly coupled with capz versions? I don't think that makes sense, this should be independent from capz releases/code.

images/capi/packer/azure/azure-default.json
```diff
@@ -0,0 +1,7 @@
+{
+  "location": "southcentralus",
+  "vm_size": "Standard_B2ms",
+  "resource_group_name": "cluster-api-images",
+  "storage_account_name": "ClusterAPIImages",
+  "shared_image_gallery_name": "ClusterAPI"
```
CecileRobertMichon: @justaugustus let me know what you think about these defaults

CecileRobertMichon:
/hold
 
doing some validation

images/capi/packer/azure/azure-default.json
```diff
@@ -0,0 +1,7 @@
+{
+  "location": "southcentralus",
```
CecileRobertMichon: Packer defaults don't work the way I thought they did... setting variables in a file will override any env var variables https://www.packer.io/docs/templates/user-variables.html#from-a-file

I need to fix this (right now the values in this file are used, no matter what the env vars are set to).

timothysc:
Only comment, please squash commits before final lgtm+approve please.

images/capi/packer/azure/azure-default.json
detiber: We handle this in AWS by setting the vars to `""` for a default, which will pick up the ENV var if set.

images/capi/packer/azure/packer.json
```diff
@@ -0,0 +1,114 @@
+{
+  "variables": {
+    "subscription_id": "{{env `AZURE_SUBSCRIPTION_ID`}}",
```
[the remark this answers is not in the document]
vincepri: IIRC, I made it up, but waiting for Cecile's input :)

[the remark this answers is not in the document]
CecileRobertMichon: They are:
https://github.com/Azure/azure-sdk-for-go#more-authentication-details
https://sourcegraph.com/github.com/Azure/go-autorest/-/blob/autorest/azure/auth/auth.go#L40

@detiber are you saying that I don't need to specify `env AZURE_SUBSCRIPTION_ID`? ie this would be `"azure_subscription_id": "",` and Packer will pick up the `AZURE_SUBSCRIPTION_ID` env variable?

images/capi/packer/azure/azure-default.json
CecileRobertMichon: won't the default value in the var file override whatever ENV var is set?

images/capi/packer/azure/packer.json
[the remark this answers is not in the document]
CecileRobertMichon: I tried that but it's not picking up my env var 😕

images/capi/packer/azure/azure-default.json
CecileRobertMichon: I verified that the location used is whatever is in the defaults file, even if the value in packer.json is `""` so not sure how this works for AWS, env var override doesn't seem to be possible...

images/capi/packer/azure/packer.json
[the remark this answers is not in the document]
detiber: well shucks, it was worth a shot at least :)

images/capi/packer/azure/azure-default.json
CecileRobertMichon: so it's not possible to make the location configurable? It will always use the value hardcoded in the defaults file if I understand correctly?

CecileRobertMichon: I guess the only way to provide different values is to pass in a different var file...

detiber: One would either need to:
- Update the defaults file
- Provide a separate defaults file

CecileRobertMichon: In which case might as well set defaults in packer.json and override values with the user provided var file no?

detiber: In general, for other providers I believe the defaults file is mostly to ease use of the Makefile, with advanced use cases expected to modify and/or consume the packer.json file in their own way

detiber: Without more of an idea of the intended workflow (or if Azure supports public images in the way that AWS does), it's hard for me to say. 

For AWS, the makefile target creates the default images that are picked up by CAPA when an imageID or organizationID is not provided.

PR #51, opened 2019-09-26
Revert #29 on packer_builder_type

figo:
Otherwise cloud-init installation been ignored,
this is probably due to packer_builder_type value is a long string,
more than `amazon` or `vmware`

cc @vincepri @detiber


images/capi/ansible/roles/providers/tasks/main.yml
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 ---
 - include_tasks: aws.yml
-  when: packer_builder_type == 'amazon'
```
codenrhoden: If I'm understanding correctly, why not change this to:
`when: packer_build_type.startswith('amazon')` ?

vincepri:
This breaks GCP given that `provider_name` is only set for aws or vsphere

figo:
> This breaks GCP given that `provider_name` is only set for aws or vsphere

@vincepri  help me to understand where is the gcp files that does cloud-init installation, i only see aws.yml and vmware.yml, thx

vincepri:
GCP doesn't have specific files there. This line
```
provider_name: "{{ 'aws' if packer_builder_type.startswith('amazon') else 'vmware' if packer_builder_type.startswith('vmware') }}"
```
assigns null if the packer provider isn't vmware or amazon.

This makes the ansible execution fail when we do `provider_name.lower() == 'aws'`

Hope that helps :)

codenrhoden:
I believe my code suggestion/alternative will work for all providers?

images/capi/ansible/roles/providers/tasks/main.yml
figo: thanks for it, this comes to my mind when i looked at the code, although i rushed to open the PR, now it is been tested and updated.

PR #52, opened 2019-09-26
[capi] Rev versions

detiber:
- Update containerd to v1.2.9
- Update kubernetes to v1.15.4
- Fix image naming for CAPI AMIs

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -52,7 +52,7 @@
         "owners": ["099720109477"],
         "most_recent": true
       },
-      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_semver`}}-{{user `build_timestamp`}}",
+      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_deb_version`}}-{{user `build_timestamp`}}",
       "ami_groups": "{{user `ami_groups`}}",
       "ami_users": "{{user `ami_users`}}",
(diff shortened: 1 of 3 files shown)


images/capi/packer/ami/packer.json
```diff
@@ -52,7 +52,7 @@
         "owners": ["099720109477"],
         "most_recent": true
       },
-      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_semver`}}-{{user `build_timestamp`}}",
+      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_deb_version`}}-{{user `build_timestamp`}}",
```
ncdc: These changes to use kubernetes_deb_version don't look right?

detiber: This is the pattern that we use for the lookups, this used to be called `kubernetes_version` and we had a filter to switch between deb and rpm version formats.

ncdc: But this means the name will have the `-00` at the end. I don't think that's necessary, is it?

ncdc: Oh, I see the lookup code expects/requires the kube semver followed by `-??`. Yuck.

detiber: Yeah, probably something that we want to clean up down the line, but would be quite a disruptive change now.

codenrhoden:
/lgtm

I'll hold off for a bit in case there is additional feedback?

ncdc:
I'm good to go

PR #54, opened 2019-09-26
kube-deploy: Move from glog to klog

justinsb:


images/kube-deploy/imagebuilder/main.go
```diff
@@ -78,13 +78,15 @@ func loadConfig(dest interface{}, src string) error {
 }
 
 func main() {
+	klog.InitFlags(nil)
+
 	rand.Seed(time.Now().UTC().UnixNano())
 
 	flag.Set("alsologtostderr", "true")
```
vincepri: This isn't needed anymore

justinsb: Ah - that is great news, thanks!  I'll send a PR to remove it then!

PR #57, opened 2019-09-26
Only download k8s images from http when user want

figo:
When user set kubernetes_source_type != pkg,
it will download kubernetes binaries and container through
http site, it will not pull through "kubeadm config images pull"
(which is pulling from container registry) anymore.

cc @akutz @jiatongw @frapposelli

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/main.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/main.yml
@@ -31,15 +31,5 @@
     dest: /etc/kubernetes-version
     src: etc/kubernetes-version
 
-- name: Create kubeadm config file
-  template:
-    dest: /etc/kubeadm.yml
-    src: etc/kubeadm.yml
-
-- name: Kubeadm pull images
-  shell: 'kubeadm config images pull --config /etc/kubeadm.yml'
-
-- name: delete kubeadm config
-  file:
-    path: /etc/kubeadm.yml
-    state: absent
+- import_tasks: kubeadmpull.yml
+  when: kubernetes_source_type == "pkg"
--- a/images/capi/ansible/roles/kubernetes/defaults/main.yml
+++ b/images/capi/ansible/roles/kubernetes/defaults/main.yml
@@ -24,3 +24,6 @@ kubernetes_imgs:
 - kube-controller-manager.tar
 - kube-scheduler.tar
 - kube-proxy.tar
+- pause.tar
+- coredns.tar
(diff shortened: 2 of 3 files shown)


images/capi/ansible/roles/kubernetes/tasks/debian.yml
```diff
@@ -32,3 +32,16 @@
       - "kubeadm={{ kubernetes_deb_version }}"
       - "kubectl={{ kubernetes_deb_version }}"
       - "kubernetes-cni={{kubernetes_cni_deb_version }}"
+
```
detiber: I'd suggest moving these steps into a different task file and importing it from `main.yml` when `kubernetes_source_type == "pkg"` to avoid duplicating across debian and redhat task files.

figo: Very good point, thanks Jason, i will update.

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -30,16 +30,3 @@
   template:
     dest: /etc/kubernetes-version
     src: etc/kubernetes-version
-
```
codenrhoden: I'm really confused by this. It looks like we took a step that was common, and copied it w/o changes to the RedHat and Debian tasks... So the end result is the same with more code? I'm missing something...

images/capi/ansible/roles/kubernetes/tasks/debian.yml
codenrhoden: Second this. I get what's happening now. Yeah, just like there is a `url.yml` for use when downloading tarballs, I would just create a `pkg.yml` with this common code in it and include it after the OS specific tasks.

figo: diff updated, thx

images/capi/ansible/roles/kubernetes/tasks/debian.yml
```diff
@@ -32,3 +32,6 @@
...
+- import_tasks: kubeadmpull.yml
+  when: kubernetes_source_type == "pkg" and ansible_os_family == "Debian"
```
detiber: I'd move this import to `main.yml` and remove the `ansible_os_family` check.

figo: @detiber yup, good point, thanks again

PR #58, opened 2019-09-26
Fix 'make clean' failure in Makefile

jiatongw:
There is a typo in `make clean` rule: `rm -fr` should be `rm -rf`

cc @figo @akutz

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -54,9 +54,9 @@ $(OVA_BUILD_TARGETS_ESX):
 	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=shell-local packer/ova/packer.json
 .PHONY: $(OVA_BUILD_TARGETS_ESX)
 
-CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES)) $(addprefix clean-,$(AMI_BUILD_NAMES)) $(addprefix clean-,$(GCE_BUILD_NAMES))
+CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES))
 $(CLEAN_TARGETS):
-	rm -fr output/$(subst clean-,,$@)*
+	rm -fr output/$(subst clean-ova-,,$@)*
 .PHONY: $(CLEAN_TARGETS)
 
 build: $(OVA_BUILD_TARGETS) $(AMI_BUILD_TARGETS) $(GCE_BUILD_TARGETS)


images/capi/Makefile
```diff
@@ -56,7 +56,7 @@ $(OVA_BUILD_TARGETS_ESX):
 
 CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES)) $(addprefix clean-,$(AMI_BUILD_NAMES)) $(addprefix clean-,$(GCE_BUILD_NAMES))
 $(CLEAN_TARGETS):
-	rm -fr output/$(subst clean-,,$@)*
+	rm -rf output/$(subst clean-,,$@)*
```
detiber: Is there a version of `rm` that requires the arguments be provided in a specific order? I would expect both versions to be equivalent.

akutz:
/hold

This is a non issue as the flag order is unimportant. I typically do `rm -fr`, but others do their thing.

jiatongw:
OK I found the reason that why `make clean` doesn't work. Actually I was crazy that thought `rm -fr` was a typo b/c I always use `rm -rf` ...

The reason that `make clean` doesn't work is because it do `rm -fr output/ova-centos-7*` and `rm -fr output/ova-ubuntu-1804*`, but the output folder, our build artifacts are named `centos-7*` and `ubuntu-1804*`, without the `ova-` prefix. (https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ova/ova-ubuntu-1804.json#L2). So the rm command cannot find the folder. @akutz @detiber

images/capi/Makefile
```diff
@@ -54,9 +54,9 @@ $(OVA_BUILD_TARGETS_ESX):
 	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=shell-local packer/ova/packer.json
 .PHONY: $(OVA_BUILD_TARGETS_ESX)
 
-CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES)) $(addprefix clean-,$(AMI_BUILD_NAMES)) $(addprefix clean-,$(GCE_BUILD_NAMES))
+CLEAN_TARGETS := $(OVA_BUILD_NAMES) $(AMI_BUILD_NAMES) $(GCE_BUILD_NAMES)
 $(CLEAN_TARGETS):
-	rm -fr output/$(subst clean-,,$@)*
+	rm -fr output/$(subst ova-,,$@)*
 .PHONY: $(CLEAN_TARGETS)
```
jiatongw: `make clean` will be 
```
rm -fr output/centos-7*
rm -fr output/ubuntu-1804*
rm -fr output/ami-default*
rm -fr output/gce-default*
```

detiber: I'm not familiar enough with the ova tooling to know if this is correct, but I don't think there is anything that needs to be cleaned up after the AMI and GCE related targets.

codenrhoden: Hmm, I'm not a big fan of this change either... But not for the reason you would expect.  On master, if I look at all the build targets, I see this:

```
$ make
Makefile                   build                      build-esx-ova-centos-7     build-gce-default          build-ova-ubuntu-1804      clean-ami-default          clean-ova-centos-7
all                        build-ami-default          build-esx-ova-ubuntu-1804  build-ova-centos-7         clean                      clean-gce-default          clean-ova-ubuntu-1804
```

With this change, that becomes:

```
$ make
Makefile                   ami-default                build-ami-default          build-esx-ova-ubuntu-1804  build-ova-centos-7         clean                      ova-centos-7
all                        build                      build-esx-ova-centos-7     build-gce-default          build-ova-ubuntu-1804      gce-default                ova-ubuntu-1804
```

You'll noticed all the `clean-` targets have disappeared. I don't think we want that. This would make it be `make clean` or nothing else, instead of the ability to clean an individual target.

**However**, do we actually need that? I think that OVA builds may be the only builds that produce something locally. I don't have a GCE or AWS setup to test with, but when those images are built, they remain hosted on the cloud provider, right? Put another way, is there ever even a `output/{ami*,gce*}` folder created? I don't think there is.

So I'm wondering if this could instead be something like:

```
CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES))
$(CLEAN_TARGETS):
	rm -fr output/$(subst clean-ova-,,$@)*
```

detiber: For AWS we do not create anything locally. Tagging in @vincepri to speak to GCE

vincepri: Correct for GCE

jiatongw: > 
> So I'm wondering if this could instead be something like:
> 
> ```
> CLEAN_TARGETS := $(addprefix clean-,$(OVA_BUILD_NAMES))
> $(CLEAN_TARGETS):
> 	rm -fr output/$(subst clean-ova-,,$@)*
> ```

For this change `make clean` would be 
```
rm -fr output/centos-7*
rm -fr output/ubuntu-1804*
rm -fr output/clean-ami-default*
rm -fr output/clean-gce-default*
```

So the `output/clean-ami-default*` and `output/clean-gce-default*` are actually fake. We'd better remove the `ami` and `gce` targets from `make clean` if we do the suggested change above

codenrhoden: Well if you look at the `CLEAN_TARGETS` line, it only lists OVA names... So that's exactly what I was suggesting. We are on the same page.

jiatongw: Addressed

codenrhoden:
Tested that this all works as expected locally.

/lgtm
/approve
/hold cancel

PR #60, opened 2019-09-27
OVA: set VMX version to 13

codenrhoden:
Explicitly set the VMX version to 13 when building the VM, and update
the OVA XML to 13 as well. Previous behavior was building a VM with
version 9, then labeling at as 11 in the OVA.

This PR includes a second commit to remove trailing spaces in the OVA EULA. My editor does this automatically, and it's going to happen every time this file is touched. Might as well just clean it up now. =)

This PR is necessary because for CAPV, we are only supporting ESXi 6.5 and above. Also, when using running K8s with the vSphere out of tree provider and CNS storage plugin, 6.5 is required. So this lines up with what we need. Version compatibility can be found [here](https://kb.vmware.com/s/article/1003746).

/assign @akutz 
/cc @andrewsykim @frapposelli

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -32,6 +32,10 @@ import tarfile
 def main():
     parser = argparse.ArgumentParser(
         description="Builds an OVA using the artifacts from a Packer build")
+    parser.add_argument('--vmx',
+                        dest='vmx_version',
+                        default='13',
+                        help='The virtual hardware version')
     parser.add_argument(dest='build_dir',
                         nargs='?',
                         metavar='BUILD_DIR',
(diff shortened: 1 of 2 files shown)


images/capi/hack/image-build-ova.py
```diff
@@ -190,7 +190,7 @@ def stream_optimize_vmdk_files(inlist):
       <System>
         <vssd:ElementName>Virtual Hardware Family</vssd:ElementName>
         <vssd:InstanceID>0</vssd:InstanceID>
-        <vssd:VirtualSystemType>vmx-11</vssd:VirtualSystemType>
+        <vssd:VirtualSystemType>vmx-13</vssd:VirtualSystemType>
```
akutz: Hi @codenrhoden,

If you're putting the VMX version in the `packer.json` file anyway, you should probably pass it to the OVA build process.

codenrhoden:
Added the version to the OVA build script...

Verified that generated OVF has `vmx-13` in it now, and also that the `.vmx` file before the OVA is built says that as well.

Also verified that if I run the Python script with a non-default value, that the given number is used instead.

codenrhoden:
@akutz can you take another look at this one?
/hold cancel

PR #61, opened 2019-09-27
OVA: parameterize network name

codenrhoden:
A recent commit hardcoded the network name to "VM Network". This
appears to work fine when building with Fusion, but if building
Workstation and that network is not found, there is no network connected
and the VM fails to boot.

Rather than hardcoding the `vmx_data` field, we can just set `network`
instead, and go back to the default value of "". Then this can be
overridden if a user needs to use a specific network by name.

/hold
/assign @jiatongw 

Jiatong, I need your help testing this to see if it still meets your needs for building on remote ESX hosts. I found that after your patch went in, I could still build locally with Fusion, but building on Linux hosts with WorkStation failed. I tracked it down to this. This patch changes it back to the previous default behavior, but should still allow you to set what ever network name you need in `esx.json`.

### DIFF
--- a/images/capi/packer/ova/esx.json
+++ b/images/capi/packer/ova/esx.json
@@ -1,6 +1,7 @@
 {
     "disk_type_id": "thin",
     "format": "ova",
+    "network": "VM Network",
     "remote_type": "esx5",
     "remote_host": "",
     "remote_datastore": "",
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -28,6 +28,7 @@
     "kubernetes_deb_gpg_key": null,
     "kubernetes_rpm_gpg_check": null,
     "kubernetes_container_registry": null,
+    "network": "",
     "reenable_public_repos": "true",
     "remove_extra_repos": "false",
     "remote_type": "",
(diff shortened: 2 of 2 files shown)


codenrhoden:
@jiatongw we should get you added to the `kubernetes-sigs` orgs so you can be assigned stuff. :) You are already in `kubernetes` so it's really easy to do.

jiatongw:
This change not work in building on ESX: I got `Network autoconfiguration failed` when building.
I think we have find a way that add something like
```"vmx_data": "{{user `vmx_data`}}"```in the builder and pass the data(which is a map structure) in `esx.json`. 

But the tricky thing is:  it seems like packer user variables cannot tolerate object. If I set ```"vmx_data": "{{user `vmx_data`}}"```  and want to pass the variable from `esx.json`, say `"vmx_data": {"name": "vm network"}`, then it will get error:
```json: cannot unmarshal object into Go value of type string```

I also tried to set ```"vmx_data": {"name": "{{user `name`}}"}``` sth like this in the builders, and set ```"name": "vm network"``` in `esx.json`, hoping that packer could pass only the `name` variable, but it failed as the same. 

I guess I have to report this issue to hashicorp/packer

jiatongw:
please refer to #62

jiatongw:
Working fine on ESX. 👍

PR #62, opened 2019-09-30
Fix vmx_data field to make build on linux host work

jiatongw:
Thanks to @codenrhoden to find this error: A previous commit hardcoded the network name to "VM Network". This work fine when building with Fusion, but not with Workstation. 

This PR #61 doesn't work on ESX host so I made this patch.
This patch fix the building failure with workstation in linux host. I tested locally with Fusion, in linux host with Workstation and remotely with ESX host and they all worked. Could you please double check @codenrhoden ?

### DIFF
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -83,7 +83,7 @@
       "vnc_disable_password": "{{user `vnc_disable_password`}}",
       "format": "{{user `format`}}",
       "vmx_data": {
-        "ethernet0.networkName": "VM Network"
+        "ethernet0.networkname": "{{user `ethernet0.networkname`}}"
       }
     }
   ],
--- a/images/capi/packer/ova/esx.json
+++ b/images/capi/packer/ova/esx.json
@@ -8,5 +8,8 @@
     "remote_password": "",
     "skip_compaction": "true",
     "vnc_bind_address": "",
-    "vnc_disable_password": "true"
+    "vnc_disable_password": "true",
+    "vmx_data": {
+        "ethernet0.networkname": "VM Network"
+    }
 }


codenrhoden:
@jiatongw thanks for continuing to work on this! I'm a little behind on testing it, but it is on my stack. :)

images/capi/packer/ova/packer.json
```diff
@@ -83,7 +83,7 @@
       "vnc_disable_password": "{{user `vnc_disable_password`}}",
       "format": "{{user `format`}}",
       "vmx_data": {
-        "ethernet0.networkName": "VM Network"
+        "ethernet0.networkname": "{{user `ethernet0.networkname`}}"
```
codenrhoden: So this isn't working for me locally...  Well, it *is*, but that's because I don't think this fix is correct either.

When running a standard `make build-ova-centos-7`, the user variable is not defined -- it's an empty string. Thus, it works because the default works. That will work on fusion as well. But if I add in the `esx.json` file, the template fails to validate:

``
$ packer validate -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"   -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/ova-centos-7.json" -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/esx.json" packer/ova/packer.json
invalid value "/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/esx.json" for flag -var-file: Error reading variables in '/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/esx.json': json: cannot unmarshal object into Go value of type string
```

Same problem as we saw before. And I'm not surprised by that. I'm not sure how the line that I'm commenting on here would pull the value out of a dictionary named `vmx_data` in a different var file. Maybe this patch is out of date?

Anyways -- I think it gave me an idea for a compromise. It would be hybrid between my patch and yours. Why don't we define a variable named `network`, default it to an empty string, and use that here in `vmx_data`. That way unless it's defined somewhere else, it will be empty. Just like it is now. Then define it however you need in `esx.json`.

It's exactly what I did in #61, but instead of using the variable in `network`, we just set it in the `vmx_data`.

Thoughts? I can update my PR if you need an example.

Comment (jiatongw) [reply images/capi/packer/ova/packer.json]: Well, this works for me 
```
$ make build-ova-centos-7 
packer build -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/cni.json"  -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/containerd.json"   -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/ova/ova-centos-7.json" packer/ova/packer.json
centos-7 output will be in this color.

==> centos-7: Retrieving ISO
```

Comment (jiatongw) [reply images/capi/packer/ova/packer.json]: Also, do you mean 
```
"vmx_data": {
    "network": "{{ user `network`}}"
}
``` 
I don't think we can use the key "network" in vmx_data. it should be exactly the key `ethernet0.networkname` . I think this is a vmx predefined key.

some of the vmx_data here:
```ethernet0.addresstype = "generated"
ethernet0.bsdname = "en0"
ethernet0.connectiontype = "nat"
ethernet0.displayname = "Ethernet"
ethernet0.linkstatepropagation.enable = "FALSE"
ethernet0.networkname = ""
ethernet0.pcislotnumber = "33"
ethernet0.present = "TRUE"
ethernet0.virtualdev = "e1000"
ethernet0.vnet = "nat"
ethernet0.wakeonpcktrcv = "FALSE"```

Comment (codenrhoden) [reply images/capi/packer/ova/packer.json]: > Well, this works for me
> 
> ```
> $ make build-ova-centos-7 
> packer build -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/cni.json"  -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/config/containerd.json"   -var-file="/home/imgbuilder-ova-01/image-builder/images/capi/packer/ova/ova-centos-7.json" packer/ova/packer.json
> centos-7 output will be in this color.
> 
> ==> centos-7: Retrieving ISO
> ```
> 
> could you please try again?

That works for me as well. Sorry I wasn't clear, but that's what I was saying with: 
> When running a standard make build-ova-centos-7, the user variable is not defined -- it's an empty string. Thus, it works because the default works. That will work on fusion as well.

Even though I did a manual `packer validate`, I was more saying that `make build-esx-ova-centos-7` doesn't work when run from this branch:

```
$ make build-esx-ova-centos-7
packer build -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"   -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/ova-centos-7.json" -var-file="packer/ova/esx.json" -except=shell-local packer/ova/packer.json
invalid value "packer/ova/esx.json" for flag -var-file: Error reading variables in 'packer/ova/esx.json': json: cannot unmarshal object into Go value of type string
```

As for what I was suggesting instead, just take a look at #61. I updated it. :) It works for Fusion and Workstation by default, and I think it will do what you need on ESX now too.

jiatongw:
PR #61 works on ESX. Thanks @codenrhoden ! 👍 Close this.

/close

PR #63, opened 2019-09-30
Add OWNERS file to images/capi

vincepri:
Adds the Cluster API admins as approvers to the subfolder

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+aliases:
+  cluster-api-admins:
+    - justinsb
+    - detiber
+    - davidewatson
+    - vincepri
+    - ncdc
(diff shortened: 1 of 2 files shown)


vincepri:
How do you mean? I’m not sure if we can reference groups across repos

timothysc:
Add the alias to the main OWNERS_ALIAS to the root of this repo, and then in this specific OWNERS file just reference this alias.

vincepri:
Sure I can try that

vincepri:
@timothysc Done, ptal

codenrhoden:
Could be a follow-on to this, but I'm wondering if @figo and myself should be moved from the top level `OWNERS` to this new one at `images/capi` instead. I know for myself, that's the only area I have knowledge and interest in.

timothysc:
@codenrhoden knock on PR ;-) 
/lgtm

PR #64, opened 2019-09-30
chore: stop and disable rpcbind{service,socket}

dennisme:
We noticed this was running during an internal security scan - 111 Portmapper. 

We were using `capa-ami-ubuntu-18.04-*` AMIs.

### DIFF
--- a/images/capi/ansible/roles/common/tasks/debian.yml
+++ b/images/capi/ansible/roles/common/tasks/debian.yml
@@ -63,3 +63,15 @@
     name: "{{ common_pinned_debs }}"
     state: present
     force: yes
+
+- name: Stop and disable rpcbind.service
+  service:
+    name: rpcbind
+    state: stopped
+    enabled: no
+
+- name: Stop and disable rpcbind.socket
+  service:
+    name: rpcbind.socket
+    state: stopped
+    enabled: no


images/capi/ansible/roles/common/tasks/debian.yml
```diff
@@ -63,3 +63,15 @@
     name: "{{ common_pinned_debs }}"
     state: present
     force: yes
+
+- name: Stop and disable rpcbind.service
```
dennisme: If we think there will be more I can happily turn this into a `with_items` dealio. Just let me know

detiber: Only drawback I see here is that this could potentially break NFS-based persistent volumes, and even CSI-based NFS volumes (https://github.com/kubernetes-csi/csi-driver-nfs/pull/14#issuecomment-502545228)

dennisme: Oh good point. Is it a fair assumption that if you wanted use CSI-based NFS volumes securing rpcbind (ie. modifying the allowed hosts in `/etc/hosts.allow`) would be a better option than leaving it on for everyone?

detiber: I'm generally good with the idea of disabling the service, especially since we give users a way to run additional commands through the KubeadmConfig, but would like to get some feedback/thoughts from others.

/assign @codenrhoden @akutz @ncdc @timothysc

ncdc: @detiber would someone use the additional commands to enable rpcbind since this change disables it by default?

detiber: @ncdc I think that is a reasonable ask for someone that knew they wanted to use NFS for volume

akutz: My primary concern with disabling this is the way machine images are currently published. This is something @detiber and I've discussed before, but we currently version machine images based on the Kubernetes version -- not the version of the image-stamper. 

In CAPV's case, if we publish new OVAs, we're essentially overwriting the old ones. Before we start making substantial, potentially breaking changes, I think we need to consider how to do versioned releases of the machine images.

codenrhoden: > My primary concern with disabling this is the way machine images are currently published. This is something @detiber and I've discussed before, but we currently version machine images based on the Kubernetes version -- not the version of the image-stamper.
> 
> In CAPV's case, if we publish new OVAs, we're essentially overwriting the old ones. Before we start making substantial, potentially breaking changes, I think we need to consider how to do versioned releases of the machine images.
> 
> @codenrhoden, thoughts?

I was just thinking about this yesterday, as I pushed new OVAs for the latest K8s patch releases. As you said, the images are versioned by K8s version, not version of the image builder, and there were some significant enough changes in the OVAs I felt uncomfortable overwriting the older OVAs. I only pushed new ones that didn't have to overwrite old ones.

My thoughts? Pretty soon we will want to only publish images that are built from a tagged version of the image stamper. I would happy tag something every time we want to build new images. But... There are multiple projects in this repo (e.g. konfigadm, kubedeploy), which is something I loathe but that's where we are. So some kind of tagging/branching strategy would have to come into play there.

detiber: One thing we could do is to branch based on a "supported version" of CAPI (at least for the capi builders), but that brings up interesting challenges around how to do an upgrade between CAPI (and provider) versions... Ideally we'd have some type of support statement for the Image builder that would say that it supports (n-y) versions of CAPI and supported providers...

dennisme: Should this be addressed in this PR or a separate one? Does changing the versioning also change how providers lookup the image?

detiber: I believe handling tagging, release, and support strategies for the image-builder and consumption of image-builder based images by cap* providers would be a separate issue that can be handled later. @codenrhoden @ncdc @akutz do you agree?

akutz: Hi @detiber,

Unfortunately this can't be handled later. @andrewsykim has hit an issue that impacts older OVAs, and we need to either re-upload those assets or agree on some versioning scheme. @andrewsykim, can you please create the GH issue to track this?

andrewsykim: Yup, will open a separate issue on this shortly

codenrhoden:
This topic discussed in today's office hours: the ability to disable this service makes sense, but since NFS is in the image by default, the worry would be about breaking NFS.

Should NFS remain in by default? (should open an issue on this)

If so, the thinking is that this can be handled as a user post-hook once the CLI is in place. Will wait to address until CLI is more mature.

ncdc:
This has been open for a while - have we made any progress on resolving the outstanding questions/concerns?

codenrhoden:
/close
This has been handled by disabling NFS by default. If users need NFS and re-enable it, they'll probably need rpcbind running anyways. :)

PR #65, opened 2019-10-03
enable building CAPI packer images using AWS_PROFILE env variable

aaroniscode:
This would allow building Cluster API images using AWS named profiles. Just a lot more convenient for those that use profiles.

/assign @detiber


images/capi/packer/ami/packer.json
```diff
@@ -60,6 +60,7 @@
       "encrypt_boot": "{{user `encrypted`}}",
       "ami_regions": "{{user `ami_regions`}}",
       "access_key": "{{user `aws_access_key`}}",
+      "profile": "{{ user `aws_profile`}}",
```
detiber: Should this be defined in the variables section above as well?

aaroniscode: I think you are right @detiber !  It was working for me without it, but I think this is the proper way. I updated the PR

aaroniscode: I think you are right @detiber ! It was working for me without it, but I think this is the proper way. I updated the PR

images/capi/packer/ami/packer.json
```diff
@@ -2,6 +2,7 @@
   "variables": {
     "aws_access_key": "",
     "aws_secret_key": "",
+    "aws_profile": "{{env `AWS_PROFILE`}}",
```
detiber: This should probably default to `""` like `aws_access_key` and `aws_secret_key`. I believe it will fallback to the env if not specified (at least the others do).

aaroniscode: updated!  I really don't know what the difference is to the code as I think it falls back to the env variable as you mentioned. But either way, this is more consistent. Thanks again!

PR #67, opened 2019-10-04
OVA: unify definition of output directory

codenrhoden:
The output directory used by the vmware-iso builder was used in 4
different places. If you changed it one place, you needed to change it
everywhere to have the post processors work. Instead, define it one
single place to avoid mistakes.

This is fixing an annoyance of mine. I've been building lots of different images lately, and often change the output directory so I can a few different versions of things. But if I didn't make the change in four different places, the post-processors break.

/assign @figo


detiber:
lgtm

PR #73, opened 2019-10-10
photon support

ppadmavilasom:


images/capi/ansible/roles/kubernetes/templates/etc/yum.repos.d/kubernetes.repo
```diff
@@ -0,0 +1,6 @@
+[kubernetes]
+name=kubernetes
+description=the kubernetes yum repo
+baseurl={{ kubernetes_rpm_repo }}
+gpgcheck=0
```
detiber: Should this use the 
```suggestion
gpgcheck={{ kubernetes_rpm_gpg_check }}
```

detiber: This should also likely add:
```
gpgkey: "{{ kubernetes_rpm_gpg_key }}"
```

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -36,6 +36,18 @@
...
+- name: Install cloud-init packages
+  command: tdnf install {{ packages }} -y
+  vars:
+    packages: "cloud-init cloud-utils python3-netifaces"
+  when: ansible_os_family == "VMware Photon OS"
+
+- name: copy cloud-init 19.1-2 rpm
+  command: curl -LO https://vmware.bintray.com/photon_dev_x86_64/noarch/cloud-init-19.1-2.ph3.noarch.rpm
+
+- name: Update cloud-init 18.4
```
detiber: This comment says 18.4
```suggestion
- name: Update cloud-init to 19.1-2
```

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -36,6 +36,18 @@
...
+  command: rpm -Uvh ./cloud-init-19.1-2.ph3.noarch.rpm
```
detiber: These commands can be wrapped in a block, to ensure they are all run with the proper when clause. Currently the cloud-init download and install would attempt to run regardless of OS, which isn't quite what we want here.

```go
- name: Install cloud-init and tools for VMware Photon OS
  block:
    - name: Install cloud-init packages
      command: tdnf install {{ packages }} -y
      vars:
        packages: "cloud-init cloud-utils python3-netifaces"
    - name: copy cloud-init 19.1-2 rpm
      command: curl -LO https://vmware.bintray.com/photon_dev_x86_64/noarch/cloud-init-19.1-2.ph3.noarch.rpm
    - name: Update cloud-init to 19.1-2
      command: rpm -Uvh ./cloud-init-19.1-2.ph3.noarch.rpm
 ```

File: images/capi/packer/goss/photon-vars.yaml [hunk 4]
```diff
@@ -0,0 +1 @@
+cloud_init_version: "18.3-2.ph3"
```
detiber: Shouldn't this match the version being installed?

I don't think goss is actually running for anything other than aws builds currently.

images/capi/ansible/roles/common/tasks/photon.yml
```diff
@@ -0,0 +1,31 @@
...
+    mode: 0755
+    owner: photon
+    group: photon
+
+- name: add bash_profile
+  template:
+    dest: /home/photon/.bash_profile
+    src: photon_bash_profile
+    owner: photon
+    group: photon
```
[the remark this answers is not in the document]
ppadmavilasom: it has dependencies with the photon user's home directory getting created here. its local enough that it is applicable to the photon specific use case.

images/capi/ansible/roles/providers/tasks/vmware.yml
ppadmavilasom: done

images/capi/ansible/roles/providers/tasks/vmware.yml
ppadmavilasom: done

images/capi/packer/goss/photon-vars.yaml
ppadmavilasom: removed unused

images/capi/ansible/roles/kubernetes/tasks/photon.yml
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: Add the kubernetes repo
+  template:
+    dest: /etc/yum.repos.d/kubernetes.repo
+    src: etc/yum.repos.d/kubernetes.repo
+
+- name: Install conntrack-tools
+  command: tdnf install conntrack-tools -y
```
detiber: Should this be added to `common_photon_rpms` isntead?

ppadmavilasom: done

images/capi/ansible/roles/kubernetes/templates/etc/yum.repos.d/kubernetes.repo
ppadmavilasom: done

images/capi/ansible/roles/common/tasks/main.yml
```diff
@@ -42,6 +45,7 @@
     state: present
     sysctl_set: yes
     reload: yes
+    sysctl_file: /etc/sysctl.d/99-sysctl.conf
```
figo: could you explain a little bit why this is required by photon? not by other OS

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -43,6 +46,7 @@
 - name: Set hostname
   hostname:
     name: localhost.local
+  when: ansible_os_family != "VMware Photon OS"
```
figo: could you comment on how photon is different from others, like no need to set hostname to localhost.local?

ppadmavilasom: errors with photon-3: TASK [sysprep : set hostname] **************************************************
    photon-3: fatal: [default]: FAILED! => {"changed": false, "msg": "hostname module cannot be used on platform Linux (Photon)"}
this is set for photon using hostnamectl in photon.yml - https://github.com/kubernetes-sigs/image-builder/pull/73/files/af84674ae5ae1a2514eb4d8b6dd1b54a1959e9b4#diff-4a7c4cb3c8e752f579b0e5d675df9e0fR21

images/capi/ansible/roles/common/tasks/main.yml
ppadmavilasom: this is because there was security hardening rules that negated the ones we set (like net.ipv4.forward=1) in lower files like /etc/sysctl.d/50-security-hardening.conf however, i see your point. i will apply these for photon only.

images/capi/ansible/roles/sysprep/tasks/main.yml
detiber: Between this error and the one related to package management, it seems like we could gain a huge benefit to adding better support for Photon to upstream Ansible.

ppadmavilasom: @detiber better upstream support for photon is ongoing. package management support is ongoing as well.

images/capi/ansible/roles/common/tasks/main.yml
ppadmavilasom: fixed using photon specific changes. photon uses systemd-sysctl and will only load files under /etc/sysctl.d by default.

figo:
@ppadmavilasom is testing ubuntu with CAPV by applying this change (since this change do change some common paths), once test passed, the PR should be ready for merge, thx @ppadmavilasom

figo:
The test passed with CAPV, i am good with the change.

/lgtm
/approve

PR #80, opened 2019-10-18
Add gsutil to GCE based images

dims:
Adding `gsutil` will help help us with the CI testing scenarios where we have to pull debs and container images from GCS buckets

Once this is in, we can remove the code in CAPG that downloads this utility.

Please see how we download/use it in CAPG:
https://cs.k8s.io/?q=gsutil&i=nope&files=.*%5C.yaml&repos=kubernetes-sigs/cluster-api-provider-gcp

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/googlecompute.yml
+++ b/images/capi/ansible/roles/providers/tasks/googlecompute.yml
@@ -1,0 +1,39 @@
+# Copyright 2019 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


images/capi/packer/gce/packer.json
```diff
@@ -58,7 +58,10 @@
       "type": "shell",
       "inline": [
         "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done",
-        "sudo apt-get -qq update && sudo DEBIAN_FRONTEND=noninteractive apt-get -qqy install python python-pip"
+        "sudo apt-get -qq update && sudo DEBIAN_FRONTEND=noninteractive apt-get -qqy install python python-pip",
+        "curl -o -s installer https://sdk.cloud.google.com",
+        "sudo bash installer --disable-prompts --install-dir=/ > /dev/null 2>&1 && rm installer",
+        "sudo ln -s /google-cloud-sdk/bin/* /bin/"
```
detiber: Would it be better to do this in the ansible bits instead? This provisioner was meant to only be a hack to ensure python was present for use by ansible.

dims: Done!

PR #84, opened 2019-10-24
Lock default user before shutdown OVA VM

figo:
Although cloud-init will lock default user, but cloud-init configure
can be override by user-data later.

This patch is to lock default user within OS, not through cloud-init.

cc @akutz @andrewsykim @jiatongw @detiber @frapposelli

### DIFF
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -74,7 +74,7 @@
         "http://{{ .HTTPIP }}:{{ .HTTPPort }}",
         "{{user `boot_command_suffix`}}"
       ],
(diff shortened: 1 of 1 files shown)


detiber:
/lgtm
/hold
Holding for further review

akutz:
/lgtm

@figo, as long as you've tested this, I'm fine with you removing the hold. It's exactly what we discussed.

akutz:
Hi @figo,

Please add `Fixes #85` to the issue description. Thanks!

figo:
Tested on ubuntu, centos, both works as expected now: only user capv can be login with ssh key.
Plan to test photon.

figo:
photon been tested

PR #87, opened 2019-10-29
🏃Bump Kubernetes version to 16.2 in packer config

nader-ziada:
**What this PR does / why we need it**:
Change the version Kubernetes to `16.1` in the packer config used when building `capi` images to be compatible with latest images needed by providers

### DIFF
--- a/images/capi/packer/config/kubernetes.json
+++ b/images/capi/packer/config/kubernetes.json
@@ -1,8 +1,8 @@
 {
-  "kubernetes_series": "v1.15",
-  "kubernetes_semver": "v1.15.4",
-  "kubernetes_rpm_version": "1.15.4-0",
-  "kubernetes_deb_version": "1.15.4-00",
+  "kubernetes_series": "v1.16",
+  "kubernetes_semver": "v1.16.2",
+  "kubernetes_rpm_version": "1.16.2-0",
+  "kubernetes_deb_version": "1.16.2-00",
   "kubernetes_source_type": "pkg",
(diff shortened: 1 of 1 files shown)


images/capi/packer/config/kubernetes.json
```diff
@@ -1,8 +1,8 @@
 {
-  "kubernetes_series": "v1.15",
-  "kubernetes_semver": "v1.15.4",
-  "kubernetes_rpm_version": "1.15.4-0",
-  "kubernetes_deb_version": "1.15.4-00",
+  "kubernetes_series": "v1.16",
+  "kubernetes_semver": "v1.16.1",
```
detiber: Any particular reason for v1.16.1 instead of v1.16.2?

nader-ziada: The two providers I was trying to run are setting the version to `v1.16.1`
 - [gcp](
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/77c358d82a5029265edd3a2ccfbe927b36c29719/examples/generate.sh#L30) 
 - [azure](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/examples/generate.sh#L32)

Should the providers be bumped to `v1.16.2`?

detiber: Probably, I'd prefer we avoid updating to a version with a known security issue if at all possible (CVE-2019-11253)

nader-ziada: ok, I'll create PRs for the providers, then come back to update this. Thanks

PR #97, opened 2019-11-11
enable ipv6 forwarding on instances

aojea:
We need to enable ipv6 forwarding on instance in order to support ipv6 k8s clusters.

### DIFF
--- a/images/capi/ansible/roles/common/tasks/main.yml
+++ b/images/capi/ansible/roles/common/tasks/main.yml
@@ -56,6 +56,24 @@
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
 
+- name: Ensure net.ipv6.conf.all.forwarding sysctl is present
+  sysctl:
+    name: net.ipv6.conf.all.forwarding
+    value: "1"
+    state: present
+    sysctl_set: yes
+    reload: yes
+    sysctl_file: "{{ sysctl_conf_file }}"
+
+- name: Ensure IPv6 is enable
+  sysctl:
+    name: net.ipv6.conf.all.disable_ipv6
+    value: "0"
+    state: present
+    sysctl_set: yes
+    reload: yes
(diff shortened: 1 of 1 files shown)


aojea:
I want to check first if something else is missing
/hold

images/capi/ansible/roles/common/tasks/main.yml
```diff
@@ -56,6 +56,15 @@
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
 
+- name: Ensure net.ipv6.conf.all.forwarding sysctl is present
+  sysctl:
+    name: net.ipv6.conf.all.forwarding
```
dims: Do we need the other `sysctl net.ipv6.conf.all.disable_ipv6=0`?

looking at https://cs.k8s.io/?q=net.ipv6.conf.all.forwarding&i=nope&files=&repos=

aojea: Iirc that's enable by default, thing is that docker disable it thus we had to enable in container environments.
Will create a cluster.manually.with the image in aws so we can get a better understanding

dims:
@aojea how's this going? need some help?

aojea:
> @aojea how's this going? need some help?

I'm able to deploy the network infra with my PR to the cluster-api-provider-aws, but I'm trying to fix the panics with the conversions. 

Although the subnets and VPC are created, I can't progress because I can't use the cloudformation template in my account :/ , need to check if I can get some way to overcome that.

aojea:
/hold cancel
 I've documented how to deploy k8s ipv6 clusters in AWS here (still missing the aws-cloud-provider integration) https://github.com/aojea/k8s-aws-ipv6

we need to enable this kernel parameters to be able to use these images
@justinsb @dims @detiber @andrewsykim

dims:
cc @jayunit100 @sedefsavas

Issue #93, opened 2019-11-08
Create gitbook and publish to sigs.k8s.io


codenrhoden:
As the image-builder is maturing, we need to have a coherent set of docs. Since this is a k8s-sigs sponsored project, we already can have  a doc page at image-builder.sigs.k8s.io. If we use gitbook and netlify, we can publish the gitbook to that URL automatically, as are done in other projects.

This is issue tracks setting up a `/doc` folder and the initial scaffolding to publish said docs.

codenrhoden:
This is in place: https://image-builder.sigs.k8s.io

/close

PR #99, opened 2019-11-13
Create gitbook docs

codenrhoden:
This patch relocates some existing READMEs into a central location,
docs/book, that can then be used with gitbook to build a static docs
site. This is the beginning of more structured documentation.

Part of: #93 

The longer term goal here is create an image-builder.sigs.k8s.io site for documentation, just like what is done at https://cloud-provider-vsphere.sigs.k8s.io/.

/assign @figo @moshloop


docs/book/GLOSSARY.md
```diff
@@ -0,0 +1,45 @@
...
+
+(source: https://github.com/kubernetes-sigs/cluster-api)
+
+## CAPV
+
+## CAPZ
+
+## ESXi
+
+ESXi is the hypervisor, or "host" OS that is used to run VMs on. ESXi provides strong separation between VMs and itself, providing strong security boundaries between the guest and host operating systems. ESXi can be used as a standalone entity, without vCenter but this is extremely uncommon and feature limited as without a higher level manager (vCenter). ESXi cannot provide its most valuable features, like High Availability, vMotion, workload balancing and vSAN (a software defined storage stack).
```
figo: ESXi is the hypervisor => ```ESXi (formerly ESX) is an enterprise-class, type-1 hypervisor developed by VMware```

docs/book/SUMMARY.md
```diff
@@ -0,0 +1,8 @@
+# Summary
+
+* Providers
+  * [CAPI](providers/capi.md)
+    * [Amazon](providers/amazon.md)
```
figo: Amazon => AWS 
amazon.md => aws.md

docs/book/SUMMARY.md
```diff
@@ -0,0 +1,8 @@
...
+    * [Azure](providers/azure.md)
+    * [VMware](providers/vsphere.md)
```
figo: VMware => vSphere

docs/book/providers/capi.md
```diff
@@ -0,0 +1,45 @@
+# CAPI Images
+
+The Image Builder can be used to build images intended for use with CAPI providers. Each provider has its own format of images that it can work with. For example, AWS instances use AMIs, and vSphere uses OVAs.
+
+## Providers
+
+* [Amazon](amazon.md)  
+* [Azure](azure.md)
+* Google *TODO*
+* [VMware](vsphere.md)
```
figo: need be consistent on using either technology name or company name, i prefer using tech name.

docs/book/GLOSSARY.md
codenrhoden: I took this straight from the Glossary in the CPI repo. I'll make the change, but we should change it there too, then.

docs/book/providers/capi.md
codenrhoden: I guess I wasn't consistent because of saying Azure instead of Microsoft. I'd like to see if others have a preference. Right now it's just a difference of what we each like. :)

codenrhoden: On second thought, I like your suggestions more.  Changed. :)

codenrhoden:
Thanks for looking @figo! I made the requested changes, and I filled out the missing parts in the glossary since I was coming back to this anyways.

figo:
/lgtm
/hold
for others to review, feel free to cancel hold anytime though.

moshloop:
Can we switch this to mdBook? The OSS gitbook is EOL
https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/install-and-build.sh

codenrhoden:
> Can we switch this to mdBook? The OSS gitbook is EOL
> https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/install-and-build.sh

Ooh, I'll take a look! I was just running with what was in the vSphere CPI repo. If gitbook is EOL, we don't want that. Thanks for pointing that out!

docs/book/providers/vsphere.md
```diff
@@ -19,9 +19,6 @@ The `vmware-iso` builder supports building against a remote VMware ESX server, b
```
detiber: Should these docs live in a subdirectory for CAPI providers rather than directly under `providers`?

codenrhoden: yeah, I was wondering this too... I don't think i"m quite clear on whether, long term, all images from this repo are meant for consumption by CAPI, or if they can be used for other things. And if they are meant for other things too, does that mean that the CAPI images are *specifically* for CAPI?

Might be a conversation for another day. Probably easier to just just them under `CAPI` for now...

codenrhoden:
Okay, I refactored this to use mdbook instead of gitbook. The main thing I notice that mdbook doesn't do is the really cool glossary integration...

I kept this simple to start with -- so no additional CSS, and none of the preprocessors that are present in the CAPI repo. I'm not opposed to them, I just couldn't find any docs on what they do so I'd rather add them later.

PR #105, opened 2019-12-08
add netlify config

moshloop:
/assign @detiber

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,24 @@
+# Netlify build instructions
+[build]
+    command = "./docs/book/build.sh"
+    publish = "docs/book/book"
+
+# Standard Netlify redirects
+[[redirects]]
+    from = "https://kubernetes-sigs-image-builder.netlify.com//*"
+    to = "https://image-builder.sigs.k8s.io/:splat"
+    status = 301
+    force = true
+
+# HTTP-to-HTTPS rules
+[[redirects]]
+    from = "http://image-builder.sigs.k8s.io/*"
+    to = "https://image-builder.sigs.k8s.io/:splat"
+    status = 301
+    force = true
+
+[[redirects]]
+    from = "http://kubernetes-sigs-image-builder.netlify.com/*"
+    to = "http://image-builder.sigs.k8s.io/:splat"
+    status = 301
+    force = true
--- a/docs/book/build.sh
+++ b/docs/book/build.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


netlify.toml
```diff
@@ -0,0 +1,24 @@
+# Netlify build instructions
+[build]
+    command = "./docs/book/build.sh"
+    publish = "docs/book/book"
+
+# Standard Netlify redirects
+[[redirects]]
+    from = "https://master--kubernetes-sigs-image-builder.netlify.com/*"
+    to = "https://master.image-builder.sigs.k8s.io/:splat"
```
detiber: Should this reference just `image-builder.sigs.k8s.io` and `kubernetes-sigs-image-builder.netlify.com` for now? At least until we have another branch to publish from?

netlify.toml
```diff
@@ -0,0 +1,24 @@
...
+# HTTP-to-HTTPS rules
+[[redirects]]
+    from = "http://master.image-builder.sigs.k8s.io/*"
+    to = "https://master.image-builder.sigs.k8s.io/:splat"
+    status = 301
+    force = true
+
+[[redirects]]
+    from = "http://master--kubernetes-sigs-image-builder.netlify.com/*"
+    to = "http://master.image-builder.sigs.k8s.io/:splat"
```
detiber: Should this reference just `image-builder.sigs.k8s.io` and `kubernetes-sigs-image-builder.netlify.com` for now? At least until we have another branch to publish from?

netlify.toml
```diff
@@ -0,0 +1,24 @@
...
+[[redirects]]
+    from = "https://master--kubernetes-sigs-image-builder.netlify.com/*"
+    to = "https://master.image-builder.sigs.k8s.io/:splat"
+    status = 301
+    force = true
+
+# HTTP-to-HTTPS rules
+[[redirects]]
+    from = "http://master.image-builder.sigs.k8s.io/*"
+    to = "https://master.image-builder.sigs.k8s.io/:splat"
```
detiber: Should this reference just `image-builder.sigs.k8s.io` for now? At least until we have another branch to publish from?

moshloop:
I think we want to release a v1 pretty soon and make sure the docs are stable because when v2 starts coming out it will be radically different.

detiber:
> I think we want to release a v1 pretty soon and make sure the docs are stable because when v2 starts coming out it will be radically different.

Completely agree, but without another branch to point to as the "default" branch, `master` is the current default branch, so the referenced URLs probably will not work.

moshloop:
@detiber done

codenrhoden:
There was already a script for building the books at `hack/build-mdbook.sh`, and it downloaded a newer version of mdbook (`0.3.5` instead of `0.3.1`, though I see now that it echo's the wrong version). Apologies that I didn't get to this earlier. I'd like to see them de-dup'd.

PR #115, opened 2020-01-06
update Photon OVA from 3.0 GA to 3.0 Rev2

MnrGreg:
- Updates ISO url to Photon OS 3.0 Rev2 

- Adds package `openssl-c_rehash` which provides `rehash_ca_certificates.sh` to regenerate the system certificate trust store (equivalent of `update-ca-trust`).

- Removes the default `/etc/cloud/cloud.cfg.d/99-disable-networking-config.cfg` file which disables cloud-init networking. Note DHCP configuration are not affected as the default OS `99-dhcp-en.network` still remains, however static IP configs from cloud-init can not be instantiated.

- Remove supurfluis packages from KS which are installed with Rev2 by default: shadow, etc

- Masks `minimal` install of docker


MnrGreg:
/assign @ncdc

++ @akutz

images/capi/packer/ova/linux/photon/http/3/ks.json
```diff
@@ -6,15 +6,15 @@
             "text": "photon"
         },
     "disk": "/dev/sda",
-    "type": "minimal",
-    "additional_packages":[
-      "minimal",
-      "shadow",
-      "ca-certificates",
+    "bootmode": "dualboot",
```
akutz: This is the default anyway.

akutz: We probably want to add:

```
"install_linux_esx": true,
```

akutz:
Hi @MnrGreg,

Do you mind also fixing the host name to be `localhost` in the kickstart file to prevent cloud-init from not setting that correctly? If not `localhost` then it's not always handled correctly.

ncdc:
Going to let Andrew own this one - he's much better suited.
/unassign
/assign @akutz 
/uncc

images/capi/packer/ova/linux/photon/http/3/ks.json
MnrGreg: - Removed bootmode.
- Added ansible sysprep photon root lock (good catch!)

MnrGreg: My understanding is that `"install_linux_esx": true,` overrides `packages: [linux]` with `packages: [linux-esx]` as seen here:

https://github.com/kubernetes-sigs/image-builder/blob/748db89e6c30a991c5ae4049ef10b029b2cc6db1/images/capi/packer/ova/linux/photon/http/3/ks.json#L9-L13

As the previous Photon OS OVA wasn't using the `linux-esx` modules specifically, I thought to keep to the generic `linux` package which includes a number of other kernel drivers. I'm actually using the Packers Qemu Builder and which fails to boot, post installing `linux-esx`. So I can change it to `linux-esx` but wouldn't be able to test.

Is the intention for the OVA `capi/packer/images` folder to be specific to ESX?

akutz: That makes sense. I didn't realize that. Sgtm.

ncdc:
/approve

@akutz you want to unhold?

PR #117, opened 2020-01-09
Update to containerd 1.3.2

dims:
v1.3.2 fixes some nasty issues including a race condition, please see:

https://github.com/containerd/containerd/releases/tag/v1.3.2

### DIFF
--- a/images/capi/packer/config/containerd.json
+++ b/images/capi/packer/config/containerd.json
@@ -1,4 +1,4 @@
 {
-    "containerd_version": "1.3.0",
-    "containerd_sha256": "47653ab55b58668ce93704e47b727b41f57d296e4048d6860daf55d6b7c2bf18"
+    "containerd_version": "1.3.2",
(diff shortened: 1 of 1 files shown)


images/capi/packer/config/containerd.json
```diff
@@ -1,4 +1,4 @@
 {
-    "containerd_version": "1.3.0",
+    "containerd_version": "1.3.2",
     "containerd_sha256": "47653ab55b58668ce93704e47b727b41f57d296e4048d6860daf55d6b7c2bf18"
```
ncdc: Shouldn't the sha change?

dims: whoops! done.

ncdc:
LGTM. I'd squash. I'm guessing we should build some new images with the new containerd version and test them out?

dims:
Done. sorry was working on something and did not want to forget this so had just created this from github UI.

PR #119, opened 2020-01-12
Photon image changes: minimal ISO, disable and mask Docker

akutz:
This PR includes the following changes for the Photon3 image:

* Commit 1
  * The minimal version of the Rev2 ISO is used instead of the full DVD as the latter is 5GiB in size with the minimal ISO around 300MiB
  * The minimal ISO requires installing packages via `tdnf` in a post-install script in the kickstart file instead of the `additional_packages` declaration as the latter does not work correctly with a Photon 3 minimal ISO because only the ISO is consulted for the packages, not the internet.
* Commit 2
  * Disables and masks the Docker systemd unit file in the Photon3 kickstart. ContainerD is the default CRI for the machine images, and so having Docker enabled in Photon could create conflicts or false-positives where it seems like the bootstrap process works, but in fact it's using Docker instead of the intended default, ContainerD.

/assign @figo 
/kind bugfix
/kind feature

cc @ppadmavilasom


ppadmavilasom:
@akutz this cannot be offline because of the installs in post reaching out to external photon repos. is this okay to do? I saw a shift towards making this part offline elsewhere in this codebase

akutz:
> @akutz this cannot be offline because of the installs in post reaching out to external photon repos. is this okay to do? I saw a shift towards making this part offline elsewhere in this codebase

Hi @ppadmavilasom,

Oh, this is definitely an issue, but until the vSphere builder plug-in from JetBrains merges into upstream Packer, it's not possible to build on ESX from the Photon 3 DVD given it's size. @figo said it fails due to the size of the ISO and temp space.

In order to continue to enable remote hypervisor support for now, I opted for the minimal ISO with network required during kickstart. Once the new Packer builder is available, perhaps things can move back to the DVD ISO.

ppadmavilasom:
i am following up on update to Photon minimal for this purpose
/lgtm

akutz:
Hi @ppadmavilasom,

Oh sure, and thank you for that. For now though we opt'd for this strategy to allow a minimal ISO, even if it requires the network. We'll definitely be updating the build once there's a new ISO to use with the fixes you're graciously driving on our behalf! Thank you again!

PR #123, opened 2020-01-13
[cluster-api Azure] Add pipeline and script to create new marketplace sku

CecileRobertMichon:
Further automate publishing new images by adding a release pipeline to create new disk versions and a new pipeline for creating SKUs. Using https://github.com/devigned/pub to script Azure Cloud Partner Portal operations.

### DIFF
--- a/images/capi/packer/azure/scripts/new-sku.sh
+++ b/images/capi/packer/azure/scripts/new-sku.sh
@@ -1,0 +1,56 @@
+#!/bin/bash -e
+
+required_env_vars=(
+    "K8S_VERSION"
+    "SKU_TEMPLATE_FILE"
+    "AZURE_TENANT_ID"
+    "AZURE_CLIENT_ID"
+    "AZURE_CLIENT_SECRET"
+    "PUBLISHER"
+    "OFFER"
+)
+
+for v in "${required_env_vars[@]}"
+do
+    if [ -z "${!v}" ]; then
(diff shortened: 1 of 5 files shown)


images/capi/packer/azure/scripts/new-sku.sh
```diff
@@ -0,0 +1,56 @@
...
+patch=${ADDR[2]}
+
+sku_id="k8s-${major}dot${minor}dot${patch}-ubuntu-1804"
+
+< $SKU_TEMPLATE_FILE sed s/{{ID}}/"$sku_id"/ | sed s/{{K8S_VERSION}}/"$K8S_VERSION/" > sku.json
+cat sku.json
+
+echo
+echo "Getting pub..."
+(set -x ; curl -fsSL https://github.com/devigned/pub/releases/download/v0.2.0/pub_v0.2.0_linux_amd64.tar.gz -o pub; tar -xzf pub)
```
marosset: If we have to update the version of pub again we should consider moving the version to a pipeline variable.

CecileRobertMichon: I think it makes sense to keep it in source control and update the version with a PR

marosset: Yea, i guess it won't change that often.

devigned:
Yay, no more UI! lgtm

Issue #39, opened 2019-09-17
CAPI OVA: support starting from existing VMX


codenrhoden:
The existing build procedure for the OVA image always starts with an ISO. Some consumers of the tool may want to start with an existing image that just needs the K8s components added to it.

This would potentially allow a user to use the `vmware-iso` builder to create an OS image from scratch, which could later be used by the `vmware-vmx` builder for additional customization.

To facilitate this, the OVA build could be broken into two stages. One to generate the base/OS image, and the second to apply the K8s artifacts.

PR #125, opened 2020-01-16
Minimize packages installed during Photon kickstart

codenrhoden:
This PR has two commits.

1.  Enable the Photon RPMs to be in a list, just like the other distros
2. Minimize the packages that are installed during the kickstart, just like the other distros.

/assign @akutz

### DIFF
--- a/images/capi/packer/ova/linux/photon/http/3/ks.json
+++ b/images/capi/packer/ova/linux/photon/http/3/ks.json
@@ -11,29 +11,12 @@
         "initramfs"
     ],
     "additional_packages": [
-        "conntrack-tools",
-        "ebtables",
-        "jq",
-        "libnetfilter_conntrack",
-        "libnetfilter_cthelper",
-        "libnetfilter_cttimeout",
-        "libnetfilter_queue",
-        "nfs-utils",
-        "ntp",
-        "openssh-clients",
         "openssh-server",
-        "python-netifaces",
-        "python-requests",
-        "python-pip",
-        "sed",
-        "socat",
-        "sudo",
-        "tar",
-        "unzip"
+        "sudo"
     ],
     "postinstall": [
         "#!/bin/sh",
-        "tdnf install -y conntrack-tools ebtables jq libnetfilter_conntrack libnetfilter_cthelper libnetfilter_cttimeout libnetfilter_queue nfs-utils ntp openssh-clients openssh-server python-netifaces python-requests python-pip sed socat sudo tar unzip",
+        "tdnf install -y openssh-server sudo",
         "useradd -U -d /home/photon -m --groups wheel photon && echo 'photon:photon' | chpasswd",
         "echo 'photon ALL=(ALL) NOPASSWD: ALL' >/etc/sudoers.d/photon",
         "chmod 440 /etc/sudoers.d/photon",
--- a/images/capi/ansible/roles/common/defaults/main.yml
+++ b/images/capi/ansible/roles/common/defaults/main.yml
@@ -25,7 +25,6 @@ common_rpms:
 - python-requests
 - socat
 - yum-utils
-common_extra_rpms: []
 common_debs:
 - openssh-client
 - openssh-server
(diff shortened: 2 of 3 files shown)


images/capi/packer/ova/linux/photon/http/3/ks.json
```diff
@@ -11,29 +11,12 @@
         "initramfs"
     ],
     "additional_packages": [
-        "conntrack-tools",
-        "ebtables",
-        "jq",
-        "libnetfilter_conntrack",
-        "libnetfilter_cthelper",
-        "libnetfilter_cttimeout",
-        "libnetfilter_queue",
```
codenrhoden: All the `libnetfilter_*` packages here are dependencies of `conntrack-tools`. I verified this. They don't need to be listed separately, so they didn't get moved over the Ansible side of things.

akutz:
Hi @codenrhoden,

Did you build this and boot it with example CAPI data to ensure it still works? Until there's automated testing, I always build an affected image and make sure that it still functions as expected. I assume you did, but as you didn't note it, I thought I should ask.

codenrhoden:
@akutz definitely always good to check. I actually do a full deploy with CAPV. Looks good, here's my workload cluster:

```
$ kubectl --kubeconfig out/trhoden-workload-cluster-1/kubeconfig get nodes -o wide
NAME                                               STATUS   ROLES    AGE     VERSION   INTERNAL-IP     EXTERNAL-IP     OS-IMAGE                 KERNEL-VERSION   CONTAINER-RUNTIME
trhoden-workload-cluster-1-controlplane-0          Ready    master   4m25s   v1.16.2   192.168.3.185   192.168.3.185   VMware Photon OS/Linux   4.19.79-2.ph3    containerd://1.3.2
trhoden-workload-cluster-1-md-0-79564cf587-q2qmd   Ready    <none>   3m3s    v1.16.2   192.168.3.208   192.168.3.208   VMware Photon OS/Linux   4.19.79-2.ph3    containerd://1.3.2
```

akutz:
Hi @codenrhoden,

This looks fine. Please feel free to remove the hold.

/lgtm
/hold

A general question -- in another PR you discussed moving packages to Ansible as a way to help support offline builds, but I don't see how that helps. It's just moving package installation from one place to the other. Could you please provide some clarity there? Thanks!

One reason I put so many packages in kickstart and pre-seeds was to eventually be able to create a base image that didn't have to be updated as often. Consider the idea that the pre-ansible part of the process could produce ISOs that could be further configured via a second, Ansible driven phase. And the more we can do in the first phase, the more useful the base ISO would be.

codenrhoden:
Hey @akutz, really good questions...

> A general question -- in another PR you discussed moving packages to Ansible as a way to help support offline builds, but I don't see how that helps. It's just moving package installation from one place to the other. Could you please provide some clarity there? Thanks!

The main thinking here is that when packages are installed via Ansible, we are at a step in the process in the build where there is an option to disable public repository access and use internal repos if so desired. That could be something like RedHat Satellite, Pulp, Artifactory, etc. Having as many packages be installed at this stage facilitates not hitting public repos if that's what you need, and allows consumers to implement package retention policies at their mirrors.

For both preseed and kickstart it *is* possible to point to a custom mirror, but not without making the preseed and kickstart files templated in some fashion. And that's just more work than it's worth, especially since we already have it handled in the 2nd, Ansible-driven phase.

> One reason I put so many packages in kickstart and pre-seeds was to eventually be able to create a base image that didn't have to be updated as often. Consider the idea that the pre-ansible part of the process could produce ISOs that could be further configured via a second, Ansible driven phase. And the more we can do in the first phase, the more useful the base ISO would be.

We completely agree on the goal here. That's why I filed #39 a few months ago. I actually think that having fewer packages installed at the ISO stage helps with this as well, though. I would want the disk image produced from the ISO to be as small as possible (in terms of packages, not bytes), so it is essentially locked to what was distributed with the ISO and never changes. It could be created once. If there is a security update to a package that is installed from the ISO, we have to re-run the Ansible step anyways to update it. We *always* have to run the Ansible step to create the final image. And since we run steps like `apt-get dist upgrade` (and the `yum` equivalent) I don't think we are saving much in the way of time by installing packages earlier.

One final thing... If you are comfortable with hit, do you think you could also take a look at #47? Jason is out on VACA and I'd like to get it in as well. It's the same process as this one...

akutz:
Hi @codenrhoden,

My thinking was to produce:

> a base image that didn't have to be updated as often

So the base image still gets built, but maybe less often? And the thing about CVEs, etc, is that, as you said:

> We always have to run the Ansible step to create the final image. 

Thus a simple `yum|tdnf|apt-get update` will update whatever packages were already built into the base image. But since most of the machine image configuration is in Ansible, it's far more likely machine images need to be rebuilt due to changes in the second phase instead of the first. It's the same reason you created a `manifests-base` image in CAPV, right?

So yes, we agree on the goal, but I think it still makes sense to include as many packages (with the exception of Kubernetes, the CRI, and a few others) in the ISO build as possible. When the base image doesn't need to be rebuilt, building a new machine image will be *much* faster since only packages requiring updates will be installed on top of what's already in the base ISO. We could also make the base ISOs available to others if they wanted to start there.

I do agree with one argument:

* Building AMIs is entirely Ansible driven, therefore...
* Package lists shared between AMIs and OVA builds make sense to maintain in Ansible and you...
* Don't want to duplicate a subset of those into the kickstart/pre-seed files

However, since the AMIs aren't built on top of minimal ISOs, they don't actually need many of the same packages we maintain in the shared lists. I'd argue that the overhead of some duplicate packages in the kickstart/pre-seed files and Ansible defaults is worth it given how much it would speed things up to be able to build on top of base ISOs we produce less often than we need to rebuild machine images due to config changes rooted in Ansible.

PR #129, opened 2020-01-17
Add RHEL 7 support

### LINKED ISSUE #78: Support running image-builder in RHEL 7
The Makefile in images/capi/packer/ami and images/capi/packer/gce uses `shasum` to verify the downloaded packer plugins.

In some distributions, such as RHEL 7, `shasum` is actually `sha256sum`. I think we can update the Makefile to detect whether `shasum` or `sha256sum` is available on the machine. If neither are there, exit with an error.

codenrhoden:
This PR has two commits.

The first commit standardizes on a common user/pass for all the OVA builders. There's really no need to have separate users for each distro, and it just adds extra code since we weren't re-using them. The user account is used for Ansible access, and then is locked and is never used again. It really doesn't matter what it is -- might as well unify and re-use.

The second commit adds the RHEL 7 support. It uses the most basic solution of of user/pass for RHSM access. Larger uses may use a different method for registration (e.g. certificate). I don't think we can cover all the bases here, but this gets folks 95% of the way there if they want to build an image on RHEL. I've called out a couple of things to be cautious of in the docs.

/hold
holding because I still need to test deployment with CAPV. Not sure if I'll have that done today or not.


images/capi/packer/ova/linux/rhel
```diff
@@ -0,0 +1 @@
+centos
```
detiber: The contents of this file feel a bit off, is this correct?

codenrhoden: GitHub doesn't show it very well... but yes. It's a symlink.

codenrhoden: I actually double-checked it a couple times, because I wondered the same thing. 😆

detiber: lol, thanks for the clarification.

codenrhoden:
well darn. When I deploy this image with CAPV, it never gets an IP address. I'll have to build a different image that doesn't have user locked down so I can get in via console and figure out why. 😞

detiber:
> well darn. When I deploy this image with CAPV, it never gets an IP address. I'll have to build a different image that doesn't have user locked down so I can get in via console and figure out why. disappointed

Is it related to a change in the mac or other identifier for the network interface? I suspect we might not be cleaning things up properly to ensure that the instance handles a new MAC properly and starts the network.

codenrhoden:
still looking into this, but at first glance it seems related to network adapter IDs. The cloud-init config is injecting something for `id0`, which does not exist. However it has the right MAC address. 🤔 
There also exists a config for `ens33`, which also does not exist, but says it is generated by dracut and initrd.

The actual device in the VM is `ens192`. No config script for it exists in `/etc/sysconfig/network-scripts/`. The config script for `id0` looks correct (correct MAC address, comes from cloud-init), but no idea why it's generated for `id0` isntead of `ens33`...

will keep digging.

codenrhoden:
I'm de-prioritizing this for now... but for posterity, it got stuck on networking. I found that the images built didn't come up with an IP address, as per previous comments. It is almost identical to the issues described in https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/583.

Why it works for CentOS and not RHEL? I don't know, and I'm no longer investigating as RHEL 7 support as a base OS hasn't been asked for or listed as anyone's requirements.

/lifecycle frozen

images/capi/ansible/roles/setup/tasks/redhat.yml
```diff
@@ -12,6 +12,16 @@
...
+- name: RHEL subscription
+  redhat_subscription:
+    state: present
+    username: "{{ lookup('env', 'RHSM_USER') }}"
+    password: "{{ lookup('env', 'RHSM_PASS') }}"
+    auto_attach: true
+  when: ansible_distribution == "RedHat"
+
+# EPEL is needed even on RHEL for python2-pip
+# An alternative could be to use SCL, but going the simpler route for now.
```
detiber, line 24: I know @voor recently had something that update RH to use pip3 instead, could the same thing be used here to avoid EPEL?

codenrhoden: So I actually looked into this a bit today, specifically if I could remove the Python 2 dependency from CentOS 7 as a base for this. Unfortunately there are many roadblocks.

CentOS 7 as a whole requires Python 2 (needed for Yum, for example). PIP is not installed by default though, so if all we wanted to do was install python3-pip and then use that, it would seem like it would work.... However...

`cloud-init` as a package on CentOS 7 requires Python 2:

```sh
# rpm -q --requires cloud-init
...
/usr/bin/python
PyYAML
libselinux-python
pyserial
python(abi) = 2.7
python-configobj
python-jinja2
python-jsonpatch
python-prettytable
python-requests
python-setuptools
python-six
...
```

So Python 2 has to stay. And additionally for running on vSphere, the install script specifically **requires** pip to install: 
https://github.com/vmware/cloud-init-vmware-guestinfo/blob/5ac2fc01e633399353726928c747943eced393b1/install.sh#L31-L60

And if it's not clear what it is doing there, it is first checking to see that cloud-init is installed as a Python module, and if it is, it also checks to see if it's in Python 2 or 3, and specifically uses that one. So since cloud-init installs for Python 2, **even if Python 3 is installed**, the VMW datasource gets installed for Python 2. Thus, we always need python2-pip. No getting around this one that I can see.

I think the stuff @voor did was looking at RH/RHEL on AWS, thus this wasn't an issue since it doesn't need to install the vmware guestinfo datasource for cloud-init.

voor: I was actually just prioritizing installing pip modules into python3 instead of python2, I didn't effectively remove python2 by any means, it's definitely an integral part of CentOS 7 -- might not be the case in CentOS 8, though.

EleanorRigby:
FYI,

RHEL7/CentOS k8s cluster fails to deploy with clusterctl : https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1847

codenrhoden:
/hold cancel
/assign @EleanorRigby 

@EleanorRigby can you take a look at this please? I was finally able to get everything to through correctly. GOSS tests pass, and I was able to remove the ISO media check to speed builds up. :)

It's now such a small PR relative to the amount of work that went into it. 😆

EleanorRigby:
@codenrhoden : I had a look at code before and now also. I agree RHEL-7 support has been long pending and a couple of supporting PRs went in to make this work. Thanks for adding this to the image builder.

/lgtm

PR #131, opened 2020-01-27
Fix imagebuilder docker build failure

guikcd:
To avoid build failure with modules, use a newer version which use go.mod

Without this, the docker build fails with "cannot find package" msgs.

See https://blog.golang.org/go1.13 for reference.


guikcd:
I signed it

codenrhoden:
/lgtm
Seems straightforward to me, but this bit of code falls to @justinsb for approval.
/unassign

justinsb:
Thanks for fixing @guikcd 

/approve
/lgtm

Issue #90, opened 2019-11-01
CAPV images for Ubuntu have unattended upgrades enabled


codenrhoden:
This must happen by default with Ubuntu, but he current images have the `unattended-upgrades` package installed and enabled. Furthermore, there are systemd units that are trying to do a daily apt-get update... These units are:

- `apt-daily.{service,timer}`
- `apt-daily-upgrade.{service,timer}`

Since the images are supposed by immutable, we *really* don't want any packages changing automatically, and we don't need any daily tasks to fetch new package lists. This should all be cleaned up.

/assign

codenrhoden:
Just to follow-up on this one, I want to pose the question: Are the images created by this tool intended to be immutable?

I believe the answer to this question is yes -- if any package update is required, the intent is to build a new image.

Other uses may not agree with this process. I know from my perspective, I have internal requirements for these images to be immutable, so at a minimum I need the _option_ to disable these services. The question then becomes -- should it be immutable by default?

/cc @akutz @detiber @figo

detiber:
@codenrhoden based on the use cases for the tool that we've described so far, I think the best option is to default to immutable. If someone wants to have mutable behavior, we shouldn't necessarily stop them, but it should definitely be the exception.

codenrhoden:
> @codenrhoden based on the use cases for the tool that we've described so far, I think the best option is to default to immutable. If someone wants to have mutable behavior, we shouldn't necessarily stop them, but it should definitely be the exception.

This is my intended approach, I just wanted to get buy-in before I do it. I'll bring it up in today's community meeting as well. I know you won't be there, so thanks for sharing your input here.

randomvariable:
FWIW, did consider adding the following, to lock the Kubernetes version but allow package updates to the rest of the system, just in case someone does do it, it shouldn't break the node.

https://github.com/randomvariable/wardroom/blob/fedora-modernization/ansible/roles/kubernetes/tasks/debian.yml#L45-L65

akutz:
So CAPV is hitting flakes all the time now because of the boot-time apt tasks. FWIW, this [is a known issue in Ubuntu](https://askubuntu.com/questions/1038923/do-i-really-need-apt-daily-service-and-apt-daily-upgrade-service), and apparently it's by mistake as these tasks [should be run post-boot](https://askubuntu.com/questions/800479/ubuntu-16-04-slow-boot-apt-daily-service).

[...]

PR #134, opened 2020-01-30
Make sure Ubuntu images remain immutable

codenrhoden:
A key tenet of the CAPI images are that they are immutable. Unfortunately, the Ubuntu image is being created with the default settings of having a few services enabled, namely `apt-daily`, `apt-daily-upgrade`, and `unattended-upgrades` enabled. Each of these is unnecessary in an immutable image.

For the `apt-daily*` services and timers, they just added unnecessary time to the boot process. For the `unattended-upgrades` service, that can actually upgrade/modify packages, and we do not want it enabled.

Furthermore, during investigating this issue, it was noticed that there was a consistent 90-second pause during the boot time waiting for the "Hyper-V KVP Protocol Daemon" to startup. This boot would [consistently fail](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1820063). I see no reason to run a Hyper-V specific daemon for OVA images running on ESXi, so disabling this service speeds boot times up by 90 seconds.

For full disclosure, there is one weird thing... According the bug linked above, the issue should have been fixed for Bionic with Linux kernel `4.15.0-74.84`. When I build a new image from scratch today (without this fix) we end up with kernel `4.15.0-76-generic`, which seems to me it should include the fix. But the problem persists, but disabling this services solves the problems. It takes VM boot time on my laptop with Fusion workstation from
```sh
$ systemd-analyze
Startup finished in 2.010s (kernel) + 1min 36.354s (userspace) = 1min 38.365s
```
to
```sh
$ systemd-analyze
Startup finished in 1.715s (kernel) + 6.638s (userspace) = 8.354s
```
Which I think is pretty great. Exactly 90 seconds less in userspace. 💪 

fixes: #90 
/assign @detiber @akutz

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -61,6 +61,18 @@
   - { path: /var/lib/apt/lists, state: absent,    mode: "0755" }
   - { path: /var/lib/apt/lists, state: directory, mode: "0755" }
 
+- name: Disable unwanted systemd services
+  systemd:
+    name: "{{ item }}"
+    state: stopped
+    enabled: false
+  loop:
+  - apt-daily
+  - apt-daily.timer
+  - apt-daily-upgrade
+  - apt-daily-upgrade.timer
+  - unattended-upgrades
+
 - name: Reset network interface IDs
   file:
     state: absent
--- a/images/capi/ansible/roles/providers/tasks/vmware.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware.yml
@@ -63,7 +63,7 @@
     path: /etc/cloud/cloud.cfg.d/99-disable-networking-config.cfg
     state: absent
   when: ansible_os_family == "VMware Photon OS"
-  
+
 #- name: Unlock password
 #  replace:
 #    path: /etc/cloud/cloud.cfg
@@ -76,3 +76,9 @@
     dest:  /usr/lib/cloud-init
     state: link
   when: ansible_os_family == "RedHat"
+
+- name: Disable Hyper-V KVP protocol daemon
+  systemd:
+    name: hv-kvp-daemon
+    state: stopped
+    enabled: false


images/capi/ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -61,6 +61,18 @@
   - { path: /var/lib/apt/lists, state: absent,    mode: "0755" }
   - { path: /var/lib/apt/lists, state: directory, mode: "0755" }
 
+- name: Disable unwanted systemd services
```
akutz: Hi @codenrhoden,

Should these also be masked? Is that what setting `enabled: false` does? It looks like there is a an actual [`masked` setting](https://docs.ansible.com/ansible/latest/modules/systemd_module.html).

codenrhoden: Yeah, let me educate myself there... I've never masked a service and am unfamiliar with it. More used to just disabling a service.

akutz: Masking a service symlinks the service in `/etc/systemd` to `/dev/null`

akutz:
/hold

I'm placing a hold so this doesn't get merged until @codenrhoden has a chance to decide if he wants to mask the services as well.

akutz:
/hold
/lgtm

I'm placing a hold so this doesn't get merged until @codenrhoden has a chance to decide if he wants to mask the services as well. Otherwise, Travis, please feel free to remove the hold and let this merge if you're good with things as is.

images/capi/ansible/roles/sysprep/tasks/debian.yml
detiber: I don't necessarily think we need to mask the services...  If someone is going to tamper with the system post-boot, they are going to tamper with it. Masking the service is more likely just to have them unmask and re-enable it if they are stubborn enough.

codenrhoden: I agree with @detiber -- if someone wants to re-start the service, they will. Either way they have to jump through the same hoops, e.g. `systemctl enable && systemctl start` or `systemctl unmask && ...`

Glad I learned about this, though!

PR #136, opened 2020-01-31
Move common ansible args to one location

codenrhoden:
Right now, each CAPI image builder (AWS, Azure, DigitalOcean, GCE,
vSphere) has a separate definition of the "extra-vars" to send to
Ansible. These are in fact mostly identical, and they should be.
Therefore, move the defintion to a common location.

The vSphere builder has a couple of unique extra vars that the other
builders don't need, so the ability to have provider-specific
additions is preserved (and used for vSphere).


codenrhoden:
This is a cleanup I've wanted to do for a while. I am about to add a new arg that needs to be passed into all the builders, and I was annoyed that I would have to add it in 5 different places. If it's common, I should only have to do it once.

With this patch, only one provider will actually "see" a difference. The Azure builder was not passing in the variables to disable public repos and provide your own. Now it is, and it was the only one not to do. Passing them in now will have no effect. Though I just realized that they are not defined, so let me add that real quick and then unhold after.

/hold

codenrhoden:
Okay, this is good now. This will also help ensure that each provider is supporting the same set of flags, with room for customization if needed.

/hold cancel
/assign @detiber @akutz

detiber:
lgtm
/cc @CecileRobertMichon 

The recursive nature of the resolution of these variables is giving me a bit of a headache, but I think it's worth it to remove the duplication.

CecileRobertMichon:
Azure changes lgtm

codenrhoden:
> The recursive nature of the resolution of these variables is giving me a bit of a headache, but I think it's worth it to remove the duplication.

Yeah, I hear you there.  All stuff that will get resolved when this is all generated by a CLI.

Also, to be clear, the AWS, DigitalOcean, and GCE builders all had identical definitions for this. The vSphere builder was the exact same, with the addition of the ability to define a location for the guestinfo datasource.

The only difference betwee the Azure builder and the other three that was "common" was the missing repo management stuff. Probably not a capability that is even needed, but having them all be common is worth it I think.

codenrhoden:
@detiber are you happy with this since Cecile is?

PR #140, opened 2020-02-04
Fix use of sha256sum for Linux vs macOS

### LINKED ISSUE #78: Support running image-builder in RHEL 7

codenrhoden:
This PR has three commits.

1. De-dupe the the Makefiles found in `capi/packer/ami` and `capi/packer/gce` -- they are identical and only do one thing -- install the GOSS plugin for Packer.
2. After adding the `packer-plugins` target to the main CAPI makefile, do a little cleanup on `make help` output
3. Fix the the call to `shasum` to work on both macOS and Linux.

The 3rd commit actually fixes the issue for not being able to use the image-builder tool itself on RHEL. In a previous comment there I had conflated *building* a RHEL image, and running the image-builder *tool* on RHEL.

fixes #78 

/assign @detiber


images/capi/Makefile
```diff
@@ -119,7 +119,21 @@ $(OVA_CLEAN_TARGETS):
...
+		$(GCE_BUILD_TARGETS) \
+		$(AZURE_BUILD_TARGETS)
+build: ## Builds all images
+
+.PHONY: clean
+clean: $(OVA_CLEAN_TARGETS)
+clean: ## Cleans all local image caches
+
+## --------------------------------------
+## Document dynamic targets
```
codenrhoden: I moved the location of these in the Makefile so they would be at the top of the list and grouped with the other build targets. Also added a newline after `make <target>`. New output looks like this:

```sh
$ make help
NOTE
  The "build-ova" targets have analogue "clean-ova" targets for
  cleaning artifacts created from building OVAs using a local
  hypervisor.

Usage:
  make <target>

  help             Display this help
  build            Builds all images
  clean            Cleans all local image caches
  build-ova-centos-7  Builds CentOS 7 OVA w local hypervisor
  build-ova-photon-3  Builds Photon 3 OVA w local hypervisor
  build-ova-ubuntu-1804  Builds Ubuntu 18.04 OVA w local hypervisor
  build-ova-ubuntu-2004  Builds Ubuntu 20.04 OVA w local hypervisor
  build-esx-centos-7  Builds CentOS 7 OVA w remote hypervisor
  build-esx-photon-3  Builds Photon 3 OVA w remote hypervisor
  build-esx-ubuntu-1804  Builds Ubuntu 18.04 OVA w remote hypervisor
  build-esx-ubuntu-2004  Builds Ubuntu 20.04 OVA w remote hypervisor
  build-ami-default  Builds the AMI default image
  build-gce-default  Builds the GCE default image
  build-azure-ubuntu-1804  Builds Ubuntu 18.04 for Azure
  build-do-default  Builds the DigitalOcean snapshot default image
  packer-plugins   Install Packer plugins for GCE and AMI builds
```

images/capi/packer/Makefile
```diff
@@ -31,7 +31,7 @@ ${HOME}/.packer.d/plugins/packer-provisioner-goss:
 	mkdir -p /tmp/packer-plugins
```
detiber: Currently it looks like goss is only used by ami/packer.json, we can remove all references to it's use by GCE at this point :)

images/capi/packer/Makefile
```diff
@@ -31,7 +31,7 @@ ${HOME}/.packer.d/plugins/packer-provisioner-goss:
...
 	curl -L -o /tmp/packer-plugins/packer-provisioner-goss ${PLUGINS_GOSS_URL}
 	echo ${PLUGINS_GOSS_SHA256-$(BINARYPATHPATTERN)}  */tmp/packer-plugins/packer-provisioner-goss > /tmp/packer-plugins/SHA256SUMS
-	shasum -c /tmp/packer-plugins/SHA256SUMS
+	{ shasum -c /tmp/packer-plugins/SHA256SUMS || sha256sum -c /tmp/packer-plugins/SHA256SUMS ; } 2>/dev/null
```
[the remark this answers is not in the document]
codenrhoden: I can flip the order, sure. I left it this way since clearly most users are using that Makefile on Mac, otherwise we'd have heard a lot more complaints. :) I muted stderr so you don't see the complaint about one of them not being found -- same pattern I've seen used in other Makefile and bash scripts that do this same thing.

images/capi/packer/Makefile
codenrhoden: hmm, I think the only remaining reference would me the comment I made in the makefile.. This begs the question of whether I should really have moved `packer/ami/Makefile` up a directory like I did, or leave it in `ami/`. Thoughts?

detiber: I think it might be good to just leave it under `ami/` considering it hasn't gained any additional adoption.

codenrhoden: Okay, agreed. And actually, I need to do more testing on the Linux side of this because as it just dawned on me, something doesn't add up. If the existing makefile is just using `shasum` on mac, that means it's doing a SHA1 sum (the default). To do do 256, you need `shasum -a 256`. But we're putting it in a file that says it's 256? I'll get it sorted out. :)

codenrhoden:
@detiber As I'm looking at this more, I also just found all the dependency stuff that is in `capi/hack/image-tools.sh`. It *also* tries to install GOSS, but also has stuff in it to check/install ansible, packer and `jq`, but those are all commented out. The SHA256 stuff for goss in `hack/image-tools.sh` already supports both commands. 🙄  I'd like to reconcile all this...

Here's what I'm thinking instead.

- Split `hack/image-tools.sh` into separate scripts that can be called individually`
- Add `<provider>-deps` make targets to call whatever scripts are needed for *them*
  - For example, Azure needs `jq`, but not goss

With this in place, you would be able to download the image-builder, then do something like: `make ami-deps && make build-ami-default`, and it would deal with packer, ansible, goss all for you. At least that would be the goal.

Thoughts?

detiber:
+1 for breaking out the scripts and adding them as dependencies to the make targets

codenrhoden:
/close

deferring to #142.

PR #147, opened 2020-02-06
Doc updates

codenrhoden:
A few general doc cleanup items. Mostly things I noticed while browsing https://image-builder.sigs.k8s.io

/assign @detiber


hack/build-mdbook.sh
```diff
@@ -1,64 +0,0 @@
-#!/bin/bash
```
detiber: It looks like this file is still called by the Makefile, should it be calling `docs/book/build.sh` instead similar to the netlify config?

codenrhoden: ah, good catch on the Makefile. Yeah, I'll fix that!

codenrhoden:
@detiber Makefile is updated!

PR #151, opened 2020-02-07
Add ability to inject extra packages

codenrhoden:
This patch exposes the ability for a user to provide a list of extra RPM
or DEB files to install during the build.

We also document the "proper" way to do this -- by providing a JSON file
to the build rather than editing existing JSON files or by the Ansible
YAML files.

/assign @detiber


codenrhoden:
/hold

A couple more tests in progress...

codenrhoden:
/hold cancel
I needed to add a `when:` clause for the Photon stuff because of the way `tdnf` works and that it's a shell call rather than an actual module...

docs/book/src/capi/capi.md
```diff
@@ -27,6 +27,34 @@ The `images/capi/packer/config` directory includes several JSON files that defin
...
+### Customization
+
+Several variables can be used to customize the image build.
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `extra_debs` | This can be set to a space delimited string containing the names of additional deb packages to install | `""` |
+| `extra_rpms` | This can be set to a space delimited string containing the names of additional RPM packages to install | `""` |
+
+The variables found in `packer/config/*.json` or `packer/<provider>/*.json` should not need to be modified directly. For customization it is better or create a JSON file with your changes and provide it via the `PACKER_VAR_FILES` environment variable. Variables set in this file will override any previous values.
```
detiber: Does `PACKER_VAR_FILES` accept multiple files? If so, is it space separated, comma separated, etc? If so we should explain how to do that for the user. If not, we should probably make it `PACKER_VAR_FILE` instead.

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -61,10 +61,10 @@ common_photon_rpms:
 - unzip
 
 common_pinned_debs: []
-common_extra_debs: []
-common_extra_rpms: []
 common_redhat_epel_rpm: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
 disable_public_repos: false
+extra_debs: ""
+extra_rpms: ""
```
detiber: Is this change needed because of the conversion from terraform variable to ansible? I think I'd prefer to leave it as an array if at all possible to avoid the string splitting we are doing elsewhere to enable this.

docs/book/src/capi/capi.md
codenrhoden: It does accept multiple. space delimited. More examples around all this are coming, for sure. I will update this to clarify behavior.

images/capi/ansible/roles/common/defaults/main.yml
codenrhoden: To be clear, these variables have never been used.

I'm putting them as strings because of how they get populated. Thought it would be easier for a user to type it out, but I think that's before I realized a user would have to put them all in a JSON file anyways... So basically, instead of doing something like this:
```json
{
  "extra_rpms": "\"nfs-utils net-tools\""
}
```

it would be:
```json
{
  "extra_rpms": ["nfs-utils", "net-tools"]
}
```

That probably does make a lot more sense...  I'll have to test it to make sure it works.

codenrhoden: @detiber So, I remember now that I had tried to do something similar before and it was a disaster. It still is. It may be worth it to someone, but right now this would turn into a **huge** time sink for me. Here's what happens...

When you put what looks like a normal JSON array into the var file, like what I put above, you get the error: 
```
Error reading variables in '/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/extra_vars.json': json: cannot unmarshal array into Go value of type string
```

Which is odd. Why would it be trying to put an array into a string? This led me to https://github.com/hashicorp/packer/issues/2278 and https://github.com/hashicorp/packer/issues/5194. That led me to try all sorts of combinations of comma separated strings.. And Packer happily took them. 

So the way you get Packet to treat it as an array is with the following:

```json
{
  "extra_rpms": "nfs-utils,net-tools"
}
```

Which I don't like because it doesn't look like JSON, but it would be manageable. But then when it calls out to Ansible, it doesn't format them as an array on the CLI. It will output something like

```
extra_rpms=nfs-utils,net-tools
```

Which again, is not an array. 😞 
I completely see this as a bug in Packer, but people have been complaining about it for years, so 🤷‍♂ 

Like I said, maybe there's a way around it, but I don't see it as time well spent at the moment. I **do** see a future where we have a CLI generate a single vars file for us, and instead of passing tons of vars to `ansible-playbook` via `--extra-vars`, we instead do `--extra-vars "@vars.json"`, which would be much much nicer.

docs/book/src/capi/capi.md
```diff
@@ -27,6 +27,50 @@ The `images/capi/packer/config` directory includes several JSON files that defin
...
+
+Several variables can be used to customize the image build.
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `extra_debs` | This can be set to a space delimited string containing the names of additional deb packages to install | `""` |
+| `extra_rpms` | This can be set to a space delimited string containing the names of additional RPM packages to install | `""` |
+
+The variables found in `packer/config/*.json` or `packer/<provider>/*.json` should not need to be modified directly. For customization it is better to create a JSON file with your changes and provide it via the `PACKER_VAR_FILES` environment variable. Variables set in this file will override any previous values. Multiple files can be passed via `PACKER_VAR_FILES`, with the last file taking precedence over any others.
+
```
codenrhoden: @detiber I added info here about passing 1 or more var files, and made example sections below.

figo:
Just a sanity check, maybe you already does, seems you have to make sure `extra_repos` works fine with `extra_rpms` and `extra_debs`, also no place to provide gpg key for the `extra_repos`.

codenrhoden:
Hi @figo,

> Just a sanity check, maybe you already does, seems you have to make sure `extra_repos` works fine with `extra_rpms` and `extra_debs`, also no place to provide gpg key for the `extra_repos`.

There's nothing **we** have to do make sure that `extra_rpms` or `extra_debs` plays nice with `extra_repos`. What I mean by that is that whatever packages you install have to be available via the repos you have active. Managing those repos is orthogonal to specifying the packages you want to install. We do make sure that all "extra" packages are installed *after* any repo management steps. So hopefully that alleviates your concern?

As for GPG keys, thanks for bringing that up. For RPM-based systems, since users are injecting a `.repo` file, they should be specifying that themselves in the `.repo` file. This would be like filling in the line yourself for [Photon](https://github.com/kubernetes-sigs/image-builder/blob/b6c9d2811da7bf1c6bb058574c4a852b59ea20a4/images/capi/ansible/roles/kubernetes/templates/etc/yum.repos.d/kubernetes.repo#L6) or [CentOS](https://github.com/kubernetes-sigs/image-builder/blob/b6c9d2811da7bf1c6bb058574c4a852b59ea20a4/images/capi/ansible/roles/kubernetes/tasks/redhat.yml#L21). Both of those result in a `.repo` file that has the GPG key listed.

As for dpkg/DEB systems... Hmm.... you do usually (always?) have to import the GPG key before hand -- it's not listed in the `.list` file. so if the packages are signed, we will need to have a step to specify the GPG key(s) to import I think. Good catch! I'll look for an example of where not having a GPG key imported first to see what happens. I might just be able to try and install the Kubernetes packages on Ubuntu but clear out the GPG key and see what happens (just to educate myself).

images/capi/ansible/roles/common/defaults/main.yml
detiber: :grimacing: sorry for sending you down that rabbit hole

codenrhoden: No, it's good to have it documented why it is the way it is. I'm mostly worried there is an obvious solution that I'm not seeing.  :/

PR #152, opened 2020-02-07
Place user supplied var files at end of command

### LINKED ISSUE #150: Make sure user-provided var files take precedence
When using a command of the form `PACKER_VAR_FILES=... make build...`, the var files provided should take precedence over what is there by default. This should be a pretty easy fix, just need to reorder how we list the files.

From the [Packer documentation](https://packer.io/docs/templates/user-variables.html):
> Combining the -var and -var-file flags together also works how you'd expect. Variables set later in the command override variables set earlier.

Right now, we are listing any user provided files first, but we really want to list them last.

codenrhoden:
This patch makes sure that any user-supplied files that will be listed
out as a -var-file for Packer will be listed at the very end of the list
of var files. This is because if a var is defined more than once, the
last supplied one takes precedence. We want users to be able to override
defaults, and that cannot happen without this change.

As an example, I created a JSON file and set the Kubernetes version in it to 1.17.2 (default is 1.16.2) and ran `PACKER_VAR_FILES=extra_vars.json make build-ova-photon-3`, but the build still used 1.16.2. After this change it correctly picked up the override and used 1.17.2 instead.

fixes #150 

/assign @detiber

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -83,20 +83,29 @@ ifeq (1,$(strip $(FOREGROUND)))
 PACKER_FLAGS += -var="headless=false"
 endif
 
+# We want the var files passed to Packer to have a specific order, because the
+# precenence of the variables they contain depends on the order. Files listed
+# later on the CLI have higher precedence. We want the common var files found in
+# packer/config to be listed first, then the var files that specific to the
+# provider, then any user-supplied var files so that a user can override what
+# they need to.
+
 # A list of variable files given to Packer to configure things like the versions
-# of Kubernetes, CNI, and ContainerD to install.
-PACKER_VAR_FILES +=	packer/config/kubernetes.json \
+# of Kubernetes, CNI, and ContainerD to install. Any additional files from the
+# environment are appended.
+COMMON_VAR_FILES :=	packer/config/kubernetes.json \
 					packer/config/cni.json \
 					packer/config/containerd.json \
 					packer/config/ansible-args.json
 
 # Initialize a list of flags to pass to Packer. This includes any existing flags
 # specified by PACKER_FLAGS, as well as prefixing the list with the variable
-# files from PACKER_VAR_FILES, with each file prefixed by -var-file=.
+# files from COMMON_VAR_FILES, with each file prefixed by -var-file=.
 #
 # Any existing values from PACKER_FLAGS take precendence over variable files.
-PACKER_FLAGS := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" ) \
+PACKER_FLAGS := $(foreach f,$(abspath $(COMMON_VAR_FILES)),-var-file="$(f)" ) \
 				$(PACKER_FLAGS)
+PACKER_VAR_FILES := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" )
 
 ## --------------------------------------
 ## Platform and version combinations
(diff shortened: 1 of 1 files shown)


images/capi/Makefile
```diff
@@ -130,26 +132,26 @@ DO_BUILD_TARGETS 	:= $(addprefix build-,$(DO_BUILD_NAMES))
 
 .PHONY: $(OVA_BUILD_TARGETS)
 $(OVA_BUILD_TARGETS):
-	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" packer/ova/packer.json
+	packer build -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" $(PACKER_FLAGS) packer/ova/packer.json
```
detiber: Would it make sense to add an additional layer of "precedence" here as well? So that "common" definitions have the least precedence, then subdirectory defined definitions, and then finally user defined definitions?

It looks like currently the subdirectory based definitions are given the least precedence.

codenrhoden: When you say "common", are you talking about what is in the `packer/config` directory?

So you are saying to list all the files from that dir first, followed by the build specific file, followed by user files... I think that makes sense.

detiber: Yes, thank you for decoding my shortcut speak as I was multi-tasking. That is exactly what I was thinking.

codenrhoden: @detiber I have implemented the ask.

You can see the results of supplying var files:

```sh
$ PACKER_VAR_FILES="extra_vars.json my_other_vars.json" make build-ova-photon-3
packer build -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/ansible-args.json"   -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/ova-photon-3.json" -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/extra_vars.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/my_other_vars.json"  packer/ova/packer.json
```
Note that the common vars at the beginning, and the user supplied ones are at the end.

And just in case I screwed up some default or anything in the makefile, I made sure everything was happy when you **didn't** supply custom files:

```sh
$ make build-ova-photon-3
packer build -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/ansible-args.json"   -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/ova-photon-3.json"  packer/ova/packer.json
```

So I think this should be good to go!

detiber:
/lgtm
This is awesome!

PR #161, opened 2020-02-14
capi: Add and document support for HTTP proxy

### LINKED ISSUE #79: Add support for building CAPI images behind an HTTP proxy
As a CAPI user, I want to be able to build my own images behind a corporate HTTP proxy.

At first glance, the following changes are needed:
- Flow http proxy vars into ansible
- Manually download goss 
- Disable the automatic download of goss in the goss plugin

codenrhoden:
fixes #79 

This patch adds support for passing `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` to Ansible. Packer already honors them.

/assign @detiber


images/capi/ansible/playbook.yml
```diff
@@ -14,9 +14,15 @@
...
     - role: common
     - role: providers
     - role: containerd
     - role: kubernetes
     - role: sysprep
+
+  environment:
+    HTTP_PROXY: "{{ http_proxy | default('') }}"
+    HTTPS_PROXY: "{{ https_proxy | default('') }}"
+    NO_PROXY: "{{ no_proxy | default('') }}"
```
detiber: Would it be possible to pass these through using the `ansible_env_vars` here: https://github.com/kubernetes-sigs/image-builder/blob/d7ac96703a90e17bff6e7dc5829725f1676dffb7/images/capi/packer/ami/packer.json#L184-L187?

codenrhoden: Not that I'm aware of. Ansible accepts env vars to change the behavior of Ansible itself, but those don't effect the plays/tasks within the remote target. This is the [documented](https://docs.ansible.com/ansible/latest/user_guide/playbooks_environment.html) way of changing the environment in the remote target.

I did look through the [config options](https://docs.ansible.com/ansible/latest/reference_appendices/config.html) and didn't see anything that seemed helpful here (though there are so many I could easily have missed it).

detiber: Ah, good to know. In that case could we leverage passing through the local env vars to the ansible vars? Or is there a use case where we would expect someone to want to use two separate sets of proxy vars that would require the separate json file definition. 

Something like
```
{
  "ansible_common_vars": "... http_proxy={{env `HTTP_PROXY`}} ..."
}
```
in `images/capi/packer/config/ansible-args.json`

codenrhoden: A tried a couple different versions of this... Short answer is still that you can't call functions (except `user`?) inside of user var files... Just like that commit I had to revert earlier...

I tried your example (definitely the cleanest) and get:

```sh
render 'packer_user_variables': template: root:1:255: executing "root" at <env `HTTP_PROXY`>: error calling env: env vars are not allowed here in:...
```

I tried making it a different variable in the same `ansible-args.json` file, but it's the same:
```sh
render 'packer_user_variables': template: root:1:2: executing "root" at <env `HTTP_PROXY`>: error calling env: env vars are not allowed here in:
```

I looked around for any examples where we are calling the `env` function outside of a base template (i.e. `packer.json`), and found this:

```sh
$ git grep "env \`" packer/*
packer/ami/packer.json:    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
packer/azure/azure-config.json:  "azure_location": "{{env `AZURE_LOCATION`}}",
packer/azure/azure-config.json:  "subscription_id": "{{env `AZURE_SUBSCRIPTION_ID`}}",
packer/azure/azure-config.json:  "client_id": "{{env `AZURE_CLIENT_ID`}}",
packer/azure/azure-config.json:  "client_secret": "{{env `AZURE_CLIENT_SECRET`}}"
packer/azure/azure-sig-ubuntu-1804.json:    "resource_group_name": "{{env `RESOURCE_GROUP_NAME`}}",
packer/azure/azure-sig-ubuntu-1804.json:    "shared_image_gallery_name": "{{env `GALLERY_NAME`}}",
packer/azure/azure-sig-ubuntu-1804.json:    "replication_regions": "{{env `AZURE_LOCATION`}}",
packer/azure/azure-vhd-ubuntu-1804.json:    "resource_group_name": "{{env `RESOURCE_GROUP_NAME`}}",
packer/azure/azure-vhd-ubuntu-1804.json:    "storage_account_name": "{{env `STORAGE_ACCOUNT_NAME`}}",
packer/azure/packer.json:    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
packer/digitalocean/packer.json:    "access_token": "{{env `DIGITALOCEAN_ACCESS_TOKEN`}}",
packer/digitalocean/packer.json:    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
packer/gce/packer.json:    "project_id": "{{env `GCP_PROJECT_ID`}}",
packer/gce/packer.json:    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
packer/ova/packer.json:    "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
```
Which if you look closely, only Azure is doing it outside of the base template... But I'm not sure if/how they got that working, because if I run `make build-azure-vhd-ubuntu-1804`, I see:

```sh
render 'packer_user_variables': template: root:1:2: executing "root" at <env `AZURE_CLIENT_SECRET`>: error calling env: env vars are not allowed here in:

{{env `AZURE_CLIENT_SECRET`}}
```

Same thing... so I don't think Azure works at all. 🤷‍♂

codenrhoden: As an aside, when I was trying to clean some of this stuff up, I *tried* to move `existing_ansible_ssh_args` to a common var file (was going to rename `ansible-args.json`), and hit this same limitation. 🙄

detiber: uggh, thanks for the background :)

PR #164, opened 2020-02-21
Update vSphere docs with SSH user and fix kubernetes test path

chrisjs:


docs/book/src/capi/providers/vsphere.md
```diff
@@ -88,11 +88,11 @@ Images may be downloaded via HTTP:
 
 #### Accessing Local VMs
 
-After the images are built, the VMs from they are built are prepped for local testing. Simply boot the VM locally with Fusion or Workstation and the machine will be initialized with cloud-init data from the `cloudinit` directory. The VMs may be accessed via SSH by using the command `hack/image-ssh.sh BUILD_DIR`.
+After the images are built, the VMs from they are built are prepped for local testing. Simply boot the VM locally with Fusion or Workstation and the machine will be initialized with cloud-init data from the `cloudinit` directory. The VMs may be accessed via SSH by using the command `hack/image-ssh.sh BUILD_DIR capv`.
```
codenrhoden: This is only true if the image is built with `FOREGROUND=1` (at least for Fusion), which is not the default. And is not documented. I need to file an issue to cover that. I actually only realized that was how things work a couple weeks ago and I've been meaning to document it ever since.

chrisjs: ah got it - i was testing against workstation and found that to work, did not test against fusion

codenrhoden: good to have that confirmed. I *thought* that was the behavior I had seen, but it'd been a while. Either way, I need to document it. :) Thanks again!

PR #166, opened 2020-02-22
Remove unused code, fix guestinfo data paths

chrisjs:


images/capi/hack/image-govc-cloudinit.sh
```diff
@@ -69,8 +57,8 @@ echo "image-govc-cloudinit: initializing cloud-init data"
 govc vm.change \
   -e "guestinfo.userdata.encoding=base64" \
   -e "guestinfo.metadata.encoding=base64" \
-  -e "guestinfo.userdata='$(base64 -w0 <build/images/cloudinit/user-data)'" \
-  -e "guestinfo.metadata='$(base64 -w0 <build/images/cloudinit/meta-data)'"
+  -e "guestinfo.userdata='$(base64 -w0 <$( cd "$( dirname "$0" )" && pwd )/../cloudinit/user-data)'" \
+  -e "guestinfo.metadata='$(base64 -w0 <$( cd "$( dirname "$0" )" && pwd )/../cloudinit/meta-data)'"
```
codenrhoden: I haven't tried it, but this looks weird to me. why the `pwd` call? Also, it seems like a couple sets of nested double-quotes without escaping...

shellcheck also does not like this:

```sh
$ shellcheck image-govc-cloudinit.sh

In image-govc-cloudinit.sh line 72:
  -e "guestinfo.userdata='$(base64 -w0 <$( cd "$( dirname "$0" )" && pwd )/../cloudinit/user-data)'" \
                                        ^-- SC2046: Quote this to prevent word splitting.


In image-govc-cloudinit.sh line 73:
  -e "guestinfo.metadata='$(base64 -w0 <$( cd "$( dirname "$0" )" && pwd )/../cloudinit/meta-data)'"
                                        ^-- SC2046: Quote this to prevent word splitting.
```

I might suggest an alternative. Rather than deleting the `cd "$(dirname "${BASH_SOURCE[0]}")/.."` above, keep it, and change `build/images/cloudinit/...` to `cloudinit/...`.

The way I see it, we've basically changed from do a `cd` to the directory above us once to doing it twice. Any compelling reason to do that?

codenrhoden: I see why the `pwd` now. But still, I think there's a much simpler way. :)

chrisjs: it was just more explicit in terms of reading for me, but the suggestion is fine and have pushed an update to reflect

codenrhoden:
/lgtm
/approve

Thanks!

It has obviously been a while someone used that script. Clearly not since the project was moved over into this repo and the directory structure changed. :)

PR #167, opened 2020-02-25
[capi/azure] add skuLongSummary to SKU template

CecileRobertMichon:
Add skuLongSummary field to the Azure create SKU template since it is now required.


marosset:
/lgtm
I had to make the same change for some aks-engine VHDs
https://github.com/Azure/aks-engine/pull/2729

PR #168, opened 2020-02-28
Move "Wants=" under Unit rather than Install

ppadmavilasom:
option `Wants=` should be under Unit. See: https://www.freedesktop.org/software/systemd/man/systemd.unit.html
Upon daemon-reload and restart, it would cause this issue: systemd unknown lvalue wants in section 'install'


codenrhoden:
Good catch!
Thanks @ppadmavilasom 

/approve

Issue #160, opened 2020-02-13
CAPI ova: need ability to customize OVF metadata


codenrhoden:
Downstream consumers of the image-builder for CAPI that building OVAs may have a need to customize the metadata included in the OVF. It's time to take a look at what is currently in the metadata, and an a generic way to customize the metadata.

This is all done in the Packer post-processing step.

/assign

MnrGreg:
@codenrhoden 

For the OS type and description to be correct, I had to include the `ovf:id` and `vmw:osType` before vSphere would recognize it.

For example from:
https://github.com/kubernetes-sigs/image-builder/blob/39ef2696339a8fa0a4f55030a77002a316e224e6/images/capi/hack/image-build-ova.py#L201

to:
```xml
    <OperatingSystemSection ovf:id="${OS_ID}" vmw:osType="${OS_TYPE}">
```

Pass through from the `ova-<osname>.json` to `image-build-ova.py` here.
```python
   # Create the OVF file.
    ovf = "%s.ovf" % build['name']
    create_ovf(ovf, {
        'BUILD_DATE': build_data['build_date'],
        'BUILD_NAME': build['name'],
        'ARTIFACT_ID': build['artifact_id'],
        'BUILD_TIMESTAMP': build_data['build_timestamp'],
        'OS_NAME': build_data['os_name'],
        'OS_TYPE': build_data['os_type'],
        'OS_ID': build_data['os_id'],
        'ISO_CHECKSUM': build_data['iso_checksum'],
        'ISO_CHECKSUM_TYPE': build_data['iso_checksum_type'],
        'ISO_URL': build_data['iso_url'],
        'POPULATED_DISK_SIZE': vmdk['size'],
        'STREAM_DISK_SIZE': vmdk['stream_size'],
        'VMX_VERSION': args.vmx_version,
    })
```

For Photon:
ovf:id="36"
vmw:osType="vmwarePhoton64Guest"

codenrhoden:
> For the OS type and description to be correct, I had to include the `ovf:id` and `vmw:osType` before vSphere would recognize it.

Ha, thanks for the pointer @MnrGreg ! You must have had a sense that I was working on that very same problem as you wrote that. I learned the same thing just a few hours before your comment. In fact, I opened a PR last night that does that very thing.

[...]

PR #171, opened 2020-03-10
capi: read OVF EULA from text file

codenrhoden:
This allows for future re-use of EULA text since it's not embedded in a
Python script. This also makes sure that utf-8 encoding of the OVF file
enforced.

This is setting up for some future/pending work. As part of #160, it was noticed that if a user is using a remote esx build host (e.g. `make build-esx-centos-7`) , since the end of that build step is the export of an OVA directly from the ESX host, we are skipping the call to `hack/image-build-ova.py`, which means we are skipping any of the **current** metadata we are setting -- like versions, timestamps, and the EULA.

#160 will change this workflow, and as part of that it needs to crack open an existing OVA and customize the OVF metadata. One of the things that will need to be done is add the EULA. So this PR pulls the EULA text into a separate file so that it can be re-used.

/assign @detiber

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -22,6 +22,7 @@
 
 import argparse
 import hashlib
+import io
 import json
 import os
 import subprocess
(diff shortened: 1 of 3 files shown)


images/capi/hack/image-build-ova.py
```diff
@@ -36,13 +37,23 @@ def main():
...
+                        help='Text file containing EULA')
     parser.add_argument(dest='build_dir',
                         nargs='?',
                         metavar='BUILD_DIR',
                         default='.',
                         help='The Packer build directory')
     args = parser.parse_args()
 
+    # Read in the EULA
+    eula = "test"
```
detiber: Any particular reason for this to not be `""`?

codenrhoden: 😞 Doh, good catch. debugging snuck it's way in!

codenrhoden: Fixed!

PR #177, opened 2020-03-13
Force cri-tools symlink

codenrhoden:
When Kubernetes is not installed via package, there is an Ansible
snippet that symlinks cri-tools from /usr/local/bin to /usr/bin. With
the latest Photon packages installed, a newer version of Docker is
pulled in, which now requires containerd. This puts a binary at
/usr/bin/ctr, and when we try installing the *newer* containerd and
symlink from /usr/local/bin, Ansible refuses to overwrite it. This patch
forces the overwrite.

/assign @detiber 
/cc @dims @figo

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/url.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/url.yml
@@ -17,6 +17,7 @@
     src:   "/usr/local/bin/{{ item }}"
     dest:  "/usr/bin/{{ item }}"
     state: link
+    force: yes
   loop:
   - ctr
   - crictl


codenrhoden:
The error you see w/o this fix:

```sh
photon-3: failed: [default] (item=ctr) => {"ansible_loop_var": "item", "changed": false, "gid": 0, "group": "root", "item": "ctr", "mode": "0755", "msg": "refusing to convert from file to symlink for /usr/bin/ctr", "owner": "root", "path": "/usr/bin/ctr", "size": 26041256, "state": "file", "uid": 0}
```

PR #179, opened 2020-03-16
Remove docker from installed packages

ppadmavilasom:
- expand photon minimal meta package, remove docker from list
- remove ansible rules to deal with docker disable


codenrhoden:
/approve

A big big thank you for this one, @ppadmavilasom! Not having Docker installed by default will make the image a lot cleaner.

A couple minor packages that could be removed in the future from the kickstart: `cloud-init` (we take care of this via Ansible already) and `vim` come to mind. We had made a concerted effort to shrink the base image as much as possible, and let end-users extend the image as they need. But since all of this stuff was previously there anyways, I don't consider it something to complain about. Thank you again!

I took the build for a test-drive and all succeeded.

PR #180, opened 2020-03-17
Add Category back to product section of metadata

figo:
It has a static value of <Category>Cluster API Provider (CAPI)</Category>

/assign @codenrhoden


codenrhoden:
I don't mind this, it's fine. I removed it in https://github.com/kubernetes-sigs/image-builder/commit/01f50634d1e2a2d1a2c774ba3e66adb18a81089f on purpose, and just wish it was driven by a clear need/requirement rather than looking for it because it used to be there. 🤷‍♂

PR #183, opened 2020-03-17
chore: install ipset for kube-proxy ipvs mode

dennisme:
Per https://github.com/kubernetes/kubernetes/tree/master/pkg/proxy/ipvs#run-kube-proxy-in-ipvs-mode

Not currently installed. 

```bash
ubuntu@ip-10-97-39-102:~$ which ipset
ubuntu@ip-10-97-39-102:~$
```


dennisme:
Holding off on this one, requires more testing.

PR #184, opened 2020-03-17
capi: Allow CNI checksum to be overridden

codenrhoden:
When installing CNI from an http source, allow the checksum to be passed
in as an ansible extra_var.

While I have been loathe to add more Ansible extra_vars, one of the goals of the capi builder is to be able to build w/o internet access (or to be able to build using internal mirrors/caches). Right now if you are installing CNI from a tarball rather than a package, there is no way to override the associated checksum. So this is pretty much required.

The checksum verification could also now be disabled by setting it to an empty string.

/assign @figo @dims 
/cc @jiatongw


images/capi/packer/config/cni.json
```diff
@@ -3,5 +3,6 @@
     "kubernetes_cni_rpm_version": "0.7.5-0",
     "kubernetes_cni_deb_version": "0.7.5-00",
     "kubernetes_cni_source_type": "pkg",
-    "kubernetes_cni_http_source": "https://github.com/containernetworking/plugins/releases/download"
-}
\ No newline at end of file
+    "kubernetes_cni_http_source": "https://github.com/containernetworking/plugins/releases/download",
+    "kubernetes_cni_http_checksum": "sha1:https://github.com/containernetworking/plugins/releases/download/v0.7.5/cni-plugins-amd64-v0.7.5.tgz.sha1"
```
dims: Can we please use sha512 or 256?

codenrhoden: Sure. FWIW, I didn't know if other checksums were published as I wasn't trying to change default behavior here, just make it overridable. But I did confirm that both SHA256 and SHA512 values are available. I put in SHA256 as default.

```sh
$ wget https://github.com/containernetworking/plugins/releases/download/v0.7.5/cni-plugins-amd64-v0.7.5.tgz.sha256
$ cat cni-plugins-amd64-v0.7.5.tgz.sha256
3ca15c0a18ee830520cf3a95408be826cbd255a1535a38e0be9608b25ad8bf64  cni-plugins-amd64-v0.7.5.tgz
```

PR #186, opened 2020-03-22
Add vsphere builder

marema31:
Since Packer has integrated a way to create image using vSphere API (and not a hack using ssh on esxi), I've created a vsphere builder to allow the creation of template directly on the vSphere.

Documentation is missing from the PR because the ova provisionner is refered as vSphere in it, in which way you prefer me to add this provisionner: a new section of vsphere or an other chapter but with which name ?

### DIFF
--- a/docs/book/src/capi/providers/vsphere.md
+++ b/docs/book/src/capi/providers/vsphere.md
@@ -8,31 +8,45 @@ The `make deps-ova` target will test that Ansible and Packer are installed and a
 
 The images may be built using one of the following hypervisors:
 
-| OS | Builder |
-|----|---------|
-| Linux | VMware Workstation |
-| macOS | VMware Fusion |
+| OS | Builder | Build target |
+|----|---------|--------------|
+| Linux | VMware Workstation | build-ova-<OS> |
+| macOS | VMware Fusion | build-ova-<OS> |
+| ESXi | ESXi | build-esx-ova-<OS> |
+| vSphere | vSphere >= 6.5 | build-vsphere-ova-<OS> |
 
-The `vmware-iso` builder supports building against a remote VMware ESX server, but is untested with this project.
 
-### Tools
+The `esxi` builder supports building against a remote VMware ESX server with [specific configuration](https://packer.io/docs/builders/vmware-iso.html#building-on-a-remote-vsphere-hypervisor) (ssh access), but is untested with this project.
+The `vsphere` builder supports building against a remote VMware vSphere using standard API.
 
-- [Packer](https://www.packer.io/intro/getting-started/install.html)
+
+### Prerequisites for all images
+
+- [Packer](https://www.packer.io/intro/getting-started/install.html) version >= 1.5.5
 - [Ansible](http://docs.ansible.com/ansible/latest/intro_installation.html) version >= 2.8.0
 
-## The `cloudinit` Directory
+### Prerequisites for vSphere builder
 
-The `cloudinit` contains files that:
+Complete the `vsphere.json` configuration file with credentials and informations specific to the remote vSphere hypervisor used to build the `ova` file.
+This file must have the following format (`cluster` can be replace by `host`):
+```
+{
+    "vcenter_server":"FQDN of vcenter",
+    "username":"vcenter_username",
+    "password":"vcenter_password",
+    "datastore":"template_datastore",
+    "folder": "template_folder_on_vcenter",
+    "cluster": "esxi_cluster_used_for_template_creation",
+    "network": "network_attached_to_template",
+    "insecure_connection": "false"
+}
+```
 
-- **Are** example data used for testing
-- Are **not** included in any of the images
-- Should **not** be used in production systems
-
-For more information about how the files in the `cloudinit` directory are used, please refer to the section on [accessing the images](#accessing-the-images).
+If you prefer to use a different configuration file, you can create it with the same format and export `PACKER_VAR_FILES` environment variable containing the full path to it.
 
 ## Building Images
 
-From the `images/capi` directory, run `make build-ova-<OS>`, where `<OS>` is the desired operating system. The available choices are listed via `make help`.
+From the `images/capi` directory, run `make build-ova-<OS>` or `make build-vsphere-ova-<OS>`, where `<OS>` is the desired operating system. The available choices are listed via `make help`.
 
 ### Configuration
 
@@ -44,6 +58,7 @@ In addition to the configuration found in `images/capi/packer/config`, the `ova`
 | `ova-centos-7.json` | The settings for the CentOS 7 image |
 | `ova-photon-3.json` | The settings for the Photon 3 image |
 | `ova-ubuntu-1804.json` | The settings for the Ubuntu 1804 image |
+| `vsphere.json` | Additional settings needed when building on a remote vSphere |
 
 
 The images are built and located in `images/capi/output/BUILD_NAME+kube-KUBERNETES_VERSION`
@@ -93,6 +108,7 @@ After the images are built, the VMs from they are built are prepped for local te
 #### Accessing Remote VMs
 
 After deploying an image to vSphere, use `hack/image-govc-cloudinit.sh VM` to snapshot the image and update it with cloud-init data from the `cloudinit` directory. Start the VM and now it may be accessed with `ssh -i cloudinit/id_rsa.capi capv@VM_IP`.
+This hack necessitate the `govc` utility from [VMmare](https://github.com/vmware/govmomi/tree/master/govc)
 
 ### Initialize a CNI
 
@@ -140,3 +156,13 @@ tar -zxvf kubernetes-test-linux-amd64.tar.gz
 cd kubernetes/test/bin
 sudo ./ginkgo --nodes=8 --flakeAttempts=2 --focus="\[Conformance\]" --skip="\[Flaky\]|\[Serial\]|\[sig-network\]|Container Lifecycle Hook" ./e2e_node.test -- --k8s-bin-dir=/usr/bin --container-runtime=remote --container-runtime-endpoint unix:///var/run/containerd/containerd.sock --container-runtime-process-name /usr/local/bin/containerd --container-runtime-pid-file= --kubelet-flags="--cgroups-per-qos=true --cgroup-root=/ --runtime-cgroups=/system.slice/containerd.service" --extra-log="{\"name\": \"containerd.log\", \"journalctl\": [\"-u\", \"containerd\"]}"
 ```
+
+## The `cloudinit` Directory
+
+The `cloudinit` contains files that:
+
+- **Are** example data used for testing
+- Are **not** included in any of the images
+- Should **not** be used in production systems
+
+For more information about how the files in the `cloudinit` directory are used, please refer to the section on [accessing the images](#accessing-the-images).
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -125,6 +125,7 @@ PLATFORMS_AND_VERSIONS	:=	$(CENTOS_VERSIONS) \
 							$(UBUNTU_VERSIONS)
 
 OVA_BUILD_NAMES			:=	$(addprefix ova-,$(PLATFORMS_AND_VERSIONS))
+VSPHERE_BUILD_NAMES		:=	$(addprefix vsphere-,$(OVA_BUILD_NAMES))
 ESX_BUILD_NAMES			:=	$(addprefix esx-,$(OVA_BUILD_NAMES))
 
 AMI_BUILD_NAMES			?=	ami-default
@@ -138,6 +139,7 @@ QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804
 ## --------------------------------------
 ## Dynamic build targets
 ## --------------------------------------
+VSPHERE_BUILD_TARGETS	:= $(addprefix build-,$(VSPHERE_BUILD_NAMES))
 OVA_BUILD_TARGETS	:= $(addprefix build-,$(OVA_BUILD_NAMES))
 ESX_BUILD_TARGETS	:= $(addprefix build-,$(ESX_BUILD_NAMES))
 AMI_BUILD_TARGETS	:= $(addprefix build-,$(AMI_BUILD_NAMES))
@@ -146,13 +148,17 @@ AZURE_BUILD_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_NAMES))
 DO_BUILD_TARGETS 	:= $(addprefix build-,$(DO_BUILD_NAMES))
 QEMU_BUILD_TARGETS	:= $(addprefix build-,$(QEMU_BUILD_NAMES))
 
+.PHONY: $(VSPHERE_BUILD_TARGETS)
+$(VSPHERE_BUILD_TARGETS):
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-vsphere-,,$@).json)" -var-file="packer/ova/vsphere.json"  -except=esx -except=local $(PACKER_VAR_FILES) -only=vsphere -force packer/ova/packer.json
+
 .PHONY: $(OVA_BUILD_TARGETS)
 $(OVA_BUILD_TARGETS):
-	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" -except=esx $(PACKER_VAR_FILES) packer/ova/packer.json
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-,,$@).json)" -except=esx -except=vsphere $(PACKER_VAR_FILES) -only=vmware-iso packer/ova/packer.json
 
 .PHONY: $(ESX_BUILD_TARGETS)
 $(ESX_BUILD_TARGETS):
-	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=local $(PACKER_VAR_FILES) packer/ova/packer.json
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=vsphere -except=local $(PACKER_VAR_FILES) -only=vmware-iso packer/ova/packer.json
 
 .PHONY: $(AMI_BUILD_TARGETS)
 $(AMI_BUILD_TARGETS):
@@ -200,6 +206,11 @@ build-azure-vhd-ubuntu-1804: ## Builds Ubuntu 18.04 VHD image for Azure
 build-do-default: ## Builds the DigitalOcean snapshot default image
 
 build-gce-default: ## Builds the GCE default image
+
+build-vsphere-ova-centos-7: ## Builds CentOS 7 template on vSphere
+build-vsphere-ova-photon-3: ## Builds Photon 3 template on vSphere
+build-vsphere-ova-ubuntu-1804: ## Builds Ubuntu 18.04 template on vSphere
+build-vsphere-ova-ubuntu-2004: ## Builds Ubuntu 20.04 template on vSphere
 
 build-ova-centos-7: ## Builds CentOS 7 OVA w local hypervisor
 build-ova-photon-3: ## Builds Photon 3 OVA w local hypervisor
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -33,6 +33,10 @@ import tarfile
 def main():
     parser = argparse.ArgumentParser(
         description="Builds an OVA using the artifacts from a Packer build")
+    parser.add_argument('--stream_vmdk',
+                        dest='stream_vmdk',
+                        action='store_true',
+                        help='Compress vmdk file')
     parser.add_argument('--vmx',
                         dest='vmx_version',
                         default='13',
@@ -72,7 +76,7 @@ def main():
     # Get the first build.
     build = data['builds'][0]
     build_data = build['custom_data']
-    print("image-build-ova: loaded %s-kube-%s" % (build['name'],
+    print("image-build-ova: loaded %s-kube-%s" % (build_data['build_name'],
                                                   build_data['kubernetes_semver']))
 
     if args.vmdk_file is None:
@@ -82,7 +86,12 @@ def main():
         vmdk_files = [{"name": args.vmdk_file, "size": os.path.getsize(args.vmdk_file)}]
 
     # Create stream-optimized versions of the VMDK files.
-    stream_optimize_vmdk_files(vmdk_files)
+    if args.stream_vmdk is True:
+        stream_optimize_vmdk_files(vmdk_files)
+    else:
+        for f in vmdk_files:
+            f['stream_name'] = f['name']
+            f['stream_size'] = os.path.getsize(f['name'])
 
     # TODO(akutz) Support multiple VMDK files in the OVF/OVA
     vmdk = vmdk_files[0]
@@ -97,7 +106,7 @@ def main():
                  "ubuntu-64": {"id": "94", "version": "", "type": "ubuntu-64"}}
 
     # Create the OVF file.
-    ovf = "%s-kube-%s.ovf" % (build['name'], build_data['kubernetes_semver'])
+    ovf = "%s-kube-%s.ovf" % (build_data['build_name'], build_data['kubernetes_semver'])
     create_ovf(ovf, {
         'BUILD_DATE': build_data['build_date'],
         'ARTIFACT_ID': build['artifact_id'],
@@ -119,11 +128,11 @@ def main():
     })
 
     # Create the OVA manifest.
-    ova_manifest = "%s-kube-%s.mf" % (build['name'], build_data['kubernetes_semver'])
+    ova_manifest = "%s-kube-%s.mf" % (build_data['build_name'], build_data['kubernetes_semver'])
     create_ova_manifest(ova_manifest, [ovf, vmdk['stream_name']])
 
     # Create the OVA.
-    ova = "%s-kube-%s.ova" % (build['name'], build_data['kubernetes_semver'])
+    ova = "%s-kube-%s.ova" % (build_data['build_name'], build_data['kubernetes_semver'])
     create_ova(ova, [ovf, ova_manifest, vmdk['stream_name']])
 
 
--- a/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
+++ b/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
@@ -98,7 +98,7 @@ d-i apt-setup/use_mirror boolean false
 d-i apt-setup/services-select multiselect none
 
 # Customize the list of packages installed.
-d-i pkgsel/include string openssh-server
+d-i pkgsel/include string openssh-server open-vm-tools unattended-upgrades
 
 
 # Ensure questions about these packages do not bother the installer.
--- a/images/capi/packer/ova/ova-ubuntu-1804.json
+++ b/images/capi/packer/ova/ova-ubuntu-1804.json
@@ -3,6 +3,7 @@
   "distro_name": "ubuntu",
   "os_display_name": "Ubuntu 18.04",
   "guest_os_type": "ubuntu-64",
+  "vsphere_guest_os_type": "ubuntu64Guest",
   "iso_url": "http://old-releases.ubuntu.com/releases/18.04.2/ubuntu-18.04.2-server-amd64.iso",
   "iso_checksum": "a2cb36dc010d98ad9253ea5ad5a07fd6b409e3412c48f1860536970b073c98f5",
   "iso_checksum_type": "sha256",
--- a/images/capi/packer/ova/packer.json
+++ b/images/capi/packer/ova/packer.json
@@ -5,20 +5,28 @@
     "boot_wait": "10s",
     "build_timestamp": "{{timestamp}}",
     "build_version": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}",
+    "cluster": "",
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "convert_to_template": "false",
+    "datastore": "",
     "disable_public_repos": "false",
     "disk_type_id": "0",
+    "disk_controller_type": "pvscsi",
+    "disk_thin_provisioned": "true",
     "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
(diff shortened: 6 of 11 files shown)


docs/book/src/capi/providers/vsphere-api.md
```diff
@@ -0,0 +1,44 @@
...
+## Prerequisites
+
+The `make deps-vsphere` target will test that Ansible and Packer are installed and available. If they are not, they will be installed to `images/capi/.bin`. This directory will need to be added to your `$PATH`.
+
+### Prerequisites for all images
+
+- [Packer](https://www.packer.io/intro/getting-started/install.html) version >= 1.5.4
+- [Ansible](http://docs.ansible.com/ansible/latest/intro_installation.html) version >= 2.8.0
+
+### Prerequisites for DigitalOcean
```
detiber: Copy pasta? Should this be vSphere instead?

docs/book/src/capi/providers/vsphere-api.md
```diff
@@ -0,0 +1,44 @@
+# Building Images for DigitalOcean
```
detiber: Copy pasta? Should this be vSphere instead?

marema31: Totally, sorry, I'll correct this, I was wondering If the term vSphere template was not more appropriate than vSphere-api. I should use this terminology instead no ?

codenrhoden:
I apologize that I haven't reviewed this yet. It's on my radar, and I will definitely get to it in the next couple of days!

marema31:
No problem

images/capi/ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -24,6 +24,7 @@
   loop: "{{ extra_repos.split() }}"
   when: remove_extra_repos and extra_repos != ""
 
+
```
codenrhoden, line 27: Please remove this extra line.

images/capi/packer/vsphere/linux/centos/http/7/ks.cfg
```diff
@@ -0,0 +1,92 @@
+# Copyright 2019 The Kubernetes Authors.
```
codenrhoden, line 1: unless this kicstart (and all the others) are different from what's already found in `capi/packer/ova`, I'd rather just symlink to them instead of duplicating them. Symlinks to work in this scenario, and you can see an example of this in a currently unmerged PR: https://github.com/kubernetes-sigs/image-builder/pull/129/files#r369790967

images/capi/packer/vsphere/linux/ubuntu/http/18.04/preseed.cfg
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 The Kubernetes Authors.
```
codenrhoden, line 1: Same applies to the preseeds. I don't think they need to be duplicated here.

images/capi/packer/vsphere/packer.json
```diff
@@ -0,0 +1,121 @@
+{
```
codenrhoden, line 1: After looking through the docs on the the `vsphere-iso` builder and reviewing this PR, I really think that the `vsphere-iso` builder can be listed as a second builder within the existing `packer/ova/packer.json` file rather than being a wholly second build type. That eliminates the duplication of the preseeds, kickstarts, and all the separate OS config. That can/should all be common.

Furthermore, the end-goal for this build is to produce an **OVA**. we need to keep that. To that end, we do need the `export` field to be filled out so that we get the files out of vCenter to create the OVA. We need to do the same post-processing on the OVF and creation of the OVA that we do today in the `ova/*` folder. There's a lot more customization coming soon, so this is a really important step.

marema31: Ok, I will rework my PR in this direction, I will have to add a post-processor to do the export of the OVA from the vsphere server.

My thought was to adress organization like tho ones I have worked on that will be more suspicious to download an OVA and push to to production infrastructure than using automation tools with source code available.

Just for your information, the vmware-iso builder will not be able to work anymore with vSphere 6.7+ due to removal of VNC capabilities, only with VMware desktop product, we may remove the mention to this builder for ESXi.

codenrhoden: > I will have to add a post-processor to do the export of the OVA from the vsphere server.

I don't think this is the case. See https://packer.io/docs/builders/vsphere-iso.html#export-configuration

marema31: I didn't view this option since I've never add the use of it, Ok, I will do this in the next few days.

figo: > Just for your information, the vmware-iso builder will not be able to work anymore with 
> vSphere 6.7+ due to removal of VNC capabilities, only with VMware desktop product, we may > remove the mention to this builder for ESXi.
Thanks for mention it, we can deprecate the ESX option once the vSphere-iso option merge and stable.

images/capi/packer/vsphere/vsphere-ubuntu-1804.json
```diff
@@ -0,0 +1,12 @@
+{
+  "build_name": "ubuntu-1804",
+  "distro_name": "ubuntu",
+  "os_display_name": "Ubuntu 18.04",
+  "guest_os_type": "ubuntu64Guest",
+  "iso_urls": "http://cdimage.ubuntu.com/releases/18.04/release/ubuntu-18.04.4-server-amd64.iso",
+  "iso_checksum": "e2ecdace33c939527cbc9e8d23576381c493b071107207d2040af72595f8990b",
+  "iso_checksum_type": "sha256",
+  "boot_command_prefix": "<esc><wait><esc><wait><enter><wait>/install/vmlinuz auto console-setup/ask_detect=false console-setup/layoutcode=us console-setup/modelcode=pc105 debconf/frontend=noninteractive debian-installer=en_US fb=false initrd=/install/initrd.gz kbd-chooser/method=us keyboard-configuration/layout=USA keyboard-configuration/variant=USA locale=en_US netcfg/get_domain=local netcfg/get_hostname=localhost grub-installer/bootdev=/dev/sda preseed/url=",
+  "boot_command_suffix": "/18.04/preseed.cfg -- <wait><enter><wait>",
```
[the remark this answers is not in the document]
jiatongw: Make sense to me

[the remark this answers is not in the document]
marema31: I let like this

marema31:
I've merged the ova builders, I manage to export the VM, but I've trouble with the post processor, @codenrhoden could you give me some help to understand what must done by it.

Without the post-processor and I obtains (no vmx file):
`
output/ubuntu-1804-kube-v1.17.3/ubuntu-1804-kube-v1.17.3-disk-0.vmdk
output/ubuntu-1804-kube-v1.17.3/packer-manifest.json
output/ubuntu-1804-kube-v1.17.3/ubuntu-1804-kube-v1.17.3.mf
output/ubuntu-1804-kube-v1.17.3/ubuntu-1804-kube-v1.17.3.ovf
`

I've updated the documentation but not merged it for the moment,  what is your preference ?

For the moment I've tested only with ubuntu 18.04, I will test the others when the post-processor will be functionnal.

images/capi/packer/vsphere/vsphere-ubuntu-2004.json
```diff
@@ -0,0 +1,14 @@
...
+  "distro_name": "ubuntu",
+  "os_display_name": "Ubuntu 20.04",
+  "guest_os_type": "ubuntu64Guest",
+  "iso_urls": "http://cdimage.ubuntu.com/ubuntu-server/daily/20200107/focal-server-amd64.iso",
+  "iso_checksum": "4901eeba3ffaa56b9356bb496d161ade5bd1e2df579552aa6c40f33984a34614",
+  "iso_checksum_type": "sha256",
+  "ssh_username": "ubuntu",
+  "ssh_password": "ubuntu",
+  "boot_command_prefix": "<esc><wait><esc><wait><enter><wait>/install/vmlinuz auto console-setup/ask_detect=false console-setup/layoutcode=us console-setup/modelcode=pc105 debconf/frontend=noninteractive debian-installer=en_US fb=false initrd=/install/initrd.gz kbd-chooser/method=us keyboard-configuration/layout=USA keyboard-configuration/variant=USA locale=en_US netcfg/get_domain=local netcfg/get_hostname=localhost grub-installer/bootdev=/dev/sda preseed/url=",
+  "boot_command_suffix": "/20.04/preseed.cfg -- <wait><enter><wait>",
```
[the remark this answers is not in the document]
marema31: I understood by the previous conversation that it is not necessary to do so

jiatongw:
@marema31 I think we expect a single OVA file other than multiple files (OVF + mf + vmdk)

marema31:
OVA is only a tar of a folder with this three files, my concern was more on the vmx file (that in my understanding is not needed on vSphere) that the python script used as post processor modify and use to generate the ovf that I already have.

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
```diff
@@ -98,7 +98,7 @@ d-i apt-setup/use_mirror boolean false
 d-i apt-setup/services-select multiselect none
 
 # Customize the list of packages installed.
-d-i pkgsel/include string openssh-server
+d-i pkgsel/include string openssh-server open-vm-tools unattended-upgrades
```
codenrhoden, line 101: This change should be removed.  We install `open-vm-tools` via Ansible (https://github.com/kubernetes-sigs/image-builder/blob/2fb8b85218bd9ba1b5d42126b6aada101b1cbd87/images/capi/ansible/roles/common/defaults/main.yml#L42), and we explictly **disable** `unattended-upgrades` (https://github.com/kubernetes-sigs/image-builder/blob/2fb8b85218bd9ba1b5d42126b6aada101b1cbd87/images/capi/ansible/roles/sysprep/tasks/debian.yml#L64-L74). All images created by this tool are meant to be immutable. FWIW, `unattended-upgrades` is installed by default in Ubuntu, but then we have to disable it.

images/capi/packer/ova/ova-ubuntu-1804.json
```diff
@@ -3,8 +3,9 @@
   "distro_name": "ubuntu",
   "os_display_name": "Ubuntu 18.04",
   "guest_os_type": "ubuntu-64",
-  "iso_url": "http://old-releases.ubuntu.com/releases/18.04.2/ubuntu-18.04.2-server-amd64.iso",
-  "iso_checksum": "a2cb36dc010d98ad9253ea5ad5a07fd6b409e3412c48f1860536970b073c98f5",
+  "vsphere_guest_os_type": "ubuntu64Guest",
+  "iso_url": "http://cdimage.ubuntu.com/releases/18.04/release/ubuntu-18.04.4-server-amd64.iso",
```
codenrhoden, line 7: Oh nice - you found a `server` image for 18.04.4 -- we had only been finding `live` versions. Regardless, I would ask for this to be a different commit or different PR. This isn't relevant to adding the new builder.

images/capi/packer/ova/packer.json
```diff
@@ -2,23 +2,32 @@
...
     "ansible_extra_vars": "guestinfo_datasource_slug={{user `guestinfo_datasource_slug`}} guestinfo_datasource_ref={{user `guestinfo_datasource_ref`}} ",
+    "boot_order": "disk,cdrom",
     "boot_wait": "10s",
     "build_timestamp": "{{timestamp}}",
     "build_version": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}",
+    "cluster": "",
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "convert_to_template": "true",
```
codenrhoden, line 13: I'm still not sure I would want to see this by default. I'll have to think about it. The end result of the run should be an OVA ready for import -- not a template within vCenter. The reason being that the OVA metadata is important.

images/capi/packer/ova/packer.json
```diff
@@ -47,15 +57,18 @@
     "skip_compaction": "false",
     "ssh_username": "builder",
     "ssh_password": "builder",
+    "vm_version": "13",
```
codenrhoden, line 60: I think the existing `vmx_version` can be reused here. I would never see overriding one but not the other.

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
marema31: For open-vm-tools, if they are not present, the ESXi does not have the Ip of the vm spinned, and therefore packer is not able to communicate with it to do the provisionner part.

For the unattended-upgrade, I'm ok to remove it, but I come to an issue that the ansible failed to disable it since it was not installed and the build failed

images/capi/packer/ova/packer.json
marema31: The OVA will contain a template and not a VM, for the moment packer provisionner is not able to remove the artifact created in vsphere. If I find time for this I may propose a patch to packer to add this option to the builder, but for now running this will ends to leave something in the hypervisor and I was thinking it was more clear to tag it as a template

images/capi/packer/ova/ova-ubuntu-1804.json
marema31: I will do so, I put the server version because I was not able to make the preseed works in vsphere with the version you are currently using

images/capi/packer/ova/packer.json
marema31: Sure, I will use it. 

For the post-processor, I will have a closer look to what it is doing. At first sight, I was puzzled by the fact it was working with a vmx file that the vsphere-iso does not create, I will see with the vExpert at dayjob if it is needed, but for my understanding it is automatically created from the information in the .ovf file contained within the ova while it is imported.

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
codenrhoden: interesting, re: `open-vm-tools`, as we haven't run into this issue before.

For `unattended-upgrades`, I'm wondering if this is a diff between using the 18.04.2 image are currently using and your change to 18.04.4. I take that as an additional reason to not make that change at the same time as this.

images/capi/packer/ova/ova-ubuntu-1804.json
codenrhoden: That's concerning. If it boots find with Desktop hypervisors and remote ESXi, it should work for this scenario too... 🤔

images/capi/packer/ova/packer.json
codenrhoden: Okay. I'll have to take this for a spin myself soon and see how things look.

images/capi/packer/ova/packer.json
codenrhoden: Can you clarify what you are looking for here? I don't think there is anything right now that is interacting with the VMX file.

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
marema31: For open-vm-tools, I'm suppose the difference is due to the way vmare-iso on Esxi and vsphere-iso retrieve the information. For vmware-iso, you have to configure a NetHack on the Esxi to allow the command-line used by ssh directly on the Esxi to retrieve the network configuration of the VM. For vsphere-iso, no such hack are necessary since it use the vSphere API, but for vSphere API to be able to provide the IP, the information must be retrieved on the vm OS by the guest tools (vmware-tools or open-vm-tools). That will explain why you didn't need the open-vm-tools during the build phase with vmware-iso.

For unattended-upgrade, I suppose this is more due to change from desktop to server version of ubuntu. I  agree that it should be added here, but the ansible part should not failed if a service it tries to disable is not installed. Maybe the best options is to make ansible ensure that the unattended-upgrade is not install at all.

images/capi/packer/ova/ova-ubuntu-1804.json
marema31: I will give again a try with 18.04.02.

I've think of openning a different PR for bumping live to server verstion of ubuntu, but I will not be able to test this since I do not have vmware desktop tool neither access to a Esxi that I could access via ssh.

images/capi/packer/ova/packer.json
marema31: You're right, I don't why I was thinking it was.

I've adjusted the post-processor to accommodate the output of vsphere-iso builder export and make optionnal the vmdk compressing that is already done by export, removing the dependencies to vmware-diskmanager.

images/capi/packer/ova/ova-ubuntu-1804.json
marema31: I've given an other try for 18.04.02, and it worked, the error I had was probably due to others modifications needed

marema31:
I've merged the documentation and done some modification in it to give precision on difference between esxi and vsphere builder

codenrhoden:
Thanks @marema31 ! I'll try to get this re-reviewed (and take it for a real spin with vCenter) over the weekend or on Monday. I think you've already figured this out, but the only thing I wanted to say based on your previous comments is that we are already using a server version of Ubuntu (not desktop). What we hadn't been able to do is find a URL for the server version of 18.04.04 instead of the "live" version.

codenrhoden:
Sorry, this is still on my to do list. Thank you for your patience!

marema31:
no problem

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
codenrhoden: Thanks for the explanation re: `open-vm-tools`. I am able to see what you are talking about. Please add `open-vm-tools` to the centos kickstart file, as I tested that build and it doesn't work without it either.

As a side note, the "minimal" CentOS image we use doesn't contain `open-vm-tools`, so the install doesn't work unless we switch to the full-size DVD. I'm going to keep the default ISO as it is now, and then document that if you want to build CentOS on vSphere, you need to use the full ISO.

As for Ubuntu, I'm still testing things out. Right now when I try to build Ubuntu on vSphere, the installation is weird (changes to a weird LOCALE and gets stuck). so I have been able to build that one successfully. It's really strange. I'm trying to build by adding `open-vm-tools` like you have, but without adding `unattended-upgrades` to see if shows the issue you have seen.

docs/book/src/capi/capi.md
```diff
@@ -14,7 +14,8 @@ The Image Builder can be used to build images intended for use with Kubernetes [
 * [DigitalOcean](./providers/digitalocean.md)
 * Google *TODO*
 * [OpenStack](./providers/openstack.md)
-* [vSphere](./providers/vsphere.md)
+* [vSphere OVA](./providers/vsphere.md)
+* [vSphere Template](./providers/vsphere-template.md)
```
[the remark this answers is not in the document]
marema31: Reverted

docs/book/src/capi/providers/vsphere.md
```diff
@@ -8,31 +8,44 @@ The `make deps-ova` target will test that Ansible and Packer are installed and a
...
 
-| OS | Builder |
-|----|---------|
-| Linux | VMware Workstation |
-| macOS | VMware Fusion |
+| OS | Builder | Build target |
+|----|---------|--------------|
+| Linux | VMware Workstation | build-ova-<OS> |
+| macOS | VMware Fusion | build-ova-<OS> |
+| esxi | Esxi | build-esx-ova-<OS> |
```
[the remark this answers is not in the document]
marema31: ok

docs/book/src/capi/providers/vsphere.md
```diff
@@ -93,6 +107,7 @@ After the images are built, the VMs from they are built are prepped for local te
 #### Accessing Remote VMs
 
 After deploying an image to vSphere, use `hack/image-govc-cloudinit.sh VM` to snapshot the image and update it with cloud-init data from the `cloudinit` directory. Start the VM and now it may be accessed with `ssh -i cloudinit/id_rsa.capi capv@VM_IP`.
+This hack necessitate the `govc` utility from [VMare](https://github.com/vmware/govmomi/tree/master/govc)
```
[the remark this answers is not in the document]
marema31: corrected

images/capi/Makefile
```diff
@@ -146,13 +148,17 @@ AZURE_BUILD_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_NAMES))
 DO_BUILD_TARGETS 	:= $(addprefix build-,$(DO_BUILD_NAMES))
 QEMU_BUILD_TARGETS	:= $(addprefix build-,$(QEMU_BUILD_NAMES))
 
+.PHONY: $(VSPHERE_BUILD_TARGETS)
+$(VSPHERE_BUILD_TARGETS):
+	packer build $(PACKER_FLAGS) -var-file="$(abspath packer/ova/$(subst build-vsphere-,,$@).json)" $(PACKER_VAR_FILES) -except=esx -except=local -only=vsphere -force packer/ova/packer.json
```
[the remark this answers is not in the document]
marema31: Since the template is not deleted on the vSphere after the export of the OVA, if I remove the -force, the build will fail because the destination template is already there and force the user to connect to vSphere to delete it. Should I really remove it ?

Added the --var-file

Ok, It's also Why I  have tried to explain the PACKER_VAR_FILES in the documentation

> If you prefer to use a different configuration file, you can create it with the same format and export PACKER_VAR_FILES environment variable containing the full path to it.

images/capi/packer/ova/packer.json
```diff
@@ -2,23 +2,32 @@
   "variables": {
     "ansible_common_vars": "",
     "ansible_extra_vars": "guestinfo_datasource_slug={{user `guestinfo_datasource_slug`}} guestinfo_datasource_ref={{user `guestinfo_datasource_ref`}} ",
+    "boot_order": "disk,cdrom",
```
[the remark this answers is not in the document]
marema31: removed

images/capi/packer/ova/packer.json
```diff
@@ -48,14 +58,16 @@
...
+    "vcenter_server": "",
     "vnc_bind_address": "127.0.0.1",
     "vnc_disable_password": "false",
     "vnc_port_min": "5900",
     "vnc_port_max": "6000"
   },
   "builders": [
     {
-      "name": "{{user `build_name`}}",
+      "name": "ova",
```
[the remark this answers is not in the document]
marema31: ok

images/capi/packer/ova/packer.json
```diff
@@ -98,6 +110,58 @@
...
+      "host":     "{{user `host`}}",
+      "convert_to_template": "true",
+      "cluster": "{{user `cluster`}}",
+      "network": "{{user `network`}}",
+
+
+
+      "vm_version": "{{user `vmx_version`}}",
+      "CPUs": 1,
+      "boot_order": "disk,cdrom",
```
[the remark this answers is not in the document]
marema31: removed

images/capi/packer/ova/packer.json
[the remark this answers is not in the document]
marema31: Also adjust the makefile for take this new name in account

images/capi/hack/image-build-ova.py
```diff
@@ -33,6 +33,10 @@
 def main():
     parser = argparse.ArgumentParser(
         description="Builds an OVA using the artifacts from a Packer build")
+    parser.add_argument('--stream_vmdk',
+                        dest='stream_vmdk',
+                        default=False,
```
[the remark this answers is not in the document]
marema31: ok

images/capi/packer/ova/packer.json
frederick-newtz: I agree for sure that the end result should be an OVA.  However, it is definitely useful to us to have the ability to create and leave a template in vCenter.  I realize that might not be the objective of this project, and the convert_to_template should be false by default, but having the ability to manually turn that option to true would be prudent in my opinion.

images/capi/packer/ova/vsphere.json
```diff
@@ -0,0 +1,9 @@
+{
+    "vcenter_server":"",
+    "username":"",
+    "password":"",
+    "datastore":"",
+    "folder": "",
+    "cluster": "",
+    "network": ""
+}
```
MaxRink, line 9:
```suggestion
   "insecure_connection": "false",
}
```
I would add this variable in here, as its basically an cluster-dependent flag, that needs to be set on most clusters

images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg
codenrhoden: @marema31 I'm still having issues with Ubuntu on my side -- would you be willing to test/check if everything works okay now if `unattended-upgrades` is removed? If so, I'd still like to not see it added. Since we are using the same ISO, I feel like it should be fine.

images/capi/packer/ova/packer.json
codenrhoden: Yeah, I can see where this would be useful. People building for vSphere may not even import the OVA -- they might just use this template instead. @marema31 can you change this default to false?

images/capi/packer/ova/vsphere.json
codenrhoden: I'm fine either way. As I mentioned in an un-related comment, I'm considering removing the `esx.json` and would also remove this `vsphere.json` file as well. It encourages people to fill in the blanks and then you have a dirty repo when you build. These JSON files should not be modified, but then they also serve no purpose in their current form. I'll probably rename this as examples and refer to them as such instead.

images/capi/packer/ova/packer.json
marema31: Just to be clear, either the value of this, the results will stay on the vSphere, but as a template or a vm. In each case, you will obtain an OVA from it.

marema31: No problem, I will put in false by default and add it in the vsphere.json

images/capi/packer/ova/vsphere.json
marema31: It's why I add it in the documentation of the builder, I'm in favor to make them example files of what can pointed by the PACKER_VAR_FILES

marema31:
I've taken the comments in account, I still have one question that I would like to be clarified: https://github.com/kubernetes-sigs/image-builder/pull/186#discussion_r410640472 

I'll test without unattended_upgrades right now and come back to you.

images/capi/Makefile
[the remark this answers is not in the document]
codenrhoden: Especially considering the use case that that people may deploy VMs from the resulting template, I think having that template overwritten so easily is a bit dangerous. One wrong `make` call and your perfect good and deployed template is destroyed (but hopefully, easily recreated).

So I still think removing it is a good idea. Thoughts?

codenrhoden:
> I've taken the comments in account, I still have one question that I would like to be clarified: [#186 (comment)](https://github.com/kubernetes-sigs/image-builder/pull/186#discussion_r410640472)
> 
> I'll test without unattended_upgrades right now and come back to you.

Thanks -- I just tried again for Ubuntu and it doesn't work in my setup. I really can't tell -- when I open the web console, it gets to the screen about choosing an installation language and then bounces around like crazy for several minutes before picking Cyrillic - then it gets stuck on some prompt. Obviously none of this should be happening because of the preseed. I'm stumped. May be my version of vCenter or something (I may be using an internal build or something, I'll have to look).

Once get the last piece in (re: `unattended-upgrades`) I'm ready to merge this in. My final ask is that you squash your commits down to one.

marema31:
I'm testing with a vSphere 6.5u3 instance.

I've tested without unnattended-upgrades, and I have this : 

vsphere: failed: [default] (item=unattended-upgrades) => {"ansible_loop_var": "item", "changed": false, "item": "unattended-upgrades", "msg": "Could not find the requested service unattended-upgrades: host"}

I only can suppose that the addition of open-vm-tools reject the automatic installation of unnattended-upgrades.

I propose you to let it for the first iteration to let you do your work, I'll have a look if I can make the ansible playbook smarter on this.

codenrhoden:
> I propose you to let it for the first iteration to let you do your work, I'll have a look if I can make the ansible playbook smarter on this.

I agree -- I'll have to get this sorted out before the next tag, and I'm sure I can figure out what's going on. But I'd rather merge this and unblock myself. Thank you very much for this effort!

/lgtm
/approve

MaxRink:
I found one thing that we all overlooked so far
https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/hack/ensure-packer.sh
this installs packer 1.4.1, at least 1.5.2 is required for vsphere-iso. So i think this should be bumped.

frederick-newtz:
The issue is with the boot_command option, Ubuntu 18.04.4 has a different keystroke combination that 18.04.2.  I had already implemented the vsphere iso builder and ran into the same issue when I attempted to use the 18.04.4 image instead of the 18.04.2 image.  I am happy to do a pull request if it will help you out.  Pretty sure the Packer project has a valid boot_command for 18.04.4 somewhere.

marema31:
Good catch @MaxRink , I forgot that packer was already available on my system, I'll do a bump PR for this

@frederick-newtz, I have tried with the 18.04.2 and 18.04.4 without changing the boot_command, but I agrees that during my first tests I had lots a strange issues of timing for the boot_command with 18.04.2.

Here an example of working packer for 18.04.4
`        "iso_urls": "http://cdimage.ubuntu.com/releases/18.04/release/ubuntu-18.04.4-server-amd64.iso",
        "iso_checksum": "e2ecdace33c939527cbc9e8d23576381c493b071107207d2040af72595f8990b",
        "boot_command": [
          "<enter><wait><f6><wait><esc><wait>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>",
          "<bs><bs><bs>",
          "/install/vmlinuz",
          " initrd=/install/initrd.gz",
          " priority=critical",
          " locale=en_US",
          " file=/media/preseed.cfg",
          "<enter>"
        ]`

frederick-newtz:
I can verify that boot_command works in my environment for 18.04.4.

codenrhoden:
Hmm, maybe I'll have to give that different boot command a try and bump to 18.04.4.  This kind of change between the two isn't very "LTS", is it? :)

What I find perplexing is that we did *not* change the ISO or the boot command. It's the same one that has been there for months now - and yet it doesn't work on vSphere for me. Very strange.

Issue #95, opened 2019-11-08
Should default images include NFS utilities


codenrhoden:
The current images install support for making NFS mounts. As brought up in #64, this does leave open ports that some users may find undesirable. Rather than explicitly disabling those ports, should it instead be left to a user to customize their image by adding NFS support if they want it?

detiber:
I'm okay with removing NFS support by default.

ncdc:
This has been open for a while - have we made any progress on a resolution?

codenrhoden:
yeah, I should pick this up again. Given the very little feedback received, I'll remove NFS by default. users can add it back if they like. That will also resolve the rpcbind issue (#64).

/assign

PR #187, opened 2020-03-22
Add nfs utils so NFS mounting works in photon3

voor:
This package was missing, so you're unable to do NFS mounts properly.  This PR just adds it.


codenrhoden:
/hold

NFS was removed by default on purpose as part of https://github.com/kubernetes-sigs/image-builder/issues/95. Anyone building their own images can install it via the `extra_rpms` or `extra_debs` flags.

voor:
Closing this to reduce clutter.

PR #188, opened 2020-03-22
Allow configuring containerd

voor:
This also paves the way for further containerd configuration, as it'll create a template that can be placed into the containerd default config location.


images/capi/packer/ami/packer.json
```diff
@@ -17,6 +17,7 @@
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "containerd_pause_image": "k8s.gcr.io/pause:3.1",
```
codenrhoden: Rather than setting this in `packer/ami/packer.json`, I'd rather set it in `packer/config/containerd.json`. Right now this would only be set for the AMI builder, but I'd rather it be global.

images/capi/packer/config/ansible-args.json
```diff
@@ -1,3 +1,3 @@
...
...etes_cni_http_source={{user `kubernetes_cni_http_source`}} kubernetes_cni_http_checksum={{user `kubernetes_cni_http_checksum`}} kubernetes_http_source={{user `kubernetes_http_source`}} kubernetes_container_registry={{user `kubernetes_container_registry`}} kubernetes_rpm_repo={{user `kubernetes_rpm_repo`}} kubernetes_rpm_gpg_key={{user `kubernetes_rpm_gpg_key`}} kubernetes_rpm_gpg_check={{user `kubernetes_rpm_gpg_check`}} kubernetes_deb_repo={{user `kubernetes_deb_repo`}} kubernetes_deb_gpg_key={{user `kubernetes_deb_gpg_key`}} kubernetes_cni_deb_version={{user `kubernetes_cni_deb_version`}} kubernetes_cni_rpm_version={{user `kubernetes_cni_rpm_version`}} kubernetes_cni_semver={{user `kubernetes_cni_semver`}} kubernetes_cni_source_type={{user `kubernetes_cni_source_type`}} kubernetes_semver={{user `kubernetes_semver`}} kubernetes_source_type={{user `kubernetes_source_type`}} kubernetes_deb_version={{user `kubernetes_deb_version`}} kubernetes_rpm_version={{user `kubernetes_rpm_version`}} no_proxy={{user `no_proxy`}} reenable_public_repos={{user `reenable_public_repos`}} remove_extra_repos={{user `remove_extra_repos`}} kubernetes_cni_http_source={{user `kubernetes_cni_http_source`}} "
```
codenrhoden: I know this long string is really hard to parse, but it looks like you were meaning to add `containerd_pause_image`, but instead duplicated `kubernetes_cni_http_source`.

codenrhoden:
LGTM. Please squash the commits into 1 and this can be merged.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -59,13 +59,27 @@
...
+    src: etc/containerd/config.toml
+
 - name: start containerd service
   systemd:
     name: containerd
     daemon_reload: yes
     enabled: True
     state: started
 
+# This step isn't really necessary since kubeadm will pull it again, but there's not much harm in making it explicit.
```
dims: let's please avoid this as it is likely to increase the footprint of the image etc.

images/capi/ansible/roles/common/defaults/main.yml
```diff
@@ -70,3 +70,5 @@ extra_repos: ""
 #photon does not have backward compatibility for legacy distro behavior for sysctl.conf by default
 #as it uses systemd-sysctl. set this var so we can use for sysctl conf file value.
 sysctl_conf_file: "{{ '/etc/sysctl.d/99-sysctl.conf' if ansible_os_family == 'VMware Photon OS' else '/etc/sysctl.conf' }}"
+
+containerd_pause_image: "k8s.gcr.io/pause:3.1"
```
dims: needs a new line?

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
```diff
@@ -0,0 +1,3 @@
+[plugins]
```
dims: Please use v2 of the config : https://github.com/containerd/cri/blob/master/docs/config.md

images/capi/ansible/roles/common/defaults/main.yml
dims: Also default is `pause:3.2`

images/capi/packer/config/containerd.json
```diff
@@ -1,4 +1,5 @@
 {
     "containerd_version": "1.3.3",
-    "containerd_sha256": "24ce7ad6b489fb25d07d2a3bb50e443fcce1ac3318f8cc0831e00668c2c9fd86"
+    "containerd_sha256": "24ce7ad6b489fb25d07d2a3bb50e443fcce1ac3318f8cc0831e00668c2c9fd86",
+    "containerd_pause_image": "k8s.gcr.io/pause:3.1"
```
dims: 3.2 please

images/capi/ansible/roles/containerd/tasks/main.yml
codenrhoden: I'm also worried that for installs that use a container side-load method:
https://github.com/kubernetes-sigs/image-builder/blob/0e039313333ec4d0ee6fef7080d70601375dec75/images/capi/ansible/roles/kubernetes/tasks/url.yml#L77-L79

This step will fail because we're never "pulling" images in that scenario.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -59,13 +59,27 @@
...
+- name: pull pause container
+  shell: "crictl pull {{ containerd_pause_image }}"
```
dims: Also if we really need to do this as something fails during testing, the right way is doc'ed here:
https://github.com/containerd/cri/blob/master/docs/crictl.md#directly-load-a-container-image

codenrhoden:
Thanks for being so responsive on this, @voor !

/lgtm
/approve

images/capi/packer/config/containerd.json
```diff
@@ -1,4 +1,5 @@
 {
     "containerd_version": "1.3.3",
-    "containerd_sha256": "24ce7ad6b489fb25d07d2a3bb50e443fcce1ac3318f8cc0831e00668c2c9fd86"
+    "containerd_sha256": "24ce7ad6b489fb25d07d2a3bb50e443fcce1ac3318f8cc0831e00668c2c9fd86",
+    "containerd_pause_image": "k8s.gcr.io/pause:3.2"
```
figo: @voor @dims @codenrhoden 
Sorry, late to this PR, the default pause version is still 3.1 since default kubernetes version is v1.16.2 at image-builder repo.

Upstream only upgraded to pause:3.2 on kubernetes 1.18, see this one:
https://github.com/kubernetes/kubernetes/blob/v1.18.0-rc.1/build/pause/Makefile#L21

kubernetes v1.17.4 still has pause:3.1
https://github.com/kubernetes/kubernetes/blob/v1.17.4/build/pause/Makefile#L23

figo: @dims confirmed pause should be guarantee to be backward compatible, then setting default to pause 3.2 is not an issue.

PR #193, opened 2020-03-31
Deploy haveged as part of base image to prevent low-entropy situations

dims:
Pick up some proactive recommendation from https://github.com/kubernetes/kubernetes/issues/60751


codenrhoden:
/hold

Hmm, this feels like one of those things were people would add it via `extra_rpms1` if they wanted it rather than having it in by default. I know i saw some things where this particular package is contentious versus others. Also, if we *did* want this, it would go in the Ansible script, not the kickstart. And also -- why just Photon? I would think all OS's.

dims:
good point, also @randomvariable had another thought as well. will request him to open a PR.

PR #195, opened 2020-04-01
Allows adding entire containerd configuration blocks

voor:
The earlier work for adding a way to change the pause image was nice, but I realize it's failing to capture all of the other things you can do to containerd inside of an image.  This rectifies that while still maintaining backwards compatibility for the pause image.

This allows adding new runtime engines and more extensive containerd configuration.

Here's an example of it in use for GPU enabled clusters: https://github.com/voor/kubernetes-gpu


images/capi/packer/config/ansible-args.json
```diff
@@ -1,3 +1,3 @@
...
...y={{user `http_proxy`}} https_proxy={{user `https_proxy`}} kubernetes_cni_http_source={{user `kubernetes_cni_http_source`}} kubernetes_cni_http_checksum={{user `kubernetes_cni_http_checksum`}} kubernetes_http_source={{user `kubernetes_http_source`}} kubernetes_container_registry={{user `kubernetes_container_registry`}} kubernetes_rpm_repo={{user `kubernetes_rpm_repo`}} kubernetes_rpm_gpg_key={{user `kubernetes_rpm_gpg_key`}} kubernetes_rpm_gpg_check={{user `kubernetes_rpm_gpg_check`}} kubernetes_deb_repo={{user `kubernetes_deb_repo`}} kubernetes_deb_gpg_key={{user `kubernetes_deb_gpg_key`}} kubernetes_cni_deb_version={{user `kubernetes_cni_deb_version`}} kubernetes_cni_rpm_version={{user `kubernetes_cni_rpm_version`}} kubernetes_cni_semver={{user `kubernetes_cni_semver`}} kubernetes_cni_source_type={{user `kubernetes_cni_source_type`}} kubernetes_semver={{user `kubernetes_semver`}} kubernetes_source_type={{user `kubernetes_source_type`}} kubernetes_deb_version={{user `kubernetes_deb_version`}} kubernetes_rpm_version={{user `kubernetes_rpm_version`}} no_proxy={{user `no_proxy`}} reenable_public_repos={{user `reenable_public_repos`}} remove_extra_repos={{user `remove_extra_repos`}} "
```
detiber, line 2: Was the addition of gpu_enabled intentional here?

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
```diff
@@ -5,3 +5,5 @@ version = 2
 [plugins]
   [plugins."io.containerd.grpc.v1.cri"]
     sandbox_image = "{{ containerd_pause_image }}"
+    
+{{containerd_additional_settings | b64decode}}
```
detiber, line 9: Will this properly handle the current setting of null or does applying the filter need to be conditionalized?

voor: That is an awesome observation, thankfully it does properly output null when run through the decode -- I just verified it on all 3 images.

images/capi/packer/config/ansible-args.json
voor: Oh nice catch, that was something else I was doing that leaked into here, I really hate this file and wish there was a better way to do this.

codenrhoden:
/lgtm
I think the only other thing I'd like to see is adding this new configuration possibility to the docs, here: https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/capi.md#customization

Right now, unless you look through the code, you don't know that the field needs to be b64 encoded.

images/capi/packer/config/ansible-args.json
codenrhoden: Oh I agree with you there -- I don't like having this long string either. I have a couple of ideas of how to get rid of it, but so far each depends on a pre-processor (i.e. a CLI that we provide args to) to dynamically generate our `packer.json` file.

voor:
Added documentation (since that's a good suggestion), also retroactively adds documentation from another PR about containerd that went through without it.

codenrhoden:
Wow, fast turn around! Thanks @voor 
/approve
I'll let @detiber have one more look for an `/lgtm`.

detiber:
lgtm, but looks like this might need a rebase

voor:
Please make sure `kubernetes_load_additional_imgs={{user `kubernetes_load_additional_imgs`}}` is still there, that was what caused the merge conflict.

voor:
Upgrade of containerd cause a merge conflict, that has been resolved.

PR #102, opened 2019-11-26
Add tutorial to the documentation gitbook for building/deploying images on vSphere and AWS

moondev:
This tutorial outlines building a machine image, as well as details for how to reference the resulting vSphere/AWS image artifact inside the cluster manifests.

I placed the markdown in `docs/book/src/tutorials` and updated the menu tree in `SUMMARY.md`. If there  is a better place for this please share!

The tutorial directs the reader to use commit `1f9002a`, ideally this should be changed to a tagged release when ready.   Thanks!

PR #198, opened 2020-04-03
WIP: Add build/deploy tutorial to gitbook

codenrhoden:
This PR is replacing #102, originally from @moondev.

It splits his work into pages, and updates minor details to avoid the modification of JSON files directly.

It still needs a bit of work, though, as it's using a v1alpha2 version of CAPI. I think it should be v1alpha3 now. I'm not sure if we want to reference those manifests directly here, or point to something in the CAPV/CAPA documentation. Thoughts on that @detiber 

/cc @moondev

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -5,4 +5,7 @@
   - [AWS](./capi/providers/aws.md)
   - [Azure](./capi/providers/azure.md)
   - [vSphere](./capi/providers/vsphere.md)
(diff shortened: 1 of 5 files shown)


codenrhoden:
I honestly think most of this info is covered in the docs already. The only stuff that isn't is making your own CAPI manifests, which I don't think is something we should cover. CAPI is under active development, and the information changes. The API revision changes and becomes out of data (v1alpha2 vs v1alpha2, for example).

/close

PR #199, opened 2020-04-04
Disable Docker and containerd services by default

### LINKED ISSUE #128: Docker is enabled on boot
At the moment, images have Docker installed and enabled by default. This is not a good idea in cases where different Docker version will be installed or some other container runtime like conainerd.

Would be a good idea to either:
1. Disable Docker auto-start on boot (will be enabled by Kops later)
2. Just copy Docker packages in the cache dir and expect Kops to install what it needs.

/cc @justinsb

hakman:
In some cases, Kops NodeUp needs to run before Docker or containerd. See https://github.com/kubernetes/kops/issues/8843.

This PR disables the Docker and containerd services by default and also stores the .deb files in the location used by Kops NodeUp for caching. This way the packages are not downloaded again when NodeUp runs.

Fixes #128


hakman:
Closing in favour of https://github.com/kubernetes-sigs/image-builder/pull/273.
/close

PR #202, opened 2020-04-13
Initial builder using containers

justinsb:
First steps towards a suite of tools for building cloud images by
reusing the container-building infrastructure.


justinsb:
Unmarking as WIP, as it seems to work reasonable well now and is tucked away in the kube-deploy directory (for now, until we can get those scripts into @moshloop 's CLI).  cc @codenrhoden and @moshloop  ... and you were right - terraform is easier than bash for the AWS "helper" instance, as well as for test instances.

codenrhoden:
/lgtm

No objections raised in today's office-hours!

Issue #197, opened 2020-04-03
Take over building of HAProxy OVA from CAPV


codenrhoden:
Just putting this here as a breadcrumb in case anyone wants to chime in on it:

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/884

codenrhoden:
Judging from reaction from the CAPV issue, and no objections raised in the community meeting today, I'm going to take this issue and implement it.

/assign

PR #203, opened 2020-04-13
Clarify what belongs to the K8s node images

codenrhoden:
With the coming inclusion of the HAProxy image that is needed by CAPV (see #197), it makes sense to clarify in the code what is "node" related. This patch renames a bunch of Ansible related roles (and playbook) to clearly call these out as "node".

Specifically for OVAs (which are the only ones right now that will be capable of building an image other than a "node" image), we clarify the `make` targets to list "node" specifically (with "haproxy" targets coming in a later PR).

We also make a more standard pattern for the OVAs to clarify how they are built using a local vs ESX hypervisor. This will make more sense when a "vcenter" target is added in #186 .

I was preparing one big patch to add the HAProxy stuff, but I kept finding more things to cleanup and change, so thought it make sense to put this up as a separate PR. The HAProxy stuff depends on this, but that may not get filed for a couple more days still and this can go up for review now.

/assign @detiber @figo 
/cc @dims


detiber:
/lgtm
/hold for additional reviews

feel free to remove the hold whenever

codenrhoden:
I'm going to keep holding on this one until after I get #186 merged in (which I'm testing now), in order to reduce the re-base burden on that contributor. :)

codenrhoden:
/hold cancel
@detiber Okay, now that the vSphere builder is merged in, this is ready for re-review!

PR #207, opened 2020-04-16
Add a flag to conditionally load additional kubernetes images when source type != pkg

### LINKED ISSUE #206: [capi] Building images using the k8s ci artifacts fails
The list of `kubernetes_imgs` that we have in `images/capi/ansible/roles/kubernetes/defaults.main.yml` includes `pause.tar`, `coredns.tar`, and `etcd.tar`, but they are not included in the k8s ci artifacts (since they are built, versioned and distributed separately).

If we want to automate loading of these images, we should likely use a separate variables for their source, type, versions rather than overloading the use of `kubernetes_http_source`, `kubernetes_semver`, and `kubernetes_source_type`

detiber:
Fixes #206 

@codenrhoden I wasn't entirely sure which behavior would be the most desired by default, so I defaulted to assuming support for the upstream k8s sources, which do not include pause, etcd, and coredns images by default.

This would require any current consumers that expect the old behavior to override the packer variable `kubernetes_load_additional_imgs` to retain that behavior.

Happy to switch up the logic if that makes more sense, though.


codenrhoden:
@detiber I'm fine with the default as you have it. I'll just call it out in the release notes on the next tag.
Love your branch name 🤣 

Changes LGTM. I notice you called it out as a WIP, but I'm guessing that was just because of your question re: default behavior.

/lgtm
/approve

I'd say just remove the WIP prefix when you are ready!

detiber:
@codenrhoden I still need to finish testing the changes, should be able to finish that up shortly.

detiber:
Ok, just wrapped up testing and it appears to be working as intended :)

Issue #174, opened 2020-03-12
Photon OVA build doesn't work with Packer 1.5.4


codenrhoden:
I'm mostly putting this here as a note to myself to dig in deeper, but after using this project weekly for well over a year, I "accidentally" upgraded my local copy of Packer through brew, and I noticed that Photon builds no longer worked consistently. In fact it didn't work >90% of the time.

When downgrading to 1.4.1 (the version installed with `make deps-ova`), things worked fine.

I was able to see why, but did not find the code change in Packer that did it.

When Photon boots and installs from an ISO, it request an IP via DHCP. I can see in the lease file (`/var/db/vmware/vmnet-dhcpd-vmnet8.leases`) that this lease has a `uid` and a `client-hostname` associated with it. I see that when Packer looks for the IP address of the VM, it grabs the first IP address in the file with the matching MAC address.

The problem arises from the fact that once the OS install completes and the VM is rebooted, it again makes a DHCP request. This as a different `uid` in it and no `client-hostname`, so the VM gets a different IP address than the one when it was doing the OS install. Packer never re-parses the lease file, and it goes into an endless loop (well, with a **4 hour** timeout) trying to reach the VM on the incorrect IP.

When using an older version of Packer, it appears to re-parse the leases file before every attempt, and it actually picks up the most recent IP address for the expected MAC address (likely by taking the last entry in the file - but again, I haven't examined this code in Packer).

Just looking at the Packer log output (e.g. `PACKER_LOG=1`) I can see that the way it tries to reach the node is very different between the two versions, and I suspect this is very much a 1.4.x vs 1.5.x thing.

This requires more digging!

hridayeshpawar:
Hey man I was banging my head with this problem since last night. You are right this seems to be an issue with Packer 1.5.4. 
I tried with Packer 1.4.4 and it works fine.
In version 1.5.4, with each ssh failure it is trying to reconnect with the same IP address.
But, In 1.4.4 with each failure it reparses /var/db/vmware/vmnet-dhcpd-vmnet8.leases  and I think picks the latest entry before trying to connect via ssh.

codenrhoden:
This does appear to be fixed with Packer 1.6.0. After first I didn't think it was, because the time it takes to pick up the updated IP address varies (see comments in above referenced Packer issues), but it does eventually work.

For that reason, I think we should bump the default Packer version in our docs and install script to 1.6.0.

PR #210, opened 2020-04-21
For vsphere-iso the minimum packer version is 1.5.2

marema31:
Reported by @MaxRink  in #186 , I've forgot to bump the packer version in the ensure-packer.sh script at the same level than documentation.
See:
https://github.com/kubernetes-sigs/image-builder/pull/186#issuecomment-616807402

### DIFF
--- a/images/capi/hack/ensure-packer.sh
+++ b/images/capi/hack/ensure-packer.sh
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


codenrhoden:
My only hesitation with this is that local Fusion builds don't work with 1.5.x (#174). So either way, a user has to know what they are building, and use an appropriate version of Packer to get everything to work.
Given that I expect more people will use the vsphere-iso builder over time, I'm okay with this.

/approve

Issue #191, opened 2020-03-30
Prepare audit log with appropriate permissions


randomvariable:
To enable auditing in Kubernetes, ideally would have this go to `/var/log/kubernetes/audit/audit*.log`

This would be done by setting the `extraVolumes` options on kubeadm, however the default permissions are not suitable for audit logs.

By prepping the file with appropriate permissions beforehand, Kubernetes API Server keeps the audit log secure.

The following should be sufficient:

``` yaml
- name: directory.logs.kubernetes
  file:
    path: /var/log/kubernetes
    state: directory
- name: kubernetes.log.audit
  file:
    path: /var/log/kubernetes/audit.log
    state: touch
    mode: "0700"
    owner: root
    group: root
```

fyang-pivotal:
@randomvariable Out of curiosity, why "0700" permission, granting executable permission as well for the root user? Would "0600" permission be sufficient?

fyangd:
@randomvariable This issue has been resolved in PR https://github.com/kubernetes-sigs/image-builder/pull/213.   If you enable audit logging and point it to the precreated `/var/log/kubernetes/audit.log` file, you should observe the appropriate permission.  

Feel free to try it out and close and the issue.

codenrhoden:
/remove-lifecycle stale

@randomvariable Can you confirm that this issue has been resolved?

randomvariable:
I think this is only resolved for Photon, and maybe affected by #314 when I just went to test on Photon.

randomvariable:
/close

After we sorted out the clean up of /var/log, this problem has gone away.

PR #213, opened 2020-04-22
Precreate `kube-apiserver` audit log

fyangd:
- Because the default permission of audit log created by kubeadm is
too unrestrictive, i.e. read for all users, an audit log file is
pre-precreated with appropriate permission. See issue https://github.com/kubernetes-sigs/image-builder/issues/191

Build an AMI image and verified that the appropriate directory and file have been created:

```
[ec2-user@ip-172-31-19-152 ~]$ ll /var/log/kubernetes/
total 0
-rw------- 1 root root 0 Apr 22 00:06 audit.log
[ec2-user@ip-172-31-19-152 ~]$ ll /var/log/
total 356
drwxr--r-x 3 root   root       17 Apr 22 00:18 amazon
-rw------- 1 root   root    14970 Apr 22 00:18 boot.log
-rw------- 1 root   utmp        0 Apr 22 00:17 btmp
drwxr-xr-x 2 chrony chrony      6 Apr 22 00:17 chrony
-rw-r--r-- 1 root   root    99716 Apr 22 00:18 cloud-init.log
-rw-r--r-- 1 root   root     3321 Apr 22 00:18 cloud-init-output.log
-rw------- 1 root   root      283 Apr 22 00:20 cron
-rw-r--r-- 1 root   root    27124 Apr 22 00:17 dmesg
drwxr-xr-x 2 root   root       23 Apr 22 00:06 kubernetes
-rw------- 1 root   root      906 Apr 22 00:20 maillog
-rw------- 1 root   root   120478 Apr 22 00:25 messages
-rw------- 1 root   root     1191 Apr 22 00:25 secure
-rw-rw-r-- 1 root   utmp     2688 Apr 22 00:25 wtmp
-rw------- 1 root   root        0 Apr 22 00:18 yum.log
```

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -131,6 +131,18 @@
   - { path: /var/log, state: absent }
   - { path: /var/log, state: directory }
 
+- name: Create a log directory for kube-apiserver audit logs
+  file:
+    path: "{{ item.path }}"
+    state: "{{ item.state }}"
+    mode: "{{ item.mode }}"
+    owner: root
+    group: root
+  loop:
+  - { path: /var/log/kubernetes, state: directory, mode: "0755" }
+  - { path: /var/log/kubernetes/audit.log, state: touch, mode: "0600" }
+  when: kubernetes_semver != '' # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
+
 - name: Truncate shell history
   file:
     state: absent


images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -131,6 +131,17 @@
...
+  file:
+    path: "{{ item.path }}"
+    state: "{{ item.state }}"
+    mode: "{{ item.mode }}"
+    owner: root
+    group: root
+  loop:
+  - { path: /var/log/kubernetes, state: directory, mode: "0755" }
+  - { path: /var/log/kubernetes/audit.log, state: touch, mode: "0600" }
+
```
codenrhoden, line 144: This might be a strange request, but can we add something like `when: "{{ kubernetes_semver }} != ''"`?

I have this sysprep task shared between the node image and the haproxy image now (well -- soon, PR coming today I hope), and right now there is nothing specific to K8s in it. I'd rather not have this audit log created in the haproxy image.

This didn't come up before because of thinking this would be in the kubernetes role instead of the sysprep role. But this sysprep role removes everything in `/var/log`. The alternative would be to find a way to not remove this file if already exists. But it's simpler to just conditionally create it I think.

fyangd: Sure, I can add the conditional.  Is the implication here that all k8s related variables, including `kubernetes_semver` will be left as empty string, when building the HA Proxy image?  

Seems like an indirect way to convey the dependency, so I will include an inline comment.

codenrhoden: Yes, that is the implication exactly. Ideal? Probably not. but we have to work with right now. :)

fyangd: ok.  I added and tested the conditional.  When the `kubernetes_semver` is set to empty, Packer skips this step:

```
...
   amazon-2: TASK [sysprep : Create a log directory for kube-apiserver audit logs] **********
    amazon-2: skipping: [default] => (item={'path': '/var/log/kubernetes', 'state': 'directory', 'mode': '0755'})
    amazon-2: skipping: [default] => (item={'path': '/var/log/kubernetes/audit.log', 'state': 'touch', 'mode': '0600'})
...
```

Resolving this issue.

codenrhoden:
/lgtm
Just needs the CLA taken care of, @fyangd

fyangd:
Sounds good!  I'm in the process of getting and signing the CLA.

codenrhoden:
/approve

Thanks @fyangd !

PR #218, opened 2020-04-24
Changes to allow air-gapped AMI builds

voor:
I have a unique use case that requires no internet access when building the AMI, this was difficult to do based on several hard-coded variables in the process.  I'm still testing and validating, but this is what would need to change so far:

- Allow using neither load method **Requires `http` variable as an option if you're loading from HTTP now**
- Use native Amazon packages and remove Cloud Formation samples (not used)
- Allow overriding goss URL instead of requiring connectivity to GitHub.
- Output a manifest for each build from AMI, allow overriding the location
- Use semver instead of debian version for AMI names.
- ~Remove epel entirely if desired, also allow overriding the URL~ (removed for compatibility reasons)
- Adds in ami description.


voor:
This is still a draft because I haven't had a chance to test it with not populating most of these variables,  hoping to get to that soon.

images/capi/ansible/roles/kubernetes/tasks/ecrpull.yml
```diff
@@ -16,12 +16,21 @@
     --output text
     --query 'authorizationData[].authorizationToken'
   register: credentials
+  # TODO: ecr_b64AuthorizationToken can be deprecated when we move to Packer v1.5 or later.
```
detiber, line 19: It appears that we already require Packer v1.5.2 or greater for the vsphere provider, I'm fine if we make that requirement across all providers.

https://github.com/kubernetes-sigs/image-builder/blob/708a4fef189216ea2288ed2ece32bc247a70fe39/images/capi/hack/ensure-packer.sh#L21

detiber: Even better, the docs state Packer version 1.5.5, so not sure which is the actual requirement :man_shrugging: 

I am wondering if we should be setting [`min_packer_version`](https://www.packer.io/docs/templates/#min_packer_version) in our packer templates to add some additional enforcement here. @codenrhoden thoughts?

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -130,6 +130,7 @@
   loop:
   - { path: /var/log, state: absent }
   - { path: /var/log, state: directory }
+  - { path: /var/log/sa, state: directory, mode: "0755" }
```
detiber, line 133: mode looks to be hardcoded above
```suggestion
  - { path: /var/log/sa, state: directory }
```

images/capi/packer/ami/packer.json
```diff
@@ -60,9 +65,10 @@
         "owners": ["099720109477"],
         "most_recent": true
       },
-      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_deb_version` | clean_resource_name}}-{{user `build_timestamp`}}",
+      "ami_name": "capa-ami-ubuntu-18.04-{{user `kubernetes_semver` | clean_resource_name}}-{{user `build_timestamp`}}",
```
detiber, line 68: I don't think we backported the image lookup logic in cluster-api-provider-aws for the v1alpha2 releases, we might need to hold off on changing the ami_Name until we are ready to break lookups for v1alpha2 based clusters.

images/capi/packer/ami/packer.json
```diff
@@ -102,9 +109,10 @@
         "owners": ["410186602215"],
         "most_recent": true
       },
-      "ami_name": "capa-ami-centos-7-{{user `kubernetes_deb_version` | clean_resource_name}}-{{user `build_timestamp`}}",
+      "ami_name": "capa-ami-centos-7-{{user `kubernetes_semver` | clean_resource_name}}-{{user `build_timestamp`}}",
```
detiber, line 112: I don't think we backported the image lookup logic in cluster-api-provider-aws for the v1alpha2 releases, we might need to hold off on changing the ami_Name until we are ready to break lookups for v1alpha2 based clusters.

images/capi/packer/ami/packer.json
```diff
@@ -145,9 +154,10 @@
         "owners": ["amazon"],
         "most_recent": true
       },
-      "ami_name": "capa-ami-amazon-2-{{user `kubernetes_deb_version` | clean_resource_name}}-{{user `build_timestamp`}}",
+      "ami_name": "capa-ami-amazon-2-{{user `kubernetes_semver` | clean_resource_name}}-{{user `build_timestamp`}}",
```
detiber, line 157: I don't think we backported the image lookup logic in cluster-api-provider-aws for the v1alpha2 releases, we might need to hold off on changing the ami_Name until we are ready to break lookups for v1alpha2 based clusters.

voor:
`redhat_epel_rpm` was renamed at some point from `common_redhat_epel_rpm` so I needed to fix that.

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -22,7 +22,7 @@
   when: kubernetes_source_type == "pkg" and ansible_os_family == "VMware Photon OS"
 
 - import_tasks: url.yml
-  when: kubernetes_source_type != "pkg" and kubernetes_cni_source_type != "pkg"
+  when: kubernetes_source_type == "http" and kubernetes_cni_source_type == "http"
```
codenrhoden, line 25: This points out that these two variables don't really serve a purpose to exist separately right now. The conditional is written (previously) as if you could do either from an HTTP source or from a package source, when in reality, the task treats them as all or nothing.

PR #220, opened 2020-04-27
Fix provider role not running on vsphere builds

MaxRink:
packer_builder_type is vsphere_iso for vsphere builds, thus cluster init never gets installed.
Thus all images build by vsphere-iso wont work with CAPV.
This fixes that.


codenrhoden:
I have this fixed in https://github.com/kubernetes-sigs/image-builder/pull/214/commits/5177e406be093937ad746a6cacd65ac31ff0c5d1 as well, but it's going to be a bit longer before that merges. Happy to drop that commit. :)

/approve

PR #222, opened 2020-04-29
Create setup role for package installation

codenrhoden:
This patch creates a "setup" role which can be used by other roles as a
role dependency. It is intended to be used as the first role run on a
new OS, dealing with initial package installation.

/assign @detiber 

Jason, I know you said in https://github.com/kubernetes-sigs/image-builder/pull/214#discussion_r417346719 that you didn't consider it blocking, but I had already took a stab at this last night and finished testing this morning. I think this is a lot cleaner. See what you think and let me know if you have any questions.

The basic premise is that the node role now depends on the setup role -- and the setup role will be run *first*. The node role passes in which packages it wants installed, and the HAProxy role will eventually do the same.


images/capi/ansible/roles/node/meta/main.yml
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+dependencies:
+  - role: setup
+    vars:
+      rpms: "{{ common_rpms if ansible_os_family == 'RedHat' else common_photon_rpms }}"
+      debs: "{{ common_debs }}"
```
detiber, line 19: nit: Any particular reason to map the vars here, since the new role still does conditional imports based on OS?

codenrhoden: By "map", you mean the conditional for Red Hat vs CentOS?

The new role does not distinguish between Red Hat vs Photon RPMs (as far as the list being held in a variable). It just takes a single list of RPMs. I didn't start this way. I originally kept the same variables, like this:

```yaml
- role: setup
  vars:
    common_rpms: "{{ common_rpms }}"
    common_photon_rpms: "{{ common_photon_rpms }}"
    common_debs: "{{ common_debs }}"
```

Ansible does *not* like this. It considers it a circular/recursive import and throws all kinds of errors. So the variable names in the "setup" role have to be different. And I just wanted to pass a list of RPMs. The new role will already use different package managers based on the OS (just like we did before), so this seemed cleaner to me.

detiber: Ah, gotcha. I'm also following a bit better with seeing the use in the other PR.

PR #225, opened 2020-05-01
Bump Ubuntu ISO 18.04.4

codenrhoden:
This updates the ubuntu ISO to 18.04.4.

As part of #186, we needed at `unattended-upgrades` to move forward without any ansible errors. This patch fixes that as well.

/assign @detiber


images/capi/ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -71,7 +71,16 @@
   - apt-daily.timer
   - apt-daily-upgrade
   - apt-daily-upgrade.timer
-  - unattended-upgrades
```
detiber, line 74: Wouldn't we still want this for to be disabled via systemd in places where we are not installing from iso?

images/capi/ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -71,7 +71,16 @@
...
+
+- name: Get installed packages
+  package_facts:
+
+- name: Disable unattended upgrades if installed
+  systemd:
+    name: unattended-upgrades
```
codenrhoden, line 80: @detiber I believe this new section here does exactly what you want?

detiber: doh, indeed, sorry about that.

PR #228, opened 2020-05-05
Fix 'make clean-ova'

codenrhoden:
With the addition of the haproxy OVA, it would be nice for the `make clean-ova` target to clean the haproxy image as well. It was missed originally.

Additionally, clarify the behavior fo `make clean` when running `make help`.

/assign @detiber


codenrhoden:
Forgot Jason is on PTO.
/assign @vincepri

Issue #231, opened 2020-05-07
Ensure AppArmor support for distros


randomvariable:
Have had reports of AppArmor not being available on PhotonOS.

Not 100% certain about this, but I believe the requirements to get it to work are:

* Ensure the AppArmor module is available / loaded
* Have some profiles available on disk, though this could be done via a daemonset like https://github.com/kubernetes/kubernetes/tree/master/test/images/apparmor-loader
* If GOSS is returning in some form, probably worth some tests to ensure AppArmor is working.

Affected distributions:
* PhotonOS
* Ubuntu
* Amazon Linux 2 (which also supports SELinux)

jessehu:
A user reported that AppArmor is not enabled.  Per K8s AppArmor doc, there should be AppArmor enabled in the node yaml, but they don’t see it:
```
kubectl get nodes -o=jsonpath=$'{range .items[*]}{@.metadata.name}: {.status.conditions[?(@.reason=="KubeletReady")].message}\n{end}'
my-cluster-control-plane-5nbhr: kubelet is posting ready status
my-cluster-md-0-86bf754697-45dp9: kubelet is posting ready status
my-cluster-md-0-86bf754697-7gdh9: kubelet is posting ready status
my-cluster-md-0-86bf754697-bzgzd: kubelet is posting ready status
```
When creating a simple “Hello AppArmor” pod with the deny-write profile https://kubernetes.io/docs/tutorials/clusters/apparmor/, the pod is in Blocked state and user see error msg "Cannot enforce AppArmor: AppArmor is not enabled on the host".

PR #232, opened 2020-05-07
apparmor: Install a default apparmor profile

randomvariable:
Part of #231 

Should install a profile you can use in Kubernetes, as well as ensuring the right  binaries exist on PhotonOS.

Been unable to get Packer to run on vSphere though.

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/apparmor.yml
+++ b/images/capi/ansible/roles/containerd/tasks/apparmor.yml
@@ -1,0 +1,12 @@
+- name: Copy in apparmor profile
+  template:
+    dest: /etc/apparmor.d/container-default
+    src: etc/apparmor.d/container-default
+  when: ansible_os_family == "Debian" || ansible_os_family == "VMware Photon OS"
+
+- name: start apparmor service
+  systemd:
+    name: apparmor
+    daemon_reload: yes
+    enabled: True
+    state: started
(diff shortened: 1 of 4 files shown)


EleanorRigby:
@randomvariable: What more needs to be done here? Do we need this support across OSs or just photon OS?

```
Been unable to get Packer to run on vSphere though.
```
Is this appArmor problem or waiting for ip timeout issue?

randomvariable:
I didn't actually test it, I think at the time I wrote this, I didn't know how to get image builder to work. Thanks for reminding me.

images/capi/ansible/roles/containerd/tasks/apparmor.yml
```diff
@@ -0,0 +1,12 @@
+- name: Copy in apparmor profile
+  template:
+    dest: /etc/apparmor.d/container-default
+    src: etc/apparmor.d/container-default
+  when: ansible_os_family == "Debian" || ansible_os_family == "VMware Photon OS"
```
detiber, line 5: Would it make sense to try and detect this?

I think there are potentially a few different approaches we could take here:
- simply check if `/etc/apparmor.d/` exists
- check `/sys/kernel/security/lsm` to see if apparmor is present
- check if `/sys/module/apparmor/parameters/enabled` exists

detiber: This conditional also seems redundant with the same conditional existing on the `import_tasks` in `iamges/capi/ansible/roles/containerd/tasks/main.yml`

PR #233, opened 2020-05-08
chore: more strict logrotate on conntrackd

mytunguyen:
conntrackd logs is taking up a lot of disk space. Logrotate needs to compress the logs.  
Before:
```
-rw-------  1 root   root    28G May  8 17:15 conntrackd-stats.log
-rw-------  1 root   root    31G May  3 06:25 conntrackd-stats.log.1
-rw-------  1 root   root   3.6G Apr 27 06:25 conntrackd-stats.log.2
```

After:
```
-rw-------  1 root   root    17M May  8 18:16 conntrackd-stats.log
-rw-------  1 root   root   2.5G May  8 17:48 conntrackd-stats.log-20200508.gz
```

### DIFF
--- a/images/capi/ansible/roles/node/files/conntrackd
+++ b/images/capi/ansible/roles/node/files/conntrackd
@@ -1,0 +1,14 @@
+/var/log/conntrackd-stats.log {
+    su root syslog
+    compress
+    dateext
+    notifempty
+    missingok
+    nocreate
+    weekly
+    rotate 2
+    copytruncate
+    postrotate
+        invoke-rc.d conntrackd restart > /dev/null
+    endscript
+}
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -73,3 +73,12 @@
     sysctl_set: yes
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
+
+- name: Ensure conntrackd logrotate is present
+  copy:
+    src: conntrackd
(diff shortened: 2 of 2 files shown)


images/capi/ansible/roles/node/files/conntrackd
```diff
@@ -0,0 +1,14 @@
...
+    compress
+    dateext
+    notifempty
+    missingok
+    nocreate
+    weekly
+    rotate 2
+    copytruncate
+    postrotate
+        invoke-rc.d conntrackd restart > /dev/null
```
detiber, line 12: Should this use `systemctl restart conntrackd` instead to support all the various OSs in this repo (which I might be overly optimistic in assuming all are using systemd at this point)?

I'm wondering if there are other os-specific tweaks that might also need to be considered in the context of this PR

codenrhoden: @mytunguyen I have the same question re: `systemctl`. All of the OS's we are building are on SystemD, so a SysV command here doesn't seem correct (even if the OS is smart enough to make it happen anyways).

ktchen14: Is this necessary at all? It seems like a better approach would be to have `conntrackd` configured with `Syslog on` so that the log rotation is invisible to it and handled by the syslog daemon. That'd also eliminate the `copytruncate` requirement.

codenrhoden:
Now that conntrackd has been completely removed (https://github.com/kubernetes-sigs/image-builder/pull/344), this is no longer necessary

/close

PR #240, opened 2020-05-21
Add chrony for time sync for Azure image

nader-ziada:
Chrony is the only way to sync time from host using PTP source according to the azure team

Related to issue [539](https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/539) in cluster-api-provider-azure

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -19,3 +19,31 @@
   vars:
     packages:
       - azure-cli
+
+- name: Ansible apt install chrony
+  apt:
+    name: chrony
+    state: present
+  when: ansible_distribution == "Ubuntu"
+  
+- name: Configure PTP
+  lineinfile:
+    path: /etc/chrony/chrony.conf
+    create: yes
+    line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
+  when: ansible_distribution == "Ubuntu"  
+
+- name: Ensure makestep parameter set as per Azure recommendation
+  lineinfile:
+    path: /etc/chrony/chrony.conf
+    regexp: '^makestep'
+    line: makestep 1.0 -1
+  when: ansible_distribution == "Ubuntu"  
+
+- name: Enable chrony.service
+  systemd:
+    enabled: yes
+    state: started
+    daemon_reload: yes
+    name: chrony.service
+  when: ansible_distribution == "Ubuntu"  


images/capi/packer/azure/packer.json
```diff
@@ -117,6 +117,12 @@
         "--extra-vars",
         "{{user `ansible_extra_vars`}}"
       ]
+    },
+    {
+      "type": "shell",
+      "inline": [
+        "sudo apt-get -qq update && sudo apt-get -qqy install chrony &&  sudo systemctl restart chrony.service"
```
detiber, line 124: Could this be moved to the Ansible code instead? I suspect there will also need to be some additional modifications to the chrony config to configure the ptp time source, as described in https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync#chrony.

nader-ziada: ok, will look into how to do that using ansible instead

randomvariable: also, if there was a daemon already present, need to make sure that's disabled and removed.

nader-ziada: The ubuntu bionic docs say to install it https://ubuntu.com/blog/ubuntu-bionic-using-chrony-to-configure-ntp but I will double check

CecileRobertMichon:
@nader-ziada how do you plan on testing this change? We should make sure the result is what we want before changing image builder (ie. building an image with these changes and using that to build a CAPZ cluster).

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -19,3 +19,19 @@
...
+
+- name: Ansible apt install chrony
+  apt:
+    name: chrony
+    state: present
+
+- name: fix chrony drift
+  shell: chronyc makestep 1.0 -1
+  args:
+    executable: /bin/bash
```
randomvariable, line 31: This step isn't relevant at the point of creating an image, as new machines are going to be created anyway with a completely different time.

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -19,3 +19,19 @@
...
+
+- name: restart chrony.service to pick up config changes
+  systemd:
+    state: restarted
```
randomvariable, line 35:
```suggestion
	enabled: yes
    state: restarted
```
Better to be explicit that we want it enabled on boot. The restart isn't technically necessary because of image creation.

images/capi/ansible/roles/providers/tasks/azure.yml
randomvariable: Does chronyc actually change the config, or just the current process config?

randomvariable: Just checked locally:
``` bash
cluster-api on  master [⇡$] at ☸️  kind-kind took 5s
➜ sudo chronyc makestep 1.0 -1
200 OK

cluster-api on  master [⇡$] at ☸️  kind-kind
➜ cat /etc/chrony.conf
# These servers were defined in the installation:
pool time.google.com iburst
# Use public servers from the pool.ntp.org project.
# Please consider joining the pool (http://www.pool.ntp.org/join.html).

# Record the rate at which the system clock gains/losses time.
driftfile /var/lib/chrony/drift

# Allow the system clock to be stepped in the first three updates
# if its offset is larger than 1 second.
makestep 1.0 3

# Enable kernel synchronization of the real-time clock (RTC).
rtcsync

# Enable hardware timestamping on all interfaces that support it.
#hwtimestamp *

# Increase the minimum number of selectable sources required to adjust
# the system clock.
#minsources 2

# Allow NTP client access from local network.
#allow 192.168.0.0/16

# Serve time even if not synchronized to a time source.
#local stratum 10

# Specify file containing keys for NTP authentication.
keyfile /etc/chrony.keys

# Get TAI-UTC offset and leap seconds from the system tz database.
leapsectz right/UTC

# Specify directory for log files.
logdir /var/log/chrony

# Select which information is logged.
#log measurements statistics tracking
```

nader-ziada: might be actually changing the current process, i need to double check

nader-ziada:
@CecileRobertMichon yes, I build the image and created a vm with it manually, next will try it with a cluster

images/capi/ansible/roles/providers/tasks/azure.yml
randomvariable: Probably want to add something like

```yaml
- name: Configure PTP
  lineinfile:
    path: /etc/chrony/chrony.conf
    create: yes
    line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0

- name: Ensure makestep parameter set as per Azure recommendation
  lineinfile:
    path: /etc/chrony/chrony.conf
    regexp: '^makestep'
    line: makestep 1.0 -1
```

nader-ziada: Thank you, I was just trying to to figure out how to do this in ansible!

CecileRobertMichon:
/cc @jackfrancis 

(related to https://github.com/Azure/aks-engine/issues/2552)

nader-ziada:
created the image and VM from it and this is what I see: 
```
$ chronyc tracking
Reference ID    : 50484330 (PHC0)
Stratum         : 1
Ref time (UTC)  : Thu May 21 23:35:23 2020
System time     : 0.000013028 seconds slow of NTP time
Last offset     : -0.000015507 seconds
RMS offset      : 0.000255382 seconds
Frequency       : 72.325 ppm slow
Residual freq   : -0.054 ppm
Skew            : 1.491 ppm
Root delay      : 0.000000001 seconds
Root dispersion : 0.000033918 seconds
Update interval : 8.0 seconds
Leap status     : Normal
```
```
$ chronyc activity
200 OK
8 sources online
0 sources offline
0 sources doing burst (return to online)
0 sources doing burst (return to offline)
0 sources with unknown address
```
```
$ chronyc sources
210 Number of sources = 9
MS Name/IP address         Stratum Poll Reach LastRx Last sample
===============================================================================
#* PHC0                          0   3   377    12    +85us[  +98us] +/-   17us
^- alphyn.canonical.com          2   6   357    22   +532us[ +548us] +/-   68ms
^- golem.canonical.com           2   6   377    16   +816us[ +834us] +/-   69ms
^- chilipepper.canonical.com     2   6   377    18   +851us[ +868us] +/-   73ms
^- pugot.canonical.com           2   6   377    19   +685us[ +702us] +/-   82ms
^- atl0.fairy.mattnordhoff.>     2   6   377    21   +542us[ +559us] +/-   29ms
^- europa.netbunker.org          2   6   377    18  +1980us[+1997us] +/-   29ms
^- t1.time.bf1.yahoo.com         2   6   377    19  +1042us[+1059us] +/-   46ms
^- 198.255.68.106                2   6   377    19  +1253us[+1270us] +/-   80ms
```

```
~$ cat /etc/chrony/chrony.conf
# Welcome to the chrony configuration file. See chrony.conf(5) for more
# information about usuable directives.

# This will use (up to):
# - 4 sources from ntp.ubuntu.com which some are ipv6 enabled
# - 2 sources from 2.ubuntu.pool.ntp.org which is ipv6 enabled as well
# - 1 source from [01].ubuntu.pool.ntp.org each (ipv4 only atm)
# This means by default, up to 6 dual-stack and up to 2 additional IPv4-only
# sources will be used.
# At the same time it retains some protection against one of the entries being
# down (compare to just using one of the lines). See (LP: #1754358) for the
# discussion.
#
# About using servers from the NTP Pool Project in general see (LP: #104525).
# Approved by Ubuntu Technical Board on 2011-02-08.
# See http://www.pool.ntp.org/join.html for more information.
pool ntp.ubuntu.com        iburst maxsources 4
pool 0.ubuntu.pool.ntp.org iburst maxsources 1
pool 1.ubuntu.pool.ntp.org iburst maxsources 1
pool 2.ubuntu.pool.ntp.org iburst maxsources 2

# This directive specify the location of the file containing ID/key pairs for
# NTP authentication.
keyfile /etc/chrony/chrony.keys

# This directive specify the file into which chronyd will store the rate
# information.
driftfile /var/lib/chrony/chrony.drift

# Uncomment the following line to turn logging on.
#log tracking measurements statistics

# Log files location.
logdir /var/log/chrony

# Stop bad estimates upsetting machine clock.
maxupdateskew 100.0

# This directive enables kernel synchronisation (every 11 minutes) of the
# real-time clock. Note that it can’t be used along with the 'rtcfile' directive.
rtcsync

# Step the system clock instead of slewing it if the adjustment is larger than
# one second, but only in the first three clock updates.
makestep 1.0 -1

refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
```

nader-ziada:
I created a few Azure clusters using the image I generated from here and everything is working fine, not sure if there is a special test we need to check on the cluster, but at least this doesn't break anything and `chrony` is running as in the above comment.

nader-ziada:
@CecileRobertMichon is it possible for you to try this out?

CecileRobertMichon:
@detiber @randomvariable do we have plans to include E2E testing as a presubmit for image-buidler PRs? Seems like it would be useful in cases like this

CecileRobertMichon:
@nader-ziada this looks good to me, but we really need a way to be able to test changes to the image in a non-manual way in the future :) 

/lgtm

nader-ziada:
@CecileRobertMichon i can work on tests but better if we can discuss what scope to cover

nader-ziada:
@detiber ready for another look, Thanks

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -19,3 +19,27 @@
   vars:
     packages:
       - azure-cli
+
+- name: Ansible apt install chrony
+  apt:
+    name: chrony
+    state: present
```
detiber, line 26: This should probably be wrapped in a conditional to only run when the OS is Ubuntu similar to: https://github.com/kubernetes-sigs/image-builder/blob/c680d2c6fc2d54b0fad0683d560d60f801dedac3/images/capi/ansible/roles/providers/tasks/aws.yml#L38-L40

Relatively minor nit, but would have an impact if/when other OS images are supported for Azure images.

nader-ziada: updated to add check for Ubuntu

nader-ziada:
Hi @codenrhoden , do you have any concerns with this change?

codenrhoden:
/approve

Thanks for this @nader-ziada! And for trying out clusters with CAPZ first since there are no blocking PR tests still...

@CecileRobertMichon:
> do we have plans to include E2E testing as a presubmit for image-buidler PRs? Seems like it would be useful in cases like this

Short answer is "yes". 😆  But definitely lacking on details so far. But I am in 100% agreement that this is necessary. It's in my near-term plans to start to focus on this.

PR #242, opened 2020-05-29
Add ethtool to photon ova(s)

dims:
ethtool is handy to have especially to tweak things like say the offload checksum when you can't get networking to work right.

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -49,6 +49,7 @@ common_photon_rpms:
 - conntrack-tools
 - distrib-compat
 - ebtables
+- ethtool
 - jq
 - net-tools
 - ntp


codenrhoden:
/hold

I'd like to have a broader discussion around this one. I 100% agree that `ethtool` is "handy", but that's never been the criterion for adding a package. We had said in the past that we only install the bare minimum required to get a K8s cluster up and running. If users want to add debugging or convenience packages, they can certainly add them when building.

I recognize this philosophy is not documented, and I'm taking it from verbal conversations at the SIG Cluster Lifecycle level in the past, so I'd like to take this as an opportunity to re-evaluate how we agree on what packages are installed.

If `ethtool` is handy, why only add it for Photon? If something is handy, wouldn't we want it in all distros? (`jq` comes to mind - it's listed for Photon and Ubuntu, but not CentOS?).
We've had other packages that were handy come up before but decided *not* to install (`yq` comes to mind).

I think we should start to have some kind of process for determining what comes in and what does not. Note that anyone who builds their own images can always add extra packages (and I personally expect that they would want to) via the `extra_rpms` or `extra_debs` parameters.

dims:
all good points @codenrhoden One thing special here with ethtool is that it is useful to fix networking issues so we could pull other interesting things from elsewhere. So helpful in bootstrapping problems.

codenrhoden:
I didn't mean for you to close the PR. Just saying I wanted to use it as a catalyst for a wider discussion.

I was more expecting the final result to be "packages that are helpful for bootstrapping in unique environments have a purpose", and then to say we should put this in for CentOS too. :) But I didn't want to taint the conversation.

detiber:
My $0.02, we should strive to keep the package set we install by default to the minimum needed to support running and bootstrapping Kubernetes.

I think trying to maintain a minimal footprint is important for a few reasons. Smaller base images will tend to bootstrap faster (COW-snapshot based images being a notable exception). They also have lower bugfix and security maintenance effort (once we get into automated building/publishing of images). 

I think we have a few options related to troubleshooting tools with a minimal image:
- Include installing the required tools as part of a troubleshooting guide prior to use, though this might require providing instructions for different OSs and package managers, which is a bit frustrating.
- Building a container image that includes all of the tools that we'd want to make available for use in a troubleshooting guide, then the first step in the troubleshooting guide could be to run the container as privileged and exec into the container to follow the rest of the guide.

Another thing we could potentially do, but would provide a bit less usability for ad-hoc troubleshooting (or for troubleshooting intermittent issues) is to have minimal and debug OS images. This would allow someone getting started (or trying to replicate an issue) to use an image that has additional tools available, but still keep the default used image minimal.

dims:
oh no worries! it's just a one liner :)

PR #246, opened 2020-06-03
Fix tar command on quickstart

diegocn:
By following the steps for `curl`, `tar`, and `cd` as they are present on docs I got a little lost because it did not create the expected `image-builder` folder. The output of tar was a folder like this `kubernetes-sigs-image-builder-<commit_hash>`.

This PR forces the output folder name for `tar` as `image-builder`.


docs/book/src/capi/quickstart.md
```diff
@@ -10,7 +10,7 @@ To get the latest image-builder source on your machine, execute the following:
 
 ```sh
 curl -L https://github.com/kubernetes-sigs/image-builder/tarball/master -o image-builder.tgz
-tar xzf image-builder.tgz
+tar xzf image-builder.tgz --one-top-level=image-builder --strip-components 1
```
codenrhoden, line 13: Thank you for catching this, @diegocn.

The `--one-top-level` option isn't available in all tar implementations (namely, not on MacOS, though that is easily replaced using `brew`).

I wonder if it would be better to be more explicit by creating the directory to extract to first, like so:
```sh
mkdir image-builder
tar xzf image-builder.tgz --strip-components 1 -C image-builder
```

What do you think of that instead?

This also makes me realize that we should really document/encourage people grabbing the source via `git` if they have that available. For some of the image targets, having the `git` executable available is assumed. And if people have it, it is much easier to grab the code with that rather than curl'ing a tarball.

diegocn: Hi @codenrhoden!

Oh, I did not know about that `tar` limitation. Both of your suggestions make sense but I'd tend to choose the second due to exactly what you've said (that we assume git as a dependency for some tasks).

I'll add a new commit with the git steps.

EleanorRigby: @diegocn : Did you find time to iterate on the second commit for suggested changes?

PR #249, opened 2020-06-08
capi: ubuntu-1804: use python3

jsen-:
`make build-qemu-ubuntu-1804` was failing on ansible with `/bin/sh: 1: /usr/bin/python`, adding the python package to `preseed.cfg` seems to help

### DIFF
--- a/images/capi/packer/qemu/packer.json
+++ b/images/capi/packer/qemu/packer.json
@@ -1,7 +1,7 @@
 {
   "variables": {
     "ansible_common_vars": "",
-    "ansible_extra_vars": "",
+    "ansible_extra_vars": "ansible_python_interpreter=/usr/bin/python3",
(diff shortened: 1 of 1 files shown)


ncdc:
@codenrhoden any concerns here?

codenrhoden:
/lgtm
/approve

> @codenrhoden any concerns here?

No concerns with this change specifically. On a broader scope, I've always had a bit of unease in that there isn't a clear cutline for different OS's on which ones are **only** Python2 or Python3. Looking at the Ubuntu DEBs we install, it appears to be only Python3 packages, but the CentOS and Photon RPMs have a mix of 2/3, and I'm not sure if that's really required, or if it's been bandage on top of bandage to keep things working.

So there's some tech debt there. :)

PR #241, opened 2020-05-22
Update OpenStack provider document

hidekazuna:
This PR reflects the issue #238 , adds Prerequisites for QCOW2 section to explain installing packages to use qemu-img and make it valid.

Issue #238, opened 2020-05-19
Failed Build QEMU Image


iputra:
Hi

I'm trying to build qemu image by using image-builder for cluster-api provider openstack, but I'm getting error in provisioning with ansible. The following error is like this :

```
. . .
    ubuntu-1804: PLAY [all] *********************************************************************
    ubuntu-1804:
    ubuntu-1804: TASK [Gathering Facts] *********************************************************
    ubuntu-1804: ok: [default]
    ubuntu-1804:
    ubuntu-1804: TASK [setup : Put templated sources.list in place] *****************************
    ubuntu-1804: fatal: [default]: FAILED! => {"changed": false, "checksum": "963e5f098c598950964617d096a2ba51d2c09734", "msg": "Destination /etc/apt not writable"}
    ubuntu-1804:
    ubuntu-1804: PLAY RECAP *********************************************************************
    ubuntu-1804: default                    : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
    ubuntu-1804:
==> ubuntu-1804: Provisioning step had errors: Running the cleanup provisioner, if present...
==> ubuntu-1804: Deleting output directory...
Build 'ubuntu-1804' errored: Error executing Ansible: Non-zero exit status: exit status 2
==> Some builds didn't complete successfully and had errors:
--> ubuntu-1804: Error executing Ansible: Non-zero exit status: exit status 2
==> Builds finished but no artifacts were created.
make: *** [build-qemu-ubuntu-1804] Error 1
. . .
```

snicholas:
Hi, I'm experiencing the same problem. Tried also with older tags, but same problem. 
I've tried on ubuntu 18.04. Ansible version is 2.8.0 and packer should be 1.5.2.

snicholas:
I resolved by adding my user to the kvm group
`sudo usermod -a -G kvm <yourusername>`
and (even if I'm not sure it is really usefull)
`sudo chown root:kvm /dev/kvm`
then exit and log back in to make the change take place.
After that run 
`make build-qemu-ubuntu-1804`
as the normal user (non root). Hope this help out!

iputra:
> Hi, I'm experiencing the same problem. Tried also with older tags, but same problem.
> I've tried on ubuntu 18.04. Ansible version is 2.8.0 and packer should be 1.5.2.
> @iputra did you rn the make command as root or normal user?

Yup, I also tried with older tags, but same problem.
[...]

PR #250, opened 2020-06-14
Use specific Ansible provisioner user for qemu build

carstenkoester:
Fixes an issue were Ansible would skip privilege escalation if
Ansible/Packer are running as root and Ansible "thinks" it is connecting
to the target as root.

### DIFF
--- a/images/capi/packer/qemu/packer.json
+++ b/images/capi/packer/qemu/packer.json
@@ -76,6 +76,7 @@
     {
       "type": "ansible",
       "playbook_file": "./ansible/node.yml",
+      "user": "builder",
       "ansible_env_vars": [
         "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'",
         "ANSIBLE_REMOTE_TEMP='/tmp/.ansible/'"


hidekazuna:
This PR is incompatible with #241 which is not merged yet.

carstenkoester:
> This PR is incompatible with #241 which is not merged yet.

Are you sure it's incompatible? What was the kind of conflict observed?

#241 certainly describes the better practice, which is running as non-root. However, this PR here ensures the build *also* works when running as root, which could be the case when running containerized or in a CI, where modifying the hosts's users or `/dev/kvm` group membership is sometimes not an option. However, this PR does not *enforce* running as root, or in any other way conflicts with users who want to run as non-root.

hidekazuna:
> > This PR is incompatible with #241 which is not merged yet.
> 
> Are you sure it's incompatible? What was the kind of conflict observed?

Sorry incompatible is over saying, I like this PR actually. #241 is based on running as root so that  I remove modifying /dev/kvm group membership after this PR merged.

/lgtm

> 
> #241 certainly describes the better practice, which is running as non-root. However, this PR here ensures the build _also_ works when running as root, which could be the case when running containerized or in a CI, where modifying the hosts's users or `/dev/kvm` group membership is sometimes not an option. However, this PR does not _enforce_ running as root, or in any other way conflicts with users who want to run as non-root.

codenrhoden:
/approve

This makes sense to me. Interestingly, it makes me wonder how non-root builds for all the other images are working. From the Packer docs:

> user (string) - The ansible_user to use. Defaults to the user running packer, NOT the user set for your communicator. If you want to use the same user as the communicator, you will need to manually set it again in this field.

The "defaults to user running Packer" part confuses me. Personally, I am running as myself on my own laptop, and that username is not present on the images. Only the `builder` user is. I'll have to enable to a verbose mode at some point and see what's really happening there.

carstenkoester:
> This makes sense to me. Interestingly, it makes me wonder how non-root builds for all the other images are working.

I wrote up a small summary here: https://github.com/kubernetes-sigs/image-builder/issues/238#issuecomment-635713892

For *non-root* builds, this isn't much of an issue. The username specified here is a completely artificial name that exists just between Ansible and Packer's SSH communicator -- it does *not* need to exist on the target image. The user which is used between the SSH communicator and the image, is the one that does need to exist on the image -- but that is a potentially different user.

The problem here really arises when Packer is run as root (and thus the ansible provisioner user defaults to "root", but the image builder user is *not* root. In that case, Ansible "thinks" it can skip privilege escalation, resulting in builds failing due to permission issues.

codenrhoden:
Thanks for that pointer, @carstenkoester, that really helped!

PR #252, opened 2020-06-15
Fix default EPEL location for CentOS 7.

codenrhoden:
A recent change (#218) made it possible to set the EPEL location to an empty
string to skip installing EPEL. Since this var is now always set by
Packer, not setting it makes it be an empty string, and since extra-vars
take precedence in Ansible, the net effect is that the var was *always*
set to an empty string, even though a default is set in the Ansible
role.

This patch sets the default value for CentOS 7 back as it was. The
addition is only the CentOS 7 file for OVAs, so it does not effect AL2,
and if needed in a future CentOS 8 distro, a release specific version
can be set there as well.

/assign @detiber 
/cc @voor

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -39,6 +39,7 @@
     "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
     "ami_groups": "all",
     "ami_users": "",
+    "redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",
     "reenable_public_repos": "true",
     "remove_extra_repos": "false",
     "snapshot_groups": "all",
--- a/images/capi/packer/ova/centos-7.json
+++ b/images/capi/packer/ova/centos-7.json
@@ -1,6 +1,7 @@
 {
   "build_name": "centos-7",
   "distro_name": "centos",
+  "redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",
   "os_display_name": "CentOS 7",
   "guest_os_type": "centos7-64",
   "vsphere_guest_os_type": "centos7_64Guest",


voor:
Well look at that, it's quietly skipping on AL2 and CentOS7 on AWS as well.  EPEL is not actually needed in those environments, so we probably need to consider whether these changes are similarly necessary in AWS as well, since it was an unexpected change.

codenrhoden:
> Well look at that, it's quietly skipping on AL2 and CentOS7 on AWS as well. EPEL is not actually needed in those environments, so we probably need to consider whether these changes are similarly necessary in AWS as well, since it was an unexpected change.

@voor If EPEL isn't need for the AMI builds, just adding it back for OVA should be enough, yes? I haven't been doing  any AWS builds, but I am no setup to do so. So I could try to do the builds with and without this change.

I was definitely thinking that this was only needed for OVAs. But I should check the AMIs!

voor:
I'm wrong, it is needed for CentOS7 in AWS because `python2-pip` is not available in the CentOS7 repositories anymore.

voor:
Building right now with `"redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",` added to `images/capi/packer/ami/packer.json` that should solve the problem for AWS

There's no CentOS7 build in Azure yet.

voor:
Yes, same change in `images/capi/packer/ami/packer.json` fixes CentOS7 on AWS.

image

codenrhoden:
Perfect. Let me know if that fixes things (I'm certain it will). That's the same issue I hit when building CentOS OVAs. I'll update this PR with that change when you give it a 👍 .

codenrhoden:
Thanks for looking, @voor! _Something something_...I need to add PR CI..._Something Something_...

voor:
Yes, we should probably also remove all python2 dependencies from these images.

codenrhoden:
> Yes, we should probably also remove all python2 dependencies from these images.

Yeah... Was just talking about that today: https://github.com/kubernetes-sigs/image-builder/pull/249#issuecomment-644354648

Once you are happy, any chance you can `/lgtm`? Not sure if you can without being in `kubernetes-sigs`...

PR #254, opened 2020-06-17
Switch to final Ubuntu 2004 ISO with the legacy installer

MaxRink:
Well, Ubuntu 2004 is out now and the ISO currently linked isnt functional anymore.
Ubuntu 2004 also switched to the new Installer, not using kickstart anymore, but luckily there is still an official image with the legacy installer that works plug-and-play with the exsiting code for 1804.
In the long term, this needs to be migrated to the new Installer tho ( https://ubuntu.com/server/docs/install/autoinstall )


codenrhoden:
Built fine for me! Thanks for updating this.
/approve

PR #259, opened 2020-06-18
CAPI: remove kubernetes-cni install

CecileRobertMichon:
As discussed in https://kubernetes.slack.com/archives/CD6U2V71N/p1592508705375700 and https://github.com/kubernetes-sigs/image-builder/pull/258, and in response to https://github.com/kubernetes/kubernetes/issues/92242, this PR removes the kubernetes-cni package install from all CAPI builders.

The current thinking is that "we don't actually need kubernetes-cni, as in general CNI providers do not consume the binaries provided by it, but vendor/use their own copies (or alternatives)" - @detiber

/hold for discussion and testing
/cc @detiber @vincepri @codenrhoden @justaugustus


detiber:
Changes look good to me

CecileRobertMichon:
Everything looks good on the Azure side. Built an image with these changes and built a cluster with CAPZ using that image.

/hold cancel

vincepri:
@detiber Should we test on AWS side as well?

detiber:
> @detiber Should we test on AWS side as well?

Yes.

@CecileRobertMichon out of curiosity, did you also verify conformance against the test image as well?

CecileRobertMichon:
@detiber running conformance right now. It will take an hour or two before I get a signal. I'll hold again until we get a green light from all providers. 

/hold

justaugustus:
Explicit hold for @kubernetes-sigs/release-engineering 
/hold

We're still discussing approaches and may end up re-introducing the kubernetes-cni package.

CecileRobertMichon:
@justaugustus can you please expand on 

> We're still discussing approaches and may end up re-introducing the kubernetes-cni package

?

If this package is not needed in CAPI images then it makes sense to remove it regardless, even if sig-release fixes the failures we are seeing because of the version incompatibility, no? Or would there be an explicit dependency on kubernetes-cni which would cause installing new k8s versions to fail without the cni there? Also, do you have an ETA for a potential fix?

This is the only workaround we have that successfully builds v1.16.11, v1.17.7 and v1.18.4 CAPI images at the moment and we are currently blocked on building those images without it.

justaugustus:
> Also, do you have an ETA for a potential fix?

@CecileRobertMichon -- The best I have is "soon", at the moment (hopefully well before week's end).
I have an idea, that will become a PR which will include more details. Will keep you posted.

justaugustus:
Opened https://github.com/kubernetes/release/pull/1375 as a potential fix, which has more details in the PR description.

justaugustus:
^^ That said, closing this in favor of https://github.com/kubernetes-sigs/image-builder/pull/258.
/close

PR #266, opened 2020-06-25
Move AMI builds to exclusively use Make

codenrhoden:
This patch updates the docs and Makefile to encourage the use of `make`
for building AMIs rather than calling `packer build` directly. Each
supported OS for AMIs now has a separate make target and can be called
individually.

Documentation is enhanced to call out the  most commonly customized
parameters, and to nudge users into using var files instead of modifying
repository files directly.

There are a couple of follow-on commits in this PR that clean a few things up. **Commits can/should be reviewed separately**, but I think it makes for 1 cohesive PR.

I've verified that I can build all 3 OS's within AWS, and that the AMIs are now tagged with the image-builder version as well (hurray). The new Make targets are:

```
  build-ami-amazon-2           Builds Amazon-2 Linux AMI
  build-ami-centos-7           Builds CentOS 7 AMI
  build-ami-ubuntu-1804        Builds Ubuntu 18.04 AMI
```

/assign @detiber 
/cc @voor @dvonthenen


codenrhoden:
@detiber Any feedback on this one? I know you build AMIs regularly, so I am most keen on your opinion.

docs/book/src/capi/providers/aws.md
```diff
@@ -26,40 +26,96 @@ make deps-ami
...
 accounts.
 
-### Limiting Images to Build
+##### Limiting AMI Regions
 
-If packer build is run without specifying which images to build, then it will attempt to build all configured images. `packer inspect packer.json` will list the configured builders. The `--only` option can be specified when running `packer build` to limit the images built.
-
-For example, to build only the Ubuntu Bionic image:
+By default images are copied to all available AWS regions. The list of all
+available regions can be obtained running:
```
detiber, line 107: So, we hardcode a list of regions to copy to currently,. I can't remember exactly why we went down that path, but it's where we are: https://github.com/kubernetes-sigs/image-builder/blob/492f35de17c76740bafdd3e7da93bda292af1779/images/capi/packer/ami/packer.json#L47

detiber: It does make me wonder if we should move the defaults here into ami-default.json to make it a bit more visible

images/capi/Makefile
```diff
@@ -136,7 +142,7 @@ HAPROXY_OVA_LOCAL_BUILD_NAMES			:=	$(addprefix haproxy-ova-local-,$(PHOTON_VERSI
 HAPROXY_OVA_ESX_BUILD_NAMES			:=	$(addprefix haproxy-ova-esx-,$(PHOTON_VERSIONS))
 HAPROXY_OVA_VSPHERE_BUILD_NAMES		:=	$(addprefix haproxy-ova-vsphere-,$(PHOTON_VERSIONS))
 
-AMI_BUILD_NAMES			?=	ami-default
+AMI_BUILD_NAMES			?=	ami-centos-7 ami-ubuntu-1804 ami-amazon-2
```
detiber, line 145: One downside to breaking these out into separate build targets is that building for all three will now be a serial operation rather than a parallel operation. This would greatly increase the runtime, since the longest part of the runtime is related to copying the AMI to the various regions.

codenrhoden: Yeah, I haven't explored if there is some `make -j` magic that can handle this or not. I did, while testing, verify that you can happily run them in parallel in different shells.

docs/book/src/capi/providers/aws.md
codenrhoden: I could put this list into the doc instead, and still point out how to get the "full lists" of regions. you're right, they may not be the same.

i think you noticed this already, but `ami-default.json` is removed in this patch. It only encourages people to edit that file, and served no real purpose to have it broken out vs putting it in `packer.json`. Either way, you should be overriding it with your own var file if you want to do so.

I'll update this section to put in the list that AMIs are copied to by default.

images/capi/Makefile
codenrhoden: One thought that I had around this - while you and I may often build multiple images for a given cloud provider, I think in the long term that won't be the most common case. Users will build a specific image they are interested in. over and over again.

Would you agree?

detiber: While I agree that most users are likely not going to be building images for multiple different target OSs and hopefully we'll be able to eventually get automation to eliminate the need for manually running the commands as we do today, I do think that it will not be uncommon to want to generate images for multiple different providers and the ability to parallelize those will greatly help those users.

That might be where @moshloop's tool comes in handy if it can do creation of the packer file on demand, then it could potentially run any supported permutations in parallel.

I'd like to avoid tripling the time required to create and verify images until we have automation in place for building the images. We already are pretty far behind publishing AMI builds for CAPA because of the time requirement, and increasing the time requirement to do so will only compound the situation.

docs/book/src/capi/providers/aws.md
codenrhoden: I already have the default value for `ami_regions` in the new `Common AWS options` section, so instead I've updated the section on limiting regions to point to that for reference, and mention how the entire list of regions can be obtained.

images/capi/Makefile
codenrhoden: Hmm, I reach a different conclusion with the same information. Since copying the AMI's takes the longest amount of time, I don't think there's much time lost in having to kick off the different builds in parallel (different tabs, screen sessions, whatever). But, I get your perspective. You're losing an ability you have now (build everything with one command), whereas I didn't have that for vSphere OVAs...  Until now!!

I've added  a commit to this PR that adds the `make build-ami-all` target for you (and similar for OVAs). If you run it with `make`, it will build all targets serially, if you run it with `make -j`, it will build them all in parallel. When you don't give a number after `-j`, there is no limit to parallel tasks.

I've run the following and it was successful:

```sh
$ cat aws.json
{
	"aws_region": "us-west-2",
	"ami_regions": ""
}
$ PACKER_VAR_FILES=aws.json make -j build-ami-all
```

/pony

PR #267, opened 2020-06-26
Adding PhotonOS to Azure.

voor:
This adds Photon 3 as an option for creating VHD output images in Azure.  I wanted to see how we could incorporate an uploaded ISO that could then turn into a VHD, since the current examples leveraged only shared gallery images as sources.

Also adds in other sovereign clouds to the VHD scripts by calling `az cloud set --name` (this defaults to `AzurePublicCloud` for backwards compatibility)

Really interested in some feedback, and completely open to further discussion on if this is the best approach.  I want to possible tackling publishing to a shared image gallery afterwards, since I know VHD is supposed to be deprecated.

/assign @CecileRobertMichon 
/cc @codenrhoden @nader-ziada

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -12,13 +12,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- name: install Azure clients
+- name: Install Azure client RPM packages on PhotonOS
+  command: tdnf install {{ packages }} --nogpgcheck -y
+  vars:
+    packages: "python-pyasn1 WALinuxAgent"
+  when: ansible_os_family == "VMware Photon OS"
+
+- name: install Azure clients through pip3
   pip:
     executable: pip3
     name: "{{ packages }}"
   vars:
     packages:
       - azure-cli
+  when: ansible_distribution == "Ubuntu" or ansible_os_family == "VMware Photon OS"
+
+- name: Add Microsoft signing key for RPM
+  rpm_key:
+    state: present
+    key: https://packages.microsoft.com/keys/microsoft.asc
+  when: ansible_os_family == "RedHat"
+
+- name: Add Azure CLI repository for RPM
+  yum_repository:
+    name: azure-cli
+    description: azure-cli
+    file: azure-cli
+    enabled: true
+    baseurl: https://packages.microsoft.com/yumrepos/azure-cli
+    gpgcheck: "{{ ansible_os_family == 'RedHat'|bool }}"
+    gpgkey: https://packages.microsoft.com/keys/microsoft.asc
+  when: ansible_os_family == "RedHat"
+
+- name: Refresh yum after repo addition since new repository is added
+  shell: 
+    cmd: yum updateinfo --enablerepo=azure-cli
+    warn: false
+  when: ansible_os_family == "RedHat"
+
+- name: Ansible yum install azure-cli, python-pyasn1, and WALinuxAgent
+  yum:
+    name:
+    - python-pyasn1
+    - WALinuxAgent
+    - azure-cli
+    enablerepo: azure-cli
+    update_cache: yes
+    state: latest
+  when: ansible_os_family == "RedHat"
 
 - name: Ansible apt install chrony
   apt:
@@ -31,14 +72,14 @@
     path: /etc/chrony/chrony.conf
     create: yes
     line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
-  when: ansible_distribution == "Ubuntu"  
+  when: ansible_distribution == "Ubuntu"
 
 - name: Ensure makestep parameter set as per Azure recommendation
   lineinfile:
     path: /etc/chrony/chrony.conf
     regexp: '^makestep'
     line: makestep 1.0 -1
-  when: ansible_distribution == "Ubuntu"  
+  when: ansible_distribution == "Ubuntu"
 
 - name: Enable chrony.service
   systemd:
@@ -46,4 +87,20 @@
     state: started
     daemon_reload: yes
     name: chrony.service
-  when: ansible_distribution == "Ubuntu"  
+  when: ansible_distribution == "Ubuntu"
+
+- name: Make sure device file exists
+  template:
+    src: etc/systemd/network/10-eth0.network.manual
+    dest: /etc/systemd/network/10-eth0.network.manual
+    owner: systemd-network
+    group: systemd-network
+    mode: '0644'
+  when: ansible_os_family == "VMware Photon OS"
+
+- name: enable waagent service
+  systemd:
+    name: waagent
+    enabled: yes
+    state: started
+  when: ansible_os_family == "RedHat" or ansible_os_family == "VMware Photon OS"
--- a/images/capi/ansible/roles/providers/templates/etc/systemd/network/10-eth0.network.manual
+++ b/images/capi/ansible/roles/providers/templates/etc/systemd/network/10-eth0.network.manual
@@ -1,0 +1,8 @@
+[Match]
+Name=eth0
+[Network]
+LinkLocalAddressing=no
+DHCP=ipv4
+[DHCP]
+UseDomains=true
+ClientIdentifier=mac
--- a/images/capi/ansible/roles/setup/defaults/main.yml
+++ b/images/capi/ansible/roles/setup/defaults/main.yml
@@ -22,3 +22,5 @@ extra_rpms: ""
 
 disable_public_repos: false
 extra_repos: ""
+
+post_reboot_delay: 30
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -8,6 +8,7 @@
     "disable_public_repos": "false",
     "extra_debs": "",
     "extra_repos": "",
+    "extra_rpms": "",
     "azure_location": null,
     "vm_size": "",
     "build_timestamp": "{{timestamp}}",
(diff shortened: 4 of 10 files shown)


images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -31,7 +72,7 @@
     path: /etc/chrony/chrony.conf
     create: yes
     line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
-  when: ansible_distribution == "Ubuntu"  
+  when: ansible_distribution == "Ubuntu"
```
detiber, line 75: If chrony is available on Red Hat and Photon, we should likely also prefer for it to be configured there as well to avoid any potential for time sync related issues on those OSs

images/capi/ansible/roles/providers/templates/etc/systemd/network/10-eth0.network.manual
```diff
@@ -0,0 +1,8 @@
+[Match]
```
detiber, line 1: Do we need the nested directory structure here, any reason not to just do `images/capi/ansible/roles/providers/templates/photon-network-config` or similar?

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -47,3 +88,28 @@
...
+    src: etc/sysconfig/network-scripts/ifcfg-eth0
+    dest: /etc/sysconfig/network-scripts/ifcfg-eth0
+    owner: root
+    group: root
+    mode: '0644'
+  when: ansible_os_family == "RedHat"
+
+- name: Make sure device file exists
+  template:
+    src: etc/systemd/network/10-eth0.network.manual
```
detiber, line 103: missing source for this template in the PR?

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -12,13 +12,54 @@
...
+  when: ansible_os_family == "VMware Photon OS"
+
 - name: install Azure clients
   pip:
     executable: pip3
     name: "{{ packages }}"
   vars:
     packages:
       - azure-cli
+  when: ansible_distribution == "Ubuntu" or ansible_os_family == "VMware Photon OS"
```
detiber, line 28: Is the azure-cli rpm package used below not compatible with Photon?

detiber: As an aside, I would love to see us use the available ubuntu package rather than pip, but that is not directly related to this PR :)

images/capi/ansible/roles/providers/tasks/azure.yml
detiber: Ah, looks like there is an open issue against photon for chrony not being available: https://github.com/vmware/photon/issues/878

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -12,13 +12,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+- name: Install Azure clients.
```
detiber, line 15: Please update the task names to include context around when they are run and not run to help with troubleshooting potential future issues.

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -12,13 +12,54 @@
...
+    file: azure-cli
+    enabled: true
+    baseurl: https://packages.microsoft.com/yumrepos/azure-cli
+    gpgcheck: "{{ ansible_os_family == 'RedHat'|bool }}"
+    gpgkey: https://packages.microsoft.com/keys/microsoft.asc
+  when: ansible_os_family == "RedHat"
+
+- name: Refresh yum after repo addition
+  shell: 
+    cmd: yum updateinfo --enablerepo=azure-cli
```
codenrhoden, line 49: Is this step explicitly necessary since the next step uses `update_cache: yes`?

voor: This is actually leftovers from when I was originally doing CentOS7 and then switched to PhotonOS.  CentOS 7 didn't seem to acknowledge the `update_cache` for new repositories -- which is what the Azure CLI is.

images/capi/ansible/roles/providers/tasks/azure.yml
voor: Yes, there is a dependency on a particular library for Python3 that is not available in the Photon distribution, originally went down the RPM route as well, was a little disappointed when it didn't work.

images/capi/ansible/roles/providers/templates/etc/systemd/network/10-eth0.network.manual
voor: I'm okay with consensus on that one, this is typically just Ansible best practices.  @codenrhoden any strong opinion either way?

images/capi/ansible/roles/providers/tasks/azure.yml
voor: That issue might be including it by default in the AMI.  I actually had to request the Photon team cut an entirely new release of Photon because the Azure default included `docker` by default.

Let me see if the package is available, you're right that earlier stuff generated some certificate issues in Azure.

voor: There is not a `chrony` package in Photon, we should find out why and probably add it unless they're using something else for time.

detiber: I suspect they probably wanted to standardize on a single time sync daemon, but considering chrony is the only one that is supports ptp devices, and that is the best source of time for Azure, and it is recommended even on AWS (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html), I think it would benefit us to be able to standardize on it.

images/capi/packer/azure/packer.json
```diff
@@ -93,6 +106,31 @@
...
+      "client_id": "{{user `client_id`}}",
+      "client_secret": "{{user `client_secret`}}",
+      "vm_size": "{{user `vm_size`}}",
+      "resource_group_name": "{{user `resource_group_name`}}",
+      "build_resource_group_name": "{{user `resource_group_name`}}",
+      "capture_container_name": "{{user `capture_container_name`}}",
+      "capture_name_prefix": "capi-{{user `build_timestamp`}}",
+      "storage_account": "{{user `storage_account_name`}}",
+      "cloud_environment_name": "{{user `cloud_environment_name`}}",
+      "ssh_username": "builder",
```
CecileRobertMichon, line 126:
```suggestion
      "ssh_username": "packer",
```

for consistency

images/capi/packer/azure/packer.json
```diff
@@ -33,8 +34,18 @@
...
+    "remove_extra_repos": "false",
+    "photon_image_url": "https://photon3.blob.core.windows.net/vhds/photon-azure-3.0-a9bc127.vhd",
+    "custom_role": "",
+    "custom_role_name": "",
+    "http_proxy": "",
+    "https_proxy": "",
+    "headless": "true",
+    "no_proxy": "",
+    "redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",
+    "build_version": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}"
```
CecileRobertMichon, line 48: where is `build_version` used?

images/capi/ansible/roles/setup/defaults/main.yml
```diff
@@ -22,3 +22,5 @@ extra_rpms: ""
 
 disable_public_repos: false
 extra_repos: ""
+
+post_reboot_delay: 30
```
CecileRobertMichon, line 26: missing newline

images/capi/ansible/roles/providers/templates/etc/systemd/network/10-eth0.network.manual
```diff
@@ -0,0 +1,8 @@
...
+Name=eth0
+[Network]
+LinkLocalAddressing=no
+DHCP=ipv4
+[DHCP]
+UseDomains=true
+ClientIdentifier=mac
```
CecileRobertMichon, line 8: missing newline

images/capi/packer/azure/packer.json
```diff
@@ -115,8 +154,29 @@
         "--extra-vars",
         "{{user `ansible_common_vars`}}",
         "--extra-vars",
-        "{{user `ansible_extra_vars`}}"
+        "{{user `ansible_extra_vars`}}",
+        "--extra-vars",
+        "azure_local=True"
```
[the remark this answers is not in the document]
voor: This is an interesting one, I want to play around more with.  This will force an Ansible variable to always have `azure_local` be `True`.  That means you can build an image on Azure, qemu, vSphere, or AWS with the eventual landing place being Azure.  I was initially doing the entire Azure build on qemu and uploading the VHD, realized that wouldn't really align with how the other Azure images are built, so I switched.

images/capi/packer/azure/packer.json
```diff
@@ -93,6 +106,31 @@
         "distribution_version": "1804",
         "kubernetes_version": "{{user `kubernetes_semver`}}"
       }
+    },
+    {
+      "name": "vhd-photon-3",
+      "type": "azure-arm",
+      "os_type": "Linux",
+      "os_disk_size_gb": "5",
```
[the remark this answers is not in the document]
voor: photon is super small, the actual VHD that is uploaded to that blobstore is like 300MB, so I needed to explicitly set the disk size or else everything would fail from running out of disk, cloud-init fires off and grows the volume from 300MB to 5GB.

images/capi/packer/azure/packer.json
voor: photon is using the same ks as vsphere, where this is hard-coded as builder.  We'd need to get @codenrhoden to agree on this change as it impacts vsphere.

images/capi/packer/azure/packer.json
voor: This is a copy pasta from vSphere, the manifest was output into that folder, but it's not necessary for Azure (will remove it)

images/capi/ansible/roles/providers/tasks/azure.yml
voor: I'll reach out and see what they're using for NTP and if they want to consider chrony.

images/capi/packer/azure/packer.json
CecileRobertMichon: I don't have a strong opinion either way but the other Azure builds use "packer" so it'd be nice to have consistency.

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -31,19 +72,35 @@
...
+    group: systemd-network
+    mode: '0644'
+  when: ansible_os_family == "VMware Photon OS"
+
+- name: enable waagent service
+  systemd:
+    name: waagent
+    enabled: yes
+    state: started
+  when: ansible_os_family == "RedHat" or ansible_os_family == "VMware Photon OS"
```
detiber, line 106: This file should probably be aligned with the similar efforts for CentOS support here: https://github.com/kubernetes-sigs/image-builder/pull/287

voor:
Closing this, this was really more for personal education and understanding of the Azure image building process than something that I see as being maintained.

PR #269, opened 2020-06-29
Add 'exclude-nics' filter in vmtools configuration file in Ansible

SanikaGawhane:
Adding an `exclude-nics` filter can solve the current problem where no NICs are returned for a VM when the NIC limit is reached (vmtools has a 16 NIC limit).

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/vmware.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware.yml
@@ -76,3 +76,11 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+ 
+- name: Create provider vmtools config drop-in file
+  copy:
+    src: files/etc/vmware-tools/tools.conf
+    dest: /etc/vmware-tools/tools.conf
+    owner: root
+    group: root
+    mode: 0644
--- a/images/capi/ansible/roles/providers/files/etc/vmware-tools/tools.conf
+++ b/images/capi/ansible/roles/providers/files/etc/vmware-tools/tools.conf
@@ -1,0 +1,3 @@
+[guestinfo]
+exclude-nics=antrea-*,ovs-system,br*,flannel*,veth*,vxlan_sys_*,genev_sys_*,gre_sys_*,stt_sys_*,????????-??????
+


SanikaGawhane:
signed a Contributor License Agreement (CLA) with VMware credentials

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -76,3 +76,16 @@
...
+
+- name: Create provider vmtools drop-in directory
+  file:
+    path: /etc/sysconfig
+    state: directory
+    
+- name: Create provider vmtools config drop-in file
+  copy:
+    src: /files/etc/sysconfig/tools.conf
+    dest: /etc/sysconfig/tools.conf
```
codenrhoden, line 88: The location for this file should be `/etc/vmware-tools/tools.conf`.

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -76,3 +76,16 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+
+- name: Create provider vmtools drop-in directory
```
codenrhoden, line 80: This step shouldn't be necessary, since `open-vm-tools` is already installed as a package, and that package will pre-create this directory.

codenrhoden:
I also didn't notice at first that there is a binary file checked in. Please remove that.

images/capi/ansible/roles/providers/tasks/vmware.yml
codenrhoden: And by "this directory", I do mean `/etc/vmware-tools`

images/capi/ansible/roles/providers/files/etc/sysconfig/tools.conf
```diff
@@ -0,0 +1,3 @@
+[guestinfo]
```
codenrhoden, line 1: Let's please also move the location of this file within the role to `files/etc/vmware-tools/tools.conf`. I think it would be confusing to have it live in `sysconfig` but not destined for that path.

Once that is in place, do please squash your commits down to 1.

akutz:
nit: fix the title of the PR :)

images/capi/ansible/roles/providers/files/etc/sysconfig/tools.conf
SanikaGawhane: Done

codenrhoden:
/lgtm
Waiting to hear from a colleague on the potential addition of one more item to the filter. If it takes a while, we'll merge and just add it later (if it's needed).

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -76,3 +76,11 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+ 
+- name: Create provider vmtools config drop-in file
+  copy:
+    src: /files/etc/vmware-tools/tools.conf
```
codenrhoden, line 82: `src` here needs to be a relative path, so it should not start with `/`. The `/` is making it be an absolute path, so when I build, I see:

```sh
    vmware-iso: fatal: [default]: FAILED! => {"changed": false, "msg": "Could not find or access '/files/etc/vmware-tools/tools.conf' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
 ```

PR #271, opened 2020-07-02
Add auditd to photon OS

EleanorRigby:
Add auditd to Node image for photon OS

  - Add auditd service
  - Enable on boot
  - Recreate log dir once sysprep cleans log dirs
  - Add containerd audit rules
  - Make sure all auditd customizations happen for photon OS only

  SYSPREP ROLE
  - Include Photon OS tasks at end.
  - Move machine mode fact out of OS specific sysprep role to set
  at begining of role execution.

### DIFF
--- a/images/capi/ansible/roles/node/tasks/photon.yml
+++ b/images/capi/ansible/roles/node/tasks/photon.yml
@@ -1,0 +1,28 @@
+# Copyright 2020 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+- name: Ensure auditd is running and comes on at reboot
+  service:
+    name: auditd
+    state: started
+    enabled: yes
+
+- name: configure auditd rules for containerd
+  copy:
+    src: photon-os/etc/audit/rules.d/containerd.rules
+    dest: /etc/audit/rules.d/containerd.rules
+    owner: root
+    group: root
+    mode: 0644
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -132,7 +132,7 @@
   - { path: /var/log, state: directory }
   - { path: /var/log/sa, state: directory }
 
-- name: Create a log directory for kube-apiserver audit logs
+- name: Create log directories for k8s
   file:
     path: "{{ item.path }}"
     state: "{{ item.state }}"
@@ -143,6 +143,16 @@
   - { path: /var/log/kubernetes, state: directory, mode: "0755" }
   - { path: /var/log/kubernetes/audit.log, state: touch, mode: "0600" }
   when: kubernetes_semver != '' # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
+
+- name: Create a log directory for auditd in Photon OS
+  file:
+    path: "/var/log/audit"
+    state: directory
+    mode: "0640"
+    owner: root
+    group: root
+  when: kubernetes_semver != '' and ansible_os_family == "VMware Photon OS" # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
+
 
 - name: Truncate shell history
   file:
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -46,6 +46,7 @@ common_debs:
 - socat
 
 common_photon_rpms:
+- audit
 - conntrack-tools
 - distrib-compat
 - ebtables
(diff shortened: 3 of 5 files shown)


EleanorRigby:
cc: @codenrhoden

images/capi/ansible/roles/setup/tasks/photon.yml
```diff
@@ -45,3 +45,16 @@
...
+    src: photon-os/etc/audit/rules.d/containerd.rules
+    dest: /etc/audit/rules.d/containerd.rules
+
+# Enable causes auditd to correctly
+# create auto-start symlink.
+# Auditd is diabled by default in Photon OS: https://github.com/vmware/photon/blob/366cf888f3b3c7e54f883c2e3223f5a752fe0e60/SPECS/audit/audit.spec#L104
+- name: configure auditd to start on boot
+  command: "{{ item }}"
+  loop:
+    - systemctl enable auditd
```
voor, line 60: I'd recommend using the service module for Ansible, 
```suggestion
- name: Ensure auditd is running and comes on at reboot
  service:
    name: auditd
    state: started
    enabled: yes
```

EleanorRigby: @voor : Thanks for the review. I want to `enable` the service to boot on the start of OS which is different from `starting` the service itself (not required). 
I think I will have to use the `enable` flag and the `state` flag can be anything. I will commit my changes on the next iteration.

EleanorRigby:
Also, I think `auditd` tasks need to go in node role rather than setup as `setup ` applies to haproxy as well. Making that change

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -0,0 +1,28 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+# Enable causes auditd to correctly
+# create auto-start symlink.
```
codenrhoden, line 17: nit: collapse these into one line. I also don't think this comment is strictly necessary -- for every service we install, we make sure that it is enabled on boot.

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -0,0 +1,28 @@
...
+# Auditd is diabled by default in Photon OS: https://github.com/vmware/photon/blob/366cf888f3b3c7e54f883c2e3223f5a752fe0e60/SPECS/audit/audit.spec#L104
+- name: Ensure auditd is running and comes on at reboot
+  service:
+    name: auditd
+    state: started
+    enabled: yes
+
+- name: configure auditd rules for containerd
+  template:
+    src: photon-os/etc/audit/rules.d/containerd.rules
```
codenrhoden, line 27: This file isn't a template (it's not a `.j2` file), so I'd rather see this placed at `roles/node/files/etc/audit/rules.d/containerd.rules` and use the `copy` module, like https://github.com/kubernetes-sigs/image-builder/blob/00377e8ea77e7ad42b29bc5853eb426fce186118/images/capi/ansible/roles/haproxy/tasks/main.yml#L52-L58

images/capi/ansible/roles/setup/tasks/photon.yml
```diff
@@ -44,4 +44,4 @@
 
 - name: install extra RPMs
   command: tdnf install {{ extra_rpms }} -y
-  when: extra_rpms != ""
+  when: extra_rpms != ""
```
codenrhoden, line 47: nit: can we skip the commit of a whitespace change?

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -12,15 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+# Used across all OS
+- name: Define file modes
+  set_fact:
+    last_log_mode:   "0644"
+    machine_id_mode: "0444"
+
```
codenrhoden, line 20: Did you run into something specific that required this change? I admit I don't know why the Ubuntu filemodes here are different than the RPM-based distros, but they *are* different.

From `debian.yml`:
```yaml
  set_fact:	
    last_log_mode:   "0664"	
    machine_id_mode: "0644"
```

And from `photon.uml` and `redhat.yml` (identical):
```yaml
    last_log_mode:   "0644"	
    machine_id_mode: "0444"
```

So for Debian systems, `last_log_mode` adds group writable, and `machine_id_mode` adds user writeable. This change would apply the RPM settings to all node types (e.g. would change it for Debian) and i don't think we need/want that without a specific reason.

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -132,7 +135,7 @@
   - { path: /var/log, state: directory }
   - { path: /var/log/sa, state: directory }
 
-- name: Create a log directory for kube-apiserver audit logs
+- name: Create a log directory for kube-apiserver, k8s audit logs
```
codenrhoden, line 138: Not sure how this change is relevant to the PR? If this directory and file are used by more than kube-apiserver, lets' just say "k8s" instead of "kube-apiserver, k8s".

images/capi/ansible/roles/sysprep/tasks/main.yml
codenrhoden: After finishing my review, I suspect this was done because of moving the Photon import to the end of the task, so for Photon these were no longer defined. I'm doubling down on my suggestion to not change that, and to use a condtiional on the OS to create the audit dir instead.

EleanorRigby: Yes I am sorry, I failed to notice the different machine mode for different OSs.

Also, I just discovered ansible's `when` can have `and` gate with conditions

images/capi/ansible/roles/node/tasks/photon.yml
EleanorRigby: Done

EleanorRigby:
> /lgtm
> /approve
> 
> Thanks @EleanorRigby! I verified that I could build this image, and booted it up and checked out the new audit log and it was full of containerd entries. I think the auditd rules as they are may be more globally applicable beyond Photon (e.g. to AL2 as well) and if we find that to be the case, we'll start dropping them in for other OS's as well.

Agreed.

PR #275, opened 2020-07-03
Add Photon 3 AMI build

voor:
This adds Photon 3 to AWS.

I noticed the AWS CLI was being installed with pip rather than pip3, so switched that when I moved Ubuntu to its own definition for installing the AWS CLI.

There's a remediation that adds a `builder` account since some of the tasks expect that username and the photon AMI only has root normally.


voor:
Closing this, PhotonOS support on AWS is less than stellar, and this would not be well maintained.

PR #276, opened 2020-07-06
Creating a container for image builder.

voor:
This is the first pass at taking a Dockerfile that has been in use for a while and adding it directly into image builder, with a hopeful eventual goal that it could be distributed with releases.

Things that could be improved here:

- Packer is installed earlier instead of with the `make deps` because `./.bin/` is not in the $PATH, so it might make more sense to just add that to the $PATH and use that instead, this would keep things in alignment.

Previously used with `custom_role` and `custom_role_name`, I'd like to figure out a better way to add this back in here, although I guess it's just mounting into the container and referencing it.

### DIFF
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:focal
+
+RUN apt-get update && apt-get install -y apt-transport-https ca-certificates python3-pip curl wget git rsync vim unzip build-essential \
+    && useradd -ms /bin/bash imagebuilder \
+    && apt-get purge --auto-remove -y \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG ARCH
+ARG PASSED_IB_VERSION
+
+COPY . /home/imagebuilder
+
+USER imagebuilder
+
+ENV PATH="/home/imagebuilder/.local/bin:${PATH}"
+
+ENV PACKER_ARGS ''
+ENV PACKER_VAR_FILES ''
+ENV IB_VERSION "${PASSED_IB_VERSION}"
+
+WORKDIR /home/imagebuilder/
+
+RUN make deps
+
+ENTRYPOINT [ "/usr/bin/make" ]
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -59,6 +59,7 @@ deps-azure:
 	hack/ensure-ansible.sh
 	hack/ensure-packer.sh
 	hack/ensure-jq.sh
+	hack/ensure-azure-cli.sh
 
 .PHONY: deps-do
 deps-do: ## Installs/checks dependencies for DigitalOcean builds
@@ -84,6 +85,15 @@ deps-qemu: ## Installs/checks dependencies for QEMU builds
 deps-qemu:
 	hack/ensure-ansible.sh
 	hack/ensure-packer.sh
+
+## --------------------------------------
+## Container variables
+## --------------------------------------
+REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+IMAGE_NAME ?= cluster-node-image-builder
+CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
+TAG ?= dev
+ARCH ?= amd64
 
 ## --------------------------------------
 ## Packer flags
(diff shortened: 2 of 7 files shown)


images/capi/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:focal
+
+ENV PACKER_ARGS ''
+ENV PACKER_VAR_FILES ''
+
+RUN apt-get update && apt-get install -y apt-transport-https ca-certificates ansible curl wget git rsync vim jq unzip build-essential \
```
moshloop, line 20: I think we should pin the ansible version as it has a tendency to break between versions

dims: 💯

images/capi/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENV PACKER_ARGS ''
+ENV PACKER_VAR_FILES ''
+
+RUN apt-get update && apt-get install -y apt-transport-https ca-certificates ansible=2.9.6+dfsg-1 curl wget git rsync vim jq unzip build-essential \
+    && useradd -ms /bin/bash ansible \
+    && curl -sL https://releases.hashicorp.com/packer/1.6.0/packer_1.6.0_linux_amd64.zip -o /tmp/packer.zip && unzip /tmp/packer.zip && mv packer /usr/local/bin/packer \
+    && rm -rf /tmp/packer.zip \
+    && apt-get purge --auto-remove -y \
+    && rm -rf /var/lib/apt/lists/*
```
CecileRobertMichon, line 25: thoughts on adding the various cloud CLIs (eg. az cli)? we need it to login for Azure builds

voor: That makes sense for Azure and is a good suggestion, you don't need the vSphere or AWS cli.

detiber: +1 to adding any cloud provider specific dependencies needed to the image

images/capi/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:focal
+
+ENV PACKER_ARGS ''
+ENV PACKER_VAR_FILES ''
+
+RUN apt-get update && apt-get install -y apt-transport-https ca-certificates ansible=2.9.6+dfsg-1 curl wget git rsync vim jq unzip build-essential \
```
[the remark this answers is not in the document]
JTarasovic: `make deps` would install packer and ansible wouldn't it? If so, then we'd just need `PATH` to include the install directory?

voor:
Okay, this ended up being a fairly big change, so please take another look.  Some things *new* since earlier today:

* Adds in Azure CLI, this uses the same mechanism, and adds it to the `deps` for Azure.
* Moves everything of note into the same make-deps process, creates uniformity around using the standard `$HOME/.local/bin` location.
* Changes the user in the container to make it less generic
* Fixes an issue I noticed where the git version is not properly put into the container due to the fact we're only copying a subdirectory

images/capi/Makefile
```diff
@@ -124,7 +134,7 @@ PACKER_NODE_FLAGS := $(foreach f,$(abspath $(COMMON_NODE_VAR_FILES)),-var-file="
 				$(PACKER_FLAGS)
 PACKER_HAPROXY_FLAGS := $(foreach f,$(abspath $(COMMON_HAPROXY_VAR_FILES)),-var-file="$(f)" ) \
 				$(PACKER_FLAGS)
-PACKER_VAR_FILES := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" )
+ABSOLUTE_PACKER_VAR_FILES := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" )
```
codenrhoden, line 137: For posterity, can you clarify why this change is needed? On the surface, it appears superficial. (I think it's because we are passing the env var in through Docker, but want to confirm)

images/capi/hack/ensure-ansible.sh
```diff
@@ -26,18 +26,18 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 if command -v ansible >/dev/null 2>&1; then exit 0; fi
 
-mkdir -p .bin && cd .bin
-
-if ! command -v python >/dev/null 2>&1; then
-  echo "Python binary must be in \$PATH" 1>&2
+if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
voor: `ubuntu:focal` would leverage python2 with `python` commands, so this is to assist in those scenarios to make sure it's leveraging python3, otherwise ansible was crapping out inside the container.

images/capi/Makefile
voor: When you pass in a make parameter from the cli it makes that read-only -- so in this example `PACKER_VAR_FILES` is write-protected as a result, by leveraging a temporary variable here instead of trying to overwrite the original we allow for those passed variables to remain read-only.  This was causing issues where `PACKER_VAR_FILES` ended up as the original passed in value `aws.json foo.json bar.json` instead of `-var-file="aws.json" -var-file="foo.json" -var-file="bar.json"`

images/capi/hack/ensure-ansible.sh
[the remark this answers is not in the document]
codenrhoden: ah, that makes sense. Okay, so I realated (but non PR-blocking) question -- why Focal instead of debian-slim? I haven't tried a different base container OS myself, but other Dockerfile's I had seen used that as a base. The resulting build from this ends up >2.2GB, which seems large to me for "just" adding Packer and Ansible.

I'm thinking ahead and wondering if after this is in place, slimming down the image is obtainable.

dims:
happy to see this effort! looks like this has a bunch of people on it.

/uncc

Issue #280, opened 2020-07-07
Enable GOSS tests for OVA builder


codenrhoden:
We should run the same tests for the OVAs that we do for the AMIs.

codenrhoden:
fixed by #286
/close

Issue #279, opened 2020-07-07
Move GOSS vars to common location


codenrhoden:
Since we want to use GOSS for more cloud providers than just AWS, I think defining `goss_url` in a different var file (e.g. in `packer/config/goss.json` or similar) would make sense so it is only defined in one place, instead of each provider's `packer.json`.

Additionally, let's turn the [tests file](https://github.com/kubernetes-sigs/image-builder/blob/00377e8ea77e7ad42b29bc5853eb426fce186118/images/capi/packer/ami/packer.json#L133) into a var (set in the same place as `goss_url`) so that it can be overridable by the user.

codenrhoden:
fixed by #286
/close

PR #286, opened 2020-07-10
Add GOSS provisioner to node OVA builder

### LINKED ISSUE #278: Update GOSS version
The GOSS version referenced at https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ami/packer.json#L23 is a very old version. Let's update it to the latest.

EleanorRigby:
- Runs goss verifier on node OVA
  - Extract out goss args in common config
  - Update GOSS version to 0.3.13

Fixes #278 #279 #280

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -20,7 +20,6 @@
     "extra_debs": "",
     "extra_repos": "",
     "extra_rpms": "",
-    "goss_url": "https://github.com/aelsabbahy/goss/releases/download/v0.3.2/goss-linux-amd64",
     "iam_instance_profile": "",
     "ib_version": "{{env `IB_VERSION`}}",
     "kms_key_id": "",
@@ -128,9 +127,12 @@
     },
     {
       "type": "goss",
+      "arch": "{{user `goss_arch`}}",
+      "format": "{{user `goss_format`}}",
+      "tests": ["packer/goss/goss.yaml"],
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
-      "tests": ["packer/goss/goss.yaml"]
+      "version": "{{user `goss_version`}}"
     }
   ],
   "post-processors": [
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -108,7 +108,8 @@ endif
 COMMON_NODE_VAR_FILES :=	packer/config/kubernetes.json \
 					packer/config/cni.json \
 					packer/config/containerd.json \
-					packer/config/ansible-args.json
+					packer/config/ansible-args.json \
+					packer/config/goss-args.json
 
 COMMON_HAPROXY_VAR_FILES := packer/ova/packer-common.json \
 					packer/config/ansible-args.json
--- a/images/capi/packer/config/goss-args.json
+++ b/images/capi/packer/config/goss-args.json
@@ -1,0 +1,6 @@
+{
+  "goss_arch": "amd64",
+  "goss_url": "",
+  "goss_version": "0.3.13",
+  "goss_format": "json"
+}
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -155,7 +155,15 @@
         "--extra-vars",
         "{{user `ansible_extra_vars`}}"
       ]
-   }
+   },{
+      "type": "goss",
+      "arch": "{{user `goss_arch`}}",
+      "format": "{{user `goss_format`}}",
+      "tests": ["packer/goss/goss.yaml"],
+      "url": "{{user `goss_url`}}",
+      "use_sudo": true,
+      "version": "{{user `goss_version`}}"
+    }
   ],
   "post-processors": [
     {


EleanorRigby:
cc: @codenrhoden

images/capi/packer/ami/packer.json
```diff
@@ -20,7 +20,8 @@
     "extra_debs": "",
     "extra_repos": "",
     "extra_rpms": "",
-    "goss_url": "https://github.com/aelsabbahy/goss/releases/download/v0.3.2/goss-linux-amd64",
+    "goss_debug": null,
```
codenrhoden, line 23: What do you think about leaving this definition out -- at letting it be set by default in `goss-args.yml`. Setting this to `null` means that this *has* to be set by a var file, which makes sense for `goss_url` since nothing works without it, but for `goss_debug`, the default behavior with an empty string is probably fine.

EleanorRigby: I can fix this. Sounds right to me.

EleanorRigby:
> @EleanorRigby I was just taking this for a spin... A couple of weird things (nothing to do with you) that I noticed...
> 
> It appears that the Goss Packer provisioner has a hardcoded path to download the binary to:
> 
> ```shell
> ==> vmware-iso: Provisioning with Goss
>     vmware-iso: Installing Goss from, https://github.com/aelsabbahy/goss/releases/download/v0.3.13/goss-linux-amd64
>     vmware-iso: Downloading Goss to /tmp/goss-0.3.9-linux-amd64
> ==> vmware-iso:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
> ==> vmware-iso:                                  Dload  Upload   Total   Spent    Left  Speed
> ==> vmware-iso: 100   627  100   627    0     0   1628      0 --:--:-- --:--:-- --:--:--  1628
> ==> vmware-iso: 100 11.6M  100 11.6M    0     0  7245k      0  0:00:01  0:00:01 --:--:-- 11.1M
>     vmware-iso: goss version v0.3.13
> ```
> 
> Notice that we do download and verify that it is 0.3.13, but it downloads it to 0.3.9. How weird. But there isn't really any problem with that (just confusing output).
> 
> The main problem I am seeing is that despite the provisioner **saying** there is a debug flag, GOSS doesn't accept one:
> 
> ```shell
> ==> vmware-iso: Running goss tests...
>     vmware-iso: Incorrect Usage. flag provided but not defined: -d
>     vmware-iso:
>     vmware-iso: NAME:
>     vmware-iso:    goss - Quick and Easy server validation
> ==> vmware-iso: 2020/07/10 20:19:09 flag provided but not defined: -d
>     vmware-iso:
>     vmware-iso: USAGE:
>     vmware-iso:    goss-0.3.9-linux-amd64 [global options] command [command options] [arguments...]
>     vmware-iso:
>     vmware-iso: VERSION:
>     vmware-iso:    v0.3.13
>     vmware-iso:
>     vmware-iso: COMMANDS:
>     vmware-iso:      validate, v  Validate system
>     vmware-iso:      serve, s     Serve a health endpoint
>     vmware-iso:      render, r    render gossfile after imports
>     vmware-iso:      autoadd, aa  automatically add all matching resource to the test suite
>     vmware-iso:      add, a       add a resource to the test suite
>     vmware-iso:      help, h      Shows a list of commands or help for one command
>     vmware-iso:
>     vmware-iso: GLOBAL OPTIONS:
>     vmware-iso:    --gossfile value, -g value  Goss file to read from / write to (default: "./goss.yaml") [$GOSS_FILE]
>     vmware-iso:    --vars value                json/yaml file containing variables for template [$GOSS_VARS]
>     vmware-iso:    --vars-inline value         json/yaml string containing variables for template (overwrites vars) [$GOSS_VARS_INLINE]
>     vmware-iso:    --package value             Package type to use [apk, dpkg, pacman, rpm]
>     vmware-iso:    --help, -h                  show help
>     vmware-iso:    --version, -v               print the version
> ```
> 
> Did you see this as well?

This happens because [here](https://github.com/YaleUniversity/packer-provisioner-goss/blob/2c65c16200441bcbc8c200f7d80b01ff877fe18a/packer-provisioner-goss.go#L119), packer-goss-provisioner deduces download path from the `version` passed in config here 
```
"provisioners" : [
  {
    "type": "goss",
    "version": "0.3.2",
    "arch": "amd64",
    "url":"https://github.com/aelsabbahy/goss/releases/download/vVERSION/goss-linux-ARCH",
    "tests": [
      "goss/goss.yaml"
    ],
    "downloadPath": "/tmp/goss-VERSION-linux-ARCH",
    "remote_folder": "/tmp",
    "remote_path": "/tmp/goss",
    "skipInstall": false,
    "skip_ssl": false,
    "use_sudo": false,
    "format": "",
    "goss_file": "",
    "vars_file": "",
    "username": "",
    "password": "",
    "debug": false,
    "retry_timeout": "0s",
    "sleep": "1s"
  }
]
```
So I am guessing it defaults to `0.3.9` somehow. And moreover, we don't have to pass `goss_url`. It puts the URL together from `version` and `arch` in [config](https://github.com/YaleUniversity/packer-provisioner-goss/blob/2c65c16200441bcbc8c200f7d80b01ff877fe18a/packer-provisioner-goss.go#L113). I am going to make that change.

Will also investigate the debug flag. 

/hold

codenrhoden:
If we can deduce the GOSS url, great. But one thing we need to be able to preserve is to **override** the URL to something else. That came in to meet a specific need from @voor in https://github.com/kubernetes-sigs/image-builder/pull/218 -- namely, to be able to download GOSS from an alternative URL in air-gapped environments.

I don't remember for sure, but I think before that PR we probably weren't using the URL directly at all. Hopefully it behaves such that it will deduce the URL, and just go with that if the URL field is empty, but if it is not empty, it will use that instead (with the side effect that the target binary target may be different version -- I'm okay with that since the logs show what the real version is and the combination that produces that should be rare if I'm understanding correctly).

images/capi/packer/ami/packer.json
```diff
@@ -20,7 +20,8 @@
...
+    "goss_url": null,
```
voor, line 24: This looks like a typo, it should probably be `"{{user `goss_url`}}"` or else we'll always use the default URL format.

EleanorRigby:
@codenrhoden : I removed `debug` totally and added a format option to be able to see goss failure form packer output. I compiled various amazon and photon images and they worked well. Please re-review.

codenrhoden:
@EleanorRigby tests work for me as well -- good work!

I have one thing to ask your opinion on. I don't think it's necessary to put the `goss_*` vars into the `packer.json` files -- since we include `goss-args.json` via the Makefile, they will always be present. I'd like to see us cutdown on clutter in those JSON files (I know the damage is done, but I'm trying to help it get cleaned up), and putting those vars in there are is a bit redundant. Only reason to have them there is if we wanted someone to run `packer inspect` to see what the options are, and since I'd really like to see everything driven my Make, it's a moot point.

But this is my opinion, and not something worth holding this up over. Do you have any thoughts on it now that you've played with it a bit?

EleanorRigby:
I was wondering if it is required to put all variables in packer.json or not. But your comment makes it clear now.  

So, I think it is not required for all vars to be in that file at all. I will make the change

EleanorRigby:
@codenrhoden addressed all comments.

codenrhoden:
This looks great! All checked out okay for me locally.

/lgtm
/approve

PR #287, opened 2020-07-11
Centos support for Azure

akhilesh2410:
This PR add :
   CentOS-7 Support for Azure cloud
     - User can create CentOS VHD and SIG image same as ubuntu 
     - To build a CentOS VHD run "make build-azure-vhd-centos-7"
     - To build a CentOS SIG image "make build-azure-sig-centos-7"


akhilesh2410:
CLA signed

jzhoucliqr:
@CecileRobertMichon

images/capi/ansible/roles/node/tasks/main.yml
```diff
@@ -76,3 +76,10 @@
     sysctl_set: yes
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
+
+- name: Disable swap memory
```
CecileRobertMichon, line 80: Does Centos depend on this change? If not and this is just a parallel improvement, can you please open a separate PR for it, given that this would impact all existing CAPI image builds.

images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -19,7 +19,7 @@ common_rpms:
 - ebtables
 - ntp
 - open-vm-tools
-- python2-pip
+- python3-pip
```
CecileRobertMichon, line 22: this might affect other CAPI builds as well, cc @detiber @codenrhoden

images/capi/ansible/roles/node/tasks/main.yml
akhilesh2410: While testing I figure out that the control plane was not up and running due to swap partition, so I added code to disable it all possible ways.

CecileRobertMichon:
/assign

I ran a quick test build of the Azure ubuntu VHD builder from this branch and it succeeded

images/capi/ansible/roles/node/defaults/main.yml
detiber: No objections from me, I believe there was a similar discussion related to work that @voor was doing to try and standardized pip-installed resources using python3 rather than python2

images/capi/ansible/roles/providers/tasks/redhat.yml
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: add epel repo
+  yum:
+    name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
+    state: present
+    lock_timeout: 60
```
detiber, line 19: There should already be a mechanism for enabling epel, I don't think we should have multiple tasks among different roles to do it otherwise it could lead to confusion.

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -12,13 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+- import_tasks: redhat.yml
+  when: ansible_os_family == "RedHat"
```
detiber, line 16: If Red Hat specific tasks are being broken out into a separate tasks file, Ubuntu ones should likely also be broken out as well.

images/capi/ansible/roles/providers/tasks/redhat.yml
```diff
@@ -0,0 +1,50 @@
...
+
+- name: prereqs for Azure CLI. Attention! This installs a compiler on the system
```
detiber, line 21: Is it not possible to use the yum repo provided by Azure? https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-yum?view=azure-cli-latest

images/capi/ansible/roles/node/tasks/main.yml
detiber: It seems like it's an issue related to the source image, generally the source images we are using elsewhere do not enable swap by default.

images/capi/packer/azure/packer.json
```diff
@@ -62,6 +62,30 @@
         "kubernetes_version": "{{user `kubernetes_semver`}}"
       }
     },
+    {
+      "name": "vhd-centos-7",
+      "type": "azure-arm",
+      "os_type": "Linux",
+      "image_publisher": "OpenLogic",
```
detiber, line 69: Is there a better official image we can use? It feels a bit off to use a CentOS image published by "OpenLogic" here. @CecileRobertMichon is there an official image that is provided by the CentOS project themselves?

images/capi/ansible/roles/node/tasks/main.yml
akhilesh2410: It could be, but what I think if we disable it from here, there is no harm on the image which already have disabled swap, and in case of swap enable, it will disable it, so would work in both cases.
wdyt ?

images/capi/ansible/roles/providers/tasks/azure.yml
akhilesh2410: Ack.

images/capi/ansible/roles/providers/tasks/redhat.yml
akhilesh2410: ok.. will check and modify code accordingly.

images/capi/ansible/roles/providers/tasks/redhat.yml
akhilesh2410: IIRC, I hit an issue with it, will double-check with yum, and update code if needed.

akhilesh2410:
@CecileRobertMichon  @detiber: code updated accordingly.

images/capi/ansible/roles/providers/tasks/redhat.yml
CecileRobertMichon: I wonder if the issue you hit was related to https://github.com/kubernetes-sigs/image-builder/pull/267#discussion_r446514203, cc @voor

images/capi/packer/azure/packer.json
CecileRobertMichon: Looks like that's the main CentOS image on the Azure marketplace https://docs.microsoft.com/en-us/azure/virtual-machines/linux/tutorial-manage-vm#understand-vm-images

```
az vm image list --output table
You are viewing an offline list of images, use --all to retrieve an up-to-date list
Offer          Publisher               Sku                 Urn                                                             UrnAlias             Version
-------------  ----------------------  ------------------  --------------------------------------------------------------  -------------------  ---------
CentOS         OpenLogic               7.5                 OpenLogic:CentOS:7.5:latest                                     CentOS               latest
CoreOS         CoreOS                  Stable              CoreOS:CoreOS:Stable:latest                                     CoreOS               latest
debian-10      Debian                  10                  Debian:debian-10:10:latest                                      Debian               latest
openSUSE-Leap  SUSE                    42.3                SUSE:openSUSE-Leap:42.3:latest                                  openSUSE-Leap        latest
RHEL           RedHat                  7-LVM               RedHat:RHEL:7-LVM:latest                                        RHEL                 latest
SLES           SUSE                    15                  SUSE:SLES:15:latest                                             SLES                 latest
UbuntuServer   Canonical               18.04-LTS           Canonical:UbuntuServer:18.04-LTS:latest                         UbuntuLTS            latest
WindowsServer  MicrosoftWindowsServer  2019-Datacenter     MicrosoftWindowsServer:WindowsServer:2019-Datacenter:latest     Win2019Datacenter    latest
WindowsServer  MicrosoftWindowsServer  2016-Datacenter     MicrosoftWindowsServer:WindowsServer:2016-Datacenter:latest     Win2016Datacenter    latest
WindowsServer  MicrosoftWindowsServer  2012-R2-Datacenter  MicrosoftWindowsServer:WindowsServer:2012-R2-Datacenter:latest  Win2012R2Datacenter  latest
WindowsServer  MicrosoftWindowsServer  2012-Datacenter     MicrosoftWindowsServer:WindowsServer:2012-Datacenter:latest     Win2012Datacenter    latest
WindowsServer  MicrosoftWindowsServer  2008-R2-SP1         MicrosoftWindowsServer:WindowsServer:2008-R2-SP1:latest         Win2008R2SP1         latest
```

images/capi/ansible/roles/providers/tasks/redhat.yml
akhilesh2410: I really can't able to recollect since I tried this couple of months ago, but at that also I follow https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-yum?view=azure-cli-latest to create the image and fails to install Azure-CLI. Maybe it is because of the reason you mentioned.
But this time is working perfectly fine.

images/capi/ansible/roles/node/tasks/main.yml
akhilesh2410: @detiber : Made changes accordingly by adding a check to see if swap partition is there.
Kindly take a look.

images/capi/ansible/roles/providers/tasks/debian.yml
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+---
+- name: install Azure clients
+  pip:
+    executable: pip3
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - azure-cli
+  when: ansible_distribution == "Ubuntu"
```
detiber, line 22: I'm not sure if there is anything actually Ubuntu specific in this file, but assuming not, then all of the `when` conditionals can likely be removed, since this file is already conditionally included based on the os family.

images/capi/ansible/roles/providers/tasks/redhat.yml
```diff
@@ -0,0 +1,29 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+- name: import the Microsoft repository key
+  command: rpm --import https://packages.microsoft.com/keys/microsoft.asc
+  become: true
```
detiber, line 17: I don't believe it's necessary to use `become: true` here
```suggestion
```

images/capi/ansible/roles/providers/tasks/redhat.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- name: copy azure-cli gpg key file
+  template:
+    src:  azure-cli.repo
+    dest: /etc/yum.repos.d/azure-cli.repo
+    mode: 0644
+  become: true
```
detiber, line 24: I don't believe it's necessary to use `become: true` here
```suggestion
```

images/capi/ansible/roles/providers/tasks/redhat.yml
```diff
@@ -0,0 +1,29 @@
...
+---
+- name: import the Microsoft repository key
+  command: rpm --import https://packages.microsoft.com/keys/microsoft.asc
+  become: true
+
+- name: copy azure-cli gpg key file
+  template:
+    src:  azure-cli.repo
+    dest: /etc/yum.repos.d/azure-cli.repo
+    mode: 0644
```
detiber, line 23: You can probably use the `yum_repository` module here rather than creating creating the file from template here:
```yaml
- name: Add the yum repository for the azure cli
  yum_repository:
    name: azure_cli
    description: Azure CLI
    baseurl: https://packages.microsoft.com/yumrepos/azure-cli
    gpgkey: https://packages.microsoft.com/keys/microsoft.asc
    gpgcheck: yes
    enabled: yes
```

images/capi/ansible/roles/providers/tasks/debian.yml
detiber: not for this PR, but we should probably follow up and install the cli from package here as well: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-apt?view=azure-cli-latest

@CecileRobertMichon any thoughts on that?

images/capi/ansible/roles/providers/tasks/redhat.yml
akhilesh2410: ack

images/capi/ansible/roles/providers/tasks/redhat.yml
akhilesh2410: Code has been modified as per review comments @detiber . Thanks

detiber:
lgtm

Thanks for all the work you've put into this @akhilesh2410

images/capi/ansible/roles/node/defaults/main.yml
codenrhoden: I know this is on me for not reviewing in time, but this breaks CentOS builds for OVAs and AMIs for sure, in more ways than one. I'll have to fix since this is already in.

PR #288, opened 2020-07-13
Add pretty print option to GOSS output

EleanorRigby:


EleanorRigby:
/hold

/blocked on https://github.com/YaleUniversity/packer-provisioner-goss/pull/30

EleanorRigby:
/unhold

@codenrhoden : this is ready for review & merge. Packer goss provisioner has a new release to support this. 

https://github.com/YaleUniversity/packer-provisioner-goss/tree/v1.1.0

EleanorRigby:
@codenrhoden : I updated goss-provisioner version and added dependency.

It fails if we use old binary with
```
Error: Failed to prepare build: "amazon-2"

1 error occurred:
	* unknown configuration key: "goss_format_options"; raws is []interface
```

codenrhoden:
/lgtm
/approve

Okay, this works for me. I have a follow-on issue i'm going to open up. I'd like to see the `ensure-goss.sh` script updated to automatically update the plugin if it is already present/installed. Right now the script sees that one is there and just exits. Instead, we need to check that it is a the correct version -- otherwise the test run will still fail.

PR #255, opened 2020-06-17
Replace deprecated configuration keys for Packer 1.6

jiatongw:
Packer version needs to upgrade to v1.6.0+ in order to use the building from remote vSphere option. This PR replaced some deprecated configuration keys for `builder type: vsphere-iso`

Fixes: #253

PR #289, opened 2020-07-14
capi: remove iso_checksum_type in qemu for packer 1.6

dongsupark:
Packer v1.6.0 or newer does not support the config `iso_checksum_type`.
So it simply fails with the following message:

```
1 error occurred:
        * Deprecated configuration key: 'iso_checksum_type'. Please call `packer fix` against your template to update your template to be compatible with the current version of Packer. Visit https://www.packer.io/docs/commands/fix/ for more detail.
```

It addresses similar issues as https://github.com/kubernetes-sigs/image-builder/pull/255, but this PR is about qemu, not ova.


codenrhoden:
/approve

This would be QEMU's equivalent to the fix that when in for OVAs here: https://github.com/kubernetes-sigs/image-builder/pull/255

Issue #283, opened 2020-07-07
Use GOSS to verify that container images are pre-pulled.


codenrhoden:
We can use GOSS to check that containerd has all the images we expect to be in the image.

ykakarap:
@codenrhoden may I work on this?

codenrhoden:
/assign @ykakarap 
yep!

EleanorRigby:
can be closed

PR #291, opened 2020-07-15
Add goss spec for checking k8s package and k8s image presence

### LINKED ISSUE #282: Use Goss to validate installed K8s packages and binaries
The beginning of this is in #277, but we want to do a final pass with GOSS to verify that all the expected K8s executables are in place. The first pass at this can be looking for the expected packages, but this doesn't cover the use case were components are installed from the HTTP source. That can be done in a follow up.

EleanorRigby:
fixes #282 #283

Also, 
  - Add skeleton to add more goss files.
  GOSS Provisioner
  - Transfer whole goss directory to the remote machine
  - Specify goss entry point file
  - Check the presence of containerd and absence of dockerd
  - Check if cloud-init is installed.

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -139,7 +139,8 @@
       "arch": "{{user `goss_arch`}}",
       "format": "{{user `goss_format`}}",
       "format_options": "{{user `goss_format_options`}}",
-      "tests": ["packer/goss/goss.yaml"],
+      "goss_file": "{{user `goss_entry_file`}}",
+      "tests": ["{{user `goss_tests_dir`}}"],
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
       "version": "{{user `goss_version`}}"
--- a/images/capi/packer/config/goss-args.json
+++ b/images/capi/packer/config/goss-args.json
@@ -1,7 +1,9 @@
 {
   "goss_arch": "amd64",
   "goss_format": "json",
-  "goss_format_options": "pretty",
+  "goss_entry_file": "goss/goss.yaml",
+  "goss_tests_dir": "packer/goss",
   "goss_url": "",
+  "goss_format_options": "pretty",
   "goss_version": "0.3.13"
 }
--- a/images/capi/packer/goss/goss-common.yaml
+++ b/images/capi/packer/goss/goss-common.yaml
@@ -1,0 +1,32 @@
+package:
+  cloud-init:
+    installed: true
+  kubeadm:
+    installed: true
+  kubelet:
+    installed: true
+  kubectl:
+    installed: true
+  kubernetes-cni:
+    installed: true
+service:
+  containerd:
+    enabled: true
+    running: true
+  dockerd:
+    enabled: false
+    running: false
+  kubelet:
+    enabled: true
+    running: false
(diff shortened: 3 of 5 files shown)


EleanorRigby:
`open-vm-tools` is a mistake and should only go in for OVA. I will remove it for now.
`curl` is a very basic utility package that might be used by other CLI or tools. We install it on all OS. First I had it in Photon specific GOSS yaml that I did not include in this commit. Since `curl` was supposed to be common I included it in the goss-common file.
I can remove it for now and cover essential packages in a different commit

EleanorRigby:
@codenrhoden : Another piece of code that I want to discuss is 
```
crictl images | grep -v 'IMAGE ID' | awk -F' ' '{print $1}' | awk -F'/' '{print $2}' | awk '!seen[$0]++' | sort
```
Here I added `awk '!seen[$0]++'` to uniquify image names if multiple versions are present. Do you think we will ever end up with multiple versions of k8s images after Ansible run? If not, I can just remove it and raise the error if multiple versions are there by mistake/custom-role.

codenrhoden:
> Here I added `awk '!seen[$0]++'` to uniquify image names if multiple versions are present. Do you think we will ever end up with multiple versions of k8s images after Ansible run? If not, I can just remove it and raise the error if multiple versions are there by mistake/custom-role.

That's a really good question. I think  you had seen a case recently where you saw two pause images -- 3.1 and 3.2, but that was in a downstream environment, I believe. I don't know the answer -- I can't think of a scenario where this would occur with the default image produced upstream with this tool, so I'm fine with keeping it as-is for now and seeing if we need to change it later. Does that make sense to you?

/lgtm

Right now I think this is ready to merge in. Want to run a couple of quick checks first, or see if there are any other comments, but I think this is a great start.

EleanorRigby:
> > Here I added `awk '!seen[$0]++'` to uniquify image names if multiple versions are present. Do you think we will ever end up with multiple versions of k8s images after Ansible run? If not, I can just remove it and raise the error if multiple versions are there by mistake/custom-role.
> 
> That's a really good question. I think you had seen a case recently where you saw two pause images -- 3.1 and 3.2, but that was in a downstream environment, I believe. I don't know the answer -- I can't think of a scenario where this would occur with the default image produced upstream with this tool, so I'm fine with keeping it as-is for now and seeing if we need to change it later. Does that make sense to you?
> 
> /lgtm
> 
> Right now I think this is ready to merge in. Want to run a couple of quick checks first, or see if there are any other comments, but I think this is a great start.

Makes sense. I will keep an eye out on this change and will follow up on `3.1` and `3.2` pause images. I thought it got there because I ran `kubeadm init phase preflight` which also pulls the images of a default version unless you specify a config file. So Ansible role pulled on version and I ran that command and pulled another. But I will work on it again to verify.

EleanorRigby:
##### UPDATE: 

@codenrhoden : Just removed a redundant `process` section from goss spec which is for now covered by `service` section on @detiber 's advice.
- Ran a few builds and they all work.

codenrhoden:
This is cool -- I didn't know GitHub was this smart. It picked up that the commits in this PR were included in a different PR, and once that other PR was merged, it marked this one as merged. Nice.

PR #293, opened 2020-07-16
Allow customizing volume_size for AWS

voor:
When adding in custom roles or additional tasks you quickly run out of space since most AMI only have 8GB of space.  This just allows a larger volume, which is already supported in Azure and vSphere.

Topic of conversation, Azure changes this with `vm_size`, vSphere changes this with `disk_size` which is a different variable from AWS that is `volume_size`, wondering if it would be confusing or offer better consistency by using the existing variable for `vm_size`?

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -48,7 +48,8 @@
     "snapshot_groups": "all",
     "snapshot_users": "",
     "subnet_id": "",
-    "vpc_id": ""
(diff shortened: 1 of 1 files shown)


detiber:
/lgtm
I'm good with the variable as named for now, we can possibly look at consolidating to a common variable name at a later time.

codenrhoden:
/approve

Agreed that it would be nice to consolidate on a common name in the future.

PR #294, opened 2020-07-16
Added cfg file for cloud-init to improve usability

SanikaGawhane:
- Adds a logging config file for cloud-init data
- Adds timestamp to cloud-init entries
- Outputs data destined for `cloud-init-output.log` to the console too.

### DIFF
--- a/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
+++ b/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
@@ -1,0 +1,67 @@
+## This yaml formated config file handles setting
+## logger information.  The values that are necessary to be set
+## are seen at the bottom.  The top '_log' are only used to remove
+## redundency in a syslog and fallback-to-file case.
+##
+## The 'log_cfgs' entry defines a list of logger configs
+## Each entry in the list is tried, and the first one that
+## works is used.  If a log_cfg list entry is an array, it will
+## be joined with '\n'.
+_log:
+ - &log_base |
+   [loggers]
+   keys=root,cloudinit
+   
+   [handlers]
+   keys=consoleHandler,cloudLogHandler
+   
+   [formatters]
+   keys=simpleFormatter,arg0Formatter
+   
+   [logger_root]
+   level=DEBUG
+   formatter=arg0Formatter
+   handlers=consoleHandler,cloudLogHandler
+   
+   [logger_cloudinit]
+   level=DEBUG
+   qualname=cloudinit
+   handlers=
+   propagate=1
+   
+   [handler_consoleHandler]
+   class=StreamHandler
+   level=WARNING
+   formatter=arg0Formatter
+   args=(sys.stderr,)
+   
+   [formatter_arg0Formatter]
+   format=%(asctime)s - %(filename)s[%(levelname)s]: %(message)s
+   
+   [formatter_simpleFormatter]
+   format=[CLOUDINIT]%(asctime)s - %(filename)s[%(levelname)s]: %(message)s
+ - &log_file |
+   [handler_cloudLogHandler]
+   class=FileHandler
+   level=DEBUG
+   formatter=arg0Formatter
+   args=('/var/log/cloud-init.log',)
+ - &log_syslog |
+   [handler_cloudLogHandler]
+   class=handlers.SysLogHandler
+   level=DEBUG
+   formatter=simpleFormatter
+   args=("/dev/log", handlers.SysLogHandler.LOG_USER)
+
+log_cfgs:
+# These will be joined into a string that defines the configuration
+ - [ *log_base, *log_syslog ]
+# These will be joined into a string that defines the configuration
+ - [ *log_base, *log_file ]
+# A file path can also be used
+# - /etc/log.conf
+
+# this tells cloud-init to redirect its stdout and stderr to
+# 'tee -a /var/log/cloud-init-output.log' so the user can see output
+# there without needing to look on the console.
+output: {all: '| python3 -c ''import sys,time;sys.stdout.write("".join(( " ".join((time.strftime("[%Y-%m-%d %H:%M:%S]", time.localtime()), line)) for line in sys.stdin )))'' | tee -a /var/log/cloud-init-output.log'}
--- a/images/capi/ansible/roles/providers/tasks/main.yml
+++ b/images/capi/ansible/roles/providers/tasks/main.yml
@@ -68,3 +68,11 @@
     - cloud-config
     - cloud-init
     - cloud-init-local
+
+- name: Create cloud-init config file
+  copy:
+    src: files/etc/cloud/cloud.cfg.d/05_logging.cfg
+    dest: /etc/cloud/cloud.cfg.d/05_logging.cfg
+    owner: root
+    group: root
+    mode: 0644


codenrhoden:
@randomvariable Can I get your help on this one? I think this builds off of some work you suggested.

A couple questions come to my mind here -- I think we can apply this change for all Node OS, not just AWS and vSphere. It's a cloud-init config that can be common. in that case, where should we put it? I'm not sure the `providers` role is appropriate at that point, but there is no obvious other place. Does anyone have opinions on putting it in `sysprep`?

When I build an OVA and boot it up, I don't see timestamps in `cloud-init-output.log`...

```sh
$ tail /var/log/cloud-init-output.log
...
Cloud-init v. 19.4 running 'modules:final' at Mon, 20 Jul 2020 16:26:56 +0000. Up 9.81 seconds.
Cloud-init v. 19.4 finished at Mon, 20 Jul 2020 16:27:46 +0000. Datasource DataSourceVMwareGuestInfo.  Up 59.53 seconds
```

So it's not obvious to me that the new config file is correct.

Additionally, there are now two cloud-init related logs in `/var/log`, but one of them is always empty, which is confusing to me:

```sh
-rw-r--r--  1 root adm       0 Jul 20 16:26 cloud-init.log
-rw-r--r--  1 root root   9233 Jul 20 16:27 cloud-init-output.log
```

EleanorRigby:
@codenrhoden : From what I understood, it seems that this line 

```
output: {all: '| tee -a /var/log/cloud-init-output.log'}
```

might cause all the output from all 3 stages of cloud-init (init, config & final) to go into one file. Other file seems useless. 

But I also found on cloud init docs 
```
Where are the logs?
Cloud-init uses two files to log to:

/var/log/cloud-init-output.log: captures the output from each stage of cloud-init when it runs
/var/log/cloud-init.log: very detailed log with debugging output, detailing each action taken
/run/cloud-init: contains logs about how cloud-init decided to enable or disable itself, as well as what platforms/datasources were detected. These logs are most useful when trying to determine what cloud-init ran or did not run.
Be aware that each time a system boots, new logs are appended to the files in /var/log. Therefore, the files may have more than one boot worth of information present.

When reviewing these logs look for any errors or Python tracebacks to check for any errors.
```

I am not able to find why that log file will be empty.

codenrhoden:
That makes sense, @EleanorRigby. I'm not really worried about the empty file, it just seemed off. If everything else is working, I think we are fine.

however, I *do* expect to see timestamps in the log, and right now I don't see them. So something is still off.

SanikaGawhane:
I think this code line -  `format=[CLOUDINIT]%(asctime)s - %(filename)s[%(levelname)s]: %(message)s` decides the format of the cloudinit logs, particularly the `(asctime)s` parameter. 
I'll try changing cloudinit format to include `(asctime)s` for `logger_cloudinit` block in the config file.

SanikaGawhane:
`formatter` argument was missing from cloud-init log block which caused the timestamps to not appear on the console. Fixed now.

images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
```diff
@@ -0,0 +1,68 @@
...
+   [logger_root]
+   level=DEBUG
+   formatter=arg0Formatter
+   handlers=consoleHandler,cloudLogHandler
+   
+   [logger_cloudinit]
+   level=DEBUG
+   qualname=cloudinit
+   formatter=arg0Formatter
+   handlers=consoleHandler,cloudLogHandler
```
codenrhoden, line 30: This can just be `handlers=`, like you originally had it, since `propagate=1`.

codenrhoden:
Just as a follow-up -- I build Ubuntu 18.04 on AWS, and was able to verify that the log entries accessible via `journalctl` has the same formatting as AL2 (which again, has some duplicative info -- but the timestamps *can* be more useful since they go to ms resolution instead of just sec). I also was able to see the timestamped entries in the console output **IF** I changed the loglevel.

randomvariable:
> Just as a follow-up -- I build Ubuntu 18.04 on AWS, and was able to verify that the log entries accessible via journalctl has the same formatting as AL2 (which again, has some duplicative info -- but the timestamps can be more useful since they go to ms resolution instead of just sec). I also was able to see the timestamped entries in the console output IF I changed the loglevel.

The original use case, which I admit should have been expanded upon was to get timestamps against kubeadm output over the duration of its running. Unfortunately, the python logging parameters don't help here as they only control the logging internal to cloud-init. After some experimentation, I have found that the following achieves what I originally wanted, which is to use the default logging configuration, but presuming gawk is available, change the `output` section to:

```
output:
  all: '| gawk ''{ print strftime("%FT%T%z:"), $0 }'' | tee -a /var/log/cloud-init-output.log'
```

codenrhoden:
@SanikaGawhane I think we can still put the proposed config file in place, with the addition of the change to `output` from @randomvariable. It's *mostly* the same as the default, so this kind of makes the implicit explicit. No harm in that.

i think the main thing we need to check is if the `awk` arguments work on all the supported distros. This would be a distro-specific thing, not a cloud provider thing. So it's really about checking CentOS 7, Ubuntu 18.04, Photon 3 on whatever is fastest for you to build and test. I think all the distros are going to have the GNU implementation of awk anyways. Photon 3 I can already verify:

```
capv@capi [ ~ ]$ ls -al /usr/bin/awk
lrwxrwxrwx 1 root root 4 Feb  4  2019 /usr/bin/awk -> gawk
capv@capi [ ~ ]$ ls -al /usr/bin/gawk
-rwxr-xr-x 2 root root 680024 Feb  4  2019 /usr/bin/gawk
```

It's worth checking just to make sure. :)

SanikaGawhane:
@randomvariable - Making this change in the logging file produced an error in some image builds because `gawk` isn't present in every base image. We were wondering if gawk needs to be added to every image or there can be a different way to handle it.

randomvariable:
@SanikaGawhane For this variant, gawk would need to be present. I have seen variants with perl, but that seems even worse. Anything that can add a timestamp to every line will work - I admit to getting the gawk variant from https://unix.stackexchange.com/questions/26728/prepending-a-timestamp-to-each-line-of-output-from-a-command , they've got a few more choices including one using bash and printf.

randomvariable:
lgtm as python3 should be inherently available everywhere

SanikaGawhane:
@randomvariable @codenrhoden Using `python` variant for logging in the output line worked best since this utility is present in all the base images. Changing the code accordingly.

codenrhoden:
I think this looks good, @SanikaGawhane !

I took this for a spin with the HAProxy OVA, because for that image there is nothing obvious that installs Python3. I felt like it's always present on Photon3, but we don't explicitly install it anywhere.  For the node images, we install `python3-pip`, which will always pull in Python3.

It worked fine on the HAProxy image -- Python3 gets pulled in by `cloud-init`. I should have known. :)

/lgtm
/approve

sample output:
```sh
[2020-08-14 17:52:47] Cloud-init v. 19.4 running 'init-local' at Fri, 14 Aug 2020 17:52:46 +0000. Up 4.70 seconds.
[2020-08-14 17:52:49] Cloud-init v. 19.4 running 'init' at Fri, 14 Aug 2020 17:52:49 +0000. Up 7.07 seconds.
[2020-08-14 17:52:49] ci-info: +++++++++++++++++++++++++++++++Net device info+++++++++++++++++++++++++++++++
[2020-08-14 17:52:49] ci-info: +--------+------+--------------+---------------+--------+-------------------+
[2020-08-14 17:52:49] ci-info: | Device |  Up  |   Address    |      Mask     | Scope  |     Hw-Address    |
[2020-08-14 17:52:49] ci-info: +--------+------+--------------+---------------+--------+-------------------+
[2020-08-14 17:52:49] ci-info: |  eth0  | True | 192.168.1.97 | 255.255.255.0 | global | 00:0c:29:09:9b:b5 |
[2020-08-14 17:52:49] ci-info: |   lo   | True |  127.0.0.1   |   255.0.0.0   |  host  |         .         |
[2020-08-14 17:52:49] ci-info: |   lo   | True |   ::1/128    |       .       |  host  |         .         |
...
```

Issue #284, opened 2020-07-07
Use GOSS to verify that expected kernel parameters are set


codenrhoden:
Verify that the kernel parameters that we are setting with Ansible haven't been undone/changed by later settings or custom roles. This is mainly the settings at https://github.com/kubernetes-sigs/image-builder/blob/00377e8ea77e7ad42b29bc5853eb426fce186118/images/capi/ansible/roles/node/tasks/main.yml#L33-L75

EleanorRigby:
@codenrhoden  Can I work on this?

EleanorRigby:
Can be closed

PR #299, opened 2020-07-20
Add kernel params to GOSS validation spec

EleanorRigby:
Add 5 kernel params that should have expected values set in sysctl.conf.

We do not need sysctl conf file location for the different OS as GOSS deals directly with sysctl CLI directly using https://github.com/lorenzosaino/go-sysctl

/hold

Should not be merged before #291 (Needs commits in there)

/assign @codenrhoden

### DIFF
--- a/images/capi/packer/goss/goss-common.yaml
+++ b/images/capi/packer/goss/goss-common.yaml
@@ -1,0 +1,32 @@
+package:
+  cloud-init:
+    installed: true
+  kubeadm:
+    installed: true
+  kubelet:
+    installed: true
+  kubectl:
+    installed: true
+  kubernetes-cni:
+    installed: true
+service:
+  containerd:
+    enabled: true
+    running: true
+  dockerd:
+    enabled: false
+    running: false
+  kubelet:
+    enabled: true
+    running: false
+command:
+  crictl ps:
+    exit-status: 0
+    stdout: []
+    stderr: []
+    timeout: 0
+  crictl images | grep -v 'IMAGE ID' | awk -F' ' '{print $1}' | awk -F'/' '{print $2}' | awk '!seen[$0]++' | sort:
+    exit-status: 0
+    stdout: ["coredns", "etcd", "kube-apiserver", "kube-controller-manager", "kube-proxy", "kube-scheduler", "pause"]
+    stderr: []
+    timeout: 0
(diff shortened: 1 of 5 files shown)


EleanorRigby:
Fixes #284

codenrhoden:
This captures everything that is in the Ansible script, so that part LGTM.

Will wait on #291.

images/capi/packer/goss/goss-common.yaml
```diff
@@ -0,0 +1,45 @@
...
+    installed: true
+  kubelet:
+    installed: true
+  kubectl:
+    installed: true
+  kubernetes-cni:
+    installed: true
+process:
+  containerd:
+    running: true
```
detiber, line 36: This seems a bit redundant with the service enabled/running check below

images/capi/packer/goss/goss-common.yaml
```diff
@@ -0,0 +1,45 @@
...
+  cloud-init:
+    installed: true
+  kubeadm:
+    installed: true
+  kubelet:
+    installed: true
+  kubectl:
+    installed: true
+  kubernetes-cni:
+    installed: true
```
detiber, line 33: Will this fail for the case that we do the binary installation of these rather than the package-based installation?

images/capi/packer/goss/goss-common.yaml
```diff
@@ -0,0 +1,45 @@
...
+process:
+  containerd:
+    running: true
+  dockerd:
+    running: false
```
detiber, line 38: Could this be moved to a service-based check like containerd/kubelet below?

detiber: If I'm reading the code correctly, I think specifying a service for docker and setting both enabled and running to false should be sufficient to verify that docker is not running.

images/capi/packer/goss/goss-common.yaml
EleanorRigby: Yes, we have not covered binary installation right now. I am building variable support and templating in GOSS. That will allow us to handle both cases. I will take that up in follow up PR confirming to Travis's suggestion here: https://github.com/kubernetes-sigs/image-builder/issues/282#issue-652754204

images/capi/packer/goss/goss-common.yaml
EleanorRigby: Yes this is redundant. I will fix it in another PR() which this PR is rebased off. (https://github.com/kubernetes-sigs/image-builder/pull/291/files)

images/capi/packer/goss/goss-common.yaml
EleanorRigby: got it.

Issue #292, opened 2020-07-15
Update ensure-goss.sh to auto-update plugin version


codenrhoden:
Right now, the `ensure-goss.sh` plugin that verifies that the Packer provisioner plugin for GOSS is installed, does not verify the version of the GOSS plugin, only it's presence. If it's not present, it is downloaded and installed and it's checksum is verified. However, if an update is made to require a newer version (which it recently was), the script will exit successfully if the plugin is present -- even though it's an older, unsupported version.

Since the plugin doesn't have a command to output it's version, we can rely on the checksum to determine the version. When the script is run, when the plugin is already present we should verify that it's the checksum of the correct version as well. If it's not, install the newer one.

/cc @EleanorRigby

PR #301, opened 2020-07-21
Make ensure-goss replace old goss provisioner version

EleanorRigby:
Right now our ensure scripts don't check the version of binary present on the machine.

This change will check the right version's presence by comparing SHASUM and install a newer version of packer provisioner goss if needed.

### DIFF
--- a/images/capi/hack/ensure-goss.sh
+++ b/images/capi/hack/ensure-goss.sh
@@ -20,18 +20,15 @@ set -o pipefail
 
 source hack/utils.sh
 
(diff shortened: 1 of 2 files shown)


EleanorRigby:
/assign @codenrhoden 
cc: @ykakarap

EleanorRigby:
fixes #292

codenrhoden:
I verified that this updates an older version to the right one, and continues to install if not present at all. nice work!

/lgtm
/approve

PR #302, opened 2020-07-21
:bug: Required volume sizes since it's now in a required position.

voor:
I broke the build, since volume_size is now required it's prepopulated.  It just so happens that all current supported images use 8 as a default, but it's placed in that file in the event one of them changes.

/assign @detiber 
/cc @codenrhoden

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -49,7 +49,7 @@
     "snapshot_users": "",
     "subnet_id": "",
     "vpc_id": "",
-    "volume_size": ""
+    "volume_size": "8"
   },
   "builders": [
     {


codenrhoden:
Interesting, I hit this too and I thought it was related to a packer version. I even ran builds on master and they all were successful. Clearly I missed something because now I see this too, off of master and with packer 1.6. weird.

Anyways, why put this in for each individual OS? We can achieve the same thing by just setting `"volume_size": "8"` in `packer.json`. That sets the global default to what is already the global default. :)

detiber:
lgtm, not overly concerned with where the defaults live, they could always be moved later if we decide to.

voor:
I moved it back into just packer.json, the intention was to make it explicit that it's a coincidence that the base images for CentOS 7, Amazon Linux 2, and Ubuntu are all the same size.

PR #303, opened 2020-07-22
Allow AMI to be private.

voor:
This is a tricky one that I'd like to discuss further.  There is *only* one valid value for this field, and that's `all`.  That means if you set this to `all` *anywhere* it cannot be overridden, because `all` is the only thing it accepts other than `nil`.  There's no unset in packer, so this is an admitted hack.  Now that we are pushing folks towards the Makefile with migration plans to a CLI, I believe it's acceptable to do something like this, which would allow defaulting an environment variable with an option to override.

Thoughts?

/cc @codenrhoden @detiber @randomvariable

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -96,6 +96,15 @@ ifeq (1,$(strip $(FOREGROUND)))
 PACKER_FLAGS += -var="headless=false"
 endif
 
+# If AMI_PRIVATE=1 then AMI_GROUPS will not be set to all, this is to workaround
+# a problem with packer only accepting one variable for ami_groups which makes it
+# impossible to override.
+ifeq (1,$(strip $(AMI_PRIVATE)))
+unexport AMI_GROUPS
+else
+export AMI_GROUPS ?= "all"
+endif
+
 # We want the var files passed to Packer to have a specific order, because the
 # precenence of the variables they contain depends on the order. Files listed
 # later on the CLI have higher precedence. We want the common var files found in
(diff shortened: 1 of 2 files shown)


codenrhoden:
I haven't tried this myself, but it sounds like you are saying the documented solution for building private AMIs is incorrect?

https://image-builder.sigs.k8s.io/capi/providers/aws.html#building-private-amis

voor:
Yes, since you cannot override a packer value with null (which are the only two accepted values -- `all` or `null`) you're stuck in a conundrum.  I'm wondering if the actual solution is to default to null, and for public images allow passing in the `all` value?  That would certainly reduce the complexity.

codenrhoden:
@voor thanks for the information. FYI, I'm playing with building private AMIs today just so that I can understand it all. :)

codenrhoden:
@voor I need a bit more help here. I'm not seeing how the currently documented method doesn't work. If I build an AMI with the default settings, and I go to that AMI in the AWS console and click on Permissions, it says "This Image is currently public". Okay, public is our default, just like the docs say.

Then, when I build with the following (first two lines are just to avoid copying the AMI all around):

```sh
$ cat aws.json
{
        "aws_region": "us-west-2",
        "ami_regions": "",
        "ami_groups": "",
        "snapshot_groups": ""
}
$ PACKER_VAR_FILES=aws.json make build-ami-amazon-2
```

When I look at Permissions for the resulting AMI, it says "This Image is currently private"., and has a interface to share the image with additional AWS account IDs.

Is the functionality you need different from this? Because it does seem to me like we can create private AMIs.

voor:
Let me take another pass at this, the issue might have happily been resolved with upgrading packer.  Historically it would look at "all" and "" and just keep "all" as the value since "" didn't override "all"

voor:
Sigh, looks like either I had been doing something wrong or updated packer fixed the issue

PR #304, opened 2020-07-22
Bump docs and scripts to Packer v1.6.0

### LINKED ISSUE #174: Photon OVA build doesn't work with Packer 1.5.4

codenrhoden:
We've merged changes recently that require Packer 1.6.0, so update
scripts and docs to reference that version.

/assign @voor 
/assign @detiber 

fixes #174

### DIFF
--- a/docs/book/src/capi/providers/vsphere.md
+++ b/docs/book/src/capi/providers/vsphere.md
@@ -1,10 +1,6 @@
 # Building Images for vSphere
 
-## Prerequisites
-
-The `make deps-ova` target will test that Ansible and Packer are installed and available. If they are not, they will be installed to `images/capi/.bin`. This directory will need to be added to your `$PATH`.
-
-### Hypervisor
+## Hypervisor
 
 The images may be built using one of the following hypervisors:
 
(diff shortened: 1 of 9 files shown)


docs/book/src/capi/providers/vsphere.md
```diff
@@ -23,13 +23,11 @@ The `vsphere` builder supports building against a remote VMware vSphere using st
 
 ### Prerequisites for all images
 
-- [Packer](https://www.packer.io/intro/getting-started/install.html) version >= 1.5.5
+- [Packer](https://www.packer.io/intro/getting-started/install.html)
 - [Ansible](http://docs.ansible.com/ansible/latest/intro_installation.html) version >= 2.8.0
```
detiber, line 27: Would it be better to remove this section from here all together, potentially in favor of linking back to the CAPI doc requirements section?

codenrhoden: I'd kind of been wondering the same thing. that's why I removed the version and add the callouts in the higher-level `capi.md`.

The truth is that each builder can get away with slightly different combinations of components, but I would sure like to move to one documented set of requirements. We are also closer to shipping a single combination with the container image.

If there are no objections, I'll remove this section from each provider-specific page and just defer to the main one. Agreed?

detiber: sgtm

codenrhoden: Done. It does make the docs more streamlined.

PR #300, opened 2020-07-21
Enable goss for Azure

EleanorRigby:
Works on VHD and Shared Image Gallery.

SIG
```
==> sig-ubuntu-1804:
Build 'sig-ubuntu-1804' finished.

==> Builds finished. The artifacts of successful builds are:
--> sig-ubuntu-1804: Azure.ResourceManagement.VMImage:
```

VHD
```
==> vhd-ubuntu-1804:
Build 'vhd-ubuntu-1804' finished.

==> Builds finished. The artifacts of successful builds are:
--> vhd-ubuntu-1804: Azure.ResourceManagement.VMImage:
```

GOSS Output
```
Running goss tests...
    vhd-ubuntu-1804: {
    vhd-ubuntu-1804:     "results": [
    vhd-ubuntu-1804:         {
    vhd-ubuntu-1804:             "duration": 14145852,
    vhd-ubuntu-1804:             "err": null,
    vhd-ubuntu-1804:             "expected": [
    vhd-ubuntu-1804:                 "true"
    vhd-ubuntu-1804:             ],
    vhd-ubuntu-1804:             "found": [
    vhd-ubuntu-1804:                 "true"
    vhd-ubuntu-1804:             ],
    vhd-ubuntu-1804:             "human": "",
    vhd-ubuntu-1804:             "meta": null,
    vhd-ubuntu-1804:             "property": "installed",
    vhd-ubuntu-1804:             "resource-id": "cloud-init",
    vhd-ubuntu-1804:             "resource-type": "Package",
    vhd-ubuntu-1804:             "result": 0,
    vhd-ubuntu-1804:             "successful": true,
    vhd-ubuntu-1804:             "summary-line": "Package: cloud-init: installed: matches expectation: [true]",
    vhd-ubuntu-1804:             "test-type": 0,
    vhd-ubuntu-1804:             "title": ""
    vhd-ubuntu-1804:         },
    vhd-ubuntu-1804:         {
    vhd-ubuntu-1804:             "duration": 7955385,
    vhd-ubuntu-1804:             "err": null,
    vhd-ubuntu-1804:             "expected": [
    vhd-ubuntu-1804:                 "true"
    vhd-ubuntu-1804:             ],
    vhd-ubuntu-1804:             "found": [
    vhd-ubuntu-1804:                 "true"
    vhd-ubuntu-1804:             ],
    vhd-ubuntu-1804:             "human": "",
    vhd-ubuntu-1804:             "meta": null,
[...]

PR #305, opened 2020-07-23
Add versions to GOSS spec

### LINKED ISSUE #285: Validate versions of components with GOSS
Related to #282 and #283

In addition to validating packages/binaries/images, it would be great if we can validate versions of those. GOSS allows for this, but since image-builder has so many customization points, it may be challenging to know *what* version we should be checking for. My thoughts around this is to make whatever versions we want to validate are in a `manifest.json` file generated by Packer, and I believe GOSS can read that and use it as a var file. Definitely something to explore and see if it is possible.

EleanorRigby:
##### Needs merge : 
 - Image Builder 
     [x] #291 and then #299
 

#### http source 
Build: (ubuntu == ubuntu1804)
- AMI
  - [x] centos 
  - [x] ubuntu
  - [x] amazon-2  

- OVA
  - [x] photon
  - [x] ubuntu
  - [x] centos

#### pkg source 
Build: (ubuntu == ubuntu1804)
- AMI
  - [x] centos 
  - [x] ubuntu
  - [x] amazon-2 

- OVA
  - [x] photon
  - [x] centos
  - [x] ubuntu

### Post Merge:
Add support to azure PR in #300


EleanorRigby:
/assign: @codenrhoden 
cc: @detiber : It supports http/pkg sources for GOSS

images/capi/packer/ami/packer.json
```diff
@@ -139,9 +139,19 @@
...
+      "goss_file": "{{user `goss_entry_file`}}",
+      "inspect": "{{user `goss_inspect_mode`}}",
+      "tests": ["{{user `goss_tests_dir`}}"],
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
+      "vars_inline": {
+        "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
+        "kubernetes_cni_source_type": "{{user `kubernetes_cni_source_type`}}",
+        "containerd_version" : "{{user `containerd_version`}}",
+        "kubernetes_cni_version": "{{user `kubernetes_cni_semver_suffix`}}",
```
detiber, line 151: Rather than introducing a new variable to use here, I think you can use:
```suggestion
        "kubernetes_cni_version": "{{user `kubernetes_cni_semver` | replace "v" "" 1}}",
```

images/capi/packer/ami/packer.json
```diff
@@ -139,9 +139,19 @@
...
+        "kubernetes_version": "{{user `kubernetes_semver_suffix`}}"
```
detiber, line 152: Similar here:
```suggestion
        "kubernetes_version": "{{user `kubernetes_semver` | replace "v" "" 1}}"
```

EleanorRigby: Excellent suggestion. was wondering how to solve this. Thanks, I will put the fix in.

EleanorRigby:
@codenrhoden @detiber : This PR has no external dependencies now. New packer-provisioner-goss([v1.2.0](https://github.com/YaleUniversity/packer-provisioner-goss/releases/tag/v1.2.0)) is released.

You can either merge (#291 and then #299) OR we can close them, and merge this instead?

images/capi/packer/goss/goss-command.yaml
```diff
@@ -0,0 +1,60 @@
...
+    timeout: 0
+    stdout: ["coredns", "etcd", "kube-apiserver", "kube-controller-manager", "kube-proxy", "kube-scheduler", "pause"]
+{{end}}
+{{if eq .Vars.kubernetes_source_type "http"}}
+  kubectl version --short --client=true -o json | jq .clientVersion.gitVersion | awk -F'[v"]' '{print $3}':
+    exit-status: 0
+    stdout: [{{ .Vars.kubernetes_version }}]
+    stderr: []
+    timeout: 0
+  kubeadm version -o json | jq .clientVersion.gitVersion | awk -F'[v"]' '{print $3}':
```
detiber, line 43: Do either (or both) of these support using `-o jsonpath`? I'm thinking it might be nice to avoid a dependency on `jq` if we can, it's a nice utility but not sure if we should require it to be present in images built with this tool.

EleanorRigby: No, they do not support `jsonpath`. I can remove jq but it becomes very ugly and kind of inflexible to changes. Like,
```
kubeadm version -o json | grep "gitVersion" | awk -F':' '{print $2}' | awk -F'["v]' '{print $3}'
```
will print `1.16.11`. Any better suggestions to parse? 

`jq` was present in all other OS except for centos. Do you consider it a PR blocker? I can work in follow up PRs to improve upon it.

detiber: Let's stick with `jq` for now, and we can always re-evaluate later, thanks!

EleanorRigby: cool.

images/capi/packer/ami/packer.json
```diff
@@ -139,9 +139,19 @@
       "arch": "{{user `goss_arch`}}",
       "format": "{{user `goss_format`}}",
       "format_options": "{{user `goss_format_options`}}",
-      "tests": ["packer/goss/goss.yaml"],
+      "goss_file": "{{user `goss_entry_file`}}",
+      "inspect": "{{user `goss_inspect_mode`}}",
+      "tests": ["{{user `goss_tests_dir`}}"],
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
+      "vars_inline": {
```
codenrhoden, line 147: I don't see `vars_inline` referenced anywhere. how is it used?

images/capi/packer/config/kubernetes.json
```diff
@@ -13,4 +13,4 @@
   "kubernetes_container_registry": "k8s.gcr.io",
   "kubernetes_load_additional_imgs": "false",
   "kubeadm_template": "etc/kubeadm.yml"
-}
+}
```
codenrhoden, line 16: Can we leave out this whitespace change since this file in unmodified otherwise?

codenrhoden:
@EleanorRigby You pushed one additional commit to remove `set -x`, which was added in the commit immediately prior. Please just combine those two commits (a git rebase with fixup would do the trick there).

images/capi/packer/ami/packer.json
EleanorRigby: This is used by goss command line 

```
goss validate --inline-vars "{JSON}"
```

The `vars_inline` are accepted by `packer-provisioner-goss`  and are added to GOSS command when executing.

images/capi/packer/config/kubernetes.json
EleanorRigby: Done

images/capi/packer/ami/packer.json
EleanorRigby: inline vars are required if we want to pass `packer` vars to goss on runtime. 
Goss does not accept multiple var files. otherwise, we could just pass more var files rather than using inline vars.

codenrhoden: Apologies here -- i misread what was going on because I thought `vars_inline` was defined up in the `"variables"` section, not directly in the goss provisioner. I was wondering how the variable(s) got into the provisioner. I just read it wrong. :)

codenrhoden:
/lgtm

I'm going to wait until the morning to see if there are any objections to merging this in all at once. If not, will `/approve`! Thanks @EleanorRigby.

PR #307, opened 2020-07-27
Add OS & provider specific GOSS validations

EleanorRigby:
#### Fix goss k8s pkg/binary version spec
   - Pass deb and rpm version for k8s and k8s-cni to goss
   - Add regex patterns (\QTEXT\E) for whole word match with escaping
#### Make crictl goss command robust
   - crictl now uses sed to replace last token delimited by `-` in container image name instead of awk.
        The last token is fixed value provided by Variable `arch`. arch is
        fixed for entire repo and is equal to amd64.
#### Improve goss command to match k8s version
  - Use trim and awk substitute instead of splitting on `v` as `v`
      can be part of valid semver too.

### Results
Works for all three builders (OVA, AMI, Azure) and all supported OSs.
GOSS is enabled in inspection mode by default, so this PR will not fail any build.

### DIFF
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -140,6 +140,9 @@
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
       "vars_inline": {
+        "OS": "{{user `distribution` | lower}}",
+        "ARCH": "amd64",
+        "PROVIDER": "azure",
         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
         "kubernetes_cni_source_type": "{{user `kubernetes_cni_source_type`}}",
         "containerd_version" : "{{user `containerd_version`}}",
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -144,7 +144,11 @@
         "kubernetes_cni_source_type": "{{user `kubernetes_cni_source_type`}}",
         "containerd_version" : "{{user `containerd_version`}}",
         "kubernetes_cni_version": "{{user `kubernetes_cni_semver` | replace \"v\" \"\" 1}}",
-        "kubernetes_version": "{{user `kubernetes_semver` | replace \"v\" \"\" 1}}"
+        "kubernetes_version": "{{user `kubernetes_semver` | replace \"v\" \"\" 1}}",
+        "kubernetes_rpm_version": "{{ split (user `kubernetes_rpm_version`) \"-\" 0  }}",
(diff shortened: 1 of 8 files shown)


EleanorRigby:
Commit of significance: https://github.com/kubernetes-sigs/image-builder/pull/307/commits/ac8eac35715c2c51f46d01339f6743f433768d41

Rest will disappear post rebase from master (after PR 305 is merged)

images/capi/packer/azure/packer.json
```diff
@@ -185,11 +185,18 @@
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
       "vars_inline": {
+        "OS": "ubuntu",
```
codenrhoden, line 188: azure has CentOS builds enabled now, so I'm not sure this should be hardcoded to Ubuntu?

EleanorRigby:
> Just the one comment on Azure so far...
> 
> I wish we didn't have to copy the list of RPMs/ DEBs into the GOSS files, but I don't think there is really an easy way to generate this list from the Ansible vars file at this time. Not without doing some pre-processing called from Make or something similar. Another thing to maybe explore with the CLI-driven approach.

I was thinking of using `ytt` to manage our input YAMLs across packer, ansible and, GOSS. Makefile can be one of the place to put them but I agree that CLI would be able to do this very cleanly. maybe we can use ytt with our cli.

images/capi/packer/azure/packer.json
EleanorRigby: I missed centos support on Azure completely. I see that Azure packer works differently from other clouds by having a builder each for different OS. Thus, there is no user variable that can be used to indicate `OS distro` to `GOSS`. I am going to add a new `packer user variable` for `OS distro`  derived from `make` command and hopefully that supports all OS on Azure.

codenrhoden: Ah, good point. I need to convert Azure to work like the others. Having a single builder (well, in this case -- two builders, one for VHD and one for SIG), but OS-specific var files. It's much easier.

EleanorRigby: I would like to correct the above statement. Azure has the same builder type but a different build name that can easily be templatized.

EleanorRigby:
/hold for the merge of #325 

The packer var files for each Azure OS type will be used to provide correct value for `OS` GOSS variable

codenrhoden:
/hold cancel
/lgtm
/approve

Thanks @EleanorRigby!  Just an FYI on the `/hold` command. It doesn't seem like the bot picks it up from edits either. 😄

PR #308, opened 2020-07-28
Fix building of ami and ova for centos

EleanorRigby:
- Adds python2-pip to vmware provider ansible riole. This is needed
  for cloud-init-vmware guestinfo to work correctly.
  - AMI now uses pip3 for downloading awscli on non amazon linux

Checked ami-centos and ova-centos for building. Both work.


codenrhoden:
Perfect, this is exactly how I was going to approach remedying the situation. Thanks @EleanorRigby !

I can confirm that before this change, builds from `master` failed for a CentOS 7 AMI and OVA, and with this change they do not. 

/lgtm
/approve

PR #309, opened 2020-07-28
Fix build name for ubuntu AMI to match cluster-api-provider-aws convention

detiber:
/assign @codenrhoden 

This fixes an issue where the Ubuntu AMI name was no longer matching against the default cluster-api-provider-aws lookup logic.


codenrhoden:
doh. then that means it was something I broke when I broke the different OS's into their own JSON files. Thanks for fixing.

/lgtm

PR #312, opened 2020-07-29
Remove potential python2 fallbacks

detiber:
I created a PR to update cluster-api-provider-gcp to use the `deps-gce` Make target for installing dependencies, and noticed that we aren't properly handling the python2 style fallbacks properly, this removes those fallbacks and defaults to explicitly using python3/pip3 instead

### DIFF
--- a/images/capi/hack/ensure-ansible.sh
+++ b/images/capi/hack/ensure-ansible.sh
@@ -26,19 +26,20 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 if command -v ansible >/dev/null 2>&1; then exit 0; fi
 
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
-  echo "Python 3 or Python binary must be in \$PATH" 1>&2
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "python3 binary must be in \$PATH" 1>&2
   exit 1
 fi
-if ! command -v pip3 || command -v pip >/dev/null 2>&1; then
+if ! command -v pip3 >/dev/null 2>&1; then
   curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-  python get-pip.py --user
+  python3 get-pip.py --user
   rm -f get-pip.py
 fi
-pip3 install --user "ansible==${_version}" || python -m pip install --user "ansible==${_version}"
+
+pip3 install --user "ansible==${_version}"
 echo $PATH
 
 if ! command -v ansible >/dev/null 2>&1; then
-  echo "User's Python binary directory must bin in \$PATH" 1>&2
+  echo "User's Python3 binary directory must bin in \$PATH" 1>&2
   exit 1
 fi
--- a/images/capi/hack/ensure-azure-cli.sh
+++ b/images/capi/hack/ensure-azure-cli.sh
@@ -26,17 +26,20 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 if command -v az >/dev/null 2>&1; then exit 0; fi
 
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
-  echo "Python 3 or Python binary must be in \$PATH" 1>&2
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "python3 binary must be in \$PATH" 1>&2
   exit 1
 fi
-if ! command -v pip3 >/dev/null 2>&1 || command -v pip >/dev/null 2>&1; then
+
+if ! command -v pip3 >/dev/null 2>&1; then
   curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-  python get-pip.py --user
+  python3 get-pip.py --user
   rm -f get-pip.py
 fi
-pip3 install --user "azure-cli==${_version}" || python -m pip install --user "azure-cli==${_version}"
-if ! command -v ansible >/dev/null 2>&1; then
-  echo "User's Python binary directory must bin in \$PATH" 1>&2
+
+pip3 install --user "azure-cli==${_version}"
+
+if ! command -v azure-cli >/dev/null 2>&1; then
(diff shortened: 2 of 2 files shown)


ncdc:
cc @codenrhoden @CecileRobertMichon

images/capi/hack/ensure-ansible.sh
```diff
@@ -26,11 +26,11 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 if command -v ansible >/dev/null 2>&1; then exit 0; fi
 
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
+if ! command -v python3 >/dev/null 2>&1; then
   echo "Python 3 or Python binary must be in \$PATH" 1>&2
```
EleanorRigby, line 30: Edit the echo to just say `python3`

images/capi/hack/ensure-azure-cli.sh
```diff
@@ -26,16 +26,16 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 
 if command -v az >/dev/null 2>&1; then exit 0; fi
 
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
+if ! command -v python3 >/dev/null 2>&1; then
   echo "Python 3 or Python binary must be in \$PATH" 1>&2
```
EleanorRigby, line 30: Edit echo below to just say python3

codenrhoden:
I know these were added when @voor added in calling these scripts within the container image, so I'd like to take a closer look at that scenario as well before removing.

voor:
I actually added in the explicit python3 calls, I left the python stuff for backwards compatibility, but I'm totally a :+1: for removing them.

images/capi/hack/ensure-ansible.sh
detiber: Updated, thanks!

images/capi/hack/ensure-azure-cli.sh
detiber: Updated, thanks!

detiber:
@codenrhoden I think this is good to go now.

images/capi/hack/ensure-azure-cli.sh
```diff
@@ -26,17 +26,17 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
...
+if ! command -v pip3 >/dev/null 2>&1; then
   curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
   python get-pip.py --user
   rm -f get-pip.py
 fi
-pip3 install --user "azure-cli==${_version}" || python -m pip install --user "azure-cli==${_version}"
-if ! command -v ansible >/dev/null 2>&1; then
-  echo "User's Python binary directory must bin in \$PATH" 1>&2
+pip3 install --user "azure-cli==${_version}"
+if ! command -v azure-cli >/dev/null 2>&1; then
```
CecileRobertMichon, line 39: good catch

CecileRobertMichon:
@codenrhoden do you have any outstanding comments on this one?

images/capi/hack/ensure-ansible.sh
```diff
@@ -26,11 +26,11 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
...
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
-  echo "Python 3 or Python binary must be in \$PATH" 1>&2
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "python3 binary must be in \$PATH" 1>&2
   exit 1
 fi
-if ! command -v pip3 || command -v pip >/dev/null 2>&1; then
+if ! command -v pip3 >/dev/null 2>&1; then
   curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
   python get-pip.py --user
```
codenrhoden, line 35: Since we are trying to explicitly use Python3 now, shouldn't this line call `python3` instead? It seems like at this line, this could possibly be pointing to Python 2 at this point.

images/capi/hack/ensure-ansible.sh
```diff
@@ -39,6 +39,6 @@ pip3 install --user "ansible==${_version}" || python -m pip install --user "ansi
 echo $PATH
```
codenrhoden, line 39: Same comment for right above this -- I feel like either we should remove the `python -m pip` fallback, or make it `python3 -m pip`. But I think we are already ensure that `pip3` is present, so the fallback no longer seems necessary.

images/capi/hack/ensure-azure-cli.sh
```diff
@@ -26,17 +26,17 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
...
-if ! command -v python3 >/dev/null || command -v python >/dev/null 2>&1; then
-  echo "Python 3 or Python binary must be in \$PATH" 1>&2
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "python3 binary must be in \$PATH" 1>&2
   exit 1
 fi
-if ! command -v pip3 >/dev/null 2>&1 || command -v pip >/dev/null 2>&1; then
+if ! command -v pip3 >/dev/null 2>&1; then
   curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
   python get-pip.py --user
```
codenrhoden, line 35: And also same here. Feels like this should be `python3`. However, I also wonder if we want to fully rely on the `Makefile` and know that this script is only called after `ensure-ansible`, which will have already done the python/pip verification. Do we really want to repeat it here?

I know that's not part of this PR, so not soemething I would hold things over. But wondering if we can eliminate some duplication here. But there may also be merit for having this work standalone...

detiber: I do think there is likely work that can be done here, but I think I would prefer scoping de-duplication efforts to a separate issue/PR

codenrhoden: Agreed - I was thinking afterwards this would be better off as a re-usable function in `images/capi/hack/utils.sh` anyways...

PR #313, opened 2020-07-29
[WIP] Add ci script for Azure

CecileRobertMichon:
This is a first attempt at adding a script that can be run as a prow presubmit for PRs to image-builder.

In order to test this with prow I will add an optional presubmit in test-infra for image-builder and run it from this PR.

/cc @codenrhoden @justaugustus


images/capi/packer/azure/scripts/ci-azure-e2e.sh
```diff
@@ -0,0 +1,59 @@
...
+: "${AZURE_CLIENT_SECRET:?Environment variable empty or not defined.}"
+
+get_random_region() {
+    local REGIONS=("eastus" "eastus2" "southcentralus" "westus2" "westeurope")
+    echo "${REGIONS[${RANDOM} % ${#REGIONS[@]}]}"
+}
+
+export AZURE_LOCATION="${AZURE_LOCATION:-$(get_random_region)}"
+export RESOURCE_GROUP_NAME="image-builder-e2e-$(head /dev/urandom | LC_ALL=C tr -dc a-z0-9 | head -c 6 ; echo '')"
+export BUILDER="${BUILDER:-ubuntu-1804}"
```
CecileRobertMichon, line 45: by default this runs the ubuntu-1804 builder but in the future we can setup multiple jobs to run different builders depending on which paths are touched

images/capi/packer/azure/scripts/ci-azure-e2e.sh
```diff
@@ -0,0 +1,59 @@
...
+export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
+
+cleanup() {
+    az group delete -n ${RESOURCE_GROUP_NAME} --yes --no-wait || true
+}
+
+trap cleanup EXIT
+
+make deps-azure
+make build-azure-vhd-${BUILDER}
```
CecileRobertMichon, line 59: I might change this to `-all` so it builds all OSes

detiber: Might also need to add `-j` if running `-all` to ensure a parallel run

images/capi/packer/azure/scripts/init-sig.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash
 
-az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
-az account set -s ${AZURE_SUBSCRIPTION_ID}
+az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID} >/dev/null 2>&1 
+az account set -s ${AZURE_SUBSCRIPTION_ID} >/dev/null 2>&1
```
detiber, line 4: We've had issues in the past where credentials have accidentally leaked because a parent script had done an `set -o xtrace`, it might be good to wrap any commands using these variables with:

```sh
tracestate="$(shopt -po xtrace)"
set +o xtrace

#sensitive code

eval "$tracestate"
```

images/capi/packer/azure/scripts/init-vhd.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
-az account set -s ${AZURE_SUBSCRIPTION_ID}
-export RESOURCE_GROUP_NAME=cluster-api-images
+az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID} >/dev/null 2>&1 
+az account set -s ${AZURE_SUBSCRIPTION_ID} >/dev/null 2>&1
```
detiber, line 4: Similar concern here related to potential credential leakage if a parent script enables xtrace

images/capi/packer/azure/scripts/parse-prow-creds.sh
```diff
@@ -0,0 +1,31 @@
...
+    grep -E -o "$1[[:blank:]]*=[[:blank:]]*\"[^[:space:]\"]+\"" | cut -d '"' -f 2
+}
+
+# for Prow we use the provided AZURE_CREDENTIALS file.
+# the file is expected to be in toml format.
+if [[ -n "${AZURE_CREDENTIALS:-}" ]]; then
+    export AZURE_SUBSCRIPTION_ID="$(cat ${AZURE_CREDENTIALS} | parse_cred SubscriptionID)"
+    export AZURE_TENANT_ID="$(cat ${AZURE_CREDENTIALS} | parse_cred TenantID)"
+    export AZURE_CLIENT_ID="$(cat ${AZURE_CREDENTIALS} | parse_cred ClientID)"
+    export AZURE_CLIENT_SECRET="$(cat ${AZURE_CREDENTIALS} | parse_cred ClientSecret)"
```
detiber, line 30: Similar concern here related to potential credential leakage if a parent script enables xtrace

images/capi/packer/azure/scripts/init-sig.sh
CecileRobertMichon: Good call, will fix. Thanks!

images/capi/packer/azure/scripts/parse-prow-creds.sh
CecileRobertMichon: Does this have any effect on environment variables? After I added it around this block of code it's silently failing

detiber: I wouldn't expect it to, it might be good enough to `set +o xtrace` above (with the other set commands) and consider it a bug if someone tries to do a `set -o xtrace` before including this script.

We could also avoid it completely for now and just try to be mindful of PRs that make changes to ensure we don't leak.

CecileRobertMichon: let me play around with this. I'd rather protect the creds programmatically if possible.

CecileRobertMichon:
ensure-ansible is failing with `Python 3 or Python binary must be in $PATH` so I guess this is where using the container image will help...

detiber:
@CecileRobertMichon hopefully the new kubekins-e2e image should resolve that issue along with https://github.com/kubernetes-sigs/image-builder/pull/312

CecileRobertMichon:
@codenrhoden any idea when we'll have the image builder container image published so I can leverage it here?

codenrhoden:
> @codenrhoden any idea when we'll have the image builder container image published so I can leverage it here?

@CecileRobertMichon apologies, I didn't realize you were held up by me! I had confused you being blocked by this with that update to the kubekins image. It's been... a weird week. I'll make that my current focus.

codenrhoden:
Just in case #312 took care of it...

/test pull-azure-all

codenrhoden:
I think #330 plus making sure that `/home/prow/.local/bin` is in `$PATH` will unblock this.

images/capi/packer/azure/scripts/ci-azure-e2e.sh
```diff
@@ -0,0 +1,58 @@
...
+export TIMESTAMP="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+export JOB_NAME="${JOB_NAME:-"image-builder-e2e"}"
+export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
+
+cleanup() {
+    az group delete -n ${RESOURCE_GROUP_NAME} --yes --no-wait || true
+}
+
+trap cleanup EXIT
+
```
codenrhoden, line 56: I think adding something here like `export PATH=$HOME/.local/bin:$PATH` will have this pick up the bins in `/home/prow/.local/bin` now...

Only needed if wanting to get this to work w/o container, I believe.

CecileRobertMichon:
@codenrhoden good news is that the test is finally running (had some trouble with the Python binary dir which was different on my local machine). Bad news is that I found out that the Azure centos build is broken due to jq package not found, looks like a regression from https://github.com/kubernetes-sigs/image-builder/pull/305/commits/4ecab1c6eb06a628aa59253bd4a95e6e43f7a49d. Do we know that this works for other centos builders? cc @EleanorRigby

nader-ziada:
tried running the azure tests target and working as expected

nader-ziada:
lgtm for me, but will let @codenrhoden and @detiber make the final call

PR #315, opened 2020-07-31
Add auditd and sysstat units for Amazon Linux 2

SanikaGawhane:
/assign @codenrhoden


images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 ---
 common_rpms:
+- audit
+- sysstat
```
ktchen14, line 17: Looks like this list is alphabetically sorted.

images/capi/ansible/roles/node/files/amazonLinux2/etc/audit/rules.d/containerd.rules
```diff
@@ -0,0 +1,10 @@
+-w /var/lib/containerd/ -p rwxa -k containerd
```
ktchen14, line 1: Should all of these be `rwxa`? For example the files in `/etc` seem like they should be `ra`, etc.?

images/capi/ansible/roles/node/tasks/amazonLinux2.yml
```diff
@@ -0,0 +1,34 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+- name: Ensure auditd is running and comes on at reboot
+  service:
+    name: auditd
+    state: started
```
ktchen14, line 19: Should this and the following service be `started` or `enabled`? This doesn't seem to match the task description.

images/capi/ansible/roles/node/tasks/amazonLinux2.yml
```diff
@@ -0,0 +1,34 @@
...
+    enabled: yes
+
+- name: Ensure sysstat is running and comes on at reboot
+  service:
+    name: sysstat
+    state: started
+    enabled: yes
+
+- name: configure auditd rules for containerd
```
ktchen14, line 28: s/configure/Configure/

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -162,6 +162,23 @@
     group: root
   when: kubernetes_semver != '' and ansible_os_family == "VMware Photon OS" # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
 
+- name: Create a log directory for auditd in Amazon Linux 2 OS
+  file:
+    path: "/var/log/audit"
```
ktchen14, line 167: Unnecessary quotes here and below

images/capi/ansible/roles/node/tasks/amazonLinux2.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Ensure auditd is running and comes on at reboot
+  service:
+    name: auditd
+    state: started
+    enabled: yes
+
+- name: Ensure sysstat is running and comes on at reboot
+  service:
+    name: sysstat
+    state: started
```
codenrhoden, line 25: The driving issue behind this PR is that it was noticed that `systatd` and `auditd` services fail to startup correctly AL2, as seen with:

```
$ sudo systemctl list-units --failed
{{ UNIT LOAD ACTIVE SUB DESCRIPTION}}
* auditd.service loaded failed failed Security Auditing Service
* sysstat.service loaded failed failed Resets System Activity Logs
```

This implies that the units/services are already installed and enabled without this change.
I think having this called out explicitly is good, it's just more evidence that we need to do a more thorough job of explicitly calling out what services we enable after install.

I'm personally in favor of having things be `state: stopped, enabled: true` for most things -- there's no reason to be running them at this stage of the build, but we want them to start up once the image is all done. However, there is an argument to be made that `state: started` is useful if our GOSS checks are looking for running services before the node is shut down... (cc @EleanorRigby).

No decision is needed here, but it's a conversation we should have!

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -153,15 +153,23 @@
...
     state: directory
     mode: "0640"
     owner: root
     group: root
-  when: kubernetes_semver != '' and ansible_os_family == "VMware Photon OS" # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
+  when: kubernetes_semver != '' and (ansible_os_family == "VMware Photon OS" or ansible_distribution == "Amazon") # Ensure audit log is not created for HA Proxy images, since this task is shared amongst build targets
 
+- name: Create a log directory for sysstat in Amazon Linux 2 OS
+  file:
+    path: /var/log/sa
```
codenrhoden, line 167: Can you look into why the code at 
https://github.com/kubernetes-sigs/image-builder/blob/d9f562889abe341b104acef4a65ba29ce72da280/images/capi/ansible/roles/sysprep/tasks/main.yml#L133
Is not sufficient for this case? That code is already creating this directory, but with a different mode. It may be worth seeing if the existing code is using a too permissive mode and it should be changed to 0640, or what exactly is happening. As it is, this step is only changing the mode on an existing dir, not creating it.

SanikaGawhane: From what I understood, line 133 recreates `sa` directory for all OSs. I was not sure if it is needed only for AL2 which I did explicitly here. 
If ` - { path: /var/log/sa, state: directory } ` is the right place to recreate the log directories, we can get rid of this code block. Similarly, we can also recreate `auditd` here and remove the auditd block too.

SanikaGawhane:
@codenrhoden - I squashed all commits into one. Probably I missed the fact that `cloud-init` commits were on master while creating and rebasing this branch and that's where the `cloud-init` commits came in. Got rid of those and the branch is clean now. Thanks for pointing this out.

images/capi/ansible/roles/sysprep/tasks/main.yml
codenrhoden: I think this PR is good to go, except for this last part.

I spun up a stock AL2 instance, and took a look at `/var/log/sa`.

```sh
# ls -al /var/log/sa
total 4
drwxr-xr-x 2 root root  18 Aug 13 22:59 .
drwxr-xr-x 7 root root 320 Aug 13 23:00 ..
-rw-r--r-- 1 root root 656 Aug 13 22:59 sa13
```

You'll note that it's mode is `0755`, which is what is created above at line 133. So, I don't think this extra section is needed right now, and the recreation of this dir can be handled as above.

codenrhoden:
/lgtm
/approve

Now I just need to figure out why this PR is labeled as a WIP... I don't see anything obvious.

PR #317, opened 2020-08-05
Specify AMI root volume device name for all distributions

randomvariable:
Amazon Linux 2 images created after #293 were non bootable on non-nvme instance types. This is because `/dev/sda1` ended up being a blank volume overriding the `/dev/xvda` root volume. This PR creates a `root_device_name` variable, which is then set for each distribution.

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -100,7 +100,7 @@
       },
       "launch_block_device_mappings": [
         {
-          "device_name": "/dev/sda1",
+          "device_name": "{{ user `root_device_name` }}",
           "volume_size": "{{ user `volume_size` }}",
           "volume_type": "gp2",
           "delete_on_termination": true
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -141,13 +141,15 @@
       "format_options": "{{user `goss_format_options`}}",
       "goss_file": "{{user `goss_entry_file`}}",
       "inspect": "{{user `goss_inspect_mode`}}",
-      "tests": ["{{user `goss_tests_dir`}}"],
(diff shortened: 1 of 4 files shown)


images/capi/packer/ami/packer.json
```diff
@@ -141,13 +141,15 @@
       "format_options": "{{user `goss_format_options`}}",
       "goss_file": "{{user `goss_entry_file`}}",
       "inspect": "{{user `goss_inspect_mode`}}",
-      "tests": ["{{user `goss_tests_dir`}}"],
+      "tests": [
```
dims, line 144: unrelated whitespace change?

images/capi/packer/ami/packer.json
```diff
@@ -141,13 +141,15 @@
...
+        "{{user `goss_tests_dir`}}"
+      ],
       "url": "{{user `goss_url`}}",
       "use_sudo": true,
       "vars_inline": {
         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
         "kubernetes_cni_source_type": "{{user `kubernetes_cni_source_type`}}",
-        "containerd_version" : "{{user `containerd_version`}}",
+        "containerd_version": "{{user `containerd_version`}}",
```
dims, line 152: same as above?

images/capi/packer/ami/packer.json
randomvariable: it got linted automatically, have been done as two commits.

dims: ah missed that :) thanks

images/capi/packer/ami/packer.json
randomvariable: same as above. i can delete the linting commit if needed.

PR #319, opened 2020-08-06
Increase Bandwidth for TCP to 1MB on Photon OS

EleanorRigby:
- Set `net.ipv4.tcp_limit_output_bytes` to `1048576 ` 
  - Photon's default kernel value for `net.ipv4.tcp_limit_output_bytes` is 256K compared to 1MB in Ubuntu

GOSS Kernel param entry for this will be added in #307 

Run result on powered on VM
```
root [ /home/capv ]# cat /etc/os-release
  NAME="VMware Photon OS" 
  VERSION="3.0"
  ID=photon 
  VERSION_ID=3.0
  PRETTY_NAME="VMware Photon OS/Linux"
  ANSI_COLOR="1;34"
  HOME_URL="https://vmware.github.io/photon/"
  BUG_REPORT_URL="https://github.com/vmware/photon/issues"
root [ /home/capv ]# sysctl -a | grep net.ipv4.tcp_limit_output_bytes
  net.ipv4.tcp_limit_output_bytes = 1000000000
root [ /home/capv ]#
```

/assign : @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/node/tasks/photon.yml
+++ b/images/capi/ansible/roles/node/tasks/photon.yml
@@ -26,3 +26,12 @@
     owner: root
     group: root
     mode: 0644
+
+- name: Ensure Bandwidth for TCP connections should 1 Mb (same as Ubuntu)
+  sysctl:
+    name: net.ipv4.tcp_limit_output_bytes
+    value: "1048576"
+    state: present
+    sysctl_set: yes
+    reload: yes
+    sysctl_file: "{{ sysctl_conf_file }}"


detiber:
Not sure if I have enough context here, is this something that we should also look to the Photon team to change in their defaults? 
Travis might have some thoughts around this as well.
/assign @codenrhoden

codenrhoden:
I believe the Photon team is sticking what is default in the base kernel, which changes in kernel 5.0. But this handles the other cases. I'll ping someone who can chime in more.

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -25,4 +25,13 @@
...
     owner: root
     group: root
-    mode: 0644
\ No newline at end of file
+    mode: 0644
+
+- name: Ensure Bandwidth for TCP connections should 1 Mb (same as Ubuntu)
+  sysctl:
+    name: net.ipv4.tcp_limit_output_bytes
+    value: "1048576"
```
codenrhoden, line 33: Is there anything we need to do to have the GOSS kernel param verification check this as well?

I know this would be a new/diff case because it's distro-specific.

suezzelur:
> Not sure if I have enough context here, is this something that we should also look to the Photon team to change in their defaults?
> Travis might have some thoughts around this as well.
> /assign @codenrhoden

Upstream recommendation for 4.19 kernel is 256kB, the default changed with code changes in 5.0 kernel version like @codenrhoden mentioned. The recommendation usually covers multiple use cases, but like in the change here it can always be tuned to conform to specific requirements.

images/capi/ansible/roles/node/tasks/photon.yml
EleanorRigby: It will be added to PR #307 if this merges first. Else later to PhotonOS kernel param section in `goss-vars.yml`

PR #320, opened 2020-08-07
Fix GOSS specs for k8s versioning

EleanorRigby:
#### Fix goss k8s pkg/binary version spec
   - Pass deb and rpm version for k8s and k8s-cni to goss
   - Add regex patterns (\QTEXT\E) for whole word match with escaping

#### Make crictl goss command robust
   - crictl now uses sed to replace last token delimited by `-` in container image name instead of awk.
        The last token is fixed value provided by Variable `arch`. arch is
        fixed for entire repo and is equal to amd64.

#### Improve goss command to match k8s version
  - Use trim and awk substitute instead of splitting on `v` as `v`
      can be part of valid semver too.


EleanorRigby:
All commits part of #307

Issue #173, opened 2020-03-11
containerd - cloud-init ordering part 2


randomvariable:
Despite our best efforts in #113 and #114, it is possible to repeatedly get containerd start up too late for kubeadm to succeed in `cloud-final`.

This can be done with a 3-node shared storage cluster with the networking deliberately slowed to 1Gb, which may not be uncommon in home lab environments and smaller organisations.

In this situation, if the backing storage is busy, then during machine boot, the following will happen:

1. systemd will start containerd prior to cloud-final
2. containerd will not have initialized until after cloud-final fails

```
Mar 11 20:09:35 test-md-0-7bf7c858b4-6p6rx systemd[1]: Starting containerd container runtime...
Mar 11 20:09:35 test-md-0-7bf7c858b4-6p6rx systemd[1]: Started containerd container runtime.
...
20 seconds of nothing, thanks
...
Mar 11 20:09:57 test-md-0-7bf7c858b4-6p6rx.home.internal.randomvariable.co.uk containerd[806]: time="2020-03-11T20:09:57.472851668Z" level=info msg="starting containerd" revision=d76c121f76a5fc8a462dc64594aea72fe18e1178 version=v1.3.3
Mar 11 20:09:57 test-md-0-7bf7c858b4-6p6rx.home.internal.randomvariable.co.uk containerd[806]: time="2020-03-11T20:09:57.504408678Z" level=info msg="loading plugin \"io.containerd.content.v1.content\"..." type=io.containerd.content.v1
Mar 11 20:09:57 test-md-0-7bf7c858b4-6p6rx.home.internal.randomvariable.co.uk containerd[806]: time="2020-03-11T20:09:57.505534329Z" level=info msg="loading plugin \"io.containerd.snapshotter.v1.btrfs\"..." type=io.containerd.snapshotter.v1
```

```
Mar 11 20:09:42 test-md-0-7bf7c858b4-6p6rx systemd[1]: Starting Execute cloud user/final scripts...
Mar 11 20:09:51 test-md-0-7bf7c858b4-6p6rx.home.internal.randomvariable.co.uk cloud-init[1035]: W0311 20:09:51.668517    1067 join.go:346] [preflight] WARNING: JoinControlPane.controlPlane settings will be ignored when control-plane flag is not set.
[...]

PR #323, opened 2020-08-11
Add notify type to containerd systemd unit file

EleanorRigby:
- Absence of **type=notify** in containerd unit file
    causes cloud-final.service trigger `kubedam init` too soon and it
    fails without detecting a container runtime on machine.
  - Adding this type=notify will enable to enforce strict boot ordering
    between cloud-final and containerd
  - Adding an `After` & `Wants` containerd systemd dependency in
  cloud-final and cloud-config boot config

Signed-off-by: Tushar Aggarwal <taggarwal@vmware.com>

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -44,6 +44,13 @@
     path: /etc/systemd/system/containerd.service.d
     state: directory
 
+- name: Configure containerd process start-up type to Notify
+  ini_file:
+    path: /etc/systemd/system/containerd.service
+    section: Service
+    option: Type
+    value: notify
+
 - name: Create containerd boot order drop in file
   template:
     dest: /etc/systemd/system/containerd.service.d/boot-order.conf
@@ -74,7 +81,7 @@
     name: containerd
     daemon_reload: yes
     enabled: True
-    state: started
+    state: restarted
 
 - name: delete tarball
   file:
--- a/images/capi/ansible/roles/containerd/templates/etc/systemd/system/containerd.service.d/boot-order.conf
+++ b/images/capi/ansible/roles/containerd/templates/etc/systemd/system/containerd.service.d/boot-order.conf
@@ -17,9 +17,4 @@
 # cloud-init boot stage managed by the cloud-final.service:
 # https://cloudinit.readthedocs.io/en/latest/topics/boot.html#final
 After=cloud-init.service
-Before=cloud-config.service
 Wants=cloud-init.service
-
-[Install]
-WantedBy=cloud-config.service
-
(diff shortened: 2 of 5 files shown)


EleanorRigby:
fixes #173

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -43,4 +43,24 @@
...
+
+- name: Creates unit file directory
+  file:
+    path: /etc/systemd/system/cloud-config.service.d
+    state: directory
+
+- name: Create cloud-final boot order drop in file
+  template:
+    dest: /etc/systemd/system/cloud-config.service.d/boot-order.conf
+    src: etc/systemd/system/cloud-config.service.d/boot-order.conf
```
codenrhoden, line 66: These new configuration pieces are all for parts of `cloud-init`, which get installed during the `providers` role. Does it makes sense to do this in the `kubernetes` role as you have it here, or would it be be better to put it at the end of the `providers` role?

images/capi/ansible/roles/kubernetes/templates/etc/systemd/system/cloud-config.service.d/boot-order.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
+After=containerd.service
+Wants=containerd.service
```
codenrhoden, line 3: This file and the `cloud-final` file aren't templates, so I'd rather see them in a files folder (e.g. `ansible/roles/kubernetes/files/...` and use the copy module so that it doesn't have to go through a templating engine.

I know we haven't been consistent about this, as I see files in `ansible/roles/containerd/templates` that are not templates (and some that are, sigh).

My preference for a template, since it runs through the Jinja2 template engine, would be to have a file that ends in `.j2` like `ansible/roles/setup/templates/etc/apt/sources.list.j2`. But now that I look through the whole repo, I see that is the *only* instance handled that way (with a `.j2` extension), so I can't complain too much. 😆

EleanorRigby: will fix this.

images/capi/ansible/roles/kubernetes/tasks/main.yml
EleanorRigby: We can do that. It will be better

EleanorRigby:
also, for future reference containerd added new ini params in containerd.service file. https://github.com/containerd/containerd/pull/4465/files#diff-5f2ec709a3860c8863a703cffa7b4921R14

We should move to installing from official containerd tarball once `containerd/pull/4465` PR merges

EleanorRigby:
@codenrhoden : addressed all comments.

PR #324, opened 2020-08-11
Enable all cloud-init services on boot (all OS)

EleanorRigby:
Signed-off-by: Tushar Aggarwal <taggarwal@vmware.com>

/assign @codenrhoden 

fixes partly #129

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/main.yml
+++ b/images/capi/ansible/roles/providers/tasks/main.yml
@@ -56,3 +56,15 @@
     owner: root
     group: root
     mode: "0755"
+
+# Some OS might disable cloud-final service on boot (rhel 7).
+# Enable all cloud-init services on boot.
+- name: Make sure all cloud init services are enabled
+  service:
+    name: "{{ item }}"
+    enabled: yes
+  with_items:
+    - cloud-final
+    - cloud-config
+    - cloud-init
+    - cloud-init-local


images/capi/ansible/roles/providers/tasks/main.yml
```diff
@@ -27,3 +27,21 @@
 
 - include_tasks: qemu.yml
   when: packer_builder_type is search('qemu')
+
+# Some OS might disable cloud-final service on boot (rhel 7).
+# Enable all cloud-init services on boot.
+# This sometimes reports service state as `UNKNOWN` and fails.
+# Retry 5 times to bypass above failure.
+- name: Make sure all cloud init services are enabled and running
```
detiber, line 35: I'm not sure that this needs to verify the services are running, I think just ensuring they are enabled should be sufficient here.

EleanorRigby:
Sounds reasonable. I can change this.

images/capi/ansible/roles/providers/tasks/main.yml
EleanorRigby: fixed.

images/capi/ansible/roles/providers/tasks/main.yml
```diff
@@ -56,3 +56,20 @@
...
+- name: Make sure all cloud init services are enabled
+  service:
+    name: "{{ item }}"
+    enabled: yes
+  with_items:
+    - cloud-final
+    - cloud-config
+    - cloud-init
+    - cloud-init-local
+  register: cloud_init_enabled
```
codenrhoden, line 73: Just one question before giving this an `/approve`. Now that this is just enabled and not trying to start/restart, is it really the case that the state comes back `UNKNOWN` and the retries are needed? I'm thinking that just making sure the service is enabled won't fail, but could be wrong.

EleanorRigby: I saw this problem happening consistently on `RHEL` but never removed retries for this PR.  I can retry a build locally and remove if it works twice consecutively for some `AMI` .

You can add it later if you face some problems on RHEL

EleanorRigby: It works. Removed retries

PR #325, opened 2020-08-11
Refactor Azure packer builder

EleanorRigby:
- Templatize building for different OS with SIG and VHD
  - Arrange azure packer vars in lexical order

/assign @codenrhoden 
cc: @CecileRobertMichon

### DIFF
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -2,18 +2,24 @@
   "variables": {
     "ansible_common_vars": "",
     "ansible_extra_vars": "",
-    "subscription_id": null,
+    "azure_location": null,
+    "build_name": null,
+    "build_timestamp": "{{timestamp}}",
     "client_id": null,
     "client_secret": null,
-    "disable_public_repos": "false",
-    "extra_debs": "",
-    "extra_repos": "",
-    "azure_location": null,
-    "vm_size": "",
-    "build_timestamp": "{{timestamp}}",
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "disable_public_repos": "false",
+    "distribution": null,
+    "distribution_release": null,
+    "distribution_version": null,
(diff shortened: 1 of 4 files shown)


codenrhoden:
This LGTM at first glance. Definitely want to hear from @CecileRobertMichon on this one. This change makes the way files are split up in line with AMI and OVA builders, and makes it much simpler to add additional OS's. There should be no discernable change in behavior when calling `make`. I still need to take this for a quick spin to confirm, though. Nice work!

CecileRobertMichon:
I like it! Just need to make sure pipelines are still working with the change (I'll run a manual test since CI isn't there yet) but looks like a good improvement.

images/capi/packer/azure/packer.json
```diff
@@ -112,49 +95,21 @@
...
   ],
   "provisioners": [
     {
       "type": "shell",
-      "only": ["vhd-ubuntu-1804", "sig-ubuntu-1804"],
+      "environment_vars": [
+        "BUILD_NAME={{user `build_name`}}"
+      ],
       "inline": [
+        "if [ $BUILD_NAME != \"ubuntu-1804\" ]; then exit 0; fi",
```
CecileRobertMichon, line 112: why is this needed? Why not just use `"only": ["vhd-ubuntu-1804", "sig-ubuntu-1804"],` since we know these are the two ubuntu-1804 builders?

CecileRobertMichon:
Verified ubuntu pipelines are still running successfully with this change

images/capi/packer/azure/packer.json
codenrhoden: This is working around a limitation in Packer. Unfortunately once you put a `user` var in the build name (e.g, in the builder section, instead of having `"name": "sig-ubuntu-1804`", we have `"name": "sig-{{user `build_name`}}",`, the `only` directives to packer no longer work, and you get errors about "build name not found". Sigh.

We had to do the same thing for AMIs: https://github.com/kubernetes-sigs/image-builder/blob/62d3c0e74b0650ebd62731a7cd558152c871c82c/images/capi/packer/ami/packer.json#L114-L118

codenrhoden:
Awesome - a huge win for consistency!

/approve

PR #326, opened 2020-08-11
Add markdown linting

codenrhoden:
This patch updates the markdown files to be compliant with
markdownlint-cli, and adds a Makefile target to make it easy to run via
Docker.

This particular container image (`gcr.io/cluster-api-provider-vsphere/extra/mdlint`) is in use with a few projects for Prow CI jobs, and once this PR is merged, a presubmit job can be added to run this check!

/assign @detiber 
/assign @CecileRobertMichon

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,3 +34,11 @@ build-book:
 .PHONY: serve-book
 serve-book: ## Build and serve the book with live-reloading enabled
 	$(MAKE) -C docs/book serve
+
+.PHONY: lint
+lint: ## Run all the lint targets
+	$(MAKE) lint-markdown
+
+.PHONY: lint-markdown
+lint-markdown: ## Lint the project's markdown
+	docker run --rm -v "$$(pwd)":/build:ro gcr.io/cluster-api-provider-vsphere/extra/mdlint:0.23.2 -- /md/lint -i images/kube-deploy .
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,0 +1,4 @@
+{
+  "default": true,
+  "line_length": false
+}
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Image Builder
 
-## Please see our [Book](https://image-builder.sigs.k8s.io) for more in-depth documentation.
+## Please see our [Book](https://image-builder.sigs.k8s.io) for more in-depth documentation
 
-## What is the Image Builder?
+## What is the Image Builder
 
 The Image Builder is a collection of cross-provider Kubernetes virtual machine image building utilities.
 
@@ -17,6 +17,7 @@ Each project is independent from each other, with the goal of eventually merging
 The `konfigadm` directory contains manifests for use with the `konfigadm CLI`.
 
 ### Useful links
+
 - [Quick Start for Cluster API Image Builder](https://image-builder.sigs.k8s.io/capi/quickstart.html)
 - [konfigadm CLI](https://github.com/flanksource/konfigadm)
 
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -34,3 +34,11 @@ build-book:
 .PHONY: serve-book
 serve-book: ## Build and serve the book with live-reloading enabled
 	$(MAKE) -C docs/book serve
+
+.PHONY: lint
```
codenrhoden, line 38: FYI, I'm going to be adding shell (`shellcheck`) and Go linting next.

Makefile
```diff
@@ -34,3 +34,11 @@ build-book:
...
+lint: ## Run all the lint targets
+	$(MAKE) lint-markdown
+
+.PHONY: lint-markdown
+lint-markdown: ## Lint the project's markdown
+	docker run --rm -v "$$(pwd)":/build:ro gcr.io/cluster-api-provider-vsphere/extra/mdlint:0.23.2 -- /md/lint -i images/kube-deploy .
```
CecileRobertMichon, line 44: could we include mdlint in our own container image and use that instead? It seems a bit odd to use a container image from a different project. If this image is being reused by several prow jobs that aren't just CAPV already then it might make sense to move it somewhere more generic.

codenrhoden: Yeah, that's an interesting idea.

This image is in use by a few different projects, but admittedly all VMware projects (CAPV, a cloud provider, and a CSI driver). I did look around defined Prow jobs to see if anyone was doing something similar yet, but I didn't find any. Examples: https://cs.k8s.io/?q=mdlint&i=nope&files=&repos=kubernetes/test-infra

So I definitely didn't want to make this VMW specific and wanted your input on it. :) I also want to promote re-use where I can.

I personally like having a dedicated image for it because it's an 80-90MB pull instead of >1GB, but I don't know if that's really a concern or not.

CecileRobertMichon: relevant issue: https://github.com/kubernetes/community/issues/3899

CecileRobertMichon:
I'm a bit hesitant about this one. The point made in https://github.com/kubernetes/community/issues/3899#issuecomment-557854713 are all valid IMO and I don't know that the added value outweighs the added barrier to entry. 

Looking at a recent run on CAPV, I don't see any logs in the prow job, is it expected to have no output? https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/989/pull-cluster-api-provider-vsphere-verify-markdown/1293308463308345344

codenrhoden:
> I'm a bit hesitant about this one. The point made in [kubernetes/community#3899 (comment)](https://github.com/kubernetes/community/issues/3899#issuecomment-557854713) are all valid IMO and I don't know that the added value outweighs the added barrier to entry.

I knew this had come up before! Thanks for finding that. Agreed it's all valid. This is where I would defer to something like sig-contribex to say what is really a better experience for newcomers. It's hard to balance that with project consistency. The given example of different ways to write a list is perfect, because while it does not change how the output is rendered and displayed, it also leads to inconsistency across the a project's documentation as a whole. This very PR shows an example where both approaches were used in the same file just a couple of lines apart from each other. Which is better? 🤷‍♂️   When it comes to things like that, I think linters are great because it enforces consistency and feedback is usually clear and actionable -- you know when you've satisfied the bot -- as opposed to a reviewer pointing out the same thing and coming off as overly pedantic and not welcoming. I may be in the minority there, so if you're hesitant, that's good enough for me to not pursue this.

/hold

> Looking at a recent run on CAPV, I don't see any logs in the prow job, is it expected to have no output? https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-vsphere/989/pull-cluster-api-provider-vsphere-verify-markdown/1293308463308345344

Yes it's expected. The linter has no output unless it finds problems.

PR #328, opened 2020-08-12
Allow multiple custom roles

xtreme-nitin-ravindran:
This PR allows user to use multiple custom roles to be run before sysprep


codenrhoden:
Hi @xtreme-nitin-ravindran, this seems like a pretty reasonable change to me. I appreciate that you updated the docs along with it. :)

I don't think this needs to be split into 4 separate commits, please squash them into 1.

xtreme-nitin-ravindran:
@codenrhoden Done.

codenrhoden:
Just took this for a spin - it works as expected.

Note that since this is space delimited (just like other args like `extra_rpms`), if specifying more than one role, you need to escape the double quotes.

For example, if you have two custom roles named `"custom1"` and `"custom2"`, and you put them in a JSON file, they'd need to be defined like this:

```json
{
  "custom_role": "true",
  "custom_role_names": "\"custom1 custom2\""
}
```

/lgtm
/approve

PR #329, opened 2020-08-12
Add Packer build color option

EleanorRigby:
- Leaves defaults unchanged

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -101,6 +101,10 @@ ARCH ?= amd64
 ## Packer flags
 ## --------------------------------------
 
+# Set Packer color to true if not already set in env variables
+PACKER_COLOR ?= true
+PACKER_FLAGS += -color=$(PACKER_COLOR)
+
 # If FOREGROUND=1 then Packer will set headless to false, causing local builds
 # to build in the foreground, with a UI. This is very useful when debugging new
 # platforms or issues with existing ones.


images/capi/Makefile
```diff
@@ -96,11 +96,12 @@ IMAGE_NAME ?= cluster-node-image-builder
 CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= dev
 ARCH ?= amd64
-
 ## --------------------------------------
 ## Packer flags
 ## --------------------------------------
 
+PACKER_COLOR=true
+
```
codenrhoden, line 104: I can think of perhaps a couple better ways to do this. I think for this to be easily overridable via env var, doesn't it have to be `PACKER_COLOR ?= true`?

Anyways, I think I'd prefer one of two different approaches.

- Knowing that `-color=true` is the default, take the same approach as `FOREGROUND` and define a new env var that when set, disables the color. something similar to:

```
ifeq (1,$(strip $(NOCOLOR)))
PACKER_FLAGS += -color=false
endif
```

- Or, just set it explicity in `PACKER_FLAGS` regardless:

```
PACKER_COLOR ?= true
...
PACKER_FLAGS += -color=$(PACKER_COLOR)
```

Both options have the benefit of not having to modify each `packer build` command because setting it in `PACKER_FLAGS` will put it in `PACKER_NODE_FLAGS` and/or `PACKER_HAPROXY_FLAGS` as needed.

This is untested, by the way. Just was my initial thought.

EleanorRigby: I agree with the approach.

codenrhoden:
I verified this works as expected. Nice work!

```sh
$ make build-node-ova-local-photon-3
hack/ensure-ansible.sh
hack/ensure-packer.sh
hack/ensure-goss.sh
Right version of binary present
packer build -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/ansible-args.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/goss-args.json"  -color=true -var-file="packer/ova/packer-common.json" -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/photon-3.json" -except=esx -except=vsphere -only=vmware-iso  packer/ova/packer-node.json
```
^ note `-color=true`

```sh
$ PACKER_COLOR=false !!
PACKER_COLOR=false make build-node-ova-local-photon-3
hack/ensure-ansible.sh
hack/ensure-packer.sh
hack/ensure-goss.sh
Right version of binary present
packer build -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/kubernetes.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/cni.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/containerd.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/ansible-args.json"  -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/config/goss-args.json"  -color=false -var-file="packer/ova/packer-common.json" -var-file="/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/packer/ova/photon-3.json" -except=esx -except=vsphere -only=vmware-iso  packer/ova/packer-node.json
```
^ `-color=false`

/lgtm
/approve

PR #330, opened 2020-08-13
Fix ensure Py3, Pip3, and azure-cli

codenrhoden:
This patch de-dups logic for making sure that Python3 and pip3 are
installed. It also corrects a corner case where pip3 could be installed
to the user local Python bin dir but that dir was not in path.

And finally, for checking that azure-cli was installed (which goes to
the user local Python bin dir), the wrong command name was used. It was
looking for `azure-cli` when it should look for `az`.

/assign @CecileRobertMichon 
@CecileRobertMichon I started looking at the container image from a CI perspective tonight, and immediately ran into this. I think this will solve your issues with running from Prow and it complaining about Python3. That is not the issue, it was that we were looking for `azure-cli` instead of `az`! 😆

### DIFF
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -67,3 +67,28 @@ get_shasum() {
   fi
   echo "$present_shasum"
 }
+
+ensure_py3_bin() {
+  # If given executable is not available, the user Python bin dir is not in path
+  # This function assumes the executable to be checked was installed with
+  # pip3 install --user ...
+  if ! command -v "${1}" >/dev/null 2>&1; then
+    echo "User's Python3 binary directory must be in \$PATH" 1>&2
+    echo "This is likely \$HOME/.local/bin" 1>&2
+    echo "\$PATH is currently: $PATH" 1>&2
+    exit 1
+  fi
+}
+
+ensure_py3() {
+  if ! command -v python3 >/dev/null 2>&1; then
+    echo "python3 binary must be in \$PATH" 1>&2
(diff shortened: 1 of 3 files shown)


images/capi/hack/utils.sh
```diff
@@ -66,4 +66,29 @@ get_shasum() {
...
+  if ! command -v python3 >/dev/null 2>&1; then
+    echo "python3 binary must be in \$PATH" 1>&2
+    exit 1
+  fi
+  if ! command -v pip3 >/dev/null 2>&1; then
+    curl -SsL https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+    python3 get-pip.py --user
+    rm -f get-pip.py
+    ensure_py3_bin pip3
+  fi
```
detiber, line 93: Should this be pulled out into it's own helper and potentially called from `ensure_py3_bin` instead? I could potentially see use cases where we might want python, but not pip

CecileRobertMichon:
> it was that we were looking for azure-cli instead of az!

Oh wow good catch, I should have caught that in review

images/capi/hack/utils.sh
codenrhoden: Hmm, good question. Right now such a usecase is not present in the code, but that doesn't mean it won't in the future. I'm inclined to say "let's cross that bridge when we get there", but if you want to see it now, I'm happy to do it. I'm already in that code and it would be simple to do. So... what would you prefer?

PR #333, opened 2020-08-17
Increase garbage collector capacity for ARP Cache Overflow issue

SanikaGawhane:
Increase garbage collector capacity to resolve the issue of ARP Cache overflowing when trying to deploy more than 1018 worker nodes in a workload cluster.

### DIFF
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -80,6 +80,51 @@
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
 
+- name: Ensure net.ipv4.neigh.default.gc_interval sysctl is present
+  sysctl:
+    name: net.ipv4.neigh.default.gc_interval
+    value: "30"
+    state: present
+    sysctl_set: yes
+    reload: yes
+    sysctl_file: "{{ sysctl_conf_file }}"
+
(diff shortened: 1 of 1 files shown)


detiber:
Are there any drawbacks to enabling this across the board by default?

Do we need to worry about decreasing the defaults for a subset of the distributions/base images we are currently using with this change?

codenrhoden:
/hold

Thanks for the feedback @detiber. This set of changes would eliminate an issue with ARP cache overflow when scaling cluster size beyond ~1024 (1024 is hard max size of table by default, so any other ARP neighbours out there plus all nodes in cluster have to be under this size).

I think this brings up an interesting question of what to tune for in the default upstream images like this. This limit was hit during scale testing for a vendor-specific K8s distribution, so what do we offer here? I'm staring to think that this kind of tuning doesn't really belong here, and would be better off taken care of using `custom_roles` for specific use cases.

To your specific point, however, I don't think changing the `_thresh` values would have negative impact to smaller clusters except for increased memory consumption by the kernel. However, i could see an impact from the `stale_time` and `interval` changes. Having ARP entries hangout in the cache and not getting marked as stale for an hour seems problematic (old value was 60s). If we do decide to modify the `_thresh` values, I'd like to see that other changes removed. I also think the `_thresh` values should remain as powers of 2, since the kernel tends to handle things in that way anways.

PR #334, opened 2020-08-19
bump default K8s to 1.16.14

codenrhoden:
This change sets the default Kubernetes version to the latest patch of n-2.

I think this is also helpful to move away from 1.16.11, since 1.16.11 had a packaging revision (`-1` instead of `-0`) which has historically been rare for the project. I've seen a couple of users lately try to modify the K8s version, and leave the `-1` in place resulting in build failures.

/assign @detiber @CecileRobertMichon


CecileRobertMichon:
lgtm

PR #339, opened 2020-08-21
capi: suppress warnings on creating files without mode

dongsupark:
Since Ansible 2.9.12 or 2.8.14, Ansible shows the following warning, when a file gets created with `mode` not being specified in the config.

```
[WARNING]: File '/tmp/kube-apiserver.tar' created with default permissions
'600'. The previous default was '666'. Specify 'mode' to avoid this warning.
```

To get rid of such a warning, specify `mode` to 0600 when creating files, as much as possible.

See also https://github.com/ansible/ansible/issues/71200 .

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/qemu.yml
+++ b/images/capi/ansible/roles/providers/tasks/qemu.yml
@@ -46,6 +46,7 @@
   file:
     src:   /usr/libexec/cloud-init
     dest:  /usr/lib/cloud-init
+    mode: 0777
     state: link
   when: ansible_os_family == "RedHat"
 
--- a/images/capi/ansible/roles/setup/tasks/debian.yml
+++ b/images/capi/ansible/roles/setup/tasks/debian.yml
@@ -16,6 +16,7 @@
   template:
     src: etc/apt/sources.list.j2
     dest: /etc/apt/sources.list
+    mode: 0644
 
 - name: Find existing repo files
   find:
@@ -36,6 +37,7 @@
   copy:
     src: "{{ item }}"
     dest: "/etc/apt/sources.list.d/{{ item | basename }}"
+    mode: 0644
   loop: "{{ extra_repos.split() }}"
   when: extra_repos != ""
 
--- a/images/capi/ansible/roles/setup/tasks/photon.yml
+++ b/images/capi/ansible/roles/setup/tasks/photon.yml
@@ -16,6 +16,7 @@
   template:
     dest: /home/builder/.bash_profile
     src: photon_bash_profile
+    mode: 0600
     owner: builder
     group: builder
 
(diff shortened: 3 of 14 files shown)


codenrhoden:
/ok-to-test

Thanks for bringing this up, @dongsupark. That issue you linked was illuminating.
I'm slightly worried that there will be unexpected behavior where something that was previously group readable no longer is, or some other version. It's a big change that we'll have to test out in full for sure.

dongsupark:
@codenrhoden Sure. Let me know if any regression was found during the full tests.

images/capi/ansible/roles/providers/tasks/qemu.yml
```diff
@@ -46,6 +46,7 @@
   file:
     src:   /usr/libexec/cloud-init
     dest:  /usr/lib/cloud-init
+    mode: 0700
```
detiber, line 49: I wouldn't expect there to be a need for setting a mode for a symlink, I would expect it to be 0777 by default: https://man7.org/linux/man-pages/man7/symlink.7.html#:~:text=On%20Linux%2C%20the%20permissions%20of,and%20can't%20be%20changed.

images/capi/ansible/roles/setup/tasks/debian.yml
```diff
@@ -16,6 +16,7 @@
   template:
     src: etc/apt/sources.list.j2
     dest: /etc/apt/sources.list
+    mode: 0600
```
detiber, line 19: Judging by the default mode of /etc/apt/sources.list on a new system, this should probably be:
```suggestion
    mode: 0644
```

images/capi/ansible/roles/setup/tasks/debian.yml
```diff
@@ -36,6 +37,7 @@
   copy:
     src: "{{ item }}"
     dest: "/etc/apt/sources.list.d/{{ item | basename }}"
+    mode: 0600
```
detiber, line 40:
```suggestion
    mode: 0644
```

images/capi/ansible/roles/setup/tasks/rpm_repos.yml
```diff
@@ -29,5 +29,6 @@
   copy:
     src: "{{ item }}"
     dest: "/etc/yum.repos.d/{{ item | basename }}"
+    mode: 0600
```
detiber, line 32:
```suggestion
    mode: 0644
```

dongsupark:
@detiber Thanks for the suggestions. Updated the PR.

images/capi/ansible/roles/setup/tasks/photon.yml
```diff
@@ -16,6 +16,7 @@
   template:
     dest: /home/builder/.bash_profile
     src: photon_bash_profile
+    mode: 0600
```
detiber, line 19: @codenrhoden This is the only mode change that I'm unsure about, can you verify that the bash profile here still works as expected?

codenrhoden: Tried this out -- works just fine. I think this is good to go now.

detiber:
/lgtm

Thanks for tackling this @dongsupark

PR #343, opened 2020-08-25
Fix call to sha256sum on Linux

codenrhoden:
In utils.sh:get_shasum(), the call used on Linux (vs MacOS) was using
the `-c` flag, which is supposed to check the sum against that found in
a file, but this function is intended to calculate the sum and return
it. Fix the function by removing `-c`.

This fixes errors seen like the following:

```
hack/ensure-goss.sh
sha256sum: /home/trhoden/.packer.d/plugins/packer-provisioner-goss: no properly formatted SHA256 checksum lines found
```

This was causing the checksum to always mismatch on Linux, and for GOSS to always be (re-)downloaded, even when it wasn't necessary.

/assign @EleanorRigby


EleanorRigby:
Nice catch.
/lgtm

PR #344, opened 2020-08-25
Disable conntrackd service to reduce CPU utilization

SanikaGawhane:
`conntrackd` service is enabled by default as a part of `conntrack-tools` package in Photon OS. This service creates short TCP connections accessing random K8s Services. The CPU utilization goes up especially in case of large number of workload clusters. 
Primary investigation suggests that `conntrackd` isn't used by Kubernetes anywhere and can be disabled to avoid high CPU utilization.

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -31,7 +31,6 @@ common_rpms:
 common_debs:
 - apt-transport-https
 - conntrack
-- conntrackd
 - curl
 - ebtables
 - jq
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -88,3 +88,10 @@
 - name: Edit fstab file to disable swap
   shell: sed -ri '/\sswap\s/s/^#?/#/' /etc/fstab
   when: ansible_memory_mb.swap.total != 0
+
+- name: Disable conntrackd service
+  systemd:
+    name: conntrackd
+    state: stopped
+    enabled: false
+  when: ansible_os_family != "Debian"
--- a/images/capi/packer/goss/goss-service.yaml
+++ b/images/capi/packer/goss/goss-service.yaml
@@ -8,6 +8,9 @@ service:
   kubelet:
     enabled: true
     running: false
+  conntrackd:
+    enabled: false
+    running: false
 {{range $name, $vers := index .Vars .Vars.OS "common-service"}}
   {{ $name }}:
   {{range $key, $val := $vers}}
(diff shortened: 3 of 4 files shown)


codenrhoden:
Thanks @SanikaGawhane. This is really interesting -- it appears that only Photon was having this service enabled by default.

Ubuntu has it as a separate package (`conntrackd`) from `conntrack` and can safely be removed (which you've done).
RPM based systems lump it all into one `conntrack-tools` package, but CentOS has it disabled by default:

CentOS:
```sh
[root@capi ~]# systemctl status conntrackd
● conntrackd.service - connection tracking daemon for debugging and High Availablity
   Loaded: loaded (/usr/lib/systemd/system/conntrackd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
```
^ Note the "vendor preset" part

Photon:
```sh
root@capi [ ~ ]# systemctl status conntrackd
● conntrackd.service - Conntrack Daemon
   Loaded: loaded (/lib/systemd/system/conntrackd.service; enabled; vendor preset: enabled)
   Active: active (running) since Tue 2020-08-04 23:17:49 UTC; 3 weeks 0 days ago
```

Yay for little idiosyncrasies.  
/approve

codenrhoden:
Good caatch @EleanorRigby. @SanikaGawhane, let's make sure that `conntrack` is removed from [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-vars.yaml#L18).

And then make sure that the service is not running on RPM distros (no need on Debian since it won't be installed) [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-vars.yaml#L163) and [here]().

@EleanorRigby since we want to make sure that the service is not running for Photon, CentOS 7, and RHEL 7, do we need to define that service check for those OS's in every provider that defines them?

SanikaGawhane:
@codenrhoden @EleanorRigby If we need `conntrackd` disabled everywhere, [this](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-service.yaml#L10) might be a good common place for the check.

codenrhoden:
> this

It's worth a shot. I was (mistakenly) thinking before that it would be harder because I thought we couldn't check if it was disabled if it wasn't installed (like it would be on Ubuntu), but that's not true. We do exactly that for `dockerd`.

EleanorRigby:
Right. Looks good to me. Yeah, I think GOSS handles missing package case for `disabled services` check.

SanikaGawhane:
Adding the check in `goss-service.yaml` seems to work. 

The change also looks good and `conntrackd` is disabled as desired.
Photon OS - 
```root [ /home/capv ]# systemctl status conntrackd
● conntrackd.service - Conntrack Daemon
   Loaded: loaded (/lib/systemd/system/conntrackd.service; disabled; vendor preset: enabled)
   Active: inactive (dead)
     Docs: man:conntrackd(8)
           man:conntrackd.conf(5)

EleanorRigby:
Thanks for finishing this up.

/lgtm

codenrhoden:
This looks good to me! @EleanorRigby can you give this one last look and LGTM if you are happy with it?

PR #351, opened 2020-09-01
Fix Azure build VHD pipeline to use manifest output

CecileRobertMichon:
Fixed the Azure DevOps pipeline to use the right storage account and resource group name outputs from the manifest.json packer output.

I accidentally broke it when removing echo statements in https://github.com/kubernetes-sigs/image-builder/commit/2ef06323f20bde597cd13ef7000c9b196e468b62#diff-765a596ea414b4b5e20a054a79d47439L12. This uses the manifest output instead which is more reliable from parsing an output log.

### DIFF
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -170,7 +170,7 @@
         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
         "os_name": "{{user `distro_name`}}",
         "resource_group_name": "{{user `resource_group_name`}}",
-        "storage_account": "{{user `storage_account_name`}}"
+        "storage_account_name": "{{user `storage_account_name`}}"
       }
     }
   ]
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
@marosset can you please review when you get a chance?

images/capi/packer/azure/packer.json
```diff
@@ -170,7 +170,7 @@
         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
         "os_name": "{{user `distro_name`}}",
         "resource_group_name": "{{user `resource_group_name`}}",
-        "storage_account": "{{user `storage_account_name`}}"
```
marosset, line 173: Did the previous PR you mentioned update this to use storage_account?
Current behavior looks correct, just wondering when this got chagned to storage_account...

CecileRobertMichon: it was always `storage_account` but it wasn't used until now.

PR #352, opened 2020-09-02
Support building base VMDKs and build OVFs from the base VMDKs

### LINKED ISSUE #39: CAPI OVA: support starting from existing VMX

kkeshavamurthy:
Breaking up the OVF image building process into 2:
- Build VMDK file from the base ISO
- Use the base VMDK image to build final OVF using `vmware-vmx` builder. 

This should help in saving some time since base image can be reused. 
I have added support for `local` building only. If there is enough interest I can work on adding support for vsphere builders as well. 

Should partially fix #39 

CC: @codenrhoden @EleanorRigby


EleanorRigby:
@kkeshavamurthy : I see that name of `make build` targets will change. That will be a breaking change. Also, will I have to always run base builder and then vmx builder to get the final image now?

Is there a way we can use the vmx builder if an existing vmx is specified else continue with the full build?

kkeshavamurthy:
@EleanorRigby none of the existing build targets are changed. 2 new targets for building base VMKD and building from VMDK files are added. This will not break any existing things.

kkeshavamurthy:
@figo, can you please approve the PR when you get a chance if everything looks good?

figo:
@kkeshavamurthy could you add document about how those two steps build works?

kkeshavamurthy:
@figo . Sure, should I include that in this PR or open up a new one?

EleanorRigby:
I think you can include usage and info in our image builder gitbook with
this PR.

kkeshavamurthy:
@figo @EleanorRigby 
Added the docs. Should be good to go.

figo:
Thanks @kkeshavamurthy for adding the usage doc.
/lgtm
/approve

PR #353, opened 2020-09-02
Update OVF spec for new hardware version

EleanorRigby:
/kind bug

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -216,15 +216,14 @@ def stream_optimize_vmdk_files(inlist):
         f['stream_name'] = outfile
         f['stream_size'] = os.path.getsize(outfile)
 
-
-_NODE_OVF_TEMPLATE = '''<?xml version='1.0' encoding='utf-8'?>
-<Envelope xmlns="http://schemas.dmtf.org/ovf/envelope/1" xmlns:ovf="http://schemas.dmtf.org/ovf/envelope/1" xmlns:vmw="http://www.vmware.com/schema/ovf" xmlns:rasd="http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/CIM_ResourceAllocationSettingData" xmlns:vssd="http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/CIM_VirtualSystemSettingData">
+_NODE_OVF_TEMPLATE = '''<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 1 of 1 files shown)


codenrhoden (commented):
Changes seem reasonable to me.

A lot of it is just re-ordering of sections, which makes it look like a bit more than it is. I see a couple things I would expect. Namely that it's adding a second IDE controller (before we just had IDE 1, now there is IDE 0 and 1). And I'm not sure how I feel about adding things like default CPU shares and a bunch of those new keys. I'm not sure if in this case we are setting reasonable defaults, or just the defaults of whatever vCenter this happened to be generated against.

3 things to follow up on for sure though:
 - What's the state of testing for this?
 - The second commit here seems to be undoing a change in the first commit, let's just squash the two
 - We will need to update the HAProxy template as well

images/capi/hack/image-build-ova.py
```diff
@@ -216,28 +216,23 @@ def stream_optimize_vmdk_files(inlist):
...
-    <Disk ovf:capacity="20" ovf:capacityAllocationUnits="byte * 2^30" ovf:format="http://www.vmware.com/interfaces/specifications/vmdk.html#streamOptimized" ovf:diskId="vmdisk1" ovf:fileRef="file1" ovf:populatedSize="${POPULATED_DISK_SIZE}"/>
+    <Info>Virtual disk information</Info>
+    <Disk ovf:capacity="20" ovf:capacityAllocationUnits="byte * 2^30" ovf:diskId="vmdisk1" ovf:fileRef="file1" ovf:format="http://www.vmware.com/interfaces/specifications/vmdk.html#streamOptimized" ovf:populatedSize="${POPULATED_DISK_SIZE}"/>
   </DiskSection>
   <NetworkSection>
     <Info>The list of logical networks</Info>
-    <Network ovf:name="nic0">
-      <Description>Please select a network</Description>
+    <Network ovf:name="VM Network">
+      <Description>The VM Network network</Description>
```
figo, line 231: I think it is make sense to keep the description of "Please select a network" as template

EleanorRigby: I will change it.

figo:
please remove the hold if nothing is pending.

EleanorRigby:
HA Proxy is pending and I have to test it with `clusterctl` once. I tried before and it worked.
Have to try again after pushing to spec.
Also, the HAProxy template needs to be reworked too.

EleanorRigby:
> 3 things to follow up on for sure though:
> 
> * What's the state of testing for this?
> * The second commit here seems to be undoing a change in the first commit, let's just squash the two
> * We will need to update the HAProxy template as well

1. Tested with 1.16.14 and HW version 15. The cluster comes up perfectly.
2. Squashed
3. HAProxy & Node OVA both use same ovf builder so both updated.

figo (approved):

PR #355, opened 2020-09-03
alphabetical sort of all json files

bavarianbidi:
currently it's a little pain to keep forks in sync if
there are lot of changes on some json files.

rebase and merging should now a little bit easier if json files
are sorted alphabetical.

sorting was done via

```bash
find . -regex ".*.json" -exec sh -c 'f={}; cat $f | jq -S "." > $f.tmp && mv $f.tmp $f' \;
```

Mario Constanti <mario.constanti@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)


codenrhoden:
I haven't scrolled through all of this yet, but a couple quick comments.

I'm a big fan of keeping variables ordered like this. In the individual `packer.json` files, however, I personally prefer to see the variables at the top of the file rather than sorted to the bottom as they would be now. I definitely would love to hear others' opinions.

@bavarianbidi You mentioned difficulty keeping things in sync with a fork. I'm wondering if you can share more about your use case and why maintaining a fork is necessary? We have tried to make as many things as configurable/customizable as possible (and are often adding more), with the intent being that you can maintain your own set of changes in a JSON file that you provide during build time, but are still using the original source for project. You can also run custom Ansible roles if needed. If there are needs you have that aren't covered, I'd be happy to hear them if you are able to share.

bavarianbidi:
> I haven't scrolled through all of this yet, but a couple quick comments.
> 
> I'm a big fan of keeping variables ordered like this. In the individual `packer.json` files, however, I personally prefer to see the variables at the top of the file rather than sorted to the bottom as they would be now. I definitely would love to hear others' opinions.
> 
> @bavarianbidi You mentioned difficulty keeping things in sync with a fork. I'm wondering if you can share more about your use case and why maintaining a fork is necessary? We have tried to make as many things as configurable/customizable as possible (and are often adding more), with the intent being that you can maintain your own set of changes in a JSON file that you provide during build time, but are still using the original source for project. You can also run custom Ansible roles if needed. If there are needs you have that aren't covered, I'd be happy to hear them if you are able to share.

@codenrhoden many thanks for your feedback!

### concrete issue with unsorted json files in our use-case
to be fair, specially for the `packer.json` files we have could been better to implement our changes. We added/changed some values directly in `.builders[]`/`.provisioners[]` instead of define the values under `.variables` and refer to them in `.builder[]`/`.provisioners[]`.

as you can see, [this diff](https://github.com/kubernetes-sigs/image-builder/files/5173648/sorted-packer-diff.txt) (generated via `git diff alphabetical-sort-json-files..caas-2004 images/capi/packer/qemu/packer.json > /tmp/sorted-packer-diff`) is much more readable/mergable than [this diff](https://github.com/kubernetes-sigs/image-builder/files/5173649/unsorted-packer-diff.txt) (generated via `git diff upstream/master..origin/caas-1910 images/capi/packer/qemu/packer.json > /tmp/unsorted-packer-diff..txt`).

i agree with you that it would be nice to have the variables on top of the file but the benefit of a sorted json is worth more to me (imho).

### concrete reason why we maintain our own fork
* we use `xfs` instead of `ext4` as filesystem ([code in forked repo](https://github.com/c445/image-builder/blob/caas-1910/images/capi/packer/os/linux/ubuntu/http/base/preseed.cfg#L53-L72))
* we [add company internal certificates](https://github.com/c445/image-builder/tree/caas/images/capi/ansible/roles/sysprep/files/certificates) there (sure, this could also be implemented in an other way)
* we don't install kubernetes during the build-process. we do this during node-provisioning before we run `kubeadm` because it's much more easier for us to build the image less often (only one reason).

I look forward to your comments :-)

EleanorRigby:
Thanks @bavarianbidi for the PR. I was trying to sort lexically bit by bit as I  raised any PR for other work. This is really good and I mostly like it. Can you rebase this again? After that,  I will probably run random local builds across provider & OS to check the validity of PR.

Is your forked repo private? I cannot open it to see `xfs` `ext4` changes.
`we don't install Kubernetes during the build-process`, I do no think we have a way to skip an ansible role right now (we can add more custom ones). Although, you can write a custom role to uninstall all k8s components and do `kubeadm reset`,  but that might be a long way to do it than forking and removing. 

Any suggestions to omit ansible roles programmatically?

bavarianbidi:
continue the work on #368 

we have created the fork of image-builder from an already forked image-builder repo (by accident).
I fixed this yesterday evening and forgot that the current open PR is created from the "fork-forked" repository.
so we have to continue the discussion here.

PR #363, opened 2020-09-09
CAPI: Update default k8s version to 1.17.11

CecileRobertMichon:
Now that 1.19.0 (and .1) has been released, updating the default k8s version to the latest 1.17 patch to stay at N-2.

### DIFF
--- a/docs/book/src/capi/capi.md
+++ b/docs/book/src/capi/capi.md
@@ -134,11 +134,3 @@ Then, execute the build (using a Photon OVA as an example) with the following:
 ```sh
 PACKER_VAR_FILES=proxy.json make build-node-ova-local-photon-3
 ```
-
-## Kubernetes versions
-| Tested Kubernetes Versions |
-|---------|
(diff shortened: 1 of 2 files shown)


docs/book/src/capi/capi.md
```diff
@@ -134,11 +134,3 @@ Then, execute the build (using a Photon OVA as an example) with the following:
 ```sh
 PACKER_VAR_FILES=proxy.json make build-node-ova-local-photon-3
 ```
-
-## Kubernetes versions
-| Tested Kubernetes Versions |
```
CecileRobertMichon, line 139: At first, I added the new versions to this and but now I think we should just remove it since we have to manually update it and "tested" doesn't really mean much here since we're not actually testing those in CI. A testgrid dashboard would tell a better story once we have one.

figo: +1 to remove this section.
the supported k8s versions should be called out in each tag version of image-builder, could be release note.

PR #365, opened 2020-09-10
Allows providing a path to the hack directory for vSphere Node Builder

voor:
This supports the use case with later versions of ansible that is executing the post shell scripts from the `/tmp` directory and not the current location, since the hacking scripts are references as relative, the `../../hack` location would go nowhere as a result.

You would need to provide an absolute path like if you're running from the container:

```
    "hack_directory": "/home/imagebuilder/hack"
```


voor:
This was not the issue, it was the fact the container does not have `python` binary and must explicitly use `python3`, will create another PR.

PR #367, opened 2020-09-10
:bug: Forces python3 for ova script post processor

voor:
The container for image builder is using a later version of Ubuntu that explicitly removes the `python` binary to enforce only supporting Python 3.x, this moves the script to likewise explicitly leverage `python3` so that it can work inside a containerized environment.

/bug
/assign @EleanorRigby


timothysc (approved):
/approve

PR #369, opened 2020-09-10
Update packer-prov-goss to version 2.0.0

EleanorRigby:


CecileRobertMichon:
`An operation is currently performing on this storage account that requires exclusive access.` 

I've seen a couple of these in recent runs, need to investigate what the issue is.

/retest

PR #371, opened 2020-09-11
capi: install crictl from upstream cri-tools repo

dongsupark:
We should download crictl directly from the upstream cri-tools repo to install it, instead of using the existing one in the cri-containerd tarball.

[Suggested](https://github.com/kubernetes-sigs/image-builder/pull/248#discussion_r485107613) by @moshloop .


ncdc:
What are the benefits of using the upstream cri-tools copy vs what's embedded in cri-containerd? Will this be problematic to keep up to date / in sync - it's another component+version that we have to match with Kube + CRI, right?

dongsupark:
> What are the benefits of using the upstream cri-tools copy vs what's embedded in cri-containerd? Will this be problematic to keep up to date / in sync - it's another component+version that we have to match with Kube + CRI, right?

Yeah, I agree. It might be another component to maintain.

However, it is necessary for Flatcar, because Flatcar does not provide `crictl` by default.
To make the [pending PR](https://github.com/kubernetes-sigs/image-builder/pull/248) simpler, I am updating the containerd part in that PR, by creating symlinking to existing binaries.
Downside of the simplification is that we need to anyway download and install `crictl`, which is not available in Flatcar.
If we would want to avoid installing `crictl` for non-Flatcar, it could require additional amount of code.
So I decided to simply install it by default.

Does it make sense?

randomvariable:
Can we add a conditional along the lines of `crictl_installation_method` with flatcar opting into this behaviour?

dongsupark:
> Can we add a conditional along the lines of crictl_pkg_source_type with flatcar opting into this behaviour?

Thanks, I will look into how I can do so.

images/capi/packer/qemu/packer.json
```diff
@@ -7,6 +7,8 @@
     "containerd_version": null,
     "containerd_sha256": null,
     "containerd_url": "https://storage.googleapis.com/cri-containerd-release/cri-containerd-{{user `containerd_version`}}.linux-amd64.tar.gz",
+    "crictl_version": null,
+    "crictl_url": "https://github.com/kubernetes-sigs/cri-tools/releases/download/v{{user `crictl_version`}}/crictl-v{{user `crictl_version`}}-linux-amd64.tar.gz",
```
EleanorRigby, line 11: This is only added to qemu builder? Will the absence of this variable in other builders cause a problem as all will run the same ansible role?

EleanorRigby (changes requested):

CecileRobertMichon:
Azure CI is failing with 

```
[0;32m    vhd-centos-7: TASK [containerd : download crictl] ********************************************[0m
[0;32m    vhd-centos-7: fatal: [default]: FAILED! => {"changed": false, "msg": "unknown url type: "}[0m
```

dongsupark:
@randomvariable 
> Can we add a conditional along the lines of crictl_pkg_source_type with flatcar opting into this behaviour?

Done.
Added a variable `crictl_pkg_source_type`, which can be either `pkg` or `http`.
Like other variables, it defaults to `pkg`.
The crictl part will be called only for `http`.

images/capi/packer/qemu/packer.json
dongsupark: > This is only added to qemu builder? Will the absence of this variable in other builders cause a problem as all will run the same ansible role?

Right.
So I added the same lines to other providers.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -40,6 +47,14 @@
     extra_opts:
       - --no-overwrite-dir
 
+- name: unpack crictl
+  unarchive:
+    remote_src: True
+    src: /tmp/crictl.tar.gz
+    dest: "{{ sysusr_prefix }}/bin"
```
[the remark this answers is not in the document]
dongsupark: > Is sysusr_prefix a pre defined Ansible variable?

Right.
Removed the variable, and set it back to `/usr`.

images/capi/packer/config/crictl.json
```diff
@@ -0,0 +1,4 @@
+{
+    "crictl_version": "1.16.1",
+    "crictl_sha256": "19fed421710fccfe58f5573383bb137c19438a9056355556f1a15da8d23b3ad1"
```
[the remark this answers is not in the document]
dongsupark: > Can we have something like crictl source type as randomvariable suggested and conditionally install it this way?

Done.

dongsupark:
@CecileRobertMichon 
> Azure CI is failing with

Fixed.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -21,6 +21,9 @@
 - import_tasks: photon.yml
   when: ansible_os_family == "VMware Photon OS"
 
+- import_tasks: crictl-url.yml
+  when: crictl_source_type == "http"
+
```
EleanorRigby, line 26: I think the task below `- name: unpack containerd` will overwrite the binary installed in `crictl-url. yml`

Containerd tar (/usr folder)
```
~/Downloads/cri-containerd-1.3.4.linux-amd64 » tree ./usr                                                               taggarwal@taggarwal-a01
./usr
└── local
    ├── bin
    │   ├── containerd
    │   ├── containerd-shim
    │   ├── containerd-shim-runc-v1
    │   ├── containerd-shim-runc-v2
    │   ├── containerd-stress
    │   ├── crictl
    │   ├── critest
    │   └── ctr
    └── sbin
        └── runc

3 directories, 9 files
```
The crictl here will be `unpack`ed, overwriting any existing binary. @dongsupark  What do you think?

EleanorRigby (changes requested):

images/capi/ansible/roles/containerd/tasks/main.yml
EleanorRigby: I think you need to perform `crictl-url` task at the end. Additionally, if you can make them independent of what order tasks are included in using conditionals `crictl_pkg_source_type `, that would be great.

dongsupark: @EleanorRigby 
> I think you need to perform crictl-url task at the end.

Good point.
I moved crictl-url to the end, and wrote comments.

>  Additionally, if you can make them independent of what order tasks are included in, by using the conditional crictl_pkg_source_type , that would be great.

Sorry, but I don't have a good idea how to do that.
We could split the containerd unpack part into multiple parts, like `pkg` or `http` in `crictl_source_type`, to conditionally unpack the crictl binary.
However, downside is, the code would become complicated.
Note that we are making this PR to simplify code, as suggested in the [Flatcar PR](https://github.com/kubernetes-sigs/image-builder/pull/248).

EleanorRigby:
Tested locally on photon OVA.


```
capv@capi [ ~ ]$ sudo crictl version
Version:  0.1.0
RuntimeName:  containerd
RuntimeVersion:  v1.3.4
RuntimeApiVersion:  v1alpha2
capv@capi [ ~ ]$
```


/lgtm

figo:
Sorry to join the review late, `crictl` is part of containerd.gz, it will be unpacked to `/usr/local/bin`, one thing i still don't understand is: why need to install separately for flatcar?  @dongsupark

ncdc:
Holding for the anser

ncdc:
From a previous comment:

> However, it is necessary for Flatcar, because Flatcar does not provide crictl by default.

images/capi/ansible/roles/containerd/tasks/crictl-url.yml
```diff
@@ -0,0 +1,41 @@
...
+  get_url:
+    url: "{{ crictl_url }}"
+    checksum: "sha256:{{ crictl_sha256 }}"
+    dest: /tmp/crictl.tar.gz
+    mode: 0600
+
+- name: Create /usr/bin directory
+  file:
+    state: directory
+    path: /usr/bin
```
EleanorRigby, line 25: Should it be `/usr/local/bin` as containerd installs it there? 

This will leave two crictl binaries in the system?

EleanorRigby (changes requested):

figo:
> From a previous comment:
> 
> > However, it is necessary for Flatcar, because Flatcar does not provide crictl by default.

Containerd always been installed from http source, https://github.com/kubernetes-sigs/image-builder/blob/8ef9f587695f5d94417ff1ecf7db8b0c19574590/images/capi/ansible/roles/containerd/tasks/main.yml
download containerd.gz from http url then unpack.

ncdc:
/approve cancel
Will re-review after discussions are resolved

images/capi/ansible/roles/containerd/tasks/crictl-url.yml
dongsupark: > Should it be /usr/local/bin as containerd installs it there?

Right. Fixed it. Thanks.

dongsupark:
@figo 
> Containerd always been installed from http source, https://github.com/kubernetes-sigs/image-builder/blob/8ef9f587695f5d94417ff1ecf7db8b0c19574590/images/capi/ansible/roles/containerd/tasks/main.yml
download containerd.gz from http url then unpack.

Yes, you are right.

Let me go through the reason why I created this PR.

Previously within the [Flatcar PR](https://github.com/kubernetes-sigs/image-builder/pull/248), it was necessary to conditionally unpack crictl only for Flatcar, i.e. under `/opt/bin`. It worked well, but its downside was the ansible code became so ugly, like a bunch of `exclude` fields for conditional installations.

That's why Moshe [suggested](https://github.com/kubernetes-sigs/image-builder/pull/248#discussion_r485107613) to split the crictl part into a separate PR for all distros. So I created this PR. After it was merged, I am going to rebase the [Flatcar PR](https://github.com/kubernetes-sigs/image-builder/pull/248) as well as a pending [simplification PR](https://github.com/kinvolk/image-builder/pull/9).

I understand, this PR alone might look redundant by overwriting the existing crictl binary. However, it would be great if you could understand that as a beginning of chain of simplifying the code base.

figo:
> @figo
> 
> > Containerd always been installed from http source, https://github.com/kubernetes-sigs/image-builder/blob/8ef9f587695f5d94417ff1ecf7db8b0c19574590/images/capi/ansible/roles/containerd/tasks/main.yml
> > download containerd.gz from http url then unpack.
> 
> Yes, you are right.
> 
> Let me go through the reason why I created this PR.
> 
> Previously within the [Flatcar PR](https://github.com/kubernetes-sigs/image-builder/pull/248), it was necessary to conditionally unpack crictl only for Flatcar, i.e. under `/opt/bin`. It worked well, but its downside was the ansible code became so ugly, like a bunch of `exclude` fields for conditional installations.
> 
> That's why Moshe [suggested](https://github.com/kubernetes-sigs/image-builder/pull/248#discussion_r485107613) to split the crictl part into a separate PR for all distros. So I created this PR. After it was merged, I am going to rebase the [Flatcar PR](https://github.com/kubernetes-sigs/image-builder/pull/248) as well as a pending [simplification PR](https://github.com/kinvolk/image-builder/pull/9).
> 
> I understand, this PR alone might look redundant by overwriting the existing crictl binary. However, it would be great if you could understand that as a beginning of chain of simplifying the code base.

Thank you for clarification, flatcar installed `containerd` and `runc` by default (ready-only), and still want user to install `crictl` and `containerd-shim`.
task `crictl-url.yml` will not be run by default, so it not affecting other OS.

could you document this dependency graph of flatcar in gitbook in a later PR?

dongsupark:
@figo 
> could you document this dependency graph of flatcar in gitbook in a later PR?

Will do so. Thanks.

images/capi/packer/config/crictl.json
```diff
@@ -0,0 +1,5 @@
+{
+    "crictl_version": "1.16.1",
+    "crictl_sha256": "19fed421710fccfe58f5573383bb137c19438a9056355556f1a15da8d23b3ad1",
+    "crictl_source_type": "pkg"
+}
```
detiber, line 5: Since this looks like it should be kept somewhat aligned (at least based on the versions of the releases attached to the repo) with the kubernetes version, should these variables be defined in `images/capi/packer/config/kubernetes.json` instead?

detiber (commented):
Based on the versioning for cri-tools, it looks like it's expected to be at least aligned to the minor version of kubernetes running, so I'm wondering if it would be good to enforce use of cri-tools independent from containerd, since we tend to apply the same version of containerd regardless of k8s version.

images/capi/packer/config/crictl.json
dongsupark: > should these variables be defined in images/capi/packer/config/kubernetes.json instead?

Good point.
Have updated it as suggested.

dongsupark:
Gentle ping.
/cc @moshloop  @ncdc @figo @EleanorRigby 

This PR is a blocker for the original Flatcar PR, https://github.com/kubernetes-sigs/image-builder/pull/248.

PR #169, opened 2020-02-28
Add Ubuntu 18.04 QEMU image for Cluster API

hidekazuna:
This PR adds Ubuntu 18.04 QEMU image for Cluster API.

Fixes #96

PR #373, opened 2020-09-14
Add auditd for Ubuntu

SanikaGawhane:
Add audit logs for Ubuntu.

### DIFF
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -13,6 +13,7 @@ common_rpms: &common_rpms
   yum-utils:
 
 common_debs: &common_debs
+  auditd:
   apt-transport-https:
   conntrack:
   curl:
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -29,6 +29,7 @@ common_rpms:
 - yum-utils
 
 common_debs:
+- auditd
 - apt-transport-https
 - conntrack
 - curl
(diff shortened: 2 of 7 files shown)


EleanorRigby (changes requested):
All current changes LGTM.
It will be a good idea to also change `goss/goss-vars.yaml` to account for the audit package in Debian based OS.

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -13,6 +13,7 @@ common_rpms: &common_rpms
   yum-utils:
 
 common_debs: &common_debs
+  auditd:
```
EleanorRigby, line 16: Since it is common to all Debian/rpm/photon OS, let's remove it entirely from here and put it in `goss-package.yaml`.  Check line#43, #169 for more presence of audit in this file. 

Please give it a run to check if GOSS passes. Editing variable files can sometimes have undesired effects on `goss render` and hence, `goss validate` too.

EleanorRigby (changes requested):

images/capi/packer/goss/goss-vars.yaml
SanikaGawhane: The names used for audit package vary - `rpms` uses `audit` whereas `debs` uses `auditd`. It will be a complex conditional statement to specify each case in `goss-package.yaml`. Hence leaving it here.

EleanorRigby:
/ok-to-test

Verified locally. Audit rules exist. Log directory was recreated and logs were present.

EleanorRigby:
Will wait for the test to finish once for lgtm

PR #375, opened 2020-09-17
WIP: Disable H/W offload for encapsulated network devices

randomvariable:


images/capi/ansible/roles/node/files/etc/udev/rules.d/90-netif-disable-hw-offload.rules
```diff
@@ -0,0 +1,3 @@
+ACTION=="add", SUBSYSTEM=="net", NAME=="veth*", TAG+="netif_hw_offload_disable"
+ACTION=="add", SUBSYSTEM=="net", NAME=="tunl", TAG+="netif_hw_offload_disable"
```
dims, line 2: is this just `"tunl"` or should it be `"tunl*"` as well?

randomvariable: I'll need to check, but I think it's just `tunl`.

vincepri:
Can we add a condition that would allow us to easily disable this patch using some configuration option?

randomvariable:
Closing this.

Basically, I am seeing other L3 routing issues even with TX/RX offload disabled when IPIP encap is on.

PR #376, opened 2020-09-17
Makefile: Introduce ON_ERROR_ASK for debugging

dongsupark:
If `ON_ERROR_ASK=1` then Packer will set `-on-error` to `ask`, causing the Packer build to pause when any error happens, instead of simply exiting.

This is useful when debugging unknown issues logging into the remote machine via ssh.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -116,6 +116,13 @@ ifeq (1,$(strip $(FOREGROUND)))
 PACKER_FLAGS += -var="headless=false"
 endif
 
+# If ON_ERROR_ASK=1 then Packer will set -on-error to ask, causing the Packer
+# build to pause when any error happens, instead of simply exiting. This is
+# useful when debugging unknown issues logging into the remote machine via ssh.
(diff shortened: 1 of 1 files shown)


EleanorRigby (approved):
/lgtm

EleanorRigby:
Verified works. see run below (notice `-on-error=ask` is added to the first run but not second)

```
~/workspace/image-builder/images/capi(do-packer) » ON_ERROR_ASK=1 make build-node-ova-local-ubuntu-1804

packer build -var-file="kubernetes.json"  ...  -color=true -on-error=ask ... -except=esx -except=vsphere -only=vmware-iso  packer/ova/packer-node.json

------------------------------------------------------------
~/workspace/image-builder/images/capi(do-packer) » make build-node-ova-local-ubuntu-1804                    
                                                                                                          
packer build -var-file="kubernetes.json"  ...  -color=true ... -except=esx -except=vsphere -only=vmware-iso  packer/ova/packer-node.json
```

PR #377, opened 2020-09-17
[WIP] Add Fedora 32 support for VMware builder

randomvariable:
Extremely WIP PR.

I basically wanted a way to test images with the latest and greatest Linux kernels, and with sets of packages I understand, so starting to add Fedora 32.

Another important change is building this straight away with UEFI secure boot. Would like to move the other distros to UEFI at some point, as it opens the path forward for better bootstrap security.

### DIFF
--- a/images/capi/packer/ova/packer-common.json
+++ b/images/capi/packer/ova/packer-common.json
@@ -25,5 +25,6 @@
   "vnc_bind_address": "127.0.0.1",
   "vnc_disable_password": "false",
   "vnc_port_min": "5900",
-  "vnc_port_max": "6000"
+  "vnc_port_max": "6000",
+  "vmware_firmware": "bios"
 }
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -52,6 +52,7 @@
       "boot_wait": "{{user `boot_wait`}}",
       "http_directory": "./packer/ova/linux/{{user `distro_name`}}/http/",
       "guest_os_type": "{{user `guest_os_type`}}",
+      "firmware": "{{ user `vmware_firmware`}}",
       "headless": "{{user `headless`}}",
       "iso_url": "{{user `iso_url`}}",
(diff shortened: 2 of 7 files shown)


images/capi/packer/ova/packer-common.json
```diff
@@ -25,5 +25,6 @@
   "vnc_bind_address": "127.0.0.1",
   "vnc_disable_password": "false",
   "vnc_port_min": "5900",
-  "vnc_port_max": "6000"
+  "vnc_port_max": "6000",
+  "vmware_firmware": "bios"
```
EleanorRigby, line 29: The firmware here is bios while we create EFI partition above. Does these both refer to different things. I know that a system can either be BIOS or EFI boot enabled but not both I guess.

EleanorRigby:
Also, this is awesome.i think EFI boot opens door for support for some PCI and GPU devices.  Not sure if it was passthrough or vGpu though

images/capi/packer/ova/packer-common.json
randomvariable: it's bios for everyone, and then overriden for fedora.

randomvariable:
Going to finish this off when Fedora 33 is out at the end of the month.

Importantly, Fedora 32 hasn't revved from Linux 5.8 to 5.9, which we need to fix the IPIP encapsulation bug that breaks Calico.

images/capi/packer/ova/packer-node.json
```diff
@@ -52,6 +52,7 @@
       "boot_wait": "{{user `boot_wait`}}",
       "http_directory": "./packer/ova/linux/{{user `distro_name`}}/http/",
       "guest_os_type": "{{user `guest_os_type`}}",
+      "firmware": "{{ user `vmware_firmware`}}",
```
MaxRink, line 55: That flag doesnt exist for vmware-iso
https://www.packer.io/docs/builders/vmware/iso

randomvariable: Yeah, going to try to move to vmx_data

PR #378, opened 2020-09-18
WIP: Add a tcpdump service for optional debugging

randomvariable:
If someone wants to do packet capture for a device on the host, they can run `systemctl enable tcpdump@<if name>`

### DIFF
--- a/images/capi/ansible/node.yml
+++ b/images/capi/ansible/node.yml
@@ -25,6 +25,8 @@
     - include_role:
         name: kubernetes
     - include_role:
+        name: debug-tools
+    - include_role:
         name: "{{ role }}"
       loop: "{{ custom_role_names.split() }}"
       loop_control:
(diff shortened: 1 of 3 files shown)


moshloop:
I am not sure this should be included in the base image-builder - Perhaps we should create a hack/examples directory of roles with a description of how they can be included for custom builds?

detiber:
> I am not sure this should be included in the base image-builder - Perhaps we should create a hack/examples directory of roles with a description of how they can be included for custom builds?

+1 I would rather see a documented way one could add this to a build using extensible mechanisms rather than including it by default in the images.

codenrhoden:
> > I am not sure this should be included in the base image-builder - Perhaps we should create a hack/examples directory of roles with a description of how they can be included for custom builds?
> 
> +1 I would rather see a documented way one could add this to a build using extensible mechanisms rather than including it by default in the images.

That makes sense to me as well. We have documented the `extra_roles` existence and  a description of what it does, but we don't have any examples.

randomvariable:
i'll redo as an example role.

/close

PR #379, opened 2020-09-18
photon: Remove appliance optimisations not suitable for container host

randomvariable:
PhotonOS has a bunch of sysctls that are appropriate for a standalone server, but not for a container host. When these are on, when using packet encapsulation based CNIs, L3 routing is broken:

This is what we had
```
# Disable response to broadcasts.
# You don't want yourself becoming a Smurf amplifier.
net.ipv4.icmp_echo_ignore_broadcasts = 1
# enable route verification on all interfaces
net.ipv4.conf.all.rp_filter = 1
# enable ipV6 forwarding
#net.ipv6.conf.all.forwarding = 1
# increase the number of possible inotify(7) watches
fs.inotify.max_user_watches = 65536
# avoid deleting secondary IPs on deleting the primary IP
net.ipv4.conf.default.promote_secondaries = 1
net.ipv4.conf.all.promote_secondaries = 1
net.ipv4.conf.default.accept_source_route=0
net.ipv4.tcp_max_syn_backlog=1280
net.ipv4.conf.all.accept_redirects=0
net.ipv4.conf.default.accept_redirects=0
net.ipv4.conf.all.send_redirects=0
net.ipv4.conf.default.send_redirects=0
net.ipv4.conf.all.secure_redirects=0
net.ipv4.conf.default.secure_redirects=0
net.ipv4.conf.all.log_martians=1
net.ipv4.conf.default.log_martians=1
net.ipv4.conf.default.rp_filter=1
net.ipv6.conf.all.accept_redirects=0
net.ipv6.conf.default.accept_redirects=0
net.ipv6.conf.all.router_solicitations=0
net.ipv6.conf.default.router_solicitations=0
kernel.printk = 3 4 1 7
```
In particular, `net.ipv4.conf.all.rp_filter = 1` breaks subnet routing in Calico.

Am OK with selectively adding these back to the sysctls

This PR brings Photon into line with other OS.

### DIFF
--- a/images/capi/ansible/roles/node/files/etc/systemd/scripts/iptables
+++ b/images/capi/ansible/roles/node/files/etc/systemd/scripts/iptables
@@ -1,0 +1,79 @@
+#!/bin/sh
+
+# Begin /etc/systemd/scripts/iptables
+
+# Insert connection-tracking modules
+# (not needed if built into the kernel)
+modprobe nf_conntrack
+modprobe xt_LOG
+
+# Enable broadcast echo Protection
+echo 1 > /proc/sys/net/ipv4/icmp_echo_ignore_broadcasts
+
+# Disable Source Routed Packets
+echo 0 > /proc/sys/net/ipv4/conf/all/accept_source_route
+echo 0 > /proc/sys/net/ipv4/conf/default/accept_source_route
+
+# Enable TCP SYN Cookie Protection
+echo 1 > /proc/sys/net/ipv4/tcp_syncookies
+
+# Disable ICMP Redirect Acceptance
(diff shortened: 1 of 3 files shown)


randomvariable:
/hold

Going to test an alternative way via Calico's manifests

randomvariable:
@EleanorRigby 
After taking a look at IgnoreLooseRPF, it's actually something different: tells Calico not to error if it sees reverse path filtering disabled. The code was eventually removed in https://github.com/projectcalico/felix/pull/2189 anyway, and I'm testing on Calico 1.16.

We therefore do need to figure out the correct set of sysctls to apply here.

randomvariable:
Found the local issues that were causing this in the end.

/close

PR #385, opened 2020-09-23
Disable H/W tx offload for ethernet devices on Photon

randomvariable:
/hold
for testing

Disables H/W offload on the transmit path for ethernet devices on Photon.

Is a workaround for https://github.com/vmware/photon/issues/1047

### DIFF
--- a/images/capi/ansible/roles/node/files/etc/udev/rules.d/90-netif-disable-hw-offload.rules
+++ b/images/capi/ansible/roles/node/files/etc/udev/rules.d/90-netif-disable-hw-offload.rules
@@ -1,0 +1,3 @@
+ACTION=="add", SUBSYSTEM=="net", KERNEL=="eth*", TAG+="netif_hw_tx_offload_disable"
+ACTION=="add", SUBSYSTEM=="net", KERNEL=="en*", TAG+="netif_hw_tx_offload_disable"
+TAG=="netif_hw_tx_offload_disable", RUN+="/usr/sbin/ethtool -K $name tx off"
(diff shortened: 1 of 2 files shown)


randomvariable:
it works

image

/unhold

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -13,7 +13,17 @@
 # limitations under the License.
 
 ---
-- name: Ensure Bandwidth for TCP connections should 1 Mb (same as Ubuntu)
```
randomvariable, line 16: That didn't really make sense as a task name and didn't describe what it did, hence renamed to "Double TCP small queue limit to be the same as Ubuntu "

images/capi/ansible/roles/node/files/etc/udev/rules.d/90-netif-disable-hw-offload.rules
```diff
@@ -0,0 +1,3 @@
+ACTION=="add", SUBSYSTEM=="net", KERNEL=="eth*", TAG+="netif_hw_tx_offload_disable"
+ACTION=="add", SUBSYSTEM=="net", KERNEL=="en*", TAG+="netif_hw_tx_offload_disable"
+TAG=="netif_hw_tx_offload_disable", RUN+="/usr/sbin/ethtool -K $name tx off"
```
EleanorRigby, line 3: Is the `ethtool` present by default on the photon image? Did you need to install it after powering on?

randomvariable: It's there by default.

PR #386, opened 2020-09-24
Default to BBR TCP congestion control algorithm

randomvariable:
See https://queue.acm.org/detail.cfm?id=3022184

This changes the default TCP congestion control algorith to TCP BBR
(Bottleneck Bandwidth and Round-trip propagation time), which is
a send-side only algorithm (doesn't require client support)
that reduces latency and increases throughput
even with mismatched bandwidths, e.g. mobile device connecting to server
with 10Gbps bandwidth.

BBR is available on Linux 4.9 and up.

This is in line with GKE that defaults to BBR.


vincepri (approved):
/approve

PR #397, opened 2020-10-05
Add chrony module for supported distros and providers

SanikaGawhane:
Use `chrony` time sync service in the following provider/OS combinations -
- AWS - Ubuntu 18.04/20.04, CentOS, Amazon Linux 2
- Azure - Ubuntu18.04/20.04, CentOS
- VMware - Ubuntu18.04/20.04, CentOS

Waiting for Chrony to be available on Photon to include it in the above list.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -17,3 +17,15 @@
 
 - import_tasks: debian.yml
   when: ansible_os_family == "Debian"
+
+- name: Configure PTP
+  lineinfile:
+    path: /etc/chrony/chrony.conf
+    create: yes
+    line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
+
+- name: Ensure makestep parameter set as per Azure recommendation
+  lineinfile:
+    path: /etc/chrony/chrony.conf
+    regexp: '^makestep'
+    line: makestep 1.0 -1
--- a/images/capi/packer/goss/goss-package.yaml
+++ b/images/capi/packer/goss/goss-package.yaml
@@ -17,6 +17,8 @@ kubernetes_cni_version: &kubernetes_cni_version
 package:
   cloud-init:
     installed: true
+  ntp:
+    installed: false
 {{if eq .Vars.kubernetes_source_type "pkg"}}
   kubeadm:
     installed: true
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -16,10 +16,10 @@ common_rpms:
 - audit
 - ca-certificates
 - conntrack-tools
+- chrony
 - curl
 - ebtables
 - jq
-- ntp
 - open-vm-tools
 - python3-pip
 - python-netifaces
@@ -32,6 +32,7 @@ common_debs:
 - auditd
 - apt-transport-https
 - conntrack
+- chrony
 - curl
 - ebtables
 - jq
@@ -41,7 +42,6 @@ common_debs:
 - libnetfilter-log1
 - linux-cloud-tools-virtual
 - linux-tools-virtual
-- ntp
 - open-vm-tools
 - python3-distutils
 - python3-netifaces
@@ -51,11 +51,11 @@ common_debs:
 common_photon_rpms:
 - audit
 - conntrack-tools
+- chrony
 - distrib-compat
 - ebtables
 - jq
 - net-tools
-- ntp
 - openssl-c_rehash
 - open-vm-tools
 - python-netifaces
--- a/images/capi/ansible/roles/providers/tasks/debian.yml
+++ b/images/capi/ansible/roles/providers/tasks/debian.yml
@@ -19,27 +19,3 @@
   vars:
     packages:
       - azure-cli
-
-- name: Ansible apt install chrony
-  apt:
-    name: chrony
-    state: present
-  
-- name: Configure PTP
-  lineinfile:
-    path: /etc/chrony/chrony.conf
-    create: yes
-    line: refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
-
-- name: Ensure makestep parameter set as per Azure recommendation
-  lineinfile:
-    path: /etc/chrony/chrony.conf
-    regexp: '^makestep'
-    line: makestep 1.0 -1
-
-- name: Enable chrony.service
-  systemd:
-    enabled: yes
-    state: started
-    daemon_reload: yes
-    name: chrony.service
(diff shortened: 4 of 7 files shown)


images/capi/ansible/roles/providers/tasks/aws.yml
```diff
@@ -64,3 +64,20 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+
+- name: Remove existing ntpd
+  yum:
+    name: ntp*
+    state: absent
+
+- name: Ansible yum install chrony
```
EleanorRigby, line 73: Should this be done when `when: ansible_os_family != "Debian"`

If there are too many ops like this, you can create files like `redhat.yml` `debian.yml` to reference them with a conditional.

images/capi/ansible/roles/providers/tasks/aws.yml
```diff
@@ -64,3 +64,20 @@
...
+  yum:
+    name: chrony
+    state: present
+
+- name: Ensure chrony is running
+  systemd:
+    enabled: yes
+    state: started
+    daemon_reload: yes
```
EleanorRigby, line 82: What is `daemon_reload` for?

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -17,3 +17,27 @@
 
 - import_tasks: debian.yml
   when: ansible_os_family == "Debian"
+
+- name: Ansible apt install chrony
+  yum:
```
EleanorRigby, line 22: Ubuntu and yum mistake. Need more conditionals to restrict this to centos like OSs

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -86,3 +86,15 @@
     owner: root
     group: root
     mode: 0644
+
+- name: Ansible yum install chrony
+  yum:
```
EleanorRigby, line 91: Needs correction. Can run for Debian system also.

EleanorRigby (changes requested):

images/capi/ansible/roles/providers/tasks/azure.yml
SanikaGawhane: Ansible has a generic package installation manager that uses yum/apt depending on which system (RPM/Debian) the package is being installed - https://docs.ansible.com/ansible/2.5/modules/package_module.html

images/capi/ansible/roles/providers/tasks/aws.yml
SanikaGawhane: Reloads systemd manager configuration. It is kind of a soft reload that reruns all generators, takes changed configurations from filesystem and regenerates dependency trees. We use it in `containerd` too - https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/containerd/tasks/main.yml#L92

EleanorRigby:
Is this still a work in progress?

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -86,3 +86,17 @@
     owner: root
     group: root
     mode: 0644
+
+- name: Ansible install chrony
```
EleanorRigby, line 90: I am wondering if we should add chrony to the base package list? 

and individual configs can remain in provider like `Ensure makestep parameter set as ..` can stay in azure provider.

Is it mandatory to remove `ntp` first? Or the ops to add chrony and delete ntpd are mutually exclusive? If they are muttually exclusive then it is  good to move it to base package list in role `node`

images/capi/ansible/roles/providers/tasks/aws.yml
```diff
@@ -64,3 +64,20 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+
+- name: Remove existing ntpd
+  package:
+    name: ntp*
```
EleanorRigby, line 70: Can we be specific about which packages to remove? Is it because ntp has different package names in different OSs?

kkeshavamurthy: Are we removing it explicitly because it's built in or can we just remove it from the pkg list?

SanikaGawhane: @EleanorRigby - Correct, the package can have names `ntp` or `ntpd` depending on the system.

@kkeshavamurthy - Right, we remove it because it is built in and my guess is it might default to `ntpd` if it is present. AWS recommends this [step](https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/).

images/capi/ansible/roles/providers/tasks/vmware.yml
SanikaGawhane: Adding chrony to the base package list is a good idea and make the configs cleaner. 

`ntpd` and `chrony` can both be installed and kept switching as needed. However, modern Fedora, CentOS, and RHEL releases have moved from NTP to Chrony as their default time-keeping implementation. That said, these are mutually exclusive and we can safely move `chrony` installation to base package list.

EleanorRigby (commented):
/lgtm

moshloop:
Can we add some goss tests for this?

images/capi/packer/goss/goss-package.yaml
```diff
@@ -18,6 +18,8 @@ package:
   cloud-init:
     installed: true
 {{if eq .Vars.kubernetes_source_type "pkg"}}
+  ntp:
```
EleanorRigby, line 21: This should not be under the above if. It is general and does not concern k8s source type.

EleanorRigby (changes requested):
just a minor GOSS change. The rest looks good.

images/capi/ansible/roles/providers/tasks/aws.yml
```diff
@@ -64,3 +64,10 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+
+- name: Ensure chrony is running
```
detiber, line 68: It seems odd to replicate this across all the different included providers, especially if we are handling the package installation more generally

SanikaGawhane: Moved this check to `main.yml`

EleanorRigby (approved):
/lgtm

The only thing I noticed is that we left QEMU builder out for chrony changes. We can further investigate and introduce support in separate PR. This looks good for now.

codenrhoden:
/approve

I do wonder if there is any extra default configuration that can/should go in for other cloud providers (namely AWS). We have Azure's recommended config put in, but does AWS have anything similar?

Nice work @SanikaGawhane !

PR #406, opened 2020-10-09
Set feature flag for Cloud Init 23 for Ubuntu systems for compatibility with Cluster API Provider AWS

randomvariable:
Latest version of cloud-init "improves" error handling. Unfortunately, this breaks what we're using to download encrypted user data in Cluster API Provider AWS.

This re-enables the old behaviour by setting the appropriate feature flag.

The CentOS & Fedora derivative operating systems are using a much older version of cloud-init.


ncdc:
@randomvariable from reading the cloud-init change, it sounds like this feature flag will be removed and the updated error handling will be the only option once cloud-init stops support Focal (20.04). Do we need to worry about that upcoming change, or is it more likely we'll have an updated method for secure bootstrap data before that happens?

/approve

randomvariable:
@ncdc The CAEP that will emerge out of https://github.com/kubernetes-sigs/cluster-api/issues/3761 should deal with that. I hope we get around to it before 2025!

ahuffman:
Validated this now runs cloud-init as expected on Ubuntu 18.04 builds successfully.

dlipovetsky:
This mechanism for overriding features was recently [removed](https://github.com/canonical/cloud-init/pull/4228) in cloud-init, and is in the latest cloud-init release of Ubuntu 20.04. To build working CAPA Ubuntu 20.04 images, we have to downgrade cloud-init. See https://github.com/kubernetes-sigs/image-builder/issues/1333

We need to revisit this in order to support Ubuntu 20.04 and newer versions with CAPA.

PR #407, opened 2020-10-12
[WIP] Add kube-vip as another preloaded Image for vmware

MaxRink:
As discussed in the last office hours.


images/capi/ansible/roles/providers/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+vmware_imgs:
+  - "docker.io/plndr/kube-vip:0.1.8"
```
detiber, line 2: I'm wondering if it would make sense to make this a bit more generic and default it to empty, and move this config into a packer variables file instead?

MaxRink: propably. 
it might also make sense to make this aviable for every provider then.

images/capi/ansible/node.yml
```diff
@@ -18,10 +18,10 @@
   tasks:
     - include_role:
         name: node
-    - include_role:
-        name: providers
     - include_role:
         name: containerd
+    - include_role:
+        name: providers
```
MaxRink, line 24: Any objections agains flipping the execution of those roles?
Doing this allows for just using containerd features in the role that already differentiates between every provider, thus slimming down on the changes.
On a quick look i havent found anythin in containerd that depens on providers being run.

images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -66,6 +66,13 @@
     state: absent
   when: ansible_os_family == "VMware Photon OS"
 
+- name: Pull vmware-specific images
+  shell: 'crictl pull {{ item }}'
```
EleanorRigby, line 70: Can we design this such that we can also pull image from an HTTP endpoint as we do for images in `url.yml` of Kubernetes ansible role? We can use `ctr` to load the downloaded image. 
We can continue to support this distinction with same flag `kubernetes_source_type` = http/pkg or we can have `kube_vip_source_type`. We have use cases where we need to get an image from a URL.

images/capi/ansible/roles/providers/defaults/main.yml
EleanorRigby: We can parameterize it through packer configs and ansible provisioner args regardless of the fact whether it is available for all providers or not. If it is just specific to vmware, we can create a packer variable in `packer/ova/packer-node.json` and pass it to ansible provisioner. Or we can also have a  new packer config `packer/config/vmware.yml or kube-vip.yml`

What do you think?

EleanorRigby (commented):
Thanks for the PR Max. I have some suggestions, please have a look.

images/capi/ansible/roles/providers/tasks/vmware.yml
MaxRink: Correct me if im wrong, but the HTTP way wont easily work with images from registries, right?
If so, wouldnt it make sense to basically have two variables, one for each sourcetype and just skip if empty? (as you might want to pull from both types)

EleanorRigby: Right, you will need a URL variable to get the image from which could be a registry or a normal http url.
Rather than skip if empty, I would prefer defaults setup in `packer/config` which defaults to open source kube-vip registry and image name+version.
The other vars like, `kube_vip_http_endpoint` are ignored even if filled up or empty if source_type is a registry.

images/capi/ansible/node.yml
EleanorRigby: If there are no dependencies, it might be a good thing to do as containerd role changes mostly with containerd version updates which aren't many. Might be a good checkpoint for someone to maintain a base image and iterate on it from provider role onwards for rapid dev.

MaxRink:
So, ive moved the var to the packer config, but im unsure how to handle the http-based download best

images/capi/ansible/node.yml
MaxRink: I havent found any and my capv clusters were able to stand up with the new images.

codenrhoden (commented):
I know from the comments that this was discussed in the latest office hours, which I was unable to make, but I do have some general feedback.

I'm not sure that `kube-vip` should be handled as a special case. I get the advantage of pre-loading , but in practice we don't pre-load several of the required images. I'm thinking of all the vSphere CSI images and the vSphere CCM as examples. CAPV+CAPI generates manifests that reference all sorts of images that aren't preloaded.

I took a look at the CAPV docs just now, and there is no documentation about running in air-gapped environments. However, I know in practice that it still requires an internal/local image registry to pull images from, rather than having all images pre-loaded. I think it is a worthwhile goal to support a deployment scenario where no image registry is needed, but I am failing to see the advantage of pre-loading `kube-vip` but none of the other images. Am I way off base here?

MaxRink:
> but I am failing to see the advantage of pre-loading `kube-vip` but none of the other images. Am I way off base here?
As i see it kube-vip is the only image ontop of the k8s-images that you actually need to even bootstrap the cluster.
Not having the CSI-Images will lead you to have an cluster without CSI and deployments in image-pull-backoff.
Not having kube-vip will not even "reward" you with a cluster, the first node just fails to bootstrap.
I think that is worth differentiating.

PR #408, opened 2020-10-12
Add subiquity-based flavour for Ubuntu 20.04 based upon UEFI

MaxRink:
As discussed in the office hours


MaxRink:
So, with subiquity we cannot run commands like swapoff -a at the end of the install process, as they actually have non-zero retun codes somehow and thus force halt the install process

MaxRink:
Besides that, everything is working as expected. 
@randomvariable switching to UEFI is something i can not fully test, as the vcenter i build upon is still on 6.5 :D

moshloop:
I am hesitant to add another ubuntu installer variant - If the subiquity installer is now recommended, we should rather just switch the existing ubuntu 20.04 to that?

MaxRink:
> If the subiquity installer is now recommended, we should rather just switch the existing ubuntu 20.04 to that?

Yeah, its now the default installer. 
To find the old, debian-installer-based, one, you need to dig pretty deep, as it isnt referenced within the normal download structure :D

randomvariable:
I'm happy to have the non subiquity variant removed.

images/capi/packer/ova/linux/ubuntu/http/20.04/user-data
```diff
@@ -0,0 +1,103 @@
...
+  keyboard:
+    layout: en
+    variant: us
+  network:
+    network:
+      version: 2
+      ethernets:
+        ens192:
+          dhcp4: true
+          dhcp-identifier: mac
```
kkeshavamurthy, line 16: Just curious, how will this work if the network interface comes up with a different name?

MaxRink: It wont. on vmware predicatble naming will always default to ens192 for the first nic in the system

MaxRink:
I've now also switched over to UEFI Images with secure boot enabled by default

MaxRink:
Btw, that secure boot point propably needs some discussion. 
Building those images with secure boot eanbled will improove security, but will also prevent them from running on esxi systems without TPM

randomvariable:
Let's remove the secure boot option by default, or make it optional. CAPV in the future could decide to enable it at VM creation time anyway.

randomvariable:
Can you add firmware = uefi to the vmx_data option of vmware-iso build, as this should get builds working on fusion / workstation.

MaxRink:
@randomvariable Both done, i cant test the vmware-builder tho.
Would be ideal if somebody else can do it

jgrexa:
@MaxRink I have tested the vmware-iso build on fusion. It was necessary to change the nic name in user-data from "ens192" to "ens33", otherwise the network in vm didnt come up.

MaxRink:
Ok, ive removed the network section as newer subiquity installers now default to dhcp on all interfaces, thus the section became redundant.
This should now also work in fusion/workstation without changes

farodin91:
I tried this out internally. I hit different issues: 
- growroot stopped working
- cloud init issue: https://kb.vmware.com/s/article/80934
- to use efi in ova template firmware have to be changed to efi
- i had to increase builder cpu count and mem size

MaxRink:
> * growroot stopped working
propably due to the switch to LVM. I have a script laying around that does that for LVM but reverting back to a plain partition is also an option.
> * cloud init issue: https://kb.vmware.com/s/article/80934
Ive included a mitigation for this
> * to use efi in ova template firmware have to be changed to efi
Yeah, ive added that, forgot to push it tho 
> * i had to increase builder cpu count and mem size
Ive havent had that issue, to what values did you have to bump it?

MaxRink:
@farodin91 ive reverted it back to a direct layout, so growroot should work again.

farodin91:
@MaxRink I didn't thought of reverting the layout. I will check it.

My plan was add grow root task to cloud init. Idea would be: grow device, run lvm commands to grow lvm, resize fs

farodin91:
I havent had that issue, to what values did you have to bump it?
> 4 cores and 4gb

farodin91:
> I havent had that issue, to what values did you have to bump it?
> 
> > 4 cores and 4gb

It mainly reduced the build time.

farodin91:
> @MaxRink I didn't thought of reverting the layout. I will check it.
> 
> My plan was add grow root task to cloud init. Idea would be: grow device, run lvm commands to grow lvm, resize fs

It hit now the same error as on the master during growroot: https://github.com/kubernetes-sigs/image-builder/issues/464

MaxRink:
so, ive rebased it to work with the ova testinfra @codenrhoden 
This is where the virtual cd shines :D

PR #418, opened 2020-10-16
Make default Azure storage account name unique and check name before create

### LINKED ISSUE #417: Azure ci flakes
Azure ci flakes out often and one of the errors I've seen is 

[pipeline1](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/414/pull-azure-all/1316898623325212672#1:build-log.txt%3A377)
[pipeline2](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/416/pull-azure-all/1317110189899386880#1:build-log.txt%3A456)

```
 An operation is currently performing on this storage account that requires exclusive access.
make: *** [Makefile:285: build-azure-vhd-ubuntu-1804] Error 1
make: *** Waiting for unfinished jobs.... 
```

Is it because ubuntu-1804 is getting built [here](https://github.com/kubernetes-sigs/image-builder/blob/05973968701f6da08a38769e5806f436add12cbc/images/capi/packer/azure/.pipelines/build-vhd.yaml#L23) & [here](https://github.com/kubernetes-sigs/image-builder/blob/05973968701f6da08a38769e5806f436add12cbc/images/capi/packer/azure/scripts/ci-azure-e2e.sh#L61) simultaneously?

CC: @CecileRobertMichon

CecileRobertMichon:
Fixes #417 

This addresses a flake in e2e where all jobs share the same resource group and the storage account name is only based on create time, which means if two builds create the storage account at the exact same time, the create operation will succeed because the storage already exists but might later result in ` An operation is currently performing on this storage account that requires exclusive access.` as two jobs will be using the same storage account.

This PR adds a random 4 letter suffix to the default storage account name and also adds a check before creating the storage account to make sure that the name is valid and that it is unique so that, if a collision does occur, it should fail with a clearer error message (although a collision could only occur two names are set at the exact same second and the same 4 letter suffix is chosen, ie. 𝑃(26,4)=> 1 in 358800).

### DIFF
--- a/images/capi/packer/azure/scripts/init-vhd.sh
+++ b/images/capi/packer/azure/scripts/init-vhd.sh
@@ -10,5 +10,7 @@ export RESOURCE_GROUP_NAME="${RESOURCE_GROUP_NAME:-cluster-api-images}"
 export AZURE_LOCATION="${AZURE_LOCATION:-southcentralus}"
 az group create -n ${RESOURCE_GROUP_NAME} -l ${AZURE_LOCATION} --tags ${TAGS}
 CREATE_TIME="$(date +%s)"
-export STORAGE_ACCOUNT_NAME="capi${CREATE_TIME}"
+RANDOM_SUFFIX="$(head /dev/urandom | LC_ALL=C tr -dc a-z | head -c 4 ; echo '')"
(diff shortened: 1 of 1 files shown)


kkeshavamurthy:
looks good to me. Thanks for the quick fix.

PR #424, opened 2020-11-04
CAPI: Add option to build Azure SIGs in CI script

CecileRobertMichon:
/hold for test-infra job

### DIFF
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -78,7 +78,7 @@
       "client_id": "{{user `client_id`}}",
       "client_secret": "{{user `client_secret`}}",
       "vm_size": "{{user `vm_size`}}",
-      "managed_image_name": "capi-{{user `build_timestamp`}}",
+      "managed_image_name": "{{user `image_name`}}-{{user `build_timestamp`}}",
       "managed_image_resource_group_name": "{{user `resource_group_name`}}",
       "shared_image_gallery_destination": {
         "resource_group": "{{user `resource_group_name`}}",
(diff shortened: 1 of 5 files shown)


kkeshavamurthy:
The tests are failing because there's not sig image-definition for centos-7

images/capi/packer/azure/packer.json
```diff
@@ -93,6 +93,7 @@
       "ssh_username": "packer",
       "azure_tags": {
         "build_timestamp": "{{user `build_timestamp`}}",
+        "creationTimestamp": "{{user `build_timestamp`}}",
```
CecileRobertMichon, line 96: this is to make sure the Azure subscription cleanup jobs don't delete the group halfway through the run. Kept the `build_timestamp` tag as well in case anything is relying on it.

images/capi/packer/azure/packer.json
```diff
@@ -78,7 +78,7 @@
       "client_id": "{{user `client_id`}}",
       "client_secret": "{{user `client_secret`}}",
       "vm_size": "{{user `vm_size`}}",
-      "managed_image_name": "capi-{{user `build_timestamp`}}",
+      "managed_image_name": "{{user `image_name`}}-{{user `build_timestamp`}}",
```
CecileRobertMichon, line 81: this makes sure there is no naming conflict between two images in the same RG (same e2e run)

CecileRobertMichon:
/hold cancel

@jsturtevant this is ready now

images/capi/packer/azure/packer.json
jsturtevant: I've added this and the `creationTimestamp` to the windows packer file as well.

PR #425, opened 2020-11-04
Add CI for OVAs

kkeshavamurthy:
Created a new CI target to build all platforms except RHEL
Added CI script. 

test-infra job [PR](https://github.com/kubernetes/test-infra/pull/19846)
CC: @codenrhoden

### DIFF
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -284,6 +284,7 @@
         "vmware-iso-base",
         "vsphere-iso-base"
       ],
+      "name": "packer-manifest",
       "output": "{{user `output_dir`}}/packer-manifest.json",
       "strip_path": true,
       "type": "manifest"
--- a/images/capi/scripts/ci-ova.sh
+++ b/images/capi/scripts/ci-ova.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit  # exits immediately on any unexpected error (does not bypass traps)
+set -o nounset  # will error if variables are used without first being defined
+set -o pipefail # any non-zero exit code in a piped command causes the pipeline to fail with that code
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
+
+TARGETS=("ubuntu-1804" "ubuntu-2004")
+
+on_exit() {
+  for target in ${TARGETS[@]};
+  do
+    echo "----------------------"
+    echo "${target} build logs"
+    echo "----------------------"
+    cat ${target}-${TIMESTAMP}.log
+  done
+
+  # kill the VPN
+  docker kill vpn
+}
+
+cleanup_build_vm() {
+  # Setup govc to delete build VM after
+  curl -L https://github.com/vmware/govmomi/releases/download/v0.23.0/govc_linux_amd64.gz | gunzip > govc
+  chmod +x govc
+  mv govc /usr/local/bin/govc
+
+  export GOVC_URL="${VSPHERE_SERVER}"
+  export GOVC_USERNAME="${VSPHERE_USERNAME}"
+  export GOVC_PASSWORD="${VSPHERE_PASSWORD}"
+  export GOVC_DATACENTER="SDDC-Datacenter"
+  export GOVC_INSECURE=true
+
+  for target in ${TARGETS[@]};
+  do
+    govc vm.destroy capv-ci-${target}-${TIMESTAMP}
+  done
+
+}
+
+trap on_exit EXIT
+
+export PATH=${PWD}/.local/bin:$PATH
+export PATH=${PYTHON_BIN_DIR:-"/root/.local/bin"}:$PATH
+export GC_KIND="false"
+export TIMESTAMP="$(date -u '+%Y%m%dT%H%M%S')"
+
+# Load VMC Creds
+if [ -f "${CONFIG_ENV-}" ]; then
+  set -o allexport && . "${CONFIG_ENV-}" && set +o allexport
+fi
+
+cat << EOF > packer/ova/vsphere.json
+{
+    "vcenter_server":"${VSPHERE_SERVER}",
+    "username":"${VSPHERE_USERNAME}",
+    "password":"${VSPHERE_PASSWORD}",
+    "datastore":"WorkloadDatastore",
+    "datacenter":"SDDC-Datacenter",
+    "cluster": "Cluster-1",
+    "network": "sddc-cgw-network-8",
+    "folder": "Workloads/ci/imagebuilder"
+}
+EOF
+
+# Since access to esxi is blocked due to firewall rules,
+# `export`, `post-processor` sections from `packer-node.json` are removed.
+cat packer/ova/packer-node.json | jq  'del(.builders[] | select( .name == "vsphere" ).export)' > packer/ova/packer-node.json.tmp && mv packer/ova/packer-node.json.tmp packer/ova/packer-node.json
+cat packer/ova/packer-node.json | jq  'del(."post-processors"[])' > packer/ova/packer-node.json.tmp && mv packer/ova/packer-node.json.tmp packer/ova/packer-node.json
+
+# Run the vpn client in container
+docker run --rm -d --name vpn -v "${HOME}/.openvpn/:${HOME}/.openvpn/" \
+  -w "${HOME}/.openvpn/" --cap-add=NET_ADMIN --net=host --device=/dev/net/tun \
+  gcr.io/cluster-api-provider-vsphere/extra/openvpn:latest
+
+# Tail the vpn logs
+docker logs vpn
+
+# install deps and build all images
+make deps-ova
+
+for target in ${TARGETS[@]};
+do
+cat << EOF > ci-${target}.json
+{
+"build_version": "capv-ci-${target}-${TIMESTAMP}"
+}
+EOF
+  PACKER_VAR_FILES="ci-${target}.json" make -j build-node-ova-vsphere-${target} > ${target}-${TIMESTAMP}.log 2>&1 &
+done
+wait
+
+cleanup_build_vm


images/capi/packer/ova/scripts/ci-ova.sh
```diff
@@ -0,0 +1,66 @@
...
+export PATH=${PWD}/.local/bin:$PATH
+export PATH=${PYTHON_BIN_DIR:-"/root/.local/bin"}:$PATH
+export GC_KIND="false"
+
+# timestamp is in RFC-3339 format to match kubetest
+export TIMESTAMP="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+export JOB_NAME="${JOB_NAME:-"image-builder-ova-e2e"}"
+export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
+
+cat << EOF > vmc.json
```
EleanorRigby, line 41: If we just edit vsphere.json , can we use the same targets to trigger this?

Any reason to create vmc.json

images/capi/Makefile
```diff
@@ -236,6 +242,10 @@ $(NODE_OVA_LOCAL_BASE_BUILD_TARGETS): deps-ova
 $(NODE_OVA_ESX_BUILD_TARGETS): deps-ova
 	packer build $(PACKER_NODE_FLAGS) -var-file="packer/ova/packer-common.json" -var-file="$(abspath packer/ova/$(subst build-node-ova-esx-,,$@).json)" -var-file="packer/ova/esx.json" -except=local -except=vsphere -only=vmware-iso $(ABSOLUTE_PACKER_VAR_FILES) packer/ova/packer-node.json
 
+.PHONY: $(NODE_OVA_VSPHERE_CI_BUILD_TARGETS)
+$(NODE_OVA_VSPHERE_CI_BUILD_TARGETS): deps-ova
+	packer build $(PACKER_NODE_FLAGS) -var-file="packer/ova/packer-common.json" -var-file="$(abspath packer/ova/$(subst build-node-ova-vsphere-,,$@).json)" -var-file="packer/ova/vsphere.json" -except=esx -except=local -only=vsphere-iso $(ABSOLUTE_PACKER_VAR_FILES) -only=vsphere packer/ova/packer-node.json
```
EleanorRigby, line 247: should this(packer/ova/vsphere.json) be vmc.json?

EleanorRigby (changes requested):
Minor questions + change

EleanorRigby (approved):
Looks good to me. will wait for one more review by someone who is familiar with CI

/lgtm

kkeshavamurthy:
ping @codenrhoden

kkeshavamurthy:
@codenrhoden can you please help in creating this folder on the VMC Cluster

```
Build 'vsphere' errored: error creating vm: folder '/SDDC-Datacenter/vm/imagebuilder' not found
```

codenrhoden:
@kkeshavamurthy yes, I was able to create a folder for you.

Try setting the folder to `/SDDC-Datacenter/vm/Workloads/ci/imagebuilder`

codenrhoden:
@kkeshavamurthy okay, some progress! It did try to create the VM. It's failing at loading the kickstart file because I don't think it has the right IP address. IN the VM console, it's trying to reach the Packer HTTP server at 10.60.115.27, which does not look like it's through the VPN to me.

If it was through the VPN, it would be looking on a **10.8.0.x** address.

I think the next actions here are:

- Get you a VMC account so you can see the VM consoles for debugging (there will probably be  more issues)
- Figure out how to get Packer to bind to the VPN address instead. Maybe there is an openvpn command that you can use to get the IP address, then use it to configure Packer with https://www.packer.io/docs/builders/vmware/vsphere-iso#http_bind_address.

codenrhoden:
I'm also going to let this job keep running for now (it's stuck) to see what the behavior is.

images/capi/packer/ova/packer-node.json
```diff
@@ -115,7 +115,7 @@
     },
     {
       "name": "vmware-iso",
-      "vm_name": "{{user `build_version`}}",
+      "vm_name": "{{user `build_version`}}-{{user `build_timestamp`}}",
```
codenrhoden, line 118: A couple things come to mine here... When we added the `vsphere-iso` builder, it originally had `-force` in the `packer build` command to prevent the error you are seeing with the VM already existing. I didn't like that idea because an errant `make` could quickly/easily destroy an existing build that may be in use as a template.

For CI, it may be appropriate to set `PACKER_FLAGS=-force` (I think that's what it would be). I think setting the VM name differently may have other effects, but now that I look through `packer-node.json` it doesn't appear to be used or passed on to the OVA scripts, so maybe it is safe. :)

Regardless, if you want to do this, I think it needs to go in `vsphere-iso` builder, not `vmware-iso`.

images/capi/Makefile
```diff
@@ -377,6 +387,7 @@ build-node-ova-vsphere-rhel-7: ## Builds RHEL 7 Node OVA and template on vSphere
 build-node-ova-vsphere-ubuntu-1804: ## Builds Ubuntu 18.04 Node OVA and template on vSphere
 build-node-ova-vsphere-ubuntu-2004: ## Builds Ubuntu 20.04 Node OVA and template on vSphere
 build-node-ova-vsphere-all: $(NODE_OVA_VSPHERE_BUILD_TARGETS) ## Builds all Node OVAs and templates on vSphere
+build-node-ova-vsphere-ci-all: $(NODE_OVA_VSPHERE_CI_BUILD_TARGETS) ## Builds all Node OVAs and templates on vSphere except RHEL
```
codenrhoden, line 390: Rather than adding additional make targets, I think I'd prefer to just keep the new `test-ova` target that calls the `ci-ova.sh` script, and control which targets are called in the script.

Meaning that in the script, we would do something like
```sh
make -j build-node-ova-vsphere-photon-3 build-node-ova-vsphere-ubuntu-2004 build-node-ova-vsphere-ubuntu-1804 build-node-ova-vsphere-centos-7
```

That way we just use the existing targets rather than creating special "CI" targets. Thoughts?

images/capi/packer/ova/packer-node.json
kkeshavamurthy: I meant to put it in `vsphere-iso`. Thanks for catching that. 
Yeah I looked thru the code and did not see the vm name being anywhere else so went this route. I prefer this method instead of force
from packer doc:
`  -force                        Force a build to continue if artifacts exist, deletes existing artifacts.`
So this means packer might destroy the vm that's started by another ci run?

images/capi/Makefile
kkeshavamurthy: This works. Will change it.

images/capi/packer/ova/packer-node.json
codenrhoden: Yeah, I agree. Since it't not being used anywhere else, seems safe.  In the job definition, we have this set:
https://github.com/kubernetes/test-infra/blob/5d64d2ba2ac533041f2328e0eaf4519799f38336/config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml#L13
So yeah, we can have concurrent builds. Which means that if we didn't do this, they would overwrite each other. :)

kkeshavamurthy:
@codenrhoden I think I'm running into this [issue](https://github.com/hashicorp/packer/issues/9973#issuecomment-728917359) 

`http_bind_address` sets the bind address correctly for packer but the boot line command issued is with the non-vpn IP which packer picks by default. There's a [PR](https://github.com/hashicorp/packer/pull/10273) open for the fix. 
Until then any idea for getting around the issue?

```
 ==> vsphere: File already uploaded; continuing
==> vsphere: Creating VM...
==> vsphere: Customizing hardware...
==> vsphere: Mounting ISO images...
2021/01/12 00:49:13 packer-builder-vsphere-iso plugin: Creating CD-ROM on controller '&{{{} 200 0xc00027af00 <nil> <nil> <nil> 0 <nil>} 0 []}' with iso '[WorkloadDatastore] packer_cache//d3e175624fc2d704975ce9a149f8f270e4768727.iso'
2021/01/12 00:49:13 packer-builder-vsphere-iso plugin: No floppy files specified. Floppy disk will not be made.
2021/01/12 00:49:13 packer-builder-vsphere-iso plugin: Found available port: 8286 on IP: 10.8.0.6
==> vsphere: Starting HTTP server on port 8286
==> vsphere: Set boot order temporary...
==> vsphere: Power on VM...
==> vsphere: Waiting 10s for boot...
==> vsphere: HTTP server is working at http://10.60.104.135:8286/
==> vsphere: Typing boot command...
2021/01/12 00:49:24 packer-builder-vsphere-iso plugin: Special code '<esc>' found, replacing with: CodeEscape
2021/01/12 00:49:24 packer-builder-vsphere-iso plugin: Waiting 1 second
2021/01/12 00:49:45 packer-builder-vsphere-iso plugin: Special code '<enter>' found, replacing with: CodeReturnEnter
2021/01/12 00:49:45 packer-builder-vsphere-iso plugin: Waiting 1 second
2021/01/12 00:49:46 packer-builder-vsphere-iso plugin: [INFO] Waiting for IP, up to total timeout: 30m0s, settle timeout: 5s
==> vsphere: Waiting for IP... 
```

codenrhoden:
While I kind of hate the idea, one way to maybe get this unstuck is to take export and post-processing out with `jq` until we come up with a solution. Maybe something like:

```
jq 'del(.builders[].export)' | jq 'del(."post-processors")'
```

For example, if I do a diff on before/after that command, it seems to only take out the needed stuff:

```sh
$ diff <(cat packer-node.json) <(cat packer-node.json | jq 'del(.builders[].export)' | jq 'del(."post-processors")')
139,142d138
<       "export": {
<         "force": true,
<         "output_directory": "{{user `base_output_dir`}}"
<       },
192,196d187
<       "export": {
<         "force": true,
<         "manifest": "{{ user `export_manifest`}}",
<         "output_directory": "{{user `output_dir`}}"
<       },
239,243d229
<       "export": {
<         "force": true,
<         "manifest": "{{ user `export_manifest`}}",
<         "output_directory": "{{user `output_dir`}}"
<       },
262,331d247
<   "post-processors": [
<     {
<       "custom_data": {
<         "build_date": "{{isotime}}",
<         "build_name": "{{user `build_name`}}",
<         "build_timestamp": "{{user `build_timestamp`}}",
<         "build_type": "node",
<         "containerd_version": "{{user `containerd_version`}}",
<         "custom_role": "{{user `custom_role`}}",
<         "disk_size": "{{user `disk_size`}}",
<         "distro_arch": "{{ user `distro_arch` }}",
<         "distro_name": "{{ user `distro_name` }}",
<         "distro_version": "{{ user `distro_version` }}",
<         "guest_os_type": "{{user `guest_os_type`}}",
<         "ib_version": "{{user `ib_version`}}",
<         "kubernetes_cni_semver": "{{user `kubernetes_cni_semver`}}",
<         "kubernetes_semver": "{{user `kubernetes_semver`}}",
<         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
<         "kubernetes_typed_version": "{{user `kubernetes_typed_version`}}",
<         "os_name": "{{user `os_display_name`}}",
<         "vsphere_guest_os_type": "{{user `vsphere_guest_os_type`}}"
<       },
<       "except": [
<         "vmware-iso-base"
<       ],
<       "output": "{{user `output_dir`}}/packer-manifest.json",
<       "strip_path": true,
<       "type": "manifest"
<     },
<     {
<       "inline": [
<         "cd {{user `output_dir`}}",
<         "tar xf {{user `build_version`}}.ova",
<         "rm {{user `build_version`}}.ova",
<         "../../hack/image-build-ova.py --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk1.vmdk"
<       ],
<       "name": "esx",
<       "type": "shell-local"
<     },
<     {
<       "inline": [
<         "cd {{user `output_dir`}}",
<         "../../hack/image-build-ova.py --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk-0.vmdk"
<       ],
<       "name": "vsphere",
<       "type": "shell-local"
<     },
<     {
<       "except": [
<         "vmware-iso-base"
<       ],
<       "inline": [
<         "./hack/image-build-ova.py --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ./hack/ovf_eula.txt --ovf_template ./hack/ovf_template.xml {{user `output_dir`}}",
<         "./hack/image-post-create-config.sh {{user `output_dir`}}"
<       ],
<       "name": "local",
<       "type": "shell-local"
<     },
<     {
<       "environment_vars": [
<         "CUSTOM_POST_PROCESSOR={{user `custom_post_processor`}}"
<       ],
<       "inline": [
<         "if [ \"$CUSTOM_POST_PROCESSOR\" != \"true\" ]; then exit 0; fi",
<         "{{user `custom_post_processor_command`}}"
<       ],
<       "name": "custom-post-processor",
<       "type": "shell-local"
<     }
<   ],
```

codenrhoden:
I feel like the long-term solution is to just create the working template in vCenter and leave it there (no export), but that means we will want to set fields in a template's vApp properties... I'm not sure what the best way to do that right now is, but it might be with `govc`. Either with a CLI (short-term) or by embedding into a a CLI (meaning I'm using this as an argument in favor of it).

codenrhoden:
/test pull-goss-populate
^ this seemed like an environment flake. It wouldn't even normally need to run for this PR, but I'd like to see if it passes now.

codenrhoden:
/lgtm
/approve

Nice work @kkeshavamurthy. You're persistence and hard work paid off! :)

Issue #18, opened 2019-08-14
CAPI - Define CI & Test approach


moshloop:
We need to be able to run tests on pull requests / periodically as well as provide an automated build & promotion process for images
/assign @figo @codenrhoden

codenrhoden:
So I've been starting to think about this, and am interested in figuring out next steps.

From the CAPI perspective, we are currently generating images for three providers -- AWS, vSphere, and GCE. Those providers use images built from this repo to deploy their clusters.

So how do we go about doing E2E tests? The only way to really prove out tests is by using the image with its associated provider. In other words, if we build a CAPA image, we need to use CAPA to deploy it to test it -- which will really be testing two things: the CAPA image, and whatever build of CAPA we used to deploy it. They are tightly coupled.

So that makes me wonder what testing we can/should do in this repo as part of CI, versus E2E tests found in the various providers' repos.

I guess the first steps no matter what is just an automated pipeline for building and publishing the images themselves. After that we can define what/how to test.

So I think that brings up the following questions:

- Where do we publish the images? (e.g. GCS buckets?)
- How ofter do we build/publish? 
  - I feel like every PR could be a bit much -- these are large images with full OS installations. Would doing this on a periodic basis be enough?
- What support matrix do we target?
  - What K8s versions do we need to build for? The latest patch release for n-2?
- How do want to handle securing credentials?
  - We'll need AWS keys for AMI building. VMC (most likely) keys for OVA building, etc.

I'm just thinking out loud here to start gathering information.

/cc @figo @timothysc @akutz @detiber 
^ feel free to ping anyone else that might be interested.

moshloop:
I really like this approach and I think it is a good step in putting a plan in place.

In terms of testing images, there is some level of testing that should be done in this repo outside of E2E testing in CAPI/KOPS/Kubespray etc :
* Basic sanity tests to ensure `kubeadm init` produces a functional control plane
* Reproducibility and bill-of-material variance. Byte level repeatable builds are out of scope, but we should have some level of confidence that repeated executions of the same build produce the same result - and critically even with the passage of time - we shouldn't find out that the build for K8 N - 2 on OS N - 2 is broken when a critical CVE needs to be released.
* Security testing, do the images have any known vulnerabilities? do they score below a threshold on CIS or similar - @randomvariable I know you were doing some work in this area?

> Where do we publish the images? (e.g. GCS buckets?)

[...]

PR #427, opened 2020-11-06
Update default k8s version 1.18.10

jsturtevant:
This updates the default k8s version as well as new crictl version.  I wasn't sure if it should be 1.18.10 or 1.19.3 just let me know which and I will update.

### DIFF
--- a/images/capi/packer/config/kubernetes.json
+++ b/images/capi/packer/config/kubernetes.json
@@ -1,8 +1,8 @@
 {
-  "kubernetes_series": "v1.17",
-  "kubernetes_semver": "v1.17.11",
-  "kubernetes_rpm_version": "1.17.11-0",
-  "kubernetes_deb_version": "1.17.11-00",
+  "kubernetes_series": "v1.18",
+  "kubernetes_semver": "v1.18.10",
+  "kubernetes_rpm_version": "1.18.10-0",
+  "kubernetes_deb_version": "1.18.10-00",
   "kubernetes_source_type": "pkg",
   "kubernetes_http_source": "https://storage.googleapis.com/kubernetes-release/release",
   "kubernetes_rpm_repo": "https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64",
@@ -13,7 +13,7 @@
   "kubernetes_container_registry": "k8s.gcr.io",
   "kubernetes_load_additional_imgs": "false",
   "kubeadm_template": "etc/kubeadm.yml",
-  "crictl_version": "1.16.1",
-  "crictl_sha256": "19fed421710fccfe58f5573383bb137c19438a9056355556f1a15da8d23b3ad1",
+  "crictl_version": "1.19.0",
+  "crictl_sha256": "87d8ef70b61f2fe3d8b4a48f6f712fd798c6e293ed3723c1e4bbb5052098f0ae",
   "crictl_source_type": "pkg"
 }


images/capi/packer/config/kubernetes.json
```diff
@@ -13,7 +13,7 @@
   "kubernetes_container_registry": "k8s.gcr.io",
   "kubernetes_load_additional_imgs": "false",
   "kubeadm_template": "etc/kubeadm.yml",
-  "crictl_version": "1.16.1",
+  "crictl_version": "1.19.0",
   "crictl_sha256": "19fed421710fccfe58f5573383bb137c19438a9056355556f1a15da8d23b3ad1",
```
detiber, line 17: Does this sha need to be updated as well?

jsturtevant: yes. I updated from https://github.com/kubernetes-sigs/cri-tools/releases/tag/v1.19.0

CecileRobertMichon:
> I wasn't sure if it should be 1.18.10 or 1.19.3 just let me know which and I will update.

I'm slightly leaning towards n-1 (so 1.18.10 at this point) and I think that's what we had discussed in the past but would like to know what others think. @moshloop / @detiber / @EleanorRigby / @kkeshavamurthy ? (Travis not tagging you on purpose but 
 feel free to chime in if you happen to read this)

kkeshavamurthy:
Leaning towards n-1 as well.

detiber:
> > I wasn't sure if it should be 1.18.10 or 1.19.3 just let me know which and I will update.
> 
> I'm slightly leaning towards n-1 (so 1.18.10 at this point) and I think that's what we had discussed in the past but would like to know what others think. @moshloop / @detiber / @EleanorRigby / @kkeshavamurthy ? (Travis not tagging you on purpose but
> feel free to chime in if you happen to read this)

I'm not overly picky on n or n-1 for the default. At some point we probably need to make it easier to build the latest image for n, n-1 and n-2, without as much work as it takes today, especially as we look to automate building/testing of images. That isn't something that needs to be solved in this PR, though.

jsturtevant:
Sounds like there is a slight preference for 1.18.  I will update for that version then

> At some point we probably need to make it easier to build the latest image for n, n-1 and n-2, without as much work as it takes today, especially as we look to automate building/testing of images. 

I believe https://github.com/kubernetes-sigs/image-builder/issues/18 is the tracking issue

images/capi/packer/config/kubernetes.json
```diff
@@ -1,8 +1,8 @@
 {
-  "kubernetes_series": "v1.17",
-  "kubernetes_semver": "v1.17.11",
-  "kubernetes_rpm_version": "1.17.11-0",
-  "kubernetes_deb_version": "1.17.11-00",
+  "kubernetes_series": "v1.18",
+  "kubernetes_semver": "v1.18.10",
```
CecileRobertMichon, line 3: should this be 1.18.12 now?

codenrhoden:
I know there is a beta out for 1.20.0 now, but we've previously agreed in office hours that our general approach would be to set the default K8s version to the latest version of the n-2 release. Since 1.19.x is still the latest, I'd like to keep this at 1.17.x.

Happy to revisit this if it's causing problems, however. And yes, we should document this. 😆

jsturtevant:
@codenrhoden sounds good. Do you think under https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/capi.md#configuration would be a good place to document it?

PR #433, opened 2020-11-18
capi: make containerd CRI socket configurable

dongsupark:
We should be able to configure path to containerd CRI socket file, `containerd_cri_socket`, for each distro.
Its main use case is to support distros with different CRI socket path, for example, `/run/docker/libcontainerd/container.sock` for Flatcar.
By default it is `/var/run/containerd/containerd.sock` as it has been in all other distros.

This PR is split out from https://github.com/kubernetes-sigs/image-builder/pull/248, as suggested by @codenrhoden.


codenrhoden (changes requested):
/ok-to-test

Thanks for breaking this out, @dongsupark.

I have one request. Rather than putting `containerd_cri_socket` into every other OS, let's instead put it in `config/containerd.json`. Since the existing value of `/var/run/containerd/containerd.sock` is indeed a default for all current OS's, there's no need to define it each place.

looking forward to flatcar support, you can then override this value in *only* your `flatcar.json` file. The makefile ensures that the OS-specific JSON file is loaded after all the "defaults" that are in the `config` folder, which for Packer means the OS-specific JSON files take precedence.

/assign @moshloop 
^ Moshe, I know you wanted to see this happen separately as well.

dongsupark:
@codenrhoden Done.
Now `containerd_cri_socket` is defined only in `packer/config/containerd.json`.

ncdc:
@fabriziopandini this is probably worth your 👀 too - thanks

codenrhoden:
/lgtm

Will wait for others to chime in.

images/capi/ansible/roles/containerd/templates/etc/crictl.yaml
```diff
@@ -0,0 +1 @@
+runtime-endpoint: unix://{{ containerd_cri_socket }}
```
detiber, line 1: Should there also be a related change to one of the task files in `images/capi/ansible/roles/containerd/tasks` to reference this file?

dongsupark: > Should there also be a related change to one of the task files in images/capi/ansible/roles/containerd/tasks to reference this file?

Good catch. Thanks.
I forgot about moving the corresponding changes from https://github.com/kubernetes-sigs/image-builder/pull/248, when splitting out the PR.
Updated this PR.

detiber (commented):
/lgtm

codenrhoden:
/approve

Thanks for the assist @detiber

PR #434, opened 2020-11-18
legacy image-builder: Pin pyrsistent to python2 compatible version

justinsb:
We still use python2 for these older images, and pyrsistent broke
python2 compatability in an update.  Pin to an older version (which is
likely more reproducible anyway).


codenrhoden:
Seems straight forward enough for even me to sign off on. 
/lgtm

PR #435, opened 2020-11-19
Update Ubuntu20 ISO image

prankul88:
This PR resolves the error which occurs while trying to build ubuntu20 image using `make build-qemu-ubuntu-2004`


hidekazuna (approved):
Thanks, good to me.

PR #438, opened 2020-11-21
Add OpenStack builder for CAPI OpenStack

hidekazuna:
Currently CAPI OpenStack user uses QEMU builder which requires local build environment. It is more useful to build an image from existing image in OpenStack environment.

### DIFF
--- a/docs/book/src/capi/providers/openstack.md
+++ b/docs/book/src/capi/providers/openstack.md
@@ -2,13 +2,18 @@
 
 ## Hypervisor
 
-The image is built using KVM hypervisor.
+The image is built using the following environments:
 
-### Prerequisites for QCOW2
+| Environment | Builder   | Build target       |
+|-------------|-----------|--------------------|
+| KVM         | QEMU      | build-qemu-ubuntu- |
+| OpenStack   | OpenStack | build-openstack    |
 
-Execute the following command to install qemu-kvm and other packages if you are running Ubuntu 18.04 LTS.
+### Prerequisites for QEMU Builder
 
 #### Installing packages to use qemu-img
+
+Execute the following command to install qemu-kvm and other packages if you are running Ubuntu 18.04 LTS.
 
 ```bash
 $ sudo -i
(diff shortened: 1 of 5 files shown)


docs/book/src/capi/providers/openstack.md
```diff
@@ -31,23 +36,35 @@ $ sudo chown root:kvm /dev/kvm
 
 Then exit and log back in to make the change take place.
 
+### Prerequisites for OpenStack Builder
+
+Complete the `openstack.json` configuration file with credentials and informations specific to the remote OpenStack environment.
+Please refer to the [Packer documentation for the Amazon AMI builder](https://www.packer.io/docs/builders/openstack) for each variables.
```
jichenjc, line 42: why it's Amazon AMI?

docs/book/src/capi/providers/openstack.md
```diff
@@ -31,23 +36,35 @@ $ sudo chown root:kvm /dev/kvm
...
-For building a ubuntu-2004 based capi image, run the following commands -
+| File                  | Description                                                    |
+|-----------------------|----------------------------------------------------------------|
+| qemu-ubuntu-1804.json | The settings for the Ubuntu 18.04 image                        |
+| qemu-ubuntu-2004.json | The settings for the Ubuntu 20.04 image                        |
+| openstack.json        | Additional settings needed when building on a remote OpenStack |
+
+### Output
+
+The image is built and located in `images/capi/output/BUILD_NAME+kube-KUBERNETES_VERSION` for QEMU Builder.
```
sbueringer, line 70: Should we really drop the instructions how to build it? I would prefer still documenting *both* variants.

images/capi/packer/qemu/openstack.json
```diff
@@ -0,0 +1,15 @@
+{
+  "build_name": "",
```
sbueringer, line 2: I'm not sure if it's intended to combine qemu in OpenStack builds in the same folder. Afaik usually AWS/Azure etc they all have their own folder. But that's something one of the maintainers has to decide :)

docs/book/src/capi/providers/openstack.md
hidekazuna: Updated to document both building images.

PR #439, opened 2020-11-27
Added ability to override variables with env vars

perithompson:
To help with automation and to avoid scripted file edits, I have added a way to create additional variables to the packer build command in the makefile by using environment variables.  With this addition, a user can now export any required variable with a `PKR_OVR_` prefix to avoid needing to edit a file.  These variables will then be appended to the end of the `packer build` command as `-var=key=value` and packer will override the contents of the user variables in configuration files.  I have updated the docs with the usage.

/kind feature

### DIFF
--- a/docs/book/src/capi/quickstart.md
+++ b/docs/book/src/capi/quickstart.md
@@ -32,27 +32,14 @@ With the CAPI image builder installed and dependencies satisfied, you are now re
 
 ## Environment Variables
 
-To assist with automation, there is also the option of overriding any variable with a corresponding environment variable prefixed with `PKR_OVR_`. A useful example of this is providing logon credentials for the vsphere.json or the kubernetes version you would like to build.  
+To assist with automation, there is also the option of overriding any variable with a corresponding environment variable prefixed with `PKR_OVR_`.
 
 ``` sh
-#Override vsphere.json
-export PKR_OVR_vcenter_server=vsphere
-export PKR_OVR_username=administrator@vsphere.local
-export PKR_OVR_password=myPassword
-export PKR_OVR_cluster=myCluster
-export PKR_OVR_datastore=myDatastore
-export PKR_OVR_datacenter=myDatacenter
-export PKR_OVR_folder=myFolder
-export PKR_OVR_resource_pool=myResourcePool
-export PKR_OVR_convert_to_template=true
-export PKR_OVR_linked_clone=false
-export PKR_OVR_create_snapshot=false
-export PKR_OVR_template=
-export PKR_OVR_insecure_connection=true
+### Any user variable can be set
+export PKR_OVR_crictl_version=v1.16.1
+export PKR_OVR_crictl_sha256=19fed421710fccfe58f5573383bb137c19438a9056355556f1a15da8d23b3ad1
+export PKR_OVR_kubernetes_semver=v1.19.4
 
 #Spaces are accepted but single quotes will not work.
-export PKR_OVR_network=VM Network
-
-### Any user variable can be set
-export PKR_OVR_kubernetes_semver=v1.19.4
+export PKR_OVR_additional_prepull_images=docker.io/sigwindowstools/flannel:0.12.0, docker.io/sigwindowstools/kube-proxy:v1.19.4
 ```
--- a/docs/book/src/capi/quickstart.md
+++ b/docs/book/src/capi/quickstart.md
@@ -29,3 +29,30 @@ export PATH=$PWD/.bin:$PATH
 ## Builds
 
 With the CAPI image builder installed and dependencies satisfied, you are now ready to build an image. In general, this is done via `make` targets, and each provider (e.g. AWS, GCE, etc.) will have different requirements for what information needs to be provided (such as cloud provider authentication credentials). Certain providers may have dependencies that are not satisfied by `make deps`, for example the vSphere provider needs access to a hypervisor (VMware Fusion on macOS, VMware Workstation on Linux). See the [specific documentation](./capi.md#providers) for your desired provider for more details.
+
+## Environment Variables
+
+To assist with automation, there is also the option of overriding any variable with a corresponding environment variable prefixed with `PKR_OVR_`. A useful example of this is providing logon credentials for the vsphere.json or the kubernetes version you would like to build.  
+
+``` sh
+#Override vsphere.json
+export PKR_OVR_vcenter_server=vsphere
+export PKR_OVR_username=administrator@vsphere.local
+export PKR_OVR_password=myPassword
+export PKR_OVR_cluster=myCluster
+export PKR_OVR_datastore=myDatastore
+export PKR_OVR_datacenter=myDatacenter
+export PKR_OVR_folder=myFolder
+export PKR_OVR_resource_pool=myResourcePool
+export PKR_OVR_convert_to_template=true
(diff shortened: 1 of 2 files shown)


jayunit100:
oh interesting, so someone might override these at boot time?

perithompson:
> oh interesting, so someone might override these at boot time?

Yep, @jayunit100 I've found it easier to iterated by running a container with environment variables rather than having to worry about going through and editing each of the various files containing user variables

jayunit100:
this is nice because we now dont have to edit json files anymore, or modify files with passwords anymore.
@perithompson is using this to run the entire build from docker , which is now easy w/ this PR

images/capi/Makefile
```diff
@@ -162,7 +170,8 @@ PACKER_NODE_FLAGS := $(foreach f,$(abspath $(COMMON_NODE_VAR_FILES)),-var-file="
 				$(PACKER_FLAGS)
 PACKER_HAPROXY_FLAGS := $(foreach f,$(abspath $(COMMON_HAPROXY_VAR_FILES)),-var-file="$(f)" ) \
 				$(PACKER_FLAGS)
-ABSOLUTE_PACKER_VAR_FILES := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" )
+ABSOLUTE_PACKER_VAR_FILES := $(foreach f,$(abspath $(PACKER_VAR_FILES)),-var-file="$(f)" ) \
```
jayunit100, line 173: can you add a comment here saying:
```
concatenate the packer_var_files with override_vars provided from the users input
```

CecileRobertMichon (commented):
/ok-to-test

docs/book/src/capi/quickstart.md
```diff
@@ -29,3 +29,30 @@ export PATH=$PWD/.bin:$PATH
 ## Builds
 
 With the CAPI image builder installed and dependencies satisfied, you are now ready to build an image. In general, this is done via `make` targets, and each provider (e.g. AWS, GCE, etc.) will have different requirements for what information needs to be provided (such as cloud provider authentication credentials). Certain providers may have dependencies that are not satisfied by `make deps`, for example the vSphere provider needs access to a hypervisor (VMware Fusion on macOS, VMware Workstation on Linux). See the [specific documentation](./capi.md#providers) for your desired provider for more details.
+
+## Environment Variables
+
+To assist with automation, there is also the option of overriding any variable with a corresponding environment variable prefixed with `PKR_OVR_`. A useful example of this is providing logon credentials for the vsphere.json or the kubernetes version you would like to build.
```
CecileRobertMichon, line 35: this doc is in the main capi quickstart but it only mentions vsphere, should the example be more generic if it applies to all providers?

CecileRobertMichon: If it's not generic and only applies to the vsphere builder (which the prefix `PKR_OVR` makes me think it is) let's more this section to the vsphere docs

jsturtevant: This will be useful for setting various values in Azure Windows build such as the windows updates patches

codenrhoden:
@perithompson Sorry it's taken me a while to look at this.

How is this different than using the existing `PACKER_FLAGS` solution?

An example of using this for RHEL builds is at https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/vsphere.md#rhel.

I think you can already accomplish what you want by setting `PACKER_FLAGS=-var=myvar=value`. No changes needed.

perithompson:
> @perithompson Sorry it's taken me a while to look at this.
> 
> How is this different than using the existing `PACKER_FLAGS` solution?
> 
> An example of using this for RHEL builds is at https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/vsphere.md#rhel.
> 
> I think you can already accomplish what you want by setting `PACKER_FLAGS=-var=myvar=value`. No changes needed.

Thanks @codenrhoden, I didn't see that option but I will test, the idea here is that if we are running in a docker container or a build script it felt more user friendly to add environment variables to override existing values from the environment or a single environment file like rather than specify 5-10 separate flags? At the moment it seems common practice to use jq to replace values in files which can make scripts hard to read.

So `docker run --env-file=envfile imagebuilder:latest 'build-node-ova-vsphere-windows-2019'`  rather than `PACKER_FLAGS=-var=windows_updates_categories="UpdateRollups SecurityUpdates CriticalUpdates" -var=username=my_username -var=password=mypassword -var=iso_path=somepath no_proxy=* make build-node-ova-vsphere-windows-2019`

I'm happy to use packer flags but I think it would be good to add the docs to make this non- RHEL specific.  I think I probably should leverage packer flags rather than add another option to the absolute files option though as I only really saw that as packer instructions rather than variables.

docs/book/src/capi/quickstart.md
perithompson: Thanks @CecileRobertMichon that is true, I'm clearly too used to just using the one builder I will try and make it more generic.

@jsturtevant yeah I have been finding the variable replacements getting longer and longer especially with changing between KBs and update categories

perithompson:
@codenrhoden any further thoughts on this?

codenrhoden:
> Thanks @codenrhoden, I didn't see that option but I will test, the idea here is that if we are running in a docker container or a build script it felt more user friendly to add environment variables to override existing values from the environment or a single environment file like rather than specify 5-10 separate flags? At the moment it seems common practice to use jq to replace values in files which can make scripts hard to read.

@perithompson Maybe I'm still not understanding? Apologies if I'm not. People should absolutely *not* be using jq to replace values in files. The files should always be left unmodified. What people *should* do is set all of their custom settings in a JSON file, and pass it with `PACKER_VAR_FILES`, which is documented [here](https://image-builder.sigs.k8s.io/capi/capi.html#examples). This would be the same idea as the env file, but does still require an extra step...

> 
> So `docker run --env-file=envfile imagebuilder:latest 'build-node-ova-vsphere-windows-2019'` rather than `PACKER_FLAGS=-var=windows_updates_categories="UpdateRollups SecurityUpdates CriticalUpdates" -var=username=my_username -var=password=mypassword -var=iso_path=somepath no_proxy=* make build-node-ova-vsphere-windows-2019`

In the Docker use case, the var file used by packer would need to be mounted into the container, and then referenced directly. So in your example, you would create something like this in a `vars.json` file:
```json
{
    "windows_updates_categories": "\"UpdateRollups SecurityUpdates CriticalUpdates\"",
    "username": "my_username",
    "password": "mypassword",
    "iso_path": "somepath",
    "no_proxy": "*"
}
```
and then run `docker run --mount type=bind,src=vars.json,dest=/tmp/vars.json PACKER_VAR_FILES=/tmp/vars.json make build-node-ova-vsphere-windows-2019`

@SanikaGawhane you are probably very interested in the above. :)

> 
> I'm happy to use packer flags but I think it would be good to add the docs to make this non- RHEL specific. I think I probably should leverage packer flags rather than add another option to the absolute files option though as I only really saw that as packer instructions rather than variables.

The example was indeed given in the RHEL section, but it is not RHEL specific. I *totally* agree the docs need improvement in that area.

perithompson:
@codenrhoden agreed, yeah I didn't notice that in the docs method but that would totally work! I'll close this, seems like that works better!

Issue #437, opened 2020-11-20
Add CI job for GCP 


eratnch:
Add a CI job for testing `GCP`  compatible `CAPI` images.

cpanato:
I will work on that if no one is against 
/assign

PR #423, opened 2020-10-29
Adding CI for AMIs

kkeshavamurthy:
Merge pending these PRs:
- [Boskos Support](https://github.com/kubernetes-sigs/boskos/pull/43) 
- [AWS sub-account creation](https://github.com/kubernetes/test-infra/pull/19645)
- [Prow job](https://github.com/kubernetes/test-infra/pull/19718)

/hold
CC: @codenrhoden

PR #445, opened 2020-12-03
gce: add scripts for CI test and nightly builds

cpanato:
The first part to add GCE image builder to run on CI

This PR adds the script that will be used in prow to build the images for CAPI/GCE

Also adds scripts to run nightly builds

Fixes: https://github.com/kubernetes-sigs/image-builder/issues/437

Will open the test-infra PR to add the job

Please let me know if there is a better way to use boskos 

/assign @detiber @vincepri

### DIFF
--- a/images/capi/packer/gce/ci/nightly/overwrite-1-18.json
+++ b/images/capi/packer/gce/ci/nightly/overwrite-1-18.json
@@ -1,0 +1,7 @@
+{
+  "build_timestamp": "nightly",
+  "kubernetes_deb_version": "1.18.18-00",
+  "kubernetes_rpm_version": "1.18.18-0",
+  "kubernetes_semver": "v1.18.18",
+  "kubernetes_series": "v1.18"
+}
--- a/images/capi/packer/gce/ci/nightly/overwrite-1-19.json
+++ b/images/capi/packer/gce/ci/nightly/overwrite-1-19.json
@@ -1,0 +1,7 @@
+{
+  "build_timestamp": "nightly",
+  "kubernetes_deb_version": "1.19.10-00",
+  "kubernetes_rpm_version": "1.19.10-0",
+  "kubernetes_semver": "v1.19.10",
+  "kubernetes_series": "v1.19"
+}
--- a/images/capi/packer/gce/ci/nightly/overwrite-1-20.json
+++ b/images/capi/packer/gce/ci/nightly/overwrite-1-20.json
@@ -1,0 +1,7 @@
+{
+  "build_timestamp": "nightly",
+  "kubernetes_deb_version": "1.20.6-00",
+  "kubernetes_rpm_version": "1.20.6-0",
+  "kubernetes_semver": "v1.20.6",
+  "kubernetes_series": "v1.20"
+}
--- a/images/capi/scripts/ci-gce-nightly.sh
+++ b/images/capi/scripts/ci-gce-nightly.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
+# usage: ci-gce-nightly.sh
+# This program build all images for capi gce for the nightly build
+################################################################################
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
+
+# Verify the required Environment Variables are present.
+: "${GOOGLE_APPLICATION_CREDENTIALS:?Environment variable empty or not defined.}"
+: "${GCP_PROJECT:?Environment variable empty or not defined.}"
+
+if [[ -z "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then
+  cat <<EOF
+GOOGLE_APPLICATION_CREDENTIALS is not set.
+Please set this to the path of the service account used to run this script.
+EOF
+  return 2
+else
+  gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
+fi
+
+# assume we are running in the CI environment as root
+# Add a user for ansible to work properly
+groupadd -r packer && useradd -m -s /bin/bash -r -g packer packer
+chown -R packer:packer /home/prow/go/src/sigs.k8s.io/image-builder
+# use the packer user to run the build
+
+# build image for 1.18
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-18.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
+
+# build image for 1.19
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-19.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
+
+# build image for 1.20
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-20.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
+
+echo "Displaying the generated image information"
+filter="name~cluster-api-ubuntu-*"
+gcloud compute images list --project "$GCP_PROJECT" \
+  --no-standard-images --filter="${filter}"
--- a/images/capi/hack/ensure-boskosctl.sh
+++ b/images/capi/hack/ensure-boskosctl.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+if [[ -z "$(command -v boskosctl)" ]]; then
+  echo "installing boskosctl"
+  GO111MODULE=on go get sigs.k8s.io/boskos/cmd/boskosctl@master
+  echo "'boskosctl' has been installed to $GOPATH/bin, make sure this directory is in your \$PATH"
+fi
+
+echo "testing boskosctl"
+boskosctl --help
--- a/images/capi/hack/ensure-go.sh
+++ b/images/capi/hack/ensure-go.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+# Ensure the go tool exists and is a viable version.
+verify_go_version() {
+  if [[ -z "$(command -v go)" ]]; then
+    if [[ "${INSTALL_GO:-"true"}" == "true" ]]; then
+      curl -sSL https://golang.org/dl/go${GO_VERSION:-"1.16.3"}.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+      export PATH=/usr/local/go/bin:$PATH
+      export PATH=$(go env GOPATH)/bin:$PATH
+    else
+      cat <<EOF
+Can't find 'go' in PATH, please fix and retry.
+See http://golang.org/doc/install for installation instructions.
+EOF
+      return 2
+    fi
+  fi
+
+  local go_version
+  IFS=" " read -ra go_version <<< "$(go version)"
+  local minimum_go_version
+  minimum_go_version=1.16.0
+  if [[ "${minimum_go_version}" != $(echo -e "${minimum_go_version}\n${go_version[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) && "${go_version[2]}" != "devel" ]]; then
+    cat <<EOF
+Detected go version: ${go_version[*]}.
+Kubernetes requires ${minimum_go_version} or greater.
+Please install ${minimum_go_version} or later.
+EOF
+    return 2
+  fi
+}
+
+echo "Checking if go is available"
+verify_go_version
+
+go version
+
+# Explicitly opt into go modules, even though we're inside a GOPATH directory
+export GO111MODULE=on
--- a/images/capi/packer/gce/ci/nightly/README.md
+++ b/images/capi/packer/gce/ci/nightly/README.md
@@ -1,0 +1,5 @@
+## Configs for nightly builds
+
+The configurations in the directory is being used for the nightly job to build the images for GCE.
+
+The script that runs is [ci-gce-nightly.sh](../../../../scripts/ci-gce-nightly.sh)
--- a/images/capi/scripts/ci-gce.sh
+++ b/images/capi/scripts/ci-gce.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################################################################
+# usage: ci-gce.sh
+# This program build all images for capi gce
+################################################################################
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
+
+# shellcheck source=ensure-go.sh
+source "./hack/ensure-go.sh"
+# shellcheck source=ensure-boskosctl.sh
+source "./hack/ensure-boskosctl.sh"
+
+# Verify the required Environment Variables are present.
+: "${GOOGLE_APPLICATION_CREDENTIALS:?Environment variable empty or not defined.}"
+
+function boskosctlwrapper() {
+  boskosctl --server-url http://"${BOSKOS_HOST}" --owner-name "cluster-api-provider-gcp" "${@}"
+}
+
+cleanup() {
+  echo "Cleaning up image"
+  filter="name~cluster-api-ubuntu-*"
+  (gcloud compute images list --project "$GCP_PROJECT" \
+    --no-standard-images --format="table[no-heading](name)" --filter="${filter}" \
+    | awk '{print "gcloud compute images delete --quiet --project '"$GCP_PROJECT"' "$1" " "\n"}' \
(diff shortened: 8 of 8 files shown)


cpanato:
Also, do we need to create any accounts?

detiber:
> Also, do we need to create any accounts?

Shouldn't we should be able to leverage existing accounts similar to what we are doing with CAPG :)

images/capi/hack/checkout_account.py
```diff
@@ -0,0 +1,60 @@
...
+
+import urllib
+import httplib
+import os
+import sys
+import json
+
+BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
+
+RESOURCE_TYPE = "gce-project"
```
detiber, line 27: Thinking ahead, this should probably be configurable to allow for other boskos resource types (such as aws accounts when they are available)

images/capi/hack/checkin_account.py
```diff
@@ -0,0 +1,38 @@
...
+
+import sys
+import httplib
+import urllib
+import os
+
+BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
+BOSKOS_RESOURCE_NAME=os.environ['BOSKOS_RESOURCE_NAME']
+
+USER = "image-builder"
```
detiber, line 25: Wondering if this should be `sigs.k8s.io/image-builder` to avoid any potential confusion with other projects named `image-builder`

images/capi/hack/checkout_account.py
```diff
@@ -0,0 +1,60 @@
...
+USER = "image-builder"
```
detiber, line 28: Wondering if this should be `sigs.k8s.io/image-builder` to avoid any potential confusion with other projects named `image-builder`

images/capi/hack/checkout_account.py
```diff
@@ -0,0 +1,60 @@
...
+    #  Afterwards we should just check for free
+    if status == 404:
+        status, reason, result = post_request(BOSKOS_HOST, "free")
+
+    if status != 200:
+        sys.exit("Got invalid response %d: %s" % (status, reason))
+
+    body = json.loads(result)
+    print 'export BOSKOS_RESOURCE_NAME="%s";' % body['name']
+    print 'export GCP_PROJECT="%s";' % body['name']
```
detiber, line 60: Ah, forgot that this bit was specific to the account type...  Maybe it would be good to call this `checkout_gcp_project.py` instead?

images/capi/hack/heartbeat_account.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+import httplib
+import urllib
+import os
+import time
+
+BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
+BOSKOS_RESOURCE_NAME=os.environ['BOSKOS_RESOURCE_NAME']
+
+USER = "image-builder"
```
detiber, line 26: Wondering if this should be `sigs.k8s.io/image-builder` to avoid any potential confusion with other projects named `image-builder`

images/capi/packer/gce/scripts/ci-gce.sh
```diff
@@ -0,0 +1,106 @@
...
+fi
+if [[ -z "$GCP_REGION" ]]; then
+  cat <<EOF
+GCP_REGION is not set.
+Please specify which the GCP region to use.
+EOF
+  return 2
+fi
+
+export GCP_PROJECT_ID="$GCP_REGION"
```
detiber, line 97: This bit doesn't seem right...  Should this be set to `GCP_PROJECT` instead?
```suggestion
export GCP_PROJECT_ID="$GCP_PROJECT"
```

detiber (commented):
It might be good to dedupe boskos management approaches with https://github.com/kubernetes-sigs/image-builder/pull/423/files#diff-1c56e5a5df7b6f697f270ac094d292b4a3929e1ebafd0a1970f35c895d4a986e

images/capi/hack/checkin_account.py
cpanato: so for that, we need to somehow create this user, right?

images/capi/packer/gce/scripts/ci-gce.sh
cpanato: nice catch!
copy/paste issues lol

images/capi/hack/checkout_account.py
cpanato: made in one script as this PR https://github.com/kubernetes-sigs/image-builder/pull/423 and make that generic for both AWS/GCP

ping @kkeshavamurthy  as well to be aware and reduce the effort

images/capi/hack/checkout_account.py
cpanato: made the user `sigs.k8s.io/image-builder` not sure if this is a valid user

images/capi/packer/gce/scripts/ci-gce.sh
cpanato: done

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,143 @@
...
+        content = r.content.decode()
+        result = json.loads(content)
+
+        print(f"export BOSKOS_RESOURCE_NAME={result['name']}")
+        if cloud.lower() == "aws":
+          print(f"export AWS_ACCESS_KEY_ID={result['userdata']['access-key-id']}")
+          print(f"export AWS_SECRET_ACCESS_KEY={result['userdata']['secret-access-key']}")
+        elif cloud.lower() == "gcp":
+          print(f"export GCP_PROJECT={result['name']}")
+
```
detiber, line 55: nit: should only be a single space here instead of double
```suggestion
```

images/capi/packer/gce/scripts/ci-gce.sh
```diff
@@ -0,0 +1,104 @@
...
+
+export GCP_REGION=${GCP_REGION:-"us-east4"}
+export GCP_B64ENCODED_CREDENTIALS=$(base64 -w0 "$GOOGLE_APPLICATION_CREDENTIALS")
+
+cleanup() {
+   # removing the image created
+  (gcloud compute images list --project "$GCP_PROJECT" \
+    --no-standard-images --filter="family:capi-ubuntu-*" --format="table[no-heading](name)" \
+        | awk '{print "gcloud compute images delete --project '"$GCP_PROJECT"' --quiet " $1 "\n"}' \
+        | bash) || true
```
detiber, line 39: Do we need to worry about this potentially deleting images that I user wouldn't want deleted, say they were running the ci script from the same account they use to actually publish images using image-builder?

Is there a way for us to uniquely mark images generated from this script vs normal usage of image-builder?

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,143 @@
...
+    parser.add_argument(
+        '--cloud', dest="cloud", type=str,
+        required=True,
+        help="Which Cloud to use. AWS/GCP"
+    )
+
+    args = parser.parse_args()
+
+    if args.checkout_account:
+        checkout_account(args.resource_type, args.user, args.cloud)
```
detiber, line 133: This call should probably be wrapped with try/except to catch the exceptions thrown above and use `sys.exit()` when the exception is caught instead of having the uncaught exception end execution.

Similar with `release_account` below

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,143 @@
...
+import os
+import time
+
+import requests
+
+BOSKOS_HOST = os.environ.get("BOSKOS_HOST", "boskos")
+BOSKOS_RESOURCE_NAME = os.environ.get('BOSKOS_RESOURCE_NAME')
+
+
+def checkout_account(resource_type, user, cloud, input_state="clean"):
```
detiber, line 30: `clean` was an artifact of an error when we initially created the aws accounts for capa, I don't think we need to keep the support here, especially since the original capa accounts are all now fixed.

```suggestion
def checkout_account(resource_type, user, cloud, input_state="free"):
```

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,143 @@
...
+        'state': input_state,
+        'dest': 'busy',
+    }
+
+    r = requests.post(url, data)
+
+    if r.status_code == 404:
+      data['state'] = "free"
+      r = requests.post(url, data)
+
```
detiber, line 44: Similar to above, we shouldn't need to worry about supporting both `free` and `clean` here.
```suggestion
```

images/capi/hack/checkin_account.py
detiber: There shouldn't be any need to create the user, it's just there for reporting purposes

images/capi/hack/boskos.py
cpanato: :) python

images/capi/hack/boskos.py
cpanato: did the change, not sure if that is correct, thanks for the review

images/capi/packer/gce/scripts/ci-gce.sh
cpanato: makes sense, I will filter for the image that we produce in the build

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,150 @@
...
+        except:
+            print("Failed to checkout account")
+            sys.exit(1)
+
+    elif args.release_account:
+        try:
+            release_account(args.user)
+        except:
+            print("Failed to release account")
+            sys.exit(1)
```
detiber, line 140: This should work to avoid losing the text from the exception raised, I haven't tested it, though:
```suggestion
        except Exception as exp:
            print("Failed to release account: ", exp)
            sys.exit(1)
```

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    args = parser.parse_args()
+
+    if args.checkout_account:
+        try:
+            checkout_account(args.resource_type, args.user, args.cloud)
+        except:
+            print("Failed to checkout account")
+            sys.exit(1)
```
detiber, line 133: This should work to avoid losing the text from the exception raised, I haven't tested it, though:
```suggestion
        except Exception as exp:
            print("Failed to checkout account: ", exp)
            sys.exit(1)
```

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,150 @@
...
+        except:
+            print("Failed to release account")
+            sys.exit(1)
+
+    elif args.send_heartbeat:
+        try:
+            send_heartbeat(args.user)
+        except:
+            print("Failed to send heartbeat")
+            sys.exit(1)
```
detiber, line 147: `send_heartbeat` doesn't raise any exceptions, so no need to wrap the call here.
```suggestion
        send_heartbeat(args.user)
```

detiber (commented):
A few suggestions around the exception handling, otherwise this is looking pretty good

images/capi/hack/boskos.py
cpanato: learning python :) thanks for the review

images/capi/packer/gce/scripts/ci-gce.sh
cpanato: added another filter, let me know what you think

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,146 @@
...
+
+"""Checks out a Image-Builder AWS or GCP account"""
+
+import argparse
+import json
+import os
+import time
+import sys
+
+import requests
```
kkeshavamurthy, line 25: `requests` module is not installed by default on kubekins images.

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,146 @@
...
+        '--user', dest="user", type=str,
+        default="sigs.k8s.io/image-builder",
+        help="username"
+    )
+
+    parser.add_argument(
+        '--cloud', dest="cloud", type=str,
+        required=True,
+        help="Which Cloud to use. AWS/GCP"
+    )
```
kkeshavamurthy, line 124: Do you think we can infer the iaas based on the resourse-type? All aws resources have `aws` in their name.

images/capi/hack/boskos.py
cpanato: good to know, thanks @kkeshavamurthy

images/capi/hack/boskos.py
cpanato: we can then :)

images/capi/hack/boskos.py
cpanato: addressed your feedback
thanks for your review

images/capi/hack/boskos.py
cpanato: done

images/capi/hack/boskos.py
```diff
@@ -0,0 +1,144 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Checks out a Image-Builder AWS or GCP account"""
+
+import argparse
+import httplib
```
kkeshavamurthy, line 20: `httplib` module is not available in python3. `http` might be the one. 

Or you can just install `requests` module from `ci` script. 
something like 
`pip3 install -r requirements.txt`
requirements.txt will contain `requests`

detiber: Oh man, just about forgot about python2/3 related issues. This made me wonder if there was anything provided by boskos that we could use to simplify things. It looks like we could use `boskosctl` to do the heavy lifting for us: https://github.com/kubernetes-sigs/boskos/tree/master/cmd/boskosctl

cpanato: sorry for the long delay :( 

I will check if any image provided contains the boskosctl or we download that on the fly

but I agree with the idea to use the command line instead of a python script

cpanato: @detiber replace to use boskosctl, now we need to make sure the image we are using in prow job have go installed 🙃 

let me know what you think

cpanato: Just saw now that boskosctl have a container image, so we don't need an extra script to install it, just need to use the extra container, will drop these changes and document

CecileRobertMichon:
Is there an associated test-infra PR to add the job as optional so we can use it to test the PR before merging?

cpanato:
@CecileRobertMichon not yet, I will work on that, Was trying to get this side in shape first :)

cpanato:
Test-infra PR: https://github.com/kubernetes/test-infra/pull/21414

cpanato:
finally, this is ready and working! :)
please review and let me know your thoughts
/assign @CecileRobertMichon @detiber @dims @codenrhoden 

Opened a PR on test-infra to add the periodic job this as well: https://github.com/kubernetes/test-infra/pull/22041

/hold cancel

cpanato:
`pull-image-builder-gcp-all — Job succeeded.` 🎉 

```
--> ubuntu-1804: A disk image was created: cluster-api-ubuntu-1804-v1-18-15-1619883016
```

dims:
@cpanato once the job finishes (either fails or succeeds, we need to clean up the image) (so we could say add a `trap` to remove the image 

see https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/a05a24469e70578b805c4d87998fb43eb53e492b/hack/ci/e2e-conformance.sh#L152-L156

cpanato:
@dims thanks for the feedback. I thought boskos clean the account after some time. but my assumption is wrong. will add the delete images.

cpanato:
Added as well a job to run the nightly builds and build for now 1.18/1.19 and 1.20, and if the image builds successfully it will replace the previous image (with the Packer flags -force)

I tried the script in my personal account and it is working and build both three images and when re-running it replace if the builds succeed.

Then I think we can use those image in the CI for CAPG :)

cc @dims

cpanato:
made the changes @dims :)
thanks for your review and feedback!

dims:
@cpanato looks good to me. will give everyone else a chance to review :)

images/capi/packer/gce/ci/nightly/overwrite-1-18.json
```diff
@@ -0,0 +1,7 @@
+{
+  "build_timestamp": "nigthly",
```
codenrhoden, line 2: s/nightly/nigthly

images/capi/packer/gce/ci/nightly/overwrite-1-19.json
```diff
@@ -0,0 +1,7 @@
+{
+  "build_timestamp": "nigthly",
```
codenrhoden, line 2: same typo here

images/capi/packer/gce/ci/nightly/overwrite-1-20.json
```diff
@@ -0,0 +1,7 @@
+{
+  "build_timestamp": "nigthly",
```
codenrhoden, line 2: ditto

images/capi/scripts/ci-gce-nightly.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# assume we are running in the CI environment as root
+# Add a user for ansible to work properly
+groupadd -r packer && useradd -m -s /bin/bash -r -g packer packer
+chown -R packer:packer /home/prow/go/src/sigs.k8s.io/image-builder
+# use the packer user to run the build
+
+# build image for 1.18
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-18.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
```
codenrhoden, line 53: `deps-gce` is already a dependency of the `gce-all` build target, so adding that explicitly ins't strictly necessary. But I have no problem with it being listed out.

images/capi/scripts/ci-gce-nightly.sh
```diff
@@ -0,0 +1,66 @@
...
+# build image for 1.19
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-19.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
+
+# build image for 1.20
+# using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
+su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=packer/gce/ci/nightly/overwrite-1-20.json PACKER_FLAGS=-force make deps-gce build-gce-all'"
+
+echo "Displaying the generated image information"
+filter="name~cluster-api-ubuntu-1804-*"
```
codenrhoden, line 64: I notice this filter includes 1804 only, but the ci-gce script does not (just has `cluster-api-ubuntu-*`. Do we want to make this one the same so that it picks up 2004? I see that 20.04 Ubuntu builds are currently commented out in the Makefile (however, not completely because it still shows up in `make help`), but if we change this filter I think it would mean that you wouldn't need to change anything when Ubuntu 20.04 *does* get reinstated?

codenrhoden (commented):
I think everything I have is a nit, but it would be nice to see the typo fixed first.

images/capi/packer/gce/ci/nightly/overwrite-1-19.json
cpanato: thanks for the review, it is clear that I don't know how to write this word 😂

images/capi/scripts/ci-gce-nightly.sh
cpanato: good catch! thanks!

cpanato: thanks for your review and feedback @codenrhoden 
made the changes requested

codenrhoden:
/test pull-azure-vhds
/lgtm

That windows Azure failure (I want to make some changes soon that are a little more sane about what tests run depending on what changed - there's no need for Azure tests in this case for example) doesn't quite look like a flake to me, but we'll see what happens.

cpanato:
I think we can override the azure-vhds tests and investigate that in a follow you because that is not part of this PR

WDYT @codenrhoden ?

and to close this PR need some final reviews/approval: 
/assign @vincepri @detiber @CecileRobertMichon

spiffxp:
I am ok merging this to unblock and finding an alternate approach as followup

spiffxp:
I'll review more closely at keyboard tomorrow

cpanato:
> I am ok merging this to unblock and finding an alternate approach as followup

I'm checking what are the changes need to apply here to use cloudbuild together with packer to build the image

cpanato:
open a 🧵  to discuss better the cloudbuild option to build CAPG GCE images: https://kubernetes.slack.com/archives/CCK68P2Q2/p1621237438253400

codenrhoden:
/lgtm
/approve

Agreed that merging this now and exploring other options as follow-up makes sense. Will see what happens with the flaky Azure test.

CecileRobertMichon:
The error in the failed pull-azure-vhds job above is:

```
[0;32m    vhd-windows-2019: fatal: [default]: FAILED! => {"changed": true, "cmd": "$exists=docker network ls -f name=host -q\nif (-not $exists) { docker network create -d nat host }", "delta": "0:00:01.455251", "end": "2021-05-16 08:12:06.223301", "msg": "non-zero return code", "rc": 1, "start": "2021-05-16 08:12:04.768050", "stderr": "error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks?filters=%7B%22name%22%3A%7B%22host%22%3Atrue%7D%7D: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.\nerror during connect: Post http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks/create: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.", "stderr_lines": ["error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks?filters=%7B%22name%22%3A%7B%22host%22%3Atrue%7D%7D: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.", "error during connect: Post http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks/create: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running."], "stdout": "", "stdout_lines": []}[0m
```

@jsturtevant @perithompson have you seen this before?

cpanato:
will override this failure

cpanato:
I cant :/

/test pull-azure-vhds

perithompson:
> The error in the failed pull-azure-vhds job above is:
> 
> ```
> �[0;32m    vhd-windows-2019: fatal: [default]: FAILED! => {"changed": true, "cmd": "$exists=docker network ls -f name=host -q\nif (-not $exists) { docker network create -d nat host }", "delta": "0:00:01.455251", "end": "2021-05-16 08:12:06.223301", "msg": "non-zero return code", "rc": 1, "start": "2021-05-16 08:12:04.768050", "stderr": "error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks?filters=%7B%22name%22%3A%7B%22host%22%3Atrue%7D%7D: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.\nerror during connect: Post http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks/create: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.", "stderr_lines": ["error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks?filters=%7B%22name%22%3A%7B%22host%22%3Atrue%7D%7D: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.", "error during connect: Post http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/networks/create: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running."], "stdout": "", "stdout_lines": []}�[0m
> ```
> 
> @jsturtevant @perithompson have you seen this before?

No this is a new one on me, looks like the docker service didn't start on this node?

PR #447, opened 2020-12-03
Document default k8s versions skew

jsturtevant:
This updates the docs to call out the default version in the kubernetes configuration as mentioned in https://github.com/kubernetes-sigs/image-builder/pull/427#issuecomment-729977732


jsturtevant:
/assign @codenrhoden 

Small PR to update the docs based on your comment in #427 (comment)

PR #444, opened 2020-12-03
Bump containerd version to 1.4.3

dims:
Link to the new release: https://github.com/containerd/containerd/releases/tag/v1.4.3

1.4.3 has fixes for CVE-2020-15257. Details are here - https://github.com/containerd/containerd/security/advisories/GHSA-36xw-fx78-c5r4

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #449, opened 2020-12-05
Align containerd version for Windows and Linux

### LINKED ISSUE #446: Containerd versions should be same value for windows and linux
Follow up to  #444  to use the same containerd version configuration file as we do with Kubernetes versions.  Example: https://github.com/kubernetes-sigs/image-builder/blob/6dcd5528554a6ffb996565395c5b93e5a1843556/images/capi/Makefile#L148-L149

Some Contained configuration will be specific to Windows such as the sha for the file download and the additional settings hence the windows specific override file

/assign

jsturtevant:
fixes #446 found in #444 

I also saw some issues with starting containerd service and setting the path which are resolved.


codenrhoden (approved):
It took me a second to see the change for ``containerd_sha256={{user `containerd_sha256_windows`}}`` in `ansible-args-windows.json`, but once I saw that I understand what is happening here.

/lgtm
Will approve in a day or two if nothing else pops up.

PR #450, opened 2020-12-08
Add image-promotion process to publish container images

SanikaGawhane:
Use GCR to publish official container images for image-builder. 

This PR

- Edits `Makefile` to build and push container image to the staging registry.
- Adds `cloudbuild.yaml` to produce a container image

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -20,6 +20,9 @@ SHELL := /usr/bin/env bash
 
 .DEFAULT_GOAL := help
 
+# This option is for running docker manifest command
+export DOCKER_CLI_EXPERIMENTAL := enabled
+
 export IB_VERSION ?= $(shell git describe --dirty)
 
 ## --------------------------------------
@@ -94,6 +97,7 @@ deps-qemu:
 ## Container variables
 ## --------------------------------------
 REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+STAGING_REGISTRY := gcr.io/k8s-staging-scl-image-builder
 IMAGE_NAME ?= cluster-node-image-builder
 CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= dev
@@ -570,9 +574,14 @@ clean-packer-cache:
 ## Docker targets
 ## --------------------------------------
 ##@ Docker
+
+.PHONY: docker-pull-prerequisites
+docker-pull-prerequisites:
+	docker pull docker/dockerfile:1.1-experimental
+
 .PHONY: docker-build
-docker-build: ## Build the docker image for controller-manager
-	docker build --pull --build-arg PASSED_IB_VERSION=$(IB_VERSION) --build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
+docker-build: docker-pull-prerequisites ## Build the docker image for controller-manager
+	DOCKER_BUILDKIT=1 docker build --build-arg PASSED_IB_VERSION=$(IB_VERSION) --build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
 
 .PHONY: docker-push
 docker-push: ## Push the docker image
@@ -585,3 +594,13 @@ docker-push: ## Push the docker image
 .PHONY: test-azure
 test-azure: ## Run the tests for Azure builders
 	$(abspath packer/azure/scripts/ci-azure-e2e.sh)
+
+## --------------------------------------
+## Release targets
+## --------------------------------------
+##@ Release
+
+.PHONY: release-staging
+release-staging: ## Builds and push container images to the staging bucket.
+	TAG=$(IB_VERSION) REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-build docker-push
+
--- a/images/capi/cloudbuild.yaml
+++ b/images/capi/cloudbuild.yaml
@@ -1,0 +1,18 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 3000s
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  # 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f'
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud@sha256:0ef22100e63c0f7cdf4758d4732008c99d51ec149baca736f8cf94d7cf9b7a1b'
+    entrypoint: make
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+      - TAG=$_GIT_TAG
+      - DOCKER_BUILDKIT=1
+    args:
+      - release-staging
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.1-experimental
+
 # Copyright 2020 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");


images/capi/cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
...
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  # 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f'
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud@sha256:0ef22100e63c0f7cdf4758d4732008c99d51ec149baca736f8cf94d7cf9b7a1b'
+    entrypoint: make
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
+      - TAG=$_GIT_TAG
+      - PULL_BASE_REF=$_PULL_BASE_REF
```
codenrhoden, line 12: After thinking about this a bit more, I think we can make our approach even simpler. I know we were taking this from CAPA, and if you look in their `MAKEFILE`, it is accepting these two vars. Our Makefile is only looking for `TAG`, but not `PULL_BASE_REF`. I also don't think we need either.

Our makefile dynamically determines the version by setting the `IB_VERSION` variable. I'd rather just continue to use that. Looking at it more, I'd rather see the Makefile not use `TAG` at all -- it should just be using `IB_VERSION`, but let's set that aside for now.

If my above comment is true (that we don't need the `DOCKER_CLI_...` var) I think we can remove the entire `env:` section, the `substitution:` section, and probably the `options:` section too? What do you think?

The way you've said up the Makefile target is perfect - a raw call to `make release-staging` should just do the right thing, no extra env vars needed.

codenrhoden (commented):
/ok-to-test

images/capi/cloudbuild.yaml
```diff
@@ -0,0 +1,19 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 3000s
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  # 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f'
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud@sha256:0ef22100e63c0f7cdf4758d4732008c99d51ec149baca736f8cf94d7cf9b7a1b'
+    entrypoint: make
+    env:
+      - DOCKER_CLI_EXPERIMENTAL=enabled
```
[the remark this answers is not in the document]
CecileRobertMichon: you can use this [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1061/files) as an example of leveraging docker build kit to build images

images/capi/cloudbuild.yaml
CecileRobertMichon: CAPA isn't doing the latest in terms of image building. I'd recommend looking at CAPI or CAPZ.

codenrhoden: @CecileRobertMichon Thanks for pointing that out, I wasn't aware.

It looks to me like the latest stuff is all about using Docker manifests and enabling multi-arch images. Is that a fair assessment? If it is, do you see implementing that as a requirement for a first pass at building and publishing an image through the promotion process?

I realize there are great examples of how to do this, since it's already done in CAPZ and CAPI. Would you be accepting of starting simple and not doing the manifest/buildkit approach, and tackling that when/if we do multi-arch images?

CecileRobertMichon: The main advantage is that the docker buildkit approach is _much_ faster. Before we switched to that approach in CAPZ recently, or image pushing job was very flaky (see the difference between before and after the [PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1061/files) merged on Dec 1st: https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/post-cluster-api-provider-azure-push-images).

To me, the improvement is worth the extra effort. Whether in this PR or a follow-up doesn't really matter but might as well get it right from the start if we aren't in a rush to merge this?

codenrhoden: Thanks @CecileRobertMichon. That is definitely a speed up that is well worth it.

I've asked @SanikaGawhane to take the buildkit + `docker manifest` approach for now. I do still think we have the tag/version of the container image solved via a separate method. If we need to change that, that's okay of course, but I don't think it needs to be part of this PR.

detiber: While `PULL_BASE_REF` might not be needed, the `_GIT_TAG` substitution is specified during the call go GCB: https://github.com/kubernetes/test-infra/tree/master/images/builder#gcb-builder

codenrhoden: Looking through that Go code (in `main.go`) I still don't understand how it is used, and the README there isn't very helpful. I think I'll have to figure how to test this locally in order to understand it.

codenrhoden:
> I've asked @SanikaGawhane to take the buildkit + docker manifest approach for now. I do still think we have the tag/version of the container image solved via a separate method. If we need to change that, that's okay of course, but I don't think it needs to be part of this PR.

@SanikaGawhane Any luck with the `buildkit` approach? I see the latest commit doesn't have it.

images/capi/cloudbuild.yaml
detiber: I think the main thing is that PROW sets the env var as `GIT_TAG`, and we are using `TAG`, so the substitution does that mapping so that we create the image with the correct tag, at least when triggered by a postsubmit job, such as: https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-cluster-api.yaml#L14

As an aside, I believe the reason we are using PULL_BASE_REF in the CAPI jobs is to do the same thing for post submit jobs that are triggered for the branch vs the tag, so that the tag name associated with the branch is updated in the staging bucket.

codenrhoden:
At first blush, this looks good to me now. I still need to see if I can run the cloudbuild stuff locally to see how the `_GIT_TAG` stuff works I guess? That's the only piece I'm not understanding. It's a bit overkill/overcomplication IMO. We just want to tag the container with a version that we already know in our Makefile. 🤷‍♂️ 

/retest

codenrhoden:
I'm going to try and play with this this afternoon, but in the meantime, any other comments are welcome. Thanks for keeping up the work on this, @SanikaGawhane!

images/capi/Makefile
```diff
@@ -548,9 +552,14 @@ clean-packer-cache:
...
+
+.PHONY: docker-pull-prerequisites
+docker-pull-prerequisites:
+	docker pull docker/dockerfile:1.1-experimental
+
 .PHONY: docker-build
-docker-build: ## Build the docker image for controller-manager
-	docker build --pull --build-arg PASSED_IB_VERSION=$(IB_VERSION) --build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
+docker-build: docker-pull-prerequisites ## Build the docker image for controller-manager
+    DOCKER_BUILDKIT=1 docker build --build-arg PASSED_IB_VERSION=$(IB_VERSION) --build-arg ARCH=$(ARCH) . -t $(CONTROLLER_IMG)-$(ARCH):$(TAG)
```
codenrhoden, line 562: Nit: this appears to have changed from a single tab to spaces. Let's make it consistent with all the others.

codenrhoden (commented):
@SanikaGawhane I took this for another testdrive, and when I run `make docker-build`, nothing actually happens for me. It pulls the `dockerfile` container, and then... nothing. Does this work for you?  Here's what I see:

```sh
[trhoden@mbp ~/git/sigs.k8s.io/image-builder/images/capi (sg-image-promoter *)]$ TAG=dev-buildkit make docker-build
docker pull docker/dockerfile:1.1-experimental
1.1-experimental: Pulling from docker/dockerfile
Digest: sha256:de85b2f3a3e8a2f7fe48e8e84a65f6fdd5cd5183afa6412fff9caa6871649c44
Status: Image is up to date for docker/dockerfile:1.1-experimental
docker.io/docker/dockerfile:1.1-experimental
[trhoden@mbp ~/git/sigs.k8s.io/image-builder/images/capi (sg-image-promoter *)]$
```

But if I run the `docker build` command directly, it works just fine with buildkit. I'm not sure what the deal is here.

images/capi/Makefile
codenrhoden: Actually, this change explans my other comment. I thought I had tested it already, but I must have done something weird. When I change this to a single tab, this works for me.

SanikaGawhane: Fixed now!

images/capi/Makefile
```diff
@@ -585,3 +594,13 @@ docker-push: ## Push the docker image
...
 	$(abspath packer/azure/scripts/ci-azure-e2e.sh)
+
+## --------------------------------------
+## Release targets
+## --------------------------------------
+##@ Release
+
+.PHONY: release-staging
+release-staging: ## Builds and push container images to the staging bucket.
+	REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-build docker-push
```
codenrhoden, line 605: @SanikaGawhane One more thing I'd like to see here, which is add `TAG=$(IB_VERSION)` at the beginning so that the container image is built with the correct tag, rather than the default `dev`. We can do this for just this Make target so it's correct if you do `release-staging`, but it is `dev` if you just do `make docker-build` yourself. Maybe that will change later, but the release target should definitely use the correct tag.

SanikaGawhane: > @SanikaGawhane I took this for another testdrive, and when I run `make docker-build`, nothing actually happens for me. It pulls the `dockerfile` container, and then... nothing. Does this work for you?

@codenrhoden There was an extra character that caused an error in the way the `CONTROLLER_IMG` variable was being set. It should work now.

Referring to more previous discussions:
> Our makefile dynamically determines the version by setting the `IB_VERSION` variable:

I am leaving the `GIT_TAG` in place as of now (till we have more clarification) even though the above statement is true.

images/capi/Makefile
```diff
@@ -94,8 +97,9 @@ deps-qemu:
 ## Container variables
 ## --------------------------------------
 REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+STAGING_REGISTRY := gcr.io/k8s-staging-scl-image-builder
 IMAGE_NAME ?= cluster-node-image-builder
-CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
+CONTROLLER_IMG ?= $(REGISTRY)$(IMAGE_NAME)
```
codenrhoden, line 102: The removed `/` here needs to be added back. Otherwise you get this for a container name:

`gcr.io/k8s-staging-scl-image-buildercluster-node-image-builder-amd64:v0.1.6-169-g7c456fd04`.

codenrhoden (changes requested):
I don't think you meant to check in the `.idea` folder here. Please remove.

codenrhoden:
/lgtm

@detiber and @CecileRobertMichon could you take a look at this again? I think this is good to go, or at least at a  place that we can merge and iterate. When running `make release-staging` now, you'd end up with a container like the following (built from Sanika's branch):

```
gcr.io/k8s-staging-scl-image-builder/cluster-node-image-builder-amd64:v0.1.6-169-g3cd852c4d
```

CecileRobertMichon (commented):
lgtm

codenrhoden (approved):
I'll merge this in Tuesday morning (since Monday is holiday) if there are no further comments.

codenrhoden:
/hold
Just waiting until Tuesday
Forgot this repo treats GitHub review approvals as `/approve`

PR #457, opened 2020-12-14
Add custom-post-processor block to packer ova building

kkeshavamurthy:
Use `custom_post_processor` & `custom_post_processor_command` variable to run custom ova image post-processing.


codenrhoden:
/lgtm

@kkeshavamurthy Can you confirm that this works fine if `custom_post_processor` and `custom_post_processor_command` are completely unset? Just want to check since they don't have default values explicitly set.

kkeshavamurthy:
Yes I checked that. Even if they are not included in the vars list, they pass.

codenrhoden:
Thank you.

/approve

PR #459, opened 2020-12-15
Fix link to ubuntu 20 iso url

caulagi:
The link to ubuntu 20 iso url seems to have changed. I am not sure if there are any downstream projects that build these images, but the changes are similar to https://github.com/kubernetes-sigs/image-builder/commit/23b03aa079705d97c0fd45840ff1d127e0d564c1


codenrhoden:
Thanks @caulagi. LGTM. Can we also please update this in the file `images/capi/packer/qemu/qemu-ubuntu-2004.json`?

Oh wait, that's already updated. Doh. I wonder when those two got out of sync...

/approve

Issue #454, opened 2020-12-11
capi packer qemu build no longer builds qemu images; unconditionally builds vagrant  boxes


carstenkoester:
Steps:

```
git clone https://github.com/kubernetes-sigs/image-builder
git checkout fda34643 # HEAD at the time this issue was reported
cd image-builder/images/capi
make build-qemu-ubuntu-2004
```

The build artifact used to be (and is expected to be) a qcow2 image in the `output/` directory, but now always builds a Vagrant box:

```
...
==> ubuntu-2004: Gracefully halting virtual machine...
==> ubuntu-2004: Converting hard drive...
==> ubuntu-2004: Running post-processor: vagrant
==> ubuntu-2004 (vagrant): Creating a dummy Vagrant box to ensure the host system can create one correctly
==> ubuntu-2004 (vagrant): Creating Vagrant box for 'libvirt' provider
    ubuntu-2004 (vagrant): Copying from artifact: output/ubuntu-2004-kube-v1.17.11/ubuntu-2004-kube-v1.17.11
    ubuntu-2004 (vagrant): Compressing: Vagrantfile
    ubuntu-2004 (vagrant): Compressing: box.img
    ubuntu-2004 (vagrant): Compressing: metadata.json
Build 'ubuntu-2004' finished after 11 minutes 49 seconds.

# ls output/
# ls *box
packer_ubuntu-2004_libvirt.box
```

Checking out an earlier revision, df55d6f0, restores the behavior of `make build-qemu-ubuntu-2004` actually building qemu/qcow2 artifacts.

It seems like the post-processor was recently added to `packer.json`,

```
$ git blame -f -s packer.json
...
d8e1b6b23 images/capi/packer/qemu/packer.json  96)     }
892282320 images/capi/packer/qemu/packer.json  97)   ],
892282320 images/capi/packer/qemu/packer.json  98)   "post-processors": [
892282320 images/capi/packer/qemu/packer.json  99)     {
6ab0a4feb images/capi/packer/qemu/packer.json 100)       "type": "vagrant"
892282320 images/capi/packer/qemu/packer.json 101)     }
d8e1b6b23 images/capi/packer/qemu/packer.json 102)   ]
d8e1b6b23 images/capi/packer/qemu/packer.json 103) }
```

with 892282320 and 6ab0a4feb both being merged yesterday (as part of #248).

[...]

PR #460, opened 2020-12-16
Fix QEMU image for OpenStack not to execute Flatcar specific provisioner and post-processor

hidekazuna:
bootstrap-flatcar.sh provisioner and vagrant post-processor are only needed for Flatcar.

Fixes: #454

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -219,7 +219,8 @@ AZURE_BUILD_SIG_NAMES	?=	azure-sig-ubuntu-1804 azure-sig-ubuntu-2004 azure-sig-c
 
 DO_BUILD_NAMES 			?=	do-centos-7 do-ubuntu-1804 do-ubuntu-2004
 
-QEMU_BUILD_NAMES			?=	qemu-flatcar qemu-ubuntu-1804 qemu-ubuntu-2004
+QEMU_FLATCAR_BUILD_NAMES	?=	qemu-flatcar
+QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804 qemu-ubuntu-2004
 
 ## --------------------------------------
 ## Dynamic build targets
@@ -246,6 +247,8 @@ AZURE_BUILD_SIG_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_SIG_NAMES))
 AZURE_VALIDATE_SIG_TARGETS	:= $(addprefix validate-,$(AZURE_BUILD_SIG_NAMES))
 DO_BUILD_TARGETS 	:= $(addprefix build-,$(DO_BUILD_NAMES))
 DO_VALIDATE_TARGETS 	:= $(addprefix validate-,$(DO_BUILD_NAMES))
+QEMU_FLATCAR_BUILD_TARGETS	:= $(addprefix build-,$(QEMU_FLATCAR_BUILD_NAMES))
+QEMU_FLATCAR_VALIDATE_TARGETS	:= $(addprefix validate-,$(QEMU_FLATCAR_BUILD_NAMES))
 QEMU_BUILD_TARGETS	:= $(addprefix build-,$(QEMU_BUILD_NAMES))
 QEMU_VALIDATE_TARGETS	:= $(addprefix validate-,$(QEMU_BUILD_NAMES))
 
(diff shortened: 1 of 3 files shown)


images/capi/packer/qemu/packer.json
```diff
@@ -42,7 +42,35 @@
...
+      ],
+      "net_device": "virtio-net",
+      "qemu_binary": "{{user `qemu_binary`}}",
+      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'usermod -L {{user `ssh_username`}} && {{user `shutdown_command`}}'",
+      "ssh_password": "{{user `ssh_password`}}",
+      "ssh_username": "{{user `ssh_username`}}",
+      "ssh_timeout": "2h"
+    },
+    {
+      "name": "flatcar",
```
dongsupark, line 73: So, before the PR, `name` was `build_name`, i.e. "flatcar-{{env `FLATCAR_CHANNEL`}}-{{env `FLATCAR_VERSION`}}", like `flatcar-stable-2605.9.0`.
Now after the PR, `name` becomes simply `flatcar`.
I see why you need to do that, the interpolation does not seem to work for post-processors.

Though can you please change also the corresponding [box_name](https://github.com/kubernetes-sigs/image-builder/blob/d3e691d1cac78793ec406a7aca0ab387562f9e5c/images/capi/hack/image-build-flatcar.sh#L49) to `packer_flatcar_libvirt.box`?
Without that, `hack/image-builder-flatcar.sh` will simply fail to run, because of a missing box file.

hidekazuna: @dongsupark Thanks, I changed  box_name to `packer_flatcar_libvirt.box`.

https://github.com/kubernetes-sigs/image-builder/pull/460/files#diff-3def1683adb98fe2f5f79c7fed643eacafea9b2c9af33e6220dcc9e2576a7cc9R49

dongsupark: @hidekazuna Thanks for the fix.
Flatcar part seems to work well.
I'll let other folks have a look as well.

jichenjc:
/lgtm

I have no background of this flatcar thing, but as @dongsupark said above seems it works, so I think we are good to go

images/capi/Makefile
```diff
@@ -337,13 +340,21 @@ $(DO_BUILD_TARGETS): deps-do
...
+	packer build $(PACKER_NODE_FLAGS) -var-file="$(abspath packer/qemu/$(subst build-,,$@).json)" $(ABSOLUTE_PACKER_VAR_FILES) -only=flatcar packer/qemu/packer.json
+
+.PHONY: $(QEMU_FLATCAR_VALIDATE_TARGETS)
+$(QEMU_FLATCAR_VALIDATE_TARGETS): deps-qemu
+	packer validate $(PACKER_NODE_FLAGS) -var-file="$(abspath packer/qemu/$(subst validate-,,$@).json)" $(ABSOLUTE_PACKER_VAR_FILES) -only=flatcar packer/qemu/packer.json
+
 .PHONY: $(QEMU_BUILD_TARGETS)
 $(QEMU_BUILD_TARGETS): deps-qemu
-	packer build $(PACKER_NODE_FLAGS) -var-file="$(abspath packer/qemu/$(subst build-,,$@).json)" $(ABSOLUTE_PACKER_VAR_FILES) packer/qemu/packer.json
+	packer build $(PACKER_NODE_FLAGS) -var-file="$(abspath packer/qemu/$(subst build-,,$@).json)" $(ABSOLUTE_PACKER_VAR_FILES) -only=local packer/qemu/packer.json
```
sbueringer, line 353: What does `local` refer to / mean in this context? I guess any other value then `flatcar` would lead to the same, right?

sbueringer: Just a bit confused because `flatcar` is an os-name, and local is not. isn't flatcar also biuld locally?

hidekazuna: I found we can use `-except=flatcar` instead of `-only=local`. I will remove `local`.

codenrhoden:
This seems fine to me since both Flatcar and QEMU contributors have taken a look. If the LGTM comes back, I'm happy to approve.

hidekazuna:
@jichenjc PTAL

PR #461, opened 2020-12-17
capi: fix Flatcar issues with versions and AMIs

dongsupark:
As soon as Flatcar stable [2605.10.0](https://www.flatcar-linux.org/releases/#release-2605.10.0) was released, the script `image-grok-latest-flatcar-version.sh` stopped working as expected.
What happens is, the json [feed](https://www.flatcar-linux.org/releases-json/releases-stable.json) returns a list of versions with `2605.9.0` at the end, not `2605.10.0`, because the list is sorted alphanumerically.

To convert the list into a semver-sorted one, we should apply a filter `sort --version-sort`, not a pure `sort`.

We need to update the default channel & version of Flatcar to in Makefile.

Also fix root_device_name of Flatcar AMI.
In case of AWS AMI, the root device of Flatcar AMI must be `/dev/xvda`.
Otherwise the instance does not boot.

### DIFF
--- a/images/capi/hack/image-grok-latest-flatcar-version.sh
+++ b/images/capi/hack/image-grok-latest-flatcar-version.sh
@@ -6,5 +6,5 @@ curl -s \
      "https://www.flatcar-linux.org/releases-json/releases-$channel.json" \
     | jq -r 'to_entries[] | "\(.key)"' \
     | grep -v "current" \
-    | sort \
+    | sort --version-sort \
     | tail -n1
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -190,8 +190,8 @@ RHEL_VERSIONS				:=	rhel-7
 UBUNTU_VERSIONS			:=	ubuntu-1804 ubuntu-2004
 
 # Set Flatcar Container Linux channel and version if not supplied
-FLATCAR_CHANNEL ?= beta
-FLATCAR_VERSION ?= 2643.1.0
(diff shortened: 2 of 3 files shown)


images/capi/hack/image-grok-latest-flatcar-version.sh
```diff
@@ -6,5 +6,5 @@ curl -s \
      "https://www.flatcar-linux.org/releases-json/releases-$channel.json" \
     | jq -r 'to_entries[] | "\(.key)"' \
     | grep -v "current" \
-    | sort \
+    | sort --version-sort \
     | tail -n1
```
pothos, line 10: Or you can do:
```
curl -s -S -f -L https://${channel}.release.flatcar-linux.net/amd64-usr/current/version.txt | grep -m 1 FLATCAR_VERSION= | cut -d = -f 2
```

dongsupark: Sure, that would also work.
I don't care much, one way or another.
Anyway I am not the original author of the code, so I would prefer to have less change if possible.

dongsupark:
Pushed another fix about rootfs of Flatcar AMIs.

All existing AMIs on AWS were deleted.
The AMIs were already built with the rootfs fix, and published.

codenrhoden:
No other comments coming in, so I'm going to take that as no objections. :)

/approve

PR #462, opened 2020-12-19
Update the Azure publishing pipeline to output result from pub

jsturtevant:
This updates the script for the case where the command errors out the error and the message is lost. This now prints out any output from pub

/assign @CecileRobertMichon

### DIFF
--- a/images/capi/packer/azure/scripts/new-sku.sh
+++ b/images/capi/packer/azure/scripts/new-sku.sh
@@ -65,7 +65,10 @@ echo "Getting pub..."
 (set -x ; curl -fsSL https://github.com/devigned/pub/releases/download/v0.2.6/pub_v0.2.6_linux_amd64.tar.gz -o pub; tar -xzf pub)
 
 echo "Creating new SKU"
-(set -x ; ./pub_linux_amd64 skus put -p $PUBLISHER -o "$OFFER" -f sku.json ; echo "")
+set -x 
+./pub_linux_amd64 skus put -p $PUBLISHER -o "$OFFER" -f sku.json 
+set +x
+echo -e "\nCreated sku"
 
 echo "Writing publishing info"
 cat <<EOF > sku-publishing-info.json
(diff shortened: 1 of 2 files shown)


images/capi/packer/azure/scripts/new-sku.sh
```diff
@@ -65,7 +65,10 @@ echo "Getting pub..."
 (set -x ; curl -fsSL https://github.com/devigned/pub/releases/download/v0.2.6/pub_v0.2.6_linux_amd64.tar.gz -o pub; tar -xzf pub)
 
 echo "Creating new SKU"
-(set -x ; ./pub_linux_amd64 skus put -p $PUBLISHER -o "$OFFER" -f sku.json ; echo "")
+set -x 
+./pub_linux_amd64 skus put -p $PUBLISHER -o "$OFFER" -f sku.json 
+set +x
+echo "Created sku"
```
CecileRobertMichon, line 71: missing newline before Create Sku

images/capi/packer/azure/scripts/new-sku.sh
```diff
@@ -65,7 +65,10 @@ echo "Getting pub..."
 (set -x ; curl -fsSL https://github.com/devigned/pub/releases/download/v0.2.6/pub_v0.2.6_linux_amd64.tar.gz -o pub; tar -xzf pub)
 
 echo "Creating new SKU"
-(set -x ; ./pub_linux_amd64 skus put -p $PUBLISHER -o "$OFFER" -f sku.json ; echo "")
```
CecileRobertMichon, line 68: Want to do the same thing for new disk version while you're here?

images/capi/packer/azure/scripts/new-sku.sh
jsturtevant: updated to echo -e "\nCreated sku"

CecileRobertMichon (approved):
/lgtm
/approve

PR #467, opened 2020-12-29
Merge ovf templates

farodin91:


codenrhoden:
/ok-to-test

Things looks interesting, @farodin91! Thanks for trying to simplify this all. I'll try to take this for a spin soon and make sure everything comes out good, but at first glance I see what you are doing and I like it.

One nice thing is that CAPV doesn't use HAProxy anymore, so I think in a future version of image-builder, this distinction between the two will go away.

images/capi/hack/image-build-ova.py
```diff
@@ -174,13 +195,14 @@ def create_ova(path, infile_paths):
         f.write(sha256(path))
 
 
-def create_ovf(path, data, type):
+def create_ovf(path, data, template_path):
     print("image-build-ova: create ovf %s" % path)
+    template = ""
+    with io.open(template_path, 'r', encoding='utf-8') as f:
```
kkeshavamurthy, line 201: This fails because we chdir to the builddir line:76. So it looks for `ovf_template.xml` inside the build_dir which is not present. So you might want to get the abspath of the template before changing the dir

farodin91: I just move it to top. I thought would be better to use the same logic as eula.

images/capi/hack/ovf_template.xml
```diff
@@ -0,0 +1,171 @@
...
+    </vmw:BootOrderSection>
+    <EulaSection>
+      <Info>An end-user license agreement</Info>
+      <License>
+${EULA}
+      </License>
+    </EulaSection>
+    <ProductSection>
+      <Info>Information about the installed software</Info>
+      <Product>${PRODUCT}</Product>
```
kkeshavamurthy, line 158: This change the product name to 
`<Product>Cluster API vSphere image - VMware Photon OS 64-bit and Kubernetes v1.17.11</Product>`

from 

`<Product>VMware Photon OS 64-bit and Kubernetes v1.17.11</Product>`

@codenrhoden not sure if matters?

codenrhoden: While I like the new description better, I'd need to do more checking to see if this breaks a couple of downstream dependencies that are parsing this vApp data. Merging these templates is a great idea, changing what the output is in the same PR perhaps is not. Let's try to keep everything the same for now. Additions are fine, but let's limit the modifications.

farodin91: Should be resolved right now.

farodin91:
I resolved merged conflicts by moving the meta data part into the xml file.

images/capi/hack/ovf_template.xml
codenrhoden: Looks to me to get the old string (and again, I agree with your proposed change, we just need to handle it separately at a later time) this would need to use `${ANNOTATION}` instead.

farodin91: I'm a bit confused which change should be handled later or is it okay this way?

codenrhoden: Everything in the `<Product>` section need to stay the same. I know of downstream consumers that are (unfortunately) doing string matches about things that are contained within. The `<Product>` section is also often referred to as the "vApp Config".

Additions (what I would call new metadata) is fine if it has a purpose. But for right now, I want to make sure that any existing fields are not modified.

farodin91: I'll move the newly added meta data only to include in the node. A part from meta data, everything should be right now the same.

kkeshavamurthy:
Thanks @farodin91 
/lgtm

images/capi/hack/ovf_template.xml
farodin91: Product should be now the same for both haproxy and node.

farodin91:
@codenrhoden I did rebase on top of the base branch. Do you review it again?

codenrhoden:
There are a couple little white space things, but nothing that I think matters at the moment. I was able to do a before and after diff on the OVF and it all looked good to me.

/lgtm

If no other comments come in, I will merge on Monday!

PR #469, opened 2021-01-08
Fix comparison for custom post processor

MaxRink:
Currently the evaluation of the condition for the custom prost processor just fails, this fixes it
before:
```
==> vsphere (shell-local): Running local shell script: /tmp/packer-shell814371810
==> vsphere (shell-local): /tmp/packer-shell814371810: 2: /tmp/packer-shell814371810: [[: not found
```

### DIFF
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -426,7 +426,7 @@
         "CUSTOM_POST_PROCESSOR={{user `custom_post_processor`}}"
       ],
       "inline": [
-        "if [[ $CUSTOM_POST_PROCESSOR != \"true\" ]]; then exit 0; fi",
+        "if [ $CUSTOM_POST_PROCESSOR != \"true\" ]; then exit 0; fi",
         "{{user `custom_post_processor_command`}}"
       ]
     }


CecileRobertMichon (commented):
/ok-to-test

codenrhoden:
/lgtm

The other places where we've done something similar use single `[ ]` brackets, so this looks right to me.

https://github.com/kubernetes-sigs/image-builder/blob/a3cc74f6a3ab041116d7b8b828efd73d44ecb253/images/capi/packer/ami/packer.json#L114
https://github.com/kubernetes-sigs/image-builder/blob/a3cc74f6a3ab041116d7b8b828efd73d44ecb253/images/capi/packer/azure/packer.json#L112

@kkeshavamurthy I think you worked on this. Care to do a quick double-check that this wasn't working before?

kkeshavamurthy:
`[[` are not supported in `sh`. Are you running on `sh` shell @MaxRink ?

MaxRink:
> `[[` are not supported in `sh`. Are you running on `sh` shell @MaxRink ?

Packer uses sh by default for shell-local ( https://www.packer.io/docs/provisioners/shell-local#inline_shebang )

codenrhoden:
I do think we should take this change. I want things to work with default settings, and our similar (previously l linked) do the same thing.

kkeshavamurthy:
lgtm

PR #470, opened 2021-01-08
add distro version & arch info to OVF Metadata

kkeshavamurthy:
OVF Metadata was missing distro version and arch fields. 

**Additional context**

/assign @codenrhoden


codenrhoden:
/lgtm

Even though this isn't also being added to the HAProxy OVA, I think I'd like to see #467 merge first. Can you check that one out too, @kkeshavamurthy ?

PR #472, opened 2021-01-13
WIP : Add GOSS check to ensure minimum kernel_version for different OSs

SanikaGawhane:
This check will ensure the kernel version with which OS image is shipped with is >= kernel version specified in OS specific CVEs.

Currently the CVE specified minimum versions are :
Photon - [https://github.com/vmware/photon/wiki/Security-Updates-3.0-140](url)
Redhat - [https://access.redhat.com/security/cve/cve-2020-14386](url)
Amazon Linux 2 - [https://alas.aws.amazon.com/AL2/ALAS-2020-1488.html](url)
Ubuntu - [https://ubuntu.com/security/CVE-2020-14386](url)

### DIFF
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -70,7 +70,7 @@ kubernetes_cni_rpm_version: ""
 # When k8s and k8s cni source is http
 kubernetes_load_additional_imgs: false
 
-# OS Specific package/Command/Kernal Params etc...
+# OS Specific package/Command/Kernel Params etc...
 # Structured in below format
 # OS_NAME
 #   common-package:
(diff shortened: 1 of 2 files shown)


images/capi/packer/goss/goss-vars.yaml
```diff
@@ -84,6 +84,12 @@ kubernetes_load_additional_imgs: false
 #  ...
 amazon linux:
   common-package: *common_rpms
+  common-kernel-param:
+    kernel.osrelease:
+      command:
+        uname -r | awk '{vers=match($0,/(4\.19\.138\-(4[0-9]*|[5-9]))|(4\.19\.(139|1[4-9][0-9]|[2-9][0-9][0-9]))|(4\.[2-9][0-9])|(5\.)/); print vers}':
```
kkeshavamurthy, line 90: Just curious, can you please explain what this exactly does?

SanikaGawhane: `uname -r` returns the kernel-release version in the image.
`match` function of `awk` returns a 1 or 0 based on whether the kernel version in the image is >= kernel-version in CVE(in this case 4.19.138-4) which is stored and printed using `vers`
The following lines check for stdout which should reflect the `vers` value printed.

EleanorRigby:
Is it possible to use something like 

```
example:
  content:
    - 1.0.1
    - 1.9.9
  matches:
     semver-constraint: ">1.0.0 <2.0.0 !=1.5.0"
```

from here: https://github.com/aelsabbahy/goss/blob/master/docs/manual.md#goss-test-creation

codenrhoden:
Is this getting refactored to use the different syntax?

PR #474, opened 2021-01-14
[WiP] Expose ContainerD mirror support for "advanced" usecases

MaxRink:
Internally we run all our Clusters airgapped, thus we make use of the mirror support in ContainerD.
This PR exposes that in the image building process so that other folx can use it as well, if it would be applicable in their environment.
This is a WiP, mostly there to gather feedback 😅 ( for example its not working for flatcar and windows right now, as its a straight copy from our fork)

### DIFF
--- a/images/capi/ansible/roles/containerd/defaults/main.yml
+++ b/images/capi/ansible/roles/containerd/defaults/main.yml
@@ -17,3 +17,47 @@ containerd_flatcar_bins:
 - containerd
 - containerd-shim
 
+# examples on how to use the mirror function
+# registry_mirrors:
+#   harbor:
+#     - harbor.yourdomain.com
+#     - harbor-alternative.yourdomain.com
+#   artifactory:
+#     - artifactory.yourdomain.com
+
+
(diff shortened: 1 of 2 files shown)


voor:
I understand this change, and don't have a problem with it, I'm just wondering if we're trending in this path of re-implementing containerd's configuration through a fairly restrictive API that exists in image-builder, and if we might want to take a step back and discuss how we could enable similar capabilities over time.  The reason I ask, is because I originally popped the `containerd_additional_settings` value in there so I could insert some containerd runtimeclasses, but I do not like the implementation, as it's very opaque and easily collides with the hard-coded way of adding in the pause image.

Again, not trying to put any sort of hold or block on this PR, just curious on other opinions on this possible slippery slope.

MaxRink:
@voor im with you on this. :D
I dont have a nice way to handle this in kind tho, as the containerd config is complex (https://github.com/containerd/containerd/blob/master/docs/cri/config.md )

moshloop:
Why not just let users specify a config file of their own?

PR #475, opened 2021-01-14
use latest so anyone building always gets the latest version

jsturtevant:
Sets the image to default to the "latest" image. This will always use the latest image published by Azure which will have the latest windows patches.  We need some more fine to control but this can be done by overriding this value in our Pipelines instead of the default value which will all for less churn on this file and ensure anyone using the project they get the latest when building at a given point in time.  

**Additional context**

/assign @CecileRobertMichon


CecileRobertMichon (approved):
/lgtm
/approve

PR #477, opened 2021-01-15
Azure: change creation timestamp tag to RFC 3339 format

CecileRobertMichon:
What this PR does / why we need it: Change the rg tag format to be compatible with kubetest cleanup jobs.

**Additional context**

### DIFF
--- a/images/capi/packer/azure/packer-windows.json
+++ b/images/capi/packer/azure/packer-windows.json
@@ -54,8 +54,9 @@
       "winrm_timeout": "10m",
       "winrm_username": "packer",
       "azure_tags": {
+        "build_date": "{{isotime}}",
         "build_timestamp": "{{user `build_timestamp`}}",
-        "creationTimestamp": "{{user `build_timestamp`}}",
+        "creationTimestamp": "{{isotime \"2006-01-02T15:04:05Z\"}}",
         "os_version": "{{user `image_sku`}}",
         "kubernetes_version": "{{user `kubernetes_semver`}}"
       }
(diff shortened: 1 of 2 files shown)


images/capi/packer/azure/packer-windows.json
```diff
@@ -89,8 +90,9 @@
       "winrm_timeout": "10m",
       "winrm_username": "packer",
       "azure_tags": {
+        "build_date": "{{isotime}}",
         "build_timestamp": "{{user `build_timestamp`}}",
-        "creationTimestamp": "{{user `build_timestamp`}}",
+        "creationTimestamp": "{{isotime \"2006-01-02T15:04:05Z\"}}",
```
jsturtevant, line 95: Interesting timestamp didn't work as it says its RFC3339 format. https://www.packer.io/docs/from-1.5/functions/datetime/timestamp

jsturtevant:
is this still used then?

 https://github.com/kubernetes-sigs/image-builder/blob/24db862cd5820218dc091aaa0c02c959517f8043/images/capi/packer/azure/scripts/ci-azure-e2e.sh#L52

images/capi/packer/azure/packer-windows.json
CecileRobertMichon: that's what I thought as well but it's actually Unix https://www.packer.io/docs/templates/engine#functions 

```
timestamp - The Unix timestamp in UTC when the Packer process was launched. Please note that if you are using a large number of builders, provisioners or post-processors, the timestamp may be slightly different for each one because it is from when the plugin is launched not the initial Packer process. In order to avoid this and make the timestamp consistent across all plugins, set it as a user variable and then access the user variable within your plugins.
```

CecileRobertMichon:
> is this still used then?

yes, that one is used for the RGs we create in E2E (for the storage account or Shared Image Gallery)

PR #479, opened 2021-01-22
fix location of cloudbuild.yaml

codenrhoden:

The file `cloudbuild.yaml`, added in #450, was accidentally located within `images/capi` rather than at the root of the repo. Can't believe I missed that!

FYI, @SanikaGawhane 
/assign @kkeshavamurthy 

**Additional context**


SanikaGawhane:
@codenrhoden Thanks for this fix!!

PR #480, opened 2021-01-22
Fix make target for release-staging

SanikaGawhane:
Fix for make target that changed because of change in `cloudbuild.yaml` location.

/assign @codenrhoden 
/assign @kkeshavamurthy

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -11,7 +11,7 @@ steps:
       - TAG=$_GIT_TAG
       - DOCKER_BUILDKIT=1
     args:
-      - release-staging
+      - -C images/capi release-staging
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -11,7 +11,7 @@ steps:
       - TAG=$_GIT_TAG
       - DOCKER_BUILDKIT=1
     args:
-      - release-staging
+      - -C images/capi release-staging
```
codenrhoden, line 14: I didn't even think about putting it here. This works!

SanikaGawhane: Thought this was a better place for the fix. Glad this works!

PR #482, opened 2021-01-26
Add EFI Support for OVA creation

z4ce:
This PR adds support for building OVAs that boot using EFI. We need EFI support for using DirectPathIO devices with more than 16GB of mapped memory. This is necessary for GPU devices.

I do wonder if EFI should be made the default, as I don't know any benefit to maintaining non-EFI images separately.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -207,7 +207,7 @@ CENTOS_VERSIONS			:=	centos-7
 FLATCAR_VERSIONS		:=	flatcar
 PHOTON_VERSIONS			:=	photon-3
 RHEL_VERSIONS			:=	rhel-7
-UBUNTU_VERSIONS			:=	ubuntu-1804 ubuntu-2004
+UBUNTU_VERSIONS			:=	ubuntu-1804 ubuntu-2004 ubuntu-2004-efi
 WINDOWS_VERSIONS		:=	windows-2019 windows-2004
 
 # Set Flatcar Container Linux channel and version if not supplied
@@ -302,6 +302,7 @@ $(NODE_OVA_ESX_BUILD_TARGETS): deps-ova
 .PHONY: $(NODE_OVA_VSPHERE_BUILD_TARGETS)
 $(NODE_OVA_VSPHERE_BUILD_TARGETS): deps-ova
 	packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS))  -var-file="packer/ova/packer-common.json" -var-file="$(abspath packer/ova/$(subst build-node-ova-vsphere-,,$@).json)" -var-file="packer/ova/vsphere.json" -except=esx -except=local -only=vsphere-iso $(ABSOLUTE_PACKER_VAR_FILES) -only=vsphere packer/ova/packer-$(if $(findstring windows,$@),windows,node).json
+
 
 .PHONY: $(NODE_OVA_VSPHERE_BASE_BUILD_TARGETS)
 $(NODE_OVA_VSPHERE_BASE_BUILD_TARGETS): deps-ova
@@ -487,7 +488,9 @@ build-node-ova-vsphere-ubuntu-1804: ## Builds Ubuntu 18.04 Node OVA and template
 build-node-ova-vsphere-ubuntu-2004: ## Builds Ubuntu 20.04 Node OVA and template on vSphere
 build-node-ova-vsphere-windows-2019: ## Builds for Windows Server 2019 and template on vSphere
 build-node-ova-vsphere-windows-2004: ## Builds for Windows Server 2004 SAC and template on vSphere
+build-node-ova-vsphere-ubuntu-2004-efi: ## Builds Ubuntu 20.04 Node OVA and template on vSphere that EFI boots
 build-node-ova-vsphere-all: $(NODE_OVA_VSPHERE_BUILD_TARGETS) ## Builds all Node OVAs and templates on vSphere
+
 build-haproxy-ova-vsphere-photon-3: ## Builds Photon 3 HAProxy OVA and template on vSphere
 
 build-node-ova-vsphere-clone-centos-7: ## Builds CentOS 7 Node OVA and template on vSphere
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -140,7 +140,8 @@ def main():
         'DISTRO_NAME': build_data['distro_name'],
         'DISTRO_VERSION': build_data['distro_version'],
         'DISTRO_ARCH': build_data['distro_arch'],
-        'NESTEDHV': "false"
+        'NESTEDHV': "false",
+        'FIRMWARE': build_data['firmware']
     }
 
     capv_url = "https://github.com/kubernetes-sigs/cluster-api-provider-vsphere"
--- a/images/capi/hack/ovf_template.xml
+++ b/images/capi/hack/ovf_template.xml
@@ -127,7 +127,7 @@
       <vmw:Config ovf:required="false" vmw:key="cpuHotAddEnabled" vmw:value="false"/>
       <vmw:Config ovf:required="false" vmw:key="cpuHotRemoveEnabled" vmw:value="false"/>
       <vmw:Config ovf:required="false" vmw:key="memoryHotAddEnabled" vmw:value="false"/>
-      <vmw:Config ovf:required="false" vmw:key="firmware" vmw:value="bios"/>
+      <vmw:Config ovf:required="false" vmw:key="firmware" vmw:value="${FIRMWARE}"/>
       <vmw:Config ovf:required="false" vmw:key="tools.syncTimeWithHost" vmw:value="false"/>
       <vmw:Config ovf:required="false" vmw:key="tools.afterPowerOn" vmw:value="true"/>
       <vmw:Config ovf:required="false" vmw:key="tools.afterResume" vmw:value="true"/>
--- a/images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
+++ b/images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
@@ -1,0 +1,128 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configure the locale
+d-i debian-installer/locale string en_US.utf8
+d-i console-setup/ask_detect boolean false
+d-i console-setup/layout string us
+
+# Configure the clock
+d-i time/zone string UTC
+d-i clock-setup/utc-auto boolean true
+d-i clock-setup/utc boolean true
+
+# Configure the keyboard
+d-i kbd-chooser/method select American English
+
+# Configure networking
+d-i netcfg/wireless_wep string
+
+# Select the kernel
+d-i base-installer/kernel/override-image string linux-virtual
+
+# Configure a non-interactive install
+debconf debconf/frontend select Noninteractive
+
+# Configure the base installation
+d-i pkgsel/install-language-support boolean false
+d-i pkgsel/language-packs multiselect
+tasksel tasksel/first multiselect # standard, ubuntu-server
+
+
+### Simple GPT configuration w/o LVM
+d-i partman-auto/disk string /dev/sda
+
+d-i partman/alignment string cylinder
+d-i partman/confirm_write_new_label boolean true
+d-i partman-basicfilesystems/choose_label string gpt
+d-i partman-basicfilesystems/default_label string gpt
+d-i partman-partitioning/choose_label string gpt
+d-i partman-partitioning/default_label string gpt
+d-i partman/choose_label string gpt
+d-i partman/default_label string gpt
+
+d-i partman-auto/method string regular
+d-i partman-auto/choose_recipe select gpt-boot-root-swap
+d-i partman-auto/expert_recipe string         \
+   gpt-boot-root-swap ::                      \
+      1 1 1 free                              \
+         $bios_boot{ }                        \
+         method{ biosgrub } .                 \
+      200 200 200 fat32                       \
+         $primary{ }                          \
+         method{ efi } format{ } .            \
+      512 512 512 ext3                        \
+         $primary{ } $bootable{ }             \
+         method{ format } format{ }           \
+         use_filesystem{ } filesystem{ ext3 } \
+         mountpoint{ /boot } .                \
+      1000 20000 -1 ext4                      \
+         $primary{ }                          \
+         method{ format } format{ }           \
+         use_filesystem{ } filesystem{ ext4 } \
+         mountpoint{ / } .
+
+d-i partman-partitioning/confirm_write_new_label boolean true
+d-i partman/choose_partition select finish
+d-i partman/confirm boolean true
+d-i partman/confirm_nooverwrite boolean true
+
+# Create the default user.
+d-i passwd/user-fullname string builder
+d-i passwd/username string builder
+d-i passwd/user-password password builder
+d-i passwd/user-password-again password builder
+d-i user-setup/encrypt-home boolean false
+d-i user-setup/allow-password-weak boolean true
+
+# Disable upgrading packages upon installation.
+d-i pkgsel/upgrade select none
+d-i grub-installer/only_debian boolean true
+d-i grub-installer/with_other_os boolean true
+d-i finish-install/reboot_in_progress note
+d-i pkgsel/update-policy select none
+
+# Disable use of the apt mirror during base install
+# This means only packages available in the ISO can be installed
+d-i apt-setup/use_mirror boolean false
+
+# Disable the security repo as well (it's on by default)
+d-i apt-setup/services-select multiselect none
+
+# Customize the list of packages installed.
+d-i pkgsel/include string openssh-server open-vm-tools
+
+
+# Ensure questions about these packages do not bother the installer.
+libssl1.1 libssl1.1/restart-without-asking boolean true
+libssl1.1:amd64  libssl1.1/restart-without-asking boolean true
+libssl1.1 libssl1.1/restart-services string
+libssl1.1:amd64 libssl1.1/restart-services string
+
+
+# This command runs after all other steps; it:
+# 1. Ensures the "builder" user doesn't require a password to use sudo
+# 2. Cleans up any packages that are no longer required
+# 3. Cleans the package cache
+# 4. Removes the cached list of packages
+# 5. Disables swapfiles
+# 6. Removes the existing swapfile
+# 7. Removes the swapfile entry from /etc/fstab
+d-i preseed/late_command string \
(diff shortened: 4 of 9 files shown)


z4ce:
CLA Verified

images/capi/Makefile
```diff
@@ -198,7 +198,7 @@ CENTOS_VERSIONS			:=	centos-7
 FLATCAR_VERSIONS		:=	flatcar
 PHOTON_VERSIONS			:=	photon-3
 RHEL_VERSIONS			:=	rhel-7
-UBUNTU_VERSIONS			:=	ubuntu-1804 ubuntu-2004
+UBUNTU_VERSIONS			:=	ubuntu-1804 ubuntu-2004 ubuntu-2004-efi
```
kkeshavamurthy, line 201: Instead of adding a new target, can `firmware` var be passed through packer vars?

z4ce: It also needs different preseed file and boot command.. so I don't think so

z4ce:
@codenrhoden wondering if you got a chance to look at this?

codenrhoden:
> @codenrhoden wondering if you got a chance to look at this?

I haven't yet, but it is certainly on my radar. Thank you for pointing it out.

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
```diff
@@ -0,0 +1,165 @@
...
+# d-i partman-md/confirm boolean true
+# d-i partman-partitioning/confirm_write_new_label boolean true
+# d-i partman/choose_partition select finish
+# d-i partman/confirm boolean true
+# d-i partman/confirm_nooverwrite boolean true
+# d-i partman-md/confirm_nooverwrite boolean true
+# d-i partman-lvm/confirm_nooverwrite boolean true
+# d-i partman-partitioning/no_bootable_gpt_biosgrub boolean true
+# d-i partman-partitioning/no_bootable_gpt_efi boolean false
+# d-i partman-efi/non_efi_system boolean false
```
kkeshavamurthy, line 76: Can these be removed?

images/capi/packer/ova/ubuntu-2004-efi.json
```diff
@@ -0,0 +1,15 @@
...
+  "distro_name": "ubuntu",
+  "os_display_name": "Ubuntu 20.04",
+  "guest_os_type": "ubuntu-64",
+  "vsphere_guest_os_type": "ubuntu64Guest",
+  "iso_url": "http://cdimage.ubuntu.com/ubuntu-legacy-server/releases/20.04/release/ubuntu-20.04.1-legacy-server-amd64.iso",
+  "iso_checksum": "f11bda2f2caed8f420802b59f382c25160b114ccc665dbac9c5046e7fceaced2",
+  "cdrom_type": "sata",
+  "iso_checksum_type": "sha256",
+  "boot_command_prefix": "<esc><wait><esc><wait><enter><wait>linux /install/vmlinuz auto console-setup/ask_detect=false console-setup/layoutcode=us console-setup/modelcode=pc105 debconf/frontend=noninteractive debian-installer=en_US fb=false kbd-chooser/method=us keyboard-configuration/layout=USA keyboard-configuration/variant=USA locale=en_US netcfg/get_domain=local netcfg/get_hostname=localhost preseed/url=",
+  "boot_command_suffix": "/20.04/preseed-efi.cfg -- <wait><enter>initrd /install/initrd.gz<enter>boot<enter><wait>",
```
kkeshavamurthy, line 12: Since non-efi Ubuntu-X building uses floppy to provide preseed file, can we keep it consistent here please?

images/capi/packer/ova/packer-node.json
```diff
@@ -402,23 +404,23 @@
         "cd {{user `output_dir`}}",
         "tar xf {{user `build_version`}}.ova",
         "rm {{user `build_version`}}.ova",
-        "../../hack/image-build-ova.py --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk1.vmdk"
+        "../../hack/image-build-ova.py --firmware={{user `efi`}}  --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk1.vmdk"
```
kkeshavamurthy, line 407: should `--firmware={{user `efi`}}` be `--firmware={{user `firmware`}}` ?

images/capi/packer/ova/packer-node.json
```diff
@@ -402,23 +404,23 @@
...
       ]
     },
     {
       "name": "vsphere",
       "type": "shell-local",
       "inline": [
         "cd {{user `output_dir`}}",
-        "../../hack/image-build-ova.py --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk-0.vmdk"
+        "../../hack/image-build-ova.py --firmware={{user `efi`}} --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk-0.vmdk"
```
kkeshavamurthy, line 415: same as above

images/capi/packer/ova/packer-node.json
```diff
@@ -402,23 +404,23 @@
...
       ]
     },
     {
       "name": "local",
       "type": "shell-local",
       "except": ["vmware-iso-base"],
       "inline": [
-        "./hack/image-build-ova.py --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ./hack/ovf_eula.txt --ovf_template ./hack/ovf_template.xml {{user `output_dir`}}",
+        "./hack/image-build-ova.py --firmware={{user `efi`}} --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ./hack/ovf_eula.txt --ovf_template ./hack/ovf_template.xml {{user `output_dir`}}",
```
kkeshavamurthy, line 423: same as above

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
z4ce: Yeah they can be

images/capi/packer/ova/ubuntu-2004-efi.json
z4ce: I'm not sure what you mean.. I didn't change the way the preseed file was delivered... did a more recent commit change the preseed from http to floppy?

images/capi/packer/ova/packer-node.json
z4ce: I'll look into these.. I probably forgot to commit this file or something when I changed the variable

images/capi/packer/ova/ubuntu-2004-efi.json
kkeshavamurthy: Yes it was changed recently in this [PR](https://github.com/kubernetes-sigs/image-builder/pull/497)

images/capi/hack/image-build-ova.py
```diff
@@ -40,6 +40,7 @@ def main():
     image_type = parser.add_mutually_exclusive_group(required=True)
     image_type.add_argument('--node', action='store_true')
     image_type.add_argument('--haproxy', action='store_true')
+    parser.add_argument('--firmware', default='bios', help="Set to bios or efi. Defaults to bios")
```
kkeshavamurthy, line 43: Since `firmware` var is added in `packer-common`, we can just stick it in the mainfest and access it similar to other vars `FIRMWARE: build_data['firmware']` ? Then there will be no need of a new python arg & no changes to how `image-builder-ova.py` is called.

z4ce: I'll try that. Should work.

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -171,13 +173,15 @@
       "inline": [
         "cd {{user `output_dir`}}",
         "../../hack/image-build-ova.py --haproxy --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk-0.vmdk"
+
```
kkeshavamurthy, line 176: nit: can we please remove the extra lines.

images/capi/packer/ova/packer-node.json
```diff
@@ -406,7 +409,7 @@
         "cd {{user `output_dir`}}",
         "tar xf {{user `build_version`}}.ova",
         "rm {{user `build_version`}}.ova",
-        "../../hack/image-build-ova.py --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk1.vmdk"
+        "../../hack/image-build-ova.py  --stream_vmdk --node --vmx {{user `vmx_version`}} --eula ../../hack/ovf_eula.txt --ovf_template ../../hack/ovf_template.xml --vmdk_file {{user `build_version`}}-disk1.vmdk"
```
kkeshavamurthy, line 412: nit: remove extra space.

images/capi/hack/image-build-ova.py
```diff
@@ -231,6 +232,8 @@ def create_ovf(path, data, ovf_template):
     print("image-build-ova: create ovf %s" % path)
     with io.open(path, 'w', encoding='utf-8') as f:
       f.write(Template(ovf_template).substitute(data))
+      f.flush()
+      os.fsync(f.fileno())
```
kkeshavamurthy, line 236: This was working fine without the flush. Any reason for adding them?

z4ce: For some reason I keep getting checksum failure on the ova.. I thought it was a race but the flush and sync didn't fix it.. I'm just now getting back to this PR to figure out what is actually going on

kkeshavamurthy:
Hi @z4ce, did you get a chance to look at the issues you were having?

kkeshavamurthy:
@z4ce #580 should fix the OVA creation issues you were seeing. Can you please rebase and confirm?

CC: @codenrhoden

perithompson:
There will be some windows parts that we need to add Windows EFI support separately? Should i create a separate issue to track that?

kkeshavamurthy:
@perithompson Yes that would be best.

images/capi/hack/ovf_template.xml
```diff
@@ -127,7 +127,7 @@
       <vmw:Config ovf:required="false" vmw:key="cpuHotAddEnabled" vmw:value="false"/>
       <vmw:Config ovf:required="false" vmw:key="cpuHotRemoveEnabled" vmw:value="false"/>
       <vmw:Config ovf:required="false" vmw:key="memoryHotAddEnabled" vmw:value="false"/>
-      <vmw:Config ovf:required="false" vmw:key="firmware" vmw:value="bios"/>
+      <vmw:Config ovf:required="false" vmw:key="firmware" vmw:value="${FIRMWARE}"/>
```
kkeshavamurthy, line 130: Also does 
```<vmw:Config ovf:required="false" vmw:key="bootOptions.efiSecureBootEnabled" vmw:value="false"/>```
need to be set to `true`?

File: images/capi/Makefile (hunk 1 above)
Comment (codenrhoden) [reply images/capi/Makefile:203]: Agreed that this would be nice, but it does seem like we are stuck with this for now. I take this as another data point in favor of creating a pre-processor that generates these files (a CLI).

File: images/capi/hack/image-build-ova.py (hunk 10 above)
Comment (codenrhoden) [reply images/capi/hack/image-build-ova.py:236]: This can go away now, checksum issue has been fixed and was unrelated.

File: images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+         $primary{ } $bootable{ }             \
+         method{ format } format{ }           \
+         use_filesystem{ } filesystem{ ext3 } \
+         mountpoint{ /boot } .                \
+      1000 20000 -1 ext4                      \
+         $primary{ }                          \
+         method{ format } format{ }           \
+         use_filesystem{ } filesystem{ ext4 } \
+         mountpoint{ / } .                    \
+      65536 65536 65536 linux-swap            \
```
codenrhoden, line 75: I think having this swap partition is my main issue with the PR right now. We make sure all the other images *don't* have any swap space since K8s doesn't want that in the nodes. I wonder if as a first attempt we can just remove the swap partition and make the root partition consume the rest of the size?

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
```diff
@@ -0,0 +1,131 @@
...
+# 2. Cleans up any packages that are no longer required
+# 3. Cleans the package cache
+# 4. Removes the cached list of packages
+# 5. Disables swapfiles
+# 6. Removes the existing swapfile
+# 7. Removes the swapfile entry from /etc/fstab
+d-i preseed/late_command string \
+    echo 'builder ALL=(ALL) NOPASSWD: ALL' >/target/etc/sudoers.d/builder ; \
+    in-target chmod 440 /etc/sudoers.d/builder ; \
+    in-target swapoff -a ; \
```
codenrhoden, line 128: FYI, here is a place where swap is removed, so partitioning for is probably wasteful.

codenrhoden (changes requested):
Hi @z4ce, thanks so much for putting this together. I think it's a capability that is becoming important. Do you think you'll have the bandwidth to address review comments and iterate on this feature?

I agree with you that it may make sense to make EFI be the default, but let's work up to that one over time. Addressing one of @kkeshavamurthy's comments, I do not think Secure Boot needs to be enabled by default with this change.

z4ce:
I've been meaning to get back to this PR, I just rebased it.. and I'll see if I can find a cluster tomorrow to give it a spin and make sure the rebase was successful. There were some files that were pretty heavily reformatted so hopefully I didn't miss a `{{user firmware}}` somewhere

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
z4ce: I'm fine with removing the swap.. honestly I just followed the partition pattern from somewhere else.. it's not necessary for the EFI support

images/capi/hack/ovf_template.xml
z4ce: No, at least not for my purposes.. maybe windows would need it?

z4ce:
@codenrhoden @kkeshavamurthy I verified this successfully built an image and no checksum errors 🎉

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
```diff
@@ -0,0 +1,128 @@
...
+    in-target rm -f /swapfile ; \
+    in-target sed -ri '/\sswap\s/s/^#?/#/' /etc/fstab ; \
+    in-target rm -f /etc/udev/rules.d/70-persistent-net.rules
```
kkeshavamurthy, line 128:
```    
in-target apt-get purge --auto-remove -y ; \
in-target rm -rf /var/lib/apt/lists/*
```
Does ^^ need to be added at the end, similar to ubuntu-2004?

kkeshavamurthy:
@z4ce Thanks for updating the PR. 
FYI, it'll fail json-sort CI test. So please run `make json-sort` and update :)

codenrhoden:
/test json-sort-check
@kkeshavamurthy I'll have to look at the regex for this test to figure out why it didn't run on this PR. it certainly should have!

images/capi/packer/ova/linux/ubuntu/http/base/preseed-efi.cfg
codenrhoden: FWIW, those two lines were added in the other PRESEED to help out a usecase where I was using apt-cacher-ng to speed up builds and running into weird stale data. It's not strictly required, but it would be nice if they were consistent. I need to check if we can have more than one `include` in a preseed. Last time I look I remember finding that we couldn't, but I think I had a different use case then so maybe that's not right.

images/capi/packer/ova/packer-common.json
```diff
@@ -24,6 +24,7 @@
   "vmx_version": "15",
   "vnc_bind_address": "127.0.0.1",
   "vnc_disable_password": "false",
+  "vnc_port_min": "5900",
   "vnc_port_max": "6000",
-  "vnc_port_min": "5900"
+  "firmware": "bios"
```
codenrhoden, line 29: as previously mentioned, please keep this in alphabetical order.

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -10,14 +10,15 @@
       "cores": 1,
       "cpus": 1,
       "disk_size": "{{user `disk_size`}}",
+      "memory": 2048,
```
codenrhoden, line 13: alphabetical order.

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -134,6 +136,7 @@
     {
       "inline": [
         "./hack/image-build-ova.py --stream_vmdk --haproxy --vmx {{user `vmx_version`}} --eula ./hack/ovf_eula.txt --ovf_template ./hack/ovf_template.xml {{user `output_dir`}}",
+
```
codenrhoden, line 139: no need for this extra space

z4ce:
Sorted the json

codenrhoden:
/ok-to-test
I hadn't noticed this was never given -- that's why the JSON sort check probably wasn't running...

I'm inclined to merge this at this point and iterate as needed. Thanks for the continuing to update, @z4ce

z4ce:
I had a quick look, and I don't see why this is failing. Variable is defined in `packer-common`, reference seem right in `packer-haproxy`.. some sort of order of operations thing?

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -11,6 +11,7 @@
       "cpus": 1,
       "disk_size": "{{user `disk_size`}}",
       "disk_type_id": "{{user `disk_type_id`}}",
+      "firmware": "{{user `firmware`}}",
```
kkeshavamurthy, line 14: `firmware` is not an option in `vmware-iso` builder. That's why the test is failing.

codenrhoden: dang, you beat me to it!

Yeah, looking at https://www.packer.io/docs/builders/vmware/vsphere-iso vs https://www.packer.io/docs/builders/vmware/iso, indeed `firmware` is not supported for the `vmware-iso` builder. This means we can't build EFI OVAs with the "local" builders, meaning Workstation or Fusion. Lame.

codenrhoden:
And just as a heads up, once the `validate` issue is fixed, you'll want to go ahead and squash this down to 1 commit. I understand the workflow of pushing changes 1-by-1, but once things are all wrapped up we want it as a single commit per "feature". Additionally, the first commit in this patchset contains [git merge conflict markers](https://github.com/kubernetes-sigs/image-builder/blob/0db0733096e01166225a63d7a7175281b57cc49a/images/capi/packer/ova/packer-haproxy.json#L132-L139), which we definitely don't want in the repo history.

z4ce:
I removed references to firmware from the vsphere-iso builders and rebased squashing the fix commits

kkeshavamurthy:
@z4ce , 
We need `firmware` references in `vsphere-iso` builders. 
We do not need `firmware` references in `vmware-iso` builders.

codenrhoden:
I think this gives us what we need to merge and iterate if/as needed. Thank you so much @z4ce !
/approve

PR #483, opened 2021-01-27
Only place required files in Docker container

codenrhoden:
This patch changes up the `Dockerfile` and `.dockerignore` to start with an
explicit deny all approach, and then add only the files/dirs that are
needed. This is most useful for developers, who may have additional
files/folders located withiin images/capi that are not part of the git
repo. Since the old Docker file was doing "COPY .", this copied anything
that wasn't explicitly ruled out in .dockerignore. This patch takes the
opposite approach.

**Additional context**
This may not be an issue in CI, but as an example, this reduced the final build size of the image from ~4.5G to ~2.2G. Also cut the build time down by about 4 minutes.

Another issue (actually the one that sent me down this path) is that if you are developing on a Mac, the old `COPY . /home/imagebuilder` meant that it would copy `.local/bin`, which may contain Mac-compiled packer binaries. Those then end up in the container image and then the image does not work. This was originally covered by the `.bin/` entry in `.dockerignore`, but at some point we started installing Packer to `.local/bin` instead, and the `.dockerignore` file never got updated.

/assign @kkeshavamurthy 
FYI @SanikaGawhane

### DIFF
--- a/images/capi/.dockerignore
+++ b/images/capi/.dockerignore
@@ -1,6 +1,9 @@
-# Ignore files that shouldn't go into the container
-.bin/
-output/
-packer_cache/
-manifest.json
-docker/
+# Ignore all files
+*
+
+# Exceptions
+!ansible
+!cloudinit
+!hack
+!packer
+!Makefile
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -24,18 +24,19 @@ RUN apt-get update && apt-get install -y apt-transport-https ca-certificates pyt
 ARG ARCH
 ARG PASSED_IB_VERSION
 
-COPY . /home/imagebuilder
-RUN chown -R imagebuilder:imagebuilder /home/imagebuilder
-
 USER imagebuilder
+WORKDIR /home/imagebuilder/
+
+COPY --chown=imagebuilder:imagebuilder ansible ansible/
+COPY --chown=imagebuilder:imagebuilder cloudinit cloudinit/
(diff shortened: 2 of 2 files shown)


codenrhoden:
@SanikaGawhane with this change in place, I can use the image to build AMIs with a command like the following:

```sh
docker run --env PACKER_VAR_FILES=var_files/aws.json -it --rm --mount type=bind,src=/Users/trhoden/.aws,dst=/home/imagebuilder/.aws,readonly --mount type=bind,src=/Users/trhoden/git/sigs.k8s.io/image-builder/images/capi/var_files,dst=/home/imagebuilder/var_files,readonly gcr.io/cloud-provider-vsphere/cluster-node-image-builder-amd64:dev build-ami-ubuntu-2004
```

@perithompson You may be interested in the above. The command I gave you in that other PR of yours wouldn't have quite worked.

Note that is required to pass env vars using `--env` args.  I also mount in my AWS cli credentials folder `~/.aws` so that it picks up my creds and default region.  The `vars_files` stuff is just a folder I have with a bunch of JSON files I use to configure things. So that shows an example of using `PACKER_VAR_FILES`.  The `-it` is required if you want to be able to cancel the build with Ctrl-C.

Just for full disclosure, my `aws.json` file just as this in it:

```json
{
	"aws_region": "us-west-2",
	"ami_regions": "",
	"ami_groups": "",
	"snapshot_groups": ""
}
```

This has the effect of not copying the resulting AMIs around the world and taking up unnecessary space (and cost).

PR #491, opened 2021-02-04
Re-add windows_updates_categories variable

perithompson:
It looks like in the last CAPA PR one of the variables was accidentally removed, this small PR is to re-add that variable


jsturtevant:
The long string is hard to catch differences.  Thanks for fixing this!

/lgtm

perithompson:
Thanks @jsturtevant I tried to think if there was a better way but it works, even though it's a tough one to see problems with

PR #493, opened 2021-02-04
[capi] Azure: Adds support for custom Shared Image Gallery Image version

shailesh91:
[capi] Azure: generalizes image version parameter, allowing using a custom value when the build target is MS Azure Shared Image gallery.

**Additional context**
Updates packer.json, packer-windows.json and azure-sig.json in "images/capi/packer/azure" folder


CecileRobertMichon (approved):
/lgtm
/approve

PR #496, opened 2021-02-05
raw image support

thebsdbox:

This PR adds the capability to generate raw images, predominantly used for bare-metal servers. It also uses the server release of the kernel for ubuntu 1804/2004.

**Additional context**

This is largely a copy&paste of the qemu workflows with modifications for output and some packages in the finished image. I'd like to tar.gz the raw output upon completion.. but I'm not sure where to do that in the `makefile`. 🤷


thebsdbox:
Added an additional change.. the raw image that is created is 6GB, which is fine as the requirements for the all packages/user-land etc. is 3GB. The raw image then compresses with gzip to ~2GB.

images/capi/packer/raw/qemu-ubuntu-2004.json
```diff
@@ -0,0 +1,12 @@
+{
```
detiber, line 1: Should this file be renamed `raw-ubuntu-2004.json`?

images/capi/packer/raw/packer.json
```diff
@@ -0,0 +1,133 @@
...
+      "type": "qemu",
+      "accelerator": "{{user `accelerator`}}",
+      "cpus": "{{user `cpus`}}",
+      "disk_size": "{{user `disk_size`}}",
+      "memory": "{{user `memory`}}",
+      "boot_wait": "{{user `boot_wait`}}",
+      "disk_interface": "virtio-scsi",
+      "format": "{{user `format`}}",
+      "headless": "{{user `headless`}}",
+      "http_directory": "./packer/qemu/linux/{{user `distro_name`}}/http/",
```
detiber, line 84:
```suggestion
      "http_directory": "./packer/raw/linux/{{user `distro_name`}}/http/",
```

images/capi/packer/raw/packer.json
```diff
@@ -0,0 +1,133 @@
...
```
detiber, line 56:
```suggestion
      "http_directory": "./packer/raw/linux/{{user `distro_name`}}/http/",
```

images/capi/packer/raw/OWNERS
```diff
@@ -0,0 +1,8 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+  - cluster-api-openstack-maintainers
+
+reviewers:
+  - cluster-api-openstack-maintainers
+  - image-builder-openstack-reviewers
```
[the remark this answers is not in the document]
thebsdbox: Not sure what these should be changed to?

images/capi/Makefile
```diff
@@ -236,6 +242,8 @@ DO_BUILD_NAMES 			?=	do-centos-7 do-ubuntu-1804 do-ubuntu-2004
 QEMU_FLATCAR_BUILD_NAMES	?=	qemu-flatcar
 QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804 qemu-ubuntu-2004
 
+RAW_BUILD_NAMES                        ?=      raw-ubuntu-1804 raw-ubuntu-2004
+
```
kkeshavamurthy, line 246: Hi @thebsdbox , 

- variables like `format`, `disk_size` etc. 
- A slightly different preseed file. 
Other than  ^^ are there any other differences between `qemu` and `raw` image-building?

If not, would it be a better idea to parameterize them and add them as additional build targets in qemu instead of creating a whole new area of builds?

CC: @codenrhoden

thebsdbox: Possibly, but I found this makefile so confusing (i'm not too good with `make`) it seemed like the easiest and quickest way.

moshloop: Long term we want to move away from Makefile to a CLI, in which case a distinct target is easier to grok, and the benefits of DRY are reduced

images/capi/packer/raw/OWNERS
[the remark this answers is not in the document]
detiber: We'd need to add new groups in https://github.com/kubernetes-sigs/image-builder/blob/master/OWNERS_ALIASES, probably starting with the two of us

detiber:
I created a branch on my fork with these changes rebased and including the OWNER_ALIASES and OWNERS file updates here: https://github.com/detiber/image-builder/tree/pr/thebsdbox/496

PR #498, opened 2021-02-09
Added a variable to configure hard disk size

### LINKED ISSUE #428: image-build-ova.py looks to hardcode 20GB disk
I successfully built Photon OS 3 OVA using K8s v1.17.11 and the only change I made from the defaults is increasing the disk capacity from 20GB to 30GB. When I tried to upload the OVA using the vSphere UI or OVFTool, it would throw an error at the very end 95% with Error 500:

```
Opening VI target: vi://administrator%40vsphere.local@192.168.30.200:443//Primp-Datacenter/host/Supermicro-Cluster/Resources/Workload
Deploying to VI: vi://administrator%40vsphere.local@192.168.30.200:443//Primp-Datacenter/host/Supermicro-Cluster/Resources/Workload
Transfer Failed
Error: Failed to transfer file [disk-0.vmdk] to host [192.168.30.14] (response code:500)
```

It took me awhile to figure out why as I have never seen this error and upon inspecting the OVA, I noticed the disk capacity was still showing 20GB. I know the Packer file did build a 30GB VMDK and browsing through the source, it looks like *image-build-ova.py* script hard codes the value of 20GB! :(

perithompson:
Add the ability to set a variable to define the disk size of ova files.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #428 

**Additional context**
I have set defaults to the initial values 20GB for Linux and 80GB for Windows but this can be overridden using PACKER_FLAGS=-var=disk_size="40926" (size is in MB)

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -29,7 +29,6 @@ import subprocess
 from string import Template
 import tarfile
 
-
 def main():
     parser = argparse.ArgumentParser(
         description="Builds an OVA using the artifacts from a Packer build")
@@ -134,7 +133,7 @@ def main():
         'OS_VERSION': OS_id_map[build_data['guest_os_type']]['version'],
         'IB_VERSION': build_data['ib_version'],
         'DISK_NAME': vmdk['stream_name'],
-        'DISK_SIZE': "20",
+        'DISK_SIZE': build_data['disk_size'],
         'POPULATED_DISK_SIZE': vmdk['size'],
         'STREAM_DISK_SIZE': vmdk['stream_size'],
         'VMX_VERSION': args.vmx_version,
@@ -175,9 +174,7 @@ def main():
                 for k, v in custom_properties.items():
                     data['PROPERTIES'] = data['PROPERTIES'] + f'''      <Property ovf:userConfigurable="false" ovf:value="{v}" ovf:type="string" ovf:key="{k}"/>\n'''
 
-        # windows nodes use nested virtualisation and require a larger hard drive
         if "windows" in OS_id_map[build_data['guest_os_type']]['type']:
-            data['DISK_SIZE'] = "80"
             if build_data['disable_hypervisor'] != "true":
                 data['NESTEDHV'] = "true"
     elif args.haproxy:
--- a/images/capi/packer/ova/packer-haproxy.json
+++ b/images/capi/packer/ova/packer-haproxy.json
@@ -14,7 +14,8 @@
     "output_dir": "./output/{{user `build_version`}}",
     "username": "",
     "vcenter_server": "",
-    "vsphere_guest_os_type": null
+    "vsphere_guest_os_type": null,
+    "disk_size": "20480"
   },
   "builders": [
     {
@@ -27,7 +28,7 @@
       "cpus": 1,
       "cores": 1,
       "memory": 2048,
-      "disk_size": 20480,
+      "disk_size": "{{user `disk_size`}}",
       "disk_type_id": "{{user `disk_type_id`}}",
       "boot_wait": "{{user `boot_wait`}}",
       "http_directory": "./packer/ova/linux/{{user `distro_name`}}/http/",
@@ -102,7 +103,7 @@
 
       "storage": [
         {
-          "disk_size": 20480,
+          "disk_size": "{{user `disk_size`}}",
           "disk_thin_provisioned": "{{user `disk_thin_provisioned`}}"
         }
       ],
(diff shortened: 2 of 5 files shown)


perithompson:
@codenrhoden I should have probably done this with my original PR can you take a look over the changes please?

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -14,7 +14,8 @@
     "output_dir": "./output/{{user `build_version`}}",
     "username": "",
     "vcenter_server": "",
-    "vsphere_guest_os_type": null
+    "vsphere_guest_os_type": null,
+    "disk_size": "20480"
```
codenrhoden, line 18: nit: this list is in alphabetical order - can we please keep it that way?

I'm not sure we need to add this for HAProxy since it's deprecated anyways, but it is good to be consistent. 👍

codenrhoden (commented):
I guess I'm not concerned about the alphabetical ordering -- too many things have snuck in already that aren't.

/lgtm

Will wait to see if there are any more comments. @kkeshavamurthy can you take a look?

images/capi/packer/ova/packer-haproxy.json
```diff
@@ -152,7 +153,8 @@
         "ib_version": "{{user `ib_version`}}",
         "vsphere_guest_os_type": "{{user `vsphere_guest_os_type`}}",
         "dataplaneapi_version": "v{{user `dataplaneapi_version`}}",
-        "os_name": "{{user `os_display_name`}}"
+        "os_name": "{{user `os_display_name`}}",
+        "disk_size": "{{user `disk_size`}}"
```
codenrhoden, line 157: so one general problem here... Now that the disk size going to packer (e.g. `20480`) is being passed to the OVA script via the manifest, we have a mismatch of units. The Packer size of `20480` is going with 20GB. But the units in the OVF metadata are already in GB (remember the old hard coded value was `20`, so what really ends up in the OVF is 20480GB, or 20TB:

```sh
$ ovftool --hideEula ubuntu-2004-kube-v1.18.15.ovf
<snip>

Deployment Sizes:
  Flat disks:   20480.00 GB
  Sparse disks: 3.77 GB
```

So I believe we need to update the OVF template to use units of `"byte * 2^20"` instead.

codenrhoden (changes requested):

images/capi/hack/image-build-ova.py
```diff
@@ -213,6 +212,16 @@ def sha256(path):
...
+# Convert byte value disk size passed to Packer into GB for ovftool
+def convert_disk_size(size_bytes_string):
+   if size_bytes_string == "0":
+       return "0"
+   size_bytes = float(size_bytes_string)
+   i = int(math.floor(math.log(size_bytes, 1024)))
+   p = math.pow(1024, 1)
+   s = round(size_bytes / p, 1)
+   return "%s" % int(s)
+
```
perithompson, line 224: `'DISK_SIZE': convert_disk_size(build_data['disk_size']),`

@codenrhoden I've added this to convert the value back to the the size in GB for the ovf

codenrhoden: This seems like a lot of work vs just changing `2^30` to `2^20` in the OVF template like I suggested. The number we're passing in has increased by a factor of 1024 (2^10), so it makes since to me to just change the template instead... A lot less work. :)

perithompson: Sorry totally missed what you meant 😄

perithompson:
sorry i messed up my rebase

images/capi/hack/image-build-ova.py
```diff
@@ -28,6 +28,7 @@
 import subprocess
 from string import Template
 import tarfile
+import math
```
codenrhoden, line 31: can we remove this import that is no longer needed?

perithompson: good spot!

images/capi/packer/ova/packer-haproxy.json
perithompson: ive sorted these alphabetically now

perithompson: ive sorted these alphabetically now 👍

images/capi/packer/ova/packer-windows.json
```diff
@@ -40,7 +40,7 @@
       "version": "{{user `vmx_version`}}",
       "cpus": 2,
       "memory": 4096,
-      "disk_size": 81920,
+      "disk_size": "{{user `disk_size`}}",
```
kkeshavamurthy, line 43: Hi Peri, Should 80G default for windows be set in `windows-X.json` or `packer-windows.json`? I do not see that in the vars list

perithompson: Yep you are right, somewhere along the refactoring I must have lost that, thanks Kiran I'll get that added!

perithompson: It should be there now!

codenrhoden:
@kkeshavamurthy care to take one more look at this before I approve?

codenrhoden:
/approve

Thanks @perithompson !

PR #499, opened 2021-02-09
Add image builder version to Azure metadata

codenrhoden:
After creating the VM image, add a tag that has the version of the image
builder that was used to create the image. This was already present for
AMIs and OVAs, this patch adds it for Azure as well.

**Additional context**

This also fixes an error for Windows node OVAs where the version would have been blank.

/assign @CecileRobertMichon

### DIFF
--- a/images/capi/packer/azure/packer-windows.json
+++ b/images/capi/packer/azure/packer-windows.json
@@ -17,6 +17,7 @@
     "cloudbase_plugins_unattend": "cloudbaseinit.plugins.common.mtu.MTUPlugin",
     "containerd_version": null,
     "containerd_url": "",
+    "ib_version": "{{env `IB_VERSION`}}",
     "image_offer": null,
     "image_publisher": null,
     "image_sku": null,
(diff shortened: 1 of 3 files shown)


images/capi/packer/azure/packer-windows.json
```diff
@@ -17,6 +17,7 @@
     "cloudbase_plugins_unattend": "cloudbaseinit.plugins.common.mtu.MTUPlugin",
     "containerd_version": null,
     "containerd_url": "",
+    "ib_version": "{{env `IB_VERSION`}}",
```
CecileRobertMichon, line 20: what sets `IB_VERSION`?

codenrhoden: The Makefile

https://github.com/kubernetes-sigs/image-builder/blob/0717551481a2720c11119ac1723bdc1da275b349/images/capi/Makefile#L26

CecileRobertMichon (commented):
/lgtm

PR #501, opened 2021-02-09
Always install kubelet env file 

### LINKED ISSUE #486: pod-infra-container-image missing from kubelet configuration (non-windows)
On the windows side, we inject `pod-infra-container-image` when the kubelet starts:
https://cs.k8s.io/?q=pod-infra-container-image&i=nope&files=&repos=kubernetes-sigs/image-builder

The same parameter must be specified in the 10-kubeadm.conf as well:
https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/kubernetes/templates/usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf#L3-L4

Note that you will see a warning from kubelet here:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L191

However, this parameter is used in the kubelet GC Manager to prevent eviction of the pause image:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/image_gc_manager.go#L212-L216

The warning just indicates that if you have an alternate runtime other than docker, you will need to specify it in that runtime's configuration and that kubelet will not be able to do that.

We will be able to avoid issues similar to that logged in https://github.com/kubernetes/kubernetes/issues/81756

kkeshavamurthy:
Always install kubelet env file with KUBELET_EXTRA_ARGS set to exclude pause image from gc

Fixes #486 

**Additional context**

/assign @codenrhoden


kkeshavamurthy:
I've built  photon-3 & ubuntu-2004 images and verify the file. 
Have not verified by spinning up a cluster yet. Hence WIP

kkeshavamurthy:
Created `photon` & `ubuntu-2004` clusters and see that `--pod-infra-container-image=k8s.gcr.io/pause:3.2` is added to the kubelet startup args.

images/capi/ansible/roles/kubernetes/files/etc/sysconfig/kubelet
```diff
@@ -1 +0,0 @@
-KUBELET_EXTRA_ARGS=
```
codenrhoden, line 1: Rather than deleting this file, I would have preferred a `git mv files/etc/sysconfig/kubelet templates/etc/sysconfig/kubelet`.

images/capi/ansible/roles/kubernetes/tasks/debian.yml
```diff
@@ -33,3 +33,11 @@
       - "kubeadm={{ kubernetes_deb_version }}"
       - "kubectl={{ kubernetes_deb_version }}"
       - "kubernetes-cni={{kubernetes_cni_deb_version }}"
+
+- name: Create kubelet default config file
```
codenrhoden, line 37: I think it is safe to put this section at the end of `kubernetes/tasks/main.yml` and only have it in 1 place. I don't see any differences between the debian, photon, redhat, or url versions of the task, so no need to duplicate it everywhere? Am I missing something there?

kkeshavamurthy: In debian, the path is `/etc/default`. Others are at `/etc/sysconfig`

codenrhoden: Hmm, You would think I would have noticed that since I was worried about that very thing in https://github.com/kubernetes-sigs/image-builder/issues/486#issuecomment-775529851. 😂 

however, this makes me thing/realize that the code in the url tasks (I'll tag it below) is invalid because it's always putting the file in `sysconfig`. I think maybe we still just put this in `main.yml` and have some logic that puts in the right place for debian vs RPM.

images/capi/ansible/roles/kubernetes/tasks/url.yml
```diff
@@ -104,8 +104,8 @@
     group: root
 
 - name: Create kubelet sysconfig file
-  copy:
-    src: files/etc/sysconfig/kubelet
+  template:
+    src: etc/sysconfig/kubelet
     dest: /etc/sysconfig/kubelet
```
codenrhoden, line 109: I know think this code is problematic for debian. It wouldn't put it in `/etc/default`. It's never been a problem before since the file was empty, but now it wouldn't be...

images/capi/ansible/roles/kubernetes/tasks/debian.yml
codenrhoden: I think you could do something like `kubelet_default_config: "{{ '/etc/default/kubelet' if ansible_os_family == 'Debian' else '/etc/sysconfig/default'}}"
` in `kubernets/defaults/main.yml` and just it with this block.

Would solve the duplication problem, and fix the URL task.

codenrhoden (commented):
/lgtm

I took another look at this focusing on whether this would also work for Flatcar, and i believe it will. Flatcar always uses the "http" path for installing, and there were no special switches in there related to Flatcar, and this set of changes accomplishes the exact same thing, but in fewer steps. So I think we are good!

Just going to wait a bit to see if any other comments come in.

One request if you have time -- can you check (and paste here) the output from the kubelet log to verify that the change is picked up?

kkeshavamurthy:
Here are the logs for photon
```
root@capi-quickstart-jj9zb:/home/capv# systemctl status kubelet.service -f
● kubelet.service - kubelet: The Kubernetes Node Agent
     Loaded: loaded (/lib/systemd/system/kubelet.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/kubelet.service.d
             └─10-kubeadm.conf
     Active: active (running) since Wed 2021-02-10 01:09:30 UTC; 1min 26s ago
       Docs: https://kubernetes.io/docs/home/
   Main PID: 827 (kubelet)
      Tasks: 17 (limit: 9521)
     Memory: 37.4M
     CGroup: /system.slice/kubelet.service
             └─827 /usr/bin/kubelet --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf --config=/var/lib/kubelet/config.yaml --cloud-provider=external --container-runtime=remote --container-runtime-endpoint=/var/run/containerd/containerd.sock --resolv-conf=/run/systemd/resolve/resolv.conf --pod-infra-container-image=k8s.gcr.io/pause:3.2
```

PR #495, opened 2021-02-05
Add documentation for container images

SanikaGawhane:
Adds a section in Image-Builder book on how to build and use container images to build your own custom images.
This tutorial lists steps for building AMIs, OVAs and Azure Images.

/cc @codenrhoden 
/cc @kkeshavamurthy

PR #509, opened 2021-02-17
Use OVFTool to create OVAs

codenrhoden:
This patch is a major change for users who are building OVAs. With
recent versions of vSphere and Fusion (and likely Workstation, but that
was untested), the OVAs created by image-builder were no longer
importable from local disk (but oddly the worked if referenced as a
URL). The root cause of this remains a bit of a mystery -- the problem
lies with the OVA file, not the OVF. A problematic OVA could be
untarred, then the OVF used as a source for import and the import was
successful.

It has long been discussed and disired that ovftool should be used to
create the OVA instead of using `tarfile` from Python. These recent
errors are reason enough to make the switch. Using ovftool has the
advantage of validating the OVF file itself as well, such that if there
are any issues with the file then the OVA will not get created.

Installing ovftool requires accepting a EULA and must be installed by a
user on their system. The `deps-ova` target has been updated to reflect
this, and directs users where to download from.

**Additional context**
Once the container docs (#495), we will need to update them clarifying that OVA builds can only happen from Linux hosts, and show how to map in the `ovftool` binary since we can't distribute the tool in the container image.

/assign @kkeshavamurthy 
/cc @perithompson

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -87,6 +87,7 @@ deps-ova:
 	hack/ensure-ansible-windows.sh
 	hack/ensure-packer.sh
 	hack/ensure-goss.sh
+	hack/ensure-ovftool.sh
 
 .PHONY: deps-qemu
 deps-qemu: ## Installs/checks dependencies for QEMU builds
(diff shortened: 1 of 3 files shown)


perithompson:
Thanks for the heads up, on Windows we can use WSL maybe to get around the Linux requirement for people without a windows laptop but does this mean that building on Mac OS is no longer an option?

codenrhoden:
> Thanks for the heads up, on Windows we can use WSL maybe to get around the Linux requirement for people without a windows laptop but does this mean that building on Mac OS is no longer an option?

I build on Mac. OVFTool is available as a .dmg

PR #510, opened 2021-02-17
No longer install cloud-initramfs-growroot Ubuntu

### LINKED ISSUE #464: After growroot new build ova ubuntu 20.04, hangs in busybox.
We are build our images based on the newest commit of the head branch using vsphere.

The build went through and afterwards cluster-api clones the image. Cluster-api powers on the machine the grow happens than it hangs on following screen: 
image

If restart the image everything works fine afterward. the machine has a correct formated disk and the kubeadm join works fine.

Any idea?

codenrhoden:
This package is not needed with kernels > 3.18, and cloud-init will
handle growing the root partition and filesystem. For Ubuntu 20.04
specifically, it was observed that recent images would hang on bootup
when the initramfs tried to grow the root partition. Leaving it to
cloud-init proceeds as expected.

Fixes #464 

**Additional context**
This was a bit weird to track down. I was able to recreate it locally with images built with Fusion, and saw the same behavior as described in #464. There is very little info out there about other people running into similar problems, but I did eventually stumble on https://bugs.launchpad.net/cloud-images/+bug/1678669/comments/1, where Scott (who maintains a lot of cloud-init related items) mentioned that the referenced package is not even needed for kernels > 3.18. Ubuntu 18.04 is currently installing a 4.15 kernel, and 20.04 is 5+, so I tried this out with both distros and they worked fine!

/assign @kkeshavamurthy 
/cc  @farodin91 @darc1

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/vmware.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware.yml
@@ -23,7 +23,6 @@
     - cloud-guest-utils
     - cloud-initramfs-copymods
     - cloud-initramfs-dyn-netconf
-    - cloud-initramfs-growroot
   when: ansible_os_family == "Debian"
 
 - name: Install cloud-init packages
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -159,7 +159,6 @@ ubuntu:
       cloud-guest-utils:
       cloud-initramfs-copymods:
       cloud-initramfs-dyn-netconf:
-      cloud-initramfs-growroot:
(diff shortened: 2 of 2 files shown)


images/capi/ansible/roles/providers/tasks/vmware.yml
```diff
@@ -23,7 +23,6 @@
     - cloud-guest-utils
     - cloud-initramfs-copymods
     - cloud-initramfs-dyn-netconf
-    - cloud-initramfs-growroot
```
kkeshavamurthy, line 26: Do we need to remove `cloud-utils-growpart` in Redhat?

PR #512, opened 2021-02-18
Makes the kubernetes typed version configurable

medoror:
* The kubernetes typed version seen in the created ovf as
  the FullVersion and Version xml tags can now be set
  on the command line in packer variable files. The original
  "kube" prefix logic still remains for backwards compatibility 

**Additional context**

### DIFF
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -37,7 +37,8 @@
     "output_dir": "./output/{{user `build_version`}}",
     "base_output_dir": "./output/{{user `base_build_version`}}",
     "username": "",
-    "vcenter_server": ""
+    "vcenter_server": "",
+    "kubernetes_typed_version": "kube-{{user `kubernetes_semver`}}"
   },
   "builders": [
     {
@@ -396,7 +397,8 @@
         "kubernetes_cni_semver": "{{user `kubernetes_cni_semver`}}",
         "kubernetes_semver": "{{user `kubernetes_semver`}}",
         "kubernetes_source_type": "{{user `kubernetes_source_type`}}",
-        "os_name": "{{user `os_display_name`}}"
+        "os_name": "{{user `os_display_name`}}",
+        "kubernetes_typed_version": "{{user `kubernetes_typed_version`}}"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


codenrhoden:
Hi @medoror,

This seems like an okay change to me. I think my preference would be just write the desired value of `TYPED_VERSION` directly into the json.manifest file like we do for many of the other values, and have the Python script read that in and add it to the XML directly, rather than having different paths depending on whether or not the string is empty/null.

In other words, in `packer.json`, set something like ```kubernetes_typed_version: kube-{{user `kubernetes-semver`}}```, and then write that into `manifest.json` file. Then it can be overriden as needed to whatever by the user.

I have *not* tested if this is allowed by Packer, though, to have on variable depend on something that defaults to null but gets filled in by a passed in var file.... 🤔

medoror:
cla me pls

medoror:
@codenrhoden Your suggestion seemed to work. Added a commit with the change

images/capi/packer/ova/packer-node.json
```diff
@@ -37,7 +37,8 @@
     "output_dir": "./output/{{user `build_version`}}",
     "base_output_dir": "./output/{{user `base_build_version`}}",
     "username": "",
-    "vcenter_server": ""
+    "vcenter_server": "",
+    "kubernetes_typed_version": "kube-{{user `kubernetes-semver`}}"
```
kkeshavamurthy, line 41: Seems like we are adding `kubernetes_typed_version` field by default to all created OVAs? Is that needed? Should it be set to null and the user can add it in their config when needed?

medoror: The main reason we did this was for backwards compatibility.  The typed version field was previously hardcoded within the script and can now be overridden in the json

kkeshavamurthy: Oh sorry did not notice that. My bad.

codenrhoden:
Hi @medoror,

Sorry to ask for a change, but can you squash the commits? This should be 1 single commit.

medoror:
@codenrhoden rebase and force pushed to make a single commit

images/capi/packer/ova/packer-node.json
codenrhoden, line 41: this needs to be `kubernetes_semver` (underscore rather than dash)

medoror: sheesh fixing now

medoror: @codenrhoden fixed!

codenrhoden:
Built this locally and all is 👍 

/lgtm
/approve

perithompson:
@codenrhoden  we need to make sure that windows is tested before merging as the addition was not added to windows

PR #514, opened 2021-02-19
Updates `ubuntu-64` OS type to `ubuntu64Guest`

SenatorSupes:

The OS type for Ubuntu 64 should be `ubuntu64Guest` for vSphere


codenrhoden:
This looks correct to me, looking at the list of types [here](https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html).

I remember there being weirdness around the types being different for Workstation/Fusion vs vSphere, which is why we have both versions in the various <OS>.json files -- so they can be set differently for the vmware builds vs the vsphere builders. But when creating the OVA, we probably want to make sure we set the vSphere version, since that's the preferred destination of the OVA. Plus, I built the OVA with this change, imported into Fusion, and it showed up as an Ubuntu 64 guest just fine.

I think the CentOS 7 and Windows targets may eventually need an update as well.

/ok-to-test

PR #515, opened 2021-02-20
Brings back mf file generation for vsphere

medoror:
* Not sure if this was intentional but image building for vsphere
  no longer generates an mf file starting from this commit 064ade4491659be11b731d3a8039225b21b1c2e4

**Additional context**

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -196,6 +196,9 @@ def main():
     # Create OVF
     create_ovf(ovf, data, ovf_template)
 
+    # Create the OVA manifest
+    create_ova_manifest(ova_manifest, [ovf, vmdk['stream_name']])
+
     # Create the OVA.
     create_ova(ova, ovf)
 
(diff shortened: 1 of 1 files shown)


codenrhoden:
It was intentional. The manifest is automatically created by OVFTool. What issue are you seeing?

medoror:
I did not see an mf file created on a recent build of image builder

Screen Shot 2021-02-22 at 9 01 34 AM

kkeshavamurthy:
Right, we do not need to create that explicitly anymore with `ovftool`. Were you using the mf file for something?

medoror:
I _think_ this file is needed for proper vCenter content library syncing. It is at least an artifact that we currently we deliver downstream

codenrhoden:
The `.mf` file is inside of the `.ova`.

When using `ovftool`, we don't want to create it before hand. `ovftool` will actually modify the `.ovf` a bit (because it renames the `.vmdk`), so any manifest we create would be invalid because the SHA would no longer be accurate. If you need the `.mf` file, you need to pull it out of the `.ova`


```
$ tar -tf ubuntu-2004-kube-v1.19.1.ova
ubuntu-2004-kube-v1.19.1.ovf
ubuntu-2004-kube-v1.19.1.mf
ubuntu-2004-kube-v1.19.1-disk1.vmdk
```

medoror:
MY BAD! this makes sense. Closing thanks for the support

PR #516, opened 2021-02-20
Allow overriding the instance type used during the AMI build process.

voor:

In the past we had an issue where this was hardcoded to t2.small and regions that didn't have the t2 generation would fail.  This goes one step further and actually allows changing the instance type used for building.  This is important for AMI that might need special underlying hardware on AWS that fail when that hardware is not present during image building -- for example GPU enabled instances.

**Additional context**

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -10,12 +10,13 @@
     "aws_profile": "",
     "aws_region": "us-east-1",
     "aws_secret_key": "",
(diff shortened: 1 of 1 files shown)


codenrhoden:
This LGTM. But the list of vars you are adding to is in alphabetical order. Can you please maintain that rather than appending at the end?

voor:
Looks like it wasn't entirely in alphabetical order, so a few additional changes involved.  Double check my commas

Issue #518, opened 2021-02-22
Generic version of debug-tools for executables


jayunit100:
**Is your feature request related to a problem? Please describe.**

feature - for windows, we need executables for CSI, CNI, and so on that are hollow , i.e. not really aware of cni or csi, but just, copying executables installed to image

**Describe the solution you'd like**

- Add an ansible step "cni-executables-install/" OR
- Add an ansible step "additional-file-copy/" 

**Describe alternatives you've considered**

- Make a container and unzip it at runtime 
- post kubeadm wget actions yuck !

**Additional context**

@perithompson and me want to ship Antrea in windows images ! 

/kind feature

jsturtevant:
A little background... Since windows doesn't have privileged containers it can't easily install and run services from pods like Linux does.  sig-windows came up with some work arounds using hostnetwork for Windows and Docker but it doesn't work as well with Containerd: https://github.com/kubernetes-sigs/sig-windows-tools/issues/128#issuecomment-744840966

For now with Containerd, in the current state of things it is necessary for containerd builds to have the CNI running on the node.  This will be temporary once [privileged containers](https://github.com/kubernetes/enhancements/pull/2288) lands and matures (which would make it closer to the way Linux install and run CNI's)

If this could be designed to be generic to install not just Antrea that would be beneficial for capz as well.

perithompson:
Completely agree with this, we were thinking as a generic way to do this, we could just start with a way to download a variable list of executables into a folder for CSI/CNI etc and let cluster API do the actual installation step? Any thoughts? I think maintaining an installation routine for each cni would be to hard to maintain...

jsturtevant:
Agreed,  Linux does something similar for the base line cnis: https://github.com/kubernetes-sigs/image-builder/blob/0b97b19a856ac257c4139b16eda76d1d43b72499/images/capi/ansible/roles/kubernetes/tasks/url.yml#L36

The tricky part might be setting up the configuraiton between containerd/CNI/kubelet.

jayunit100:
im ok w ANY solution here, generic or CNI specific , but feel like generic gives max flexibility to vendors w/o making upstream opinionated

PR #526, opened 2021-03-01
Added some info to docs for building windows nodes

### LINKED ISSUE #523: Allow building windows images without wins installed
**Is your feature request related to a problem?**
Wins is used to workaround the need for privileged containers in windows and allows for kube-proxy and other cnis to run on host.  However because this allows for scheduling workloads on the host, this can be considered insecure by some user and it would be useful to be able to skip the install of this 

Provide a variable to skip the wins installation

/kind feature
/sig windows

perithompson:
This adds some information on building windows nodes and helping to skip installing wins and configuring windows services with alternate service manager

Which issue(s) this PR fixes: Fixes #523

### DIFF
--- a/docs/book/src/capi/windows/windows.md
+++ b/docs/book/src/capi/windows/windows.md
@@ -12,6 +12,18 @@ The `images/capi/packer/config/windows` directory includes several JSON files th
 | `packer/config/windows/kubernetes.json` | The version of Kubernetes to install and it's install path |
 | `packer/config/windows/containerd.json` | The version of containerd to install |
 
+## Service Manager
+
+Image-builder provides you two ways to configure windows services. The default is setup using [nssm](https://nssm.cc/) which configures a windows service for kubelet by running `{{ kubernetes_install_path }}\StartKubelet.ps1` allowing easy editing of command arguments in the startup file.  The alternate is to use the windows native [sc.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/sc-config) which uses the kubelet argument `--windows-service` to install kubelet as a native windows service with the command line arguments configured on the service. Nssm handles service restarts, if you are using sc.exe you may wish to configure the service restart options on kubelet. To avoid starting kubelet to early, image-builder sets the kubelet service to *manual* which you should consider changing once the node has joined a cluster.
+
+**Important: sc.exe does not support kubeadm KUBELET_KUBEADM_ARGS which is used by Cluster API to pass extra user args**
+
+## Wins (by rancher)
+
+As a workaround for the lack of privileged containers on windows nodes, SIG Windows utilise [Wins](https://github.com/rancher/wins/) to allow pods to run processes on hosts, an example is when you are using kube-proxy or a cni provider in a pod. This means that by default we install wins using image-builder. This may be undesirable, if you do not need pod to have the ability to run processes on the host. Additionally, Alpha support for Windows [Host Processes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support), which is the windows equivalent for privileged containers, will be introduced in 1.22 which negates the need for wins entirely. To skip installing wins on the host you can add `"wins_url": ""` to your variables.  
+
+When using containerd, due to lack of host networking, there is currently no way to run cni inside a pod.   To work around this, containerd needs to be installed with an additional custom Ansible module or after the image has been created.  With CAPI, the addtional configuration can be done with pre/postKubeadm scripts when the node is created.
+
 ## Windows Updates
 
 When building Windows images it is necessary to install OS and Security updates.  Image Builder provides two variables to allow choosing which updates get installed which can be used together or seperately (with individual KBs installed first).


perithompson:
@CecileRobertMichon @jsturtevant is this missing anything that you think would be good to add?

docs/book/src/capi/windows/windows.md
```diff
@@ -12,6 +12,16 @@ The `images/capi/packer/config/windows` directory includes several JSON files th
...
+
+**Important: sc.exe does not support kubeadm KUBELET_KUBEADM_ARGS which is used by Cluster API to pass extra user args**
+
+## Wins (by rancher)
+
+As a workaround for the lack of privilidged containers on windows nodes, SIG Windows utilise [Wins](https://github.com/rancher/wins/) to allow pods to run processes on hosts, an example is when you are using kube-proxy or a cni provider in a pod. This means that by default we install wins using image-builder. This may be undesirable, if you do not need pod to have the ability to run processes on the host. To skip installing wins on the host you can add `"wins_url": ""` to your variables.
```
jsturtevant, line 23: Maybe link to the privileged kep?  We should also add a sentence that tells users they will be required to install cni and kubeproxy manually on the node. There isn't a great way to do that yet besides using pre/post kubeadm scripts: https://github.com/kubernetes-sigs/image-builder/issues/518

perithompson:
@jsturtevant are you able to take another look please?

docs/book/src/capi/windows/windows.md
```diff
@@ -12,6 +12,18 @@ The `images/capi/packer/config/windows` directory includes several JSON files th
 | `packer/config/windows/kubernetes.json` | The version of Kubernetes to install and it's install path |
 | `packer/config/windows/containerd.json` | The version of containerd to install |
 
+## Service Manager
+
+Image-builder provides you two ways to configure windows services. The default is setup using [nssm](https://nssm.cc/) which configures a windows service for kubelet by running `{{ kubernetes_install_path }}\StartKubelet.ps1` allowing easy editing of command arguments in the startup file.  The alternate is to use the windows native [sc.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/sc-config) which uses the kubelet argument `--windows-service` to install kubelet as a native windows service with the command line arguments configured on the service. Nssm handles service restarts, if you are using sc.exe you may wish to configure the service restart options on kubelet. To avoid starting kubelet to early, image-builder sets the kubelet service to *manual* which you should consider changing once the node has joined a cluster.
```
jsturtevant, line 17: should we consider moving the service automatic?  I think I moved it to manual because otherwise there were a bunch of kubelet logs from packer setup.  But we could clean those up...

docs/book/src/capi/windows/windows.md
```diff
@@ -12,6 +12,18 @@ The `images/capi/packer/config/windows` directory includes several JSON files th
...
+
+**Important: sc.exe does not support kubeadm KUBELET_KUBEADM_ARGS which is used by Cluster API to pass extra user args**
+
+## Wins (by rancher)
+
+As a workaround for the lack of privileged containers on windows nodes, SIG Windows utilise [Wins](https://github.com/rancher/wins/) to allow pods to run processes on hosts, an example is when you are using kube-proxy or a cni provider in a pod. This means that by default we install wins using image-builder. This may be undesirable, if you do not need pod to have the ability to run processes on the host. Additionally, Alpha support for Windows [Host Processes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support), which is the windows equivalent for privileged containers, will be introduced in 1.21 which negates the need for wins entirely. To skip installing wins on the host you can add `"wins_url": ""` to your variables.
```
jsturtevant, line 23:
```suggestion
As a workaround for the lack of privileged containers on windows nodes, SIG Windows utilise [Wins](https://github.com/rancher/wins/) to allow pods to run processes on hosts, an example is when you are using kube-proxy or a cni provider in a pod. This means that by default we install wins using image-builder. This may be undesirable, if you do not need pod to have the ability to run processes on the host. Additionally, Alpha support for Windows [Host Processes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support), which is the windows equivalent for privileged containers, will be introduced in 1.22 which negates the need for wins entirely. To skip installing wins on the host you can add `"wins_url": ""` to your variables.  
```

docs/book/src/capi/windows/windows.md
```diff
@@ -12,6 +12,18 @@ The `images/capi/packer/config/windows` directory includes several JSON files th
...
+
+## Wins (by rancher)
+
+As a workaround for the lack of privileged containers on windows nodes, SIG Windows utilise [Wins](https://github.com/rancher/wins/) to allow pods to run processes on hosts, an example is when you are using kube-proxy or a cni provider in a pod. This means that by default we install wins using image-builder. This may be undesirable, if you do not need pod to have the ability to run processes on the host. Additionally, Alpha support for Windows [Host Processes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support), which is the windows equivalent for privileged containers, will be introduced in 1.21 which negates the need for wins entirely. To skip installing wins on the host you can add `"wins_url": ""` to your variables.  
+
+On containerd, due to the way host networking works, there is currently no way to run cni inside a pod so needs to be installed, along with kube-proxy, after the image has been created.  With CAPI, this is done with pre/postKubeadm scripts as the node is created.
```
jsturtevant, line 25:
```suggestion
When using containerd, due to lack of host networking, there is currently no way to run cni inside a pod.   To work around this, containerd needs to be installed with an additional custom Ansible module or after the image has been created.  With CAPI, the addtional configuration can be done with pre/postKubeadm scripts when the node is created.
```

docs/book/src/capi/windows/windows.md
perithompson: good spot!

docs/book/src/capi/windows/windows.md
perithompson: better wording that my rambling! thank you!

docs/book/src/capi/windows/windows.md
perithompson: I'm currently settings this to auto in preKubeadm but yeah we could set it to auto? I think it's worth leaving as long as its pointed out, kubelet does create a bit of noise, but I'm happy either way!

jsturtevant: created an issue to track it: https://github.com/kubernetes-sigs/image-builder/issues/554

jsturtevant:
Thanks for documenting all of this.
/lgtm

PR #368, opened 2020-09-10
alphabetical sort of all json files

bavarianbidi:
currently it's a little pain to keep forks in sync if
there are lot of changes on some json files.

rebase and merging should now a little bit easier if json files
are sorted alphabetical.

sorting was done via

```bash
find . -regex ".*.json" -exec sh -c 'f={}; cat $f | jq -S "." > $f.tmp && mv $f.tmp $f' \;
```

Mario Constanti <mario.constanti@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

PR #527, opened 2021-03-02
Sort json files

kkeshavamurthy:
- `make json-sort` to sort all the json files
- `./hack/ci-json-sort.sh` to run from CI to check if all json files are sorted. 
- presubmit job [PR](https://github.com/kubernetes/test-infra/pull/21112). 

**Additional context**

### DIFF
--- a/images/capi/hack/ci-json-sort.sh
+++ b/images/capi/hack/ci-json-sort.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###############################################################################
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/../../..
+cd "${CAPI_ROOT}" || exit 1
+
+cleanup() {
+    returnCode="$?"
+    exit "${returnCode}"
+}
+
+trap cleanup EXIT
+
+json_files=$(find . -type f -name "*.json" | sort -u)
+for f in ${json_files}
+do
+  if ! diff <(jq -S . ${f}) ${f} >> /dev/null; then
+    echo "json files are not sorted!! Please sort them with \"make sort\" before commit"
+    echo "Unsorted file: ${f}"
+    exit 1
+  fi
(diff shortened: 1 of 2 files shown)


codenrhoden:
So originally when this was attempted in https://github.com/kubernetes-sigs/image-builder/pull/368 (which is still open), my main objection had been that the `variables` section gets moved to the bottom of the `packer.json` files rather than the top. I could be convinced that the benefits of this outweigh that drawback, though.

The other thing that came up in that PR was that sorting once wasn't very helpful if we didn't have a corresponding CI check, which this PR does add, so that would be good. I'm inclined to approve this. I know @CecileRobertMichon weighed in on the other PR (which has been stalled for a couple of months waiting on a corresponding CI check). Thoughts?

@kkeshavamurthy one thing i wonder, why isn't there a second commit with the files actually sorted? Waiting to see if people think this is a good idea?

images/capi/hack/ci-json-sort.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
```
codenrhoden, line 3: as a new file, this (c) date would need to be 2021.

kkeshavamurthy:
As you mentioned, the missing parts from #368 are the CI job and a make target to do the sort, which this PR adds. 
Since #368 was already committing the sorted files, I did not want to step on his work.

images/capi/hack/ci-json-sort.sh
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###############################################################################
+
+# This script is executed by presubmit `pull-cluster-api-provider-azure-e2e`
+# To run locally, set AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, AZURE_SUBSCRIPTION_ID, AZURE_TENANT_ID
```
codenrhoden, line 20: These comments are copy pasta I think. :)

images/capi/hack/ci-json-sort.sh
```diff
@@ -0,0 +1,50 @@
...
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/../../..
+cd "${CAPI_ROOT}" || exit 1
+
+# timestamp is in RFC-3339 format to match kubetest
+export TIMESTAMP="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+export JOB_NAME="${JOB_NAME:-"check-json"}"
+export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
```
codenrhoden, line 34: I don't think these exports are needed? Now that the test is enabled in prow (optionally) I'd try running the test (which will always fail, but still) with and without these. None of our others have this? I guess the Azure CI does, but I'm not familiar with this being required.

codenrhoden:
Looking at the logs at https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-image-builder#pr-json-sort-check, I can't see any meaningful diff w/o those export values in place.

I think this is good. :)

/lgtm
/approve

PR #531, opened 2021-03-03
Do not display skipped ansible tasks

kkeshavamurthy:
Ansible output is very cluttered by displaying all the skipped tasks. This PR add `ANSIBLE_DISPLAY_SKIPPED_HOSTS=no` env variable to ansible provisioner and the output is much better

**Additional context**

Before

```
    amazon-2:
    amazon-2: PLAY [all] *********************************************************************
    amazon-2:
    amazon-2: TASK [Gathering Facts] *********************************************************
    amazon-2: ok: [default]
    amazon-2: [WARNING]: Platform linux on host default is using the discovered Python
    amazon-2: interpreter at /usr/bin/python, but future installation of another Python
    amazon-2: interpreter could change the meaning of that path. See https://docs.ansible.com
    amazon-2: /ansible/2.10/reference_appendices/interpreter_discovery.html for more
    amazon-2: information.
    amazon-2:
    amazon-2: TASK [include_role : node] *****************************************************
    amazon-2:
    amazon-2: TASK [setup : Put templated sources.list in place] *****************************
    amazon-2: skipping: [default]
    amazon-2:
    amazon-2: TASK [setup : Put templated apt.conf.d/90proxy in place when defined] **********
    amazon-2: skipping: [default]
30 more lines
    .....
    amazon-2: TASK [setup : add epel repo] ***************************************************
    amazon-2: changed: [default]
```

After
```
    amazon-2:
    amazon-2: PLAY [all] *********************************************************************
    amazon-2:
    amazon-2: [WARNING]: Platform linux on host default is using the discovered Python
    amazon-2: TASK [Gathering Facts] *********************************************************
    amazon-2: interpreter at /usr/bin/python, but future installation of another Python
    amazon-2: ok: [default]
    amazon-2: interpreter could change the meaning of that path. See https://docs.ansible.com
    amazon-2: /ansible/2.10/reference_appendices/interpreter_discovery.html for more
    amazon-2: information.
    amazon-2:
    amazon-2: TASK [setup : add epel repo] ***************************************************
    amazon-2: changed: [default]
    amazon-2:
    amazon-2: TASK [setup : perform a yum update] ********************************************
    amazon-2: changed: [default]
```

### DIFF
--- a/images/capi/ansible.cfg
+++ b/images/capi/ansible.cfg
@@ -14,6 +14,7 @@
 
 [defaults]
 remote_tmp = /tmp/.ansible
+display_skipped_hosts = False
 
 [ssh_connection]
 pipelining = False


jsturtevant:
Since this is being applied to all configurations could we set it in the `ansible.cfg` [file](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#envvar-ANSIBLE_DISPLAY_SKIPPED_HOSTS)?

kkeshavamurthy:
@jsturtevant I added `display_skipped_hosts = False` under `[defaults]` in the ansible.cfg and could not get it to work, so added it in the provisioner. Not sure what I was doing wrong. If it works, `ansible.cfg` would be the best place to add it, agree.

codenrhoden:
Let's think about this one for a bit.

The reason editing `ansible.cfg` that is found in `images/capi/ansible` doesn't work is because that [isn't read](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#the-configuration-file). I had forgotten that it was there, but it isn't in one of the paths Ansible is going to read. We are running Ansible from a PWD of `images/capi`, so it would have to be there, or we would need to set `ANSIBLE_CONFIG=$(pwd)/ansible` or something similar.

We've never really visited this setup.

I see a few of options to consider

- Set this explicit env var in every Packer file (current proposal, lots of file changes)
- Change every Packer file to point to `ansible.cfg` correctly and add it there (lots of changes, only has to be done once and future changes can be made to `ansible.cfg)
- export the env var from the Makefile (1 file to change, haven't confirmed this works)

I'm personally a fan of adding it to the makefile, so that you can set/change/override it without modifying any other files once it is in place. Basically, while this may make a sensible default, I would want an easy way to change this behavior.

Seperately from that decision, we should figure out what to do with the existing `ansible.cfg`, because I sure don't think it is being used at all. Happy to see evidence otherwise (I can test this sometime in the future, but not at the moment).

kkeshavamurthy:
Tried the `Makefile` method, doesn't work. 
I tried manually exporting the `ANSIBLE_CONFIG` env var before running `make`, did not work either. 
I did not investigate it further.

codenrhoden:
@kkeshavamurthy  Witrh #584 merged in, can you please rebase this and add the needed line to `ansible.cfg`?

kkeshavamurthy:
@codenrhoden Done.

codenrhoden:
/lgtm
/approve

woohoo!

PR #533, opened 2021-03-04
GOSS test fix for load_additional_imgs flag

SanikaGawhane:
Fixes an error in GOSS test conditions that caused a wrong GOSS test to run and fail when `load_additional_imgs="true"` 

/cc @kkeshavamurthy

### DIFF
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -18,7 +18,7 @@ command:
     stdout: ["coredns", "etcd", "kube-apiserver", "kube-controller-manager", "kube-proxy", "kube-scheduler", "pause"]
 {{end}}
 {{end}}
-{{if and (eq .Vars.kubernetes_source_type "http") (eq .Vars.kubernetes_cni_source_type "http") (not .Vars.kubernetes_load_additional_imgs)}}
+{{if and (eq .Vars.kubernetes_source_type "http") (eq .Vars.kubernetes_cni_source_type "http") (eq .Vars.kubernetes_load_additional_imgs "false")}}
 # The second last pipe of awk is to take out arch from kube-apiserver-amd64 (i.e. amd64 or any other arch)
   crictl images | grep -v 'IMAGE ID' | awk -F'[ /]' '{print $2}' | sed 's/-{{ .Vars.arch }}//g' | sort:
     exit-status: 0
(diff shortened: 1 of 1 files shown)


kkeshavamurthy:
/ok-to-test

lgtm from me, @EleanorRigby do you mind taking a look?

images/capi/packer/goss/goss-command.yaml
```diff
@@ -18,15 +18,15 @@ command:
...
+{{if and (eq .Vars.kubernetes_source_type "http") (eq .Vars.kubernetes_cni_source_type "http") (eq .Vars.kubernetes_load_additional_imgs "false")}}
 # The second last pipe of awk is to take out arch from kube-apiserver-amd64 (i.e. amd64 or any other arch)
   crictl images | grep -v 'IMAGE ID' | awk -F'[ /]' '{print $2}' | sed 's/-{{ .Vars.arch }}//g' | sort:
     exit-status: 0
     stderr: []
     timeout: 0
     stdout: ["kube-apiserver", "kube-controller-manager", "kube-proxy", "kube-scheduler"]
 {{end}}
-{{if and (eq .Vars.kubernetes_source_type "http") (eq .Vars.kubernetes_cni_source_type "http") (.Vars.kubernetes_load_additional_imgs)}}
+{{if and (eq .Vars.kubernetes_source_type "http") (eq .Vars.kubernetes_cni_source_type "http") (eq .Vars.kubernetes_load_additional_imgs "true")}}
```
codenrhoden, line 29: So is the takeaway here that this needed to be treated as a string instead of a bool?

@kkeshavamurthy thinking ahead, I wonder how these will be affected when you can load extra/arbitrary containers like you are proposing in another PR...

kkeshavamurthy: True, need to think about that.

codenrhoden:
@SanikaGawhane I know you have some internal builds you run that depend on this logic. I think you showed me some results that may mean this change isn't necessary? Can you confirm?

SanikaGawhane:
Yes, this change isn't necessary. We were running the correct test as opposed to what we thought earlier. 
The error was due to `awk -F'[ /]' '{print $2}'` part of the `crictl` command printing the wrong string in our internal builds because of the way they're structured/named.

PR #535, opened 2021-03-04
Standardise pause image for windows and linux

perithompson:

As Linux now enforces the --pod-infra-container-image meaning this is no longer containerd specific and Windows uses the same multi-arch pause image, I have standardised the naming convention and image for both variables.  This is a breaking change for anyone using containerd_pause_image in their pipelines or build process

**Additional context**


images/capi/packer/config/common.json
```diff
@@ -8,5 +8,6 @@
   "no_proxy": "",
   "redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm",
   "reenable_public_repos": "true",
-  "remove_extra_repos": "false"
+  "remove_extra_repos": "false",
+  "pause_image": "k8s.gcr.io/pause:3.2"
```
jsturtevant, line 12: for windows the pause image must be at 3.4.1+ (https://github.com/kubernetes/kubernetes/pull/91452).

@codenrhoden Do you have any reservations bumping this?

images/capi/packer/config/kubernetes.json
```diff
@@ -1,22 +1,22 @@
...
-  "kubernetes_series": "v1.18",
-  "kubernetes_semver": "v1.18.15",
-  "kubernetes_rpm_version": "1.18.15-0",
+  "crictl_sha256": "44d5f550ef3f41f9b53155906e0229ffdbee4b19452b4df540265e29572b899c",
+  "crictl_source_type": "pkg",
+  "crictl_version": "1.20.0",
+  "kubeadm_template": "etc/kubeadm.yml",
+  "kubernetes_container_registry": "k8s.gcr.io",
+  "kubernetes_deb_gpg_key": "https://packages.cloud.google.com/apt/doc/apt-key.gpg",
+  "kubernetes_deb_repo": "\"https://apt.kubernetes.io/ kubernetes-xenial\"",
```
jsturtevant, line 8: while re-ordering this is a good thing it make it hard to see the diff here...  Maybe there should be linter on the files to enforce order?

images/capi/packer/config/containerd.json
```diff
@@ -2,7 +2,6 @@
     "containerd_version": "1.4.3",
     "containerd_sha256": "2697a342e3477c211ab48313e259fd7e32ad1f5ded19320e6a559f50a82bff3d",
     "containerd_sha256_windows": "4464dc16978b6e984c442e17cfcc0826e1aebd8e12cb64a3ce2bdb89affa2138",
-    "containerd_pause_image": "k8s.gcr.io/pause:3.2",
```
jsturtevant, line 5: if keep this value and set it to `pause_image` we could avoid breaking change.  @codenrhoden what is our depreciation policy for various fields?

images/capi/packer/config/kubernetes.json
perithompson: I originally added pause_image in here hence the reorder but I can revert. I agree on this though, it is hard to see the differences not sure if @codenrhoden has any thoughts, I think a linter of some kind would be good!

images/capi/packer/config/containerd.json
perithompson: I'd be concerned someone would set containerd_pause_image potentially at a more recent version and not set pause_image, you could then potentially have a mismatch with --pod-infrastructure-image and containerd's toml, unless I am missing something?

images/capi/packer/config/kubernetes.json
kkeshavamurthy: I have a [PR](https://github.com/kubernetes/test-infra/pull/21112) open to check if json files are sorted.

images/capi/packer/config/containerd.json
jsturtevant: Yea, I think I mixed this up.  I had a different scenario in mind. Still interested in how we should handle the depreciation.

docs/book/src/capi/capi.md
```diff
@@ -57,7 +57,7 @@ Several variables can be used to customize the image build.
 | `no_proxy` | This can be set to a comma-delimited list of domains that should be exluded from proxying during the Ansible stage of building | `""` |
 | `reenable_public_repos` | If set to `"false"`, the package repositories disabled by setting `disable_public_repos` will remain disabled at the end of the build. | `"true"` |
 | `remove_extra_repos` | If set to `"true"`, the package repositories added to the OS through the use of `extra_repos` will be removed at the end of the build. | `"false"` |
-| `containerd_pause_image` | This can be used to override the default containerd pause image used to hold the network namespace and IP for the pod. | `"k8s.gcr.io/pause:3.2"` |
+| `pause_image` | This can be used to override the default containerd pause image used to hold the network namespace and IP for the pod. | `"k8s.gcr.io/pause:3.2"` |
```
detiber, line 60:
```suggestion
| `pause_image` | This can be used to override the default pause image used to hold the network namespace and IP for the pod. | `"k8s.gcr.io/pause:3.2"` |
```

images/capi/packer/config/common.json
detiber: Makes sense to bump the version to me, from my understanding the pause image should be backwards compatible

perithompson:
I've rebased and bumped the pause version too

images/capi/packer/config/containerd.json
```diff
@@ -1,8 +1,8 @@
 {
   "containerd_additional_settings": null,
   "containerd_cri_socket": "/var/run/containerd/containerd.sock",
-  "containerd_pause_image": "k8s.gcr.io/pause:3.2",
   "containerd_sha256": "96641849cb78a0a119223a427dfdc1ade88412ef791a14193212c8c8e29d447b",
   "containerd_sha256_windows": "0b668d28ec7ffda3d1930ed729ff8591540278a56539ae0bd682dea8c9964483",
-  "containerd_version": "1.4.4"
+  "containerd_version": "1.4.4",
+  "pause_image": "k8s.gcr.io/pause:3.4.1"
```
jsturtevant, line 7: I don't think this is needed since it is set in common?

jsturtevant (commented):
Generally looks good.  Minor comment and one open question is on how to handle the breaking change.

images/capi/packer/config/containerd.json
perithompson: Oh good point!

perithompson:
> Generally looks good.  Minor comment and one open question is on how to handle the breaking change.  

Thanks @jsturtevant! I changed from WIP to put it onto @codenrhoden s radar to address that :)

perithompson:
@codenrhoden are you able to review this for me, and let me know if there is anything we need to do about swapping the containerd_pause_image to just pause_image and standardise on one, as I know this is a breaking change

jsturtevant:
/lgtm 
/test pull-azure-vhds

We discussed this at office hours and we are good to go on the breaking change.  There have been others like this in the past due to being still 0.0.x version.  Discussed that it might be possible to hide these changes behind a cli in the future.

detiber:
It looks like there is consensus to move forward despite the breaking change.
/approve

PR #536, opened 2021-03-05
Set kernel param for strict reverse packet filtering for Ubuntu images

SanikaGawhane:
Sets `net.ipv4.conf.all.rp_filter=1` for Ubuntu images.
Includes corresponding GOSS test.

/cc @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -83,3 +83,12 @@
     group: root
     mode: 0644
   when: ansible_os_family != "Flatcar"
+
+- name: Ensure reverse packet filtering is set as strict
+  sysctl:
+    name: net.ipv4.conf.all.rp_filter
+    value: "1"
+    state: present
+    sysctl_set: yes
+    reload: yes
+  when: ansible_distribution == "Ubuntu"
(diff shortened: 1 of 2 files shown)


codenrhoden:
/ok-to-test

For background, this is addressing a regression in Ubuntu 20.04 from 18.04 when trying to run Calico as the CNI. Ubuntu 20.04 has `rp_filter` set to `2` by default, which Calico does not like:

```
2021-02-28 22:46:00.444 [FATAL][45] int_dataplane.go 1035: Kernel's RPF check is set to 'loose'.  This would allow endpoints to spoof their IP address.  Calico requires net.ipv4.conf.all.rp_filter to be set to 0
or 1. If you require loose RPF and you are not concerned about spoofing, this check can be disabled by setting the IgnoreLooseRPF configuration parameter to 'true'.
```

This has also been seen in KIND: https://github.com/kubernetes-sigs/kind/issues/891

Ubuntu 18.04 had this set to `1` by default.

Setting to `1` by default in the image is the more secure option. If someone really wants to disable all reverse path filtering for their environment, they can change their image. But right now, Calico tends not to work at all with Ubuntu 20.04 without this change.

/lgtm

PR #540, opened 2021-03-08
Set systemd cgroup driver for K8s >= 1.21

### LINKED ISSUE #471: containerd should be using systemd cgroup hierarchy

As per https://kubernetes.io/docs/setup/production-environment/container-runtimes/, there should be only be a single cgroup hiearchy running on the OS. We do not configure containerd to run using the systemd hierarchy, such that Kubernetes will have an incorrect view of resource utilisation.

ContainerD should be configured to use systemd hierarchy

ContainerD config.toml should have something along the lines of

``` toml
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
  ...
  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
    SystemdCgroup = true
```

**Environment:**

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`):
- Packer Version:
- Packer Provider:
- Ansible Version: 

/kind bug

codenrhoden:
When Kubernetes is 1.21.0 or greater, we want to make sure we configure
containerd to use the SystemD cgroup driver, not the default cgroupfs.

Fixes #471

**Additional context**
The built-in `version` comparison for Ansible seems to handle this just fine.

The resulting config file with this change looks like this:

```toml
## template: jinja

# Use config version 2 to enable new configuration fields.
# Config file is parsed as version 1 by default.
version = 2

[plugins]
  [plugins."io.containerd.grpc.v1.cri"]
    sandbox_image = "k8s.gcr.io/pause:3.2"
    [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
      SystemdCgroup = true
```

/cc @neolit123 @fabriziopandini 

I've verified that this gets populated correctly with < 1.21.0, and with >= 1.21.0

### DIFF
--- a/images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
+++ b/images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
@@ -37,7 +37,11 @@ version = 2
 [plugins]
   [plugins."io.containerd.grpc.v1.cri"]
     sandbox_image = "{{ containerd_pause_image }}"
+  {% if kubernetes_semver is version('v1.21.0', '>=') %}
+  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+      SystemdCgroup = true
+  {% endif %}
 
 {% endif %}
-    
+
 {{containerd_additional_settings | b64decode}}


neolit123 (commented):
/lgtm
/hold

i remember us discussing in the image-builder meeting that maybe the new value should be templatable. is this still needed?

you can `/hold cancel` when you thing this is ready to merge.
thanks

codenrhoden:
@neolit123 can you refresh my memory on this?

I know what I meant when I said it was "templatable" was that I could put a switch in the existing `config.toml` *template*, which is what this PR does. I believe the exact entries in the `config.toml` file are fixed depending on the Kubernetes version, so I'm not sure that there is additional flexibility needed here.

neolit123:
> @neolit123 can you refresh my memory on this?
> 
> I know what I meant when I said it was "templatable" was that I could put a switch in the existing `config.toml` _template_, which is what this PR does. I believe the exact entries in the `config.toml` file are fixed depending on the Kubernetes version, so I'm not sure that there is additional flexibility needed here.

i don't know if there is a case where this config will not be used in a setup where kubeadm is not used and the kubelet will be run without systemd (the init system).

if there is no such use case this seems fine and whoever needs it they can file a PR to template the "systemdCgroup" value.

detiber:
> i don't know if there is a case where this config will not be used in a setup where kubeadm is not used and the kubelet will be run without systemd (the init system).
> 
> if there is no such use case this seems fine and whoever needs it they can file a PR to template the "systemdCgroup" value.

I believe all the supported OSs that are built for Linux here all support systemd, so I don't think there will be an issue there.

codenrhoden:
/hold cancel

Let's go for it! :)

fabriziopandini:
many thanks @codenrhoden !

PR #546, opened 2021-03-11
Fix validate make targets

codenrhoden:
There were a few problems with the validate make targets. Flatcar had
been added to the list of OS's supported for OVA builds, which is
incorrect. The Azure windows-sig target had a typo in it. The QEMU
target was missing flatcar, and the AMI target wasn't documented. This
patch fixes all those things and `make validate-all` can work again,
presubming you supply the needed ENV vars.

**Additional context**
To run this on my local machine, I use the following:
```commandline
$ AZURE_LOCATION=fake RESOURCE_GROUP_NAME=fake STORAGE_ACCOUNT_NAME=fake DIGITALOCEAN_ACCESS_TOKEN=fake GCP_PROJECT_ID=fake make validate-all
```

That provides fake vars that will allow the check to pass. I'm going to add this test to CI.

@perithompson I know you were interested in this. Unfortunately it would not have caught the Windows+OVA issue you say, as this doesn't test the interface between the Packer manifest file and the Python script that generates the OVA.

### DIFF
--- a/images/capi/scripts/ci-packer-validate.sh
+++ b/images/capi/scripts/ci-packer-validate.sh
@@ -1,6 +1,6 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
-# Copyright 2020 The Kubernetes Authors.
+# Copyright 2021 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,23 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+###############################################################################
+
 set -o errexit
 set -o nounset
 set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-source hack/utils.sh
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
 
-_version="2.10.0"
+export PATH=${PWD}/.local/bin:$PATH
+export PATH=${PYTHON_BIN_DIR:-"${HOME}/.local/bin"}:$PATH
 
-# Change directories to the parent directory of the one in which this
-# script is located.
-cd "$(dirname "${BASH_SOURCE[0]}")/.."
-
-if command -v ansible >/dev/null 2>&1; then exit 0; fi
-
-ensure_py3
-pip3 install --user "ansible==${_version}"
-ensure_py3_bin ansible
-ensure_py3_bin ansible-playbook
+AZURE_LOCATION=fake RESOURCE_GROUP_NAME=fake STORAGE_ACCOUNT_NAME=fake \
+  DIGITALOCEAN_ACCESS_TOKEN=fake GCP_PROJECT_ID=fake \
+  make validate-all
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -216,7 +216,6 @@ FLATCAR_VERSION ?= 2605.10.0
 export FLATCAR_CHANNEL FLATCAR_VERSION
 
 PLATFORMS_AND_VERSIONS	:=	$(CENTOS_VERSIONS) \
-							$(FLATCAR_VERSIONS) \
 							$(PHOTON_VERSIONS) \
 							$(RHEL_VERSIONS) \
 							$(UBUNTU_VERSIONS) \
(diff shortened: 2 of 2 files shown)


perithompson:
I can only think that the earlier problem would have been picked up by validating the post-processor match the required ova data fields?

codenrhoden:
> I can only think that the earlier problem would have been picked up by validating the post-processor match the required ova data fields?

Yeah, right now we don't have a way to verify that without building the image. :/

I have opened https://github.com/kubernetes/test-infra/pull/21325 to at least run the static validation tests for Packer.

codenrhoden:
This also makes good on the promise I made in https://github.com/kubernetes-sigs/image-builder/pull/336 almost 6 months ago. 😆

codenrhoden:
Hi @CecileRobertMichon,

I could use your advise on how best to fix this. The current testing error is happening because I don't have something like https://github.com/kubernetes-sigs/image-builder/blob/3819049d878203e591e9495ca88166698a750615/images/capi/packer/azure/scripts/ci-azure-e2e.sh#L46 inside of a script to run these tests since it is just running a `make` target.

Do you think it's better to have every test inside of a script (so we can modify the test independent of Prow), or is it better to try and modify `PATH` inside of the Prow job definition?

codenrhoden:
> Do you think it's better to have every test inside of a script (so we can modify the test independent of Prow), or is it better to try and modify `PATH` inside of the Prow job definition?

@detiber Do you have an opinion on this one? Trying to decide if its better to fix this in a script or in the Prow job definition. If the answer is a script, I'm more of a fan of having a dedicated `ci` folder for all CI scripts/jobs, rather than having them deep down in something like `packer/azure/scripts` like Azure is now, but I suspect that's more of a matter of preference.

CecileRobertMichon:
> I'm more of a fan of having a dedicated ci folder for all CI scripts/jobs, rather than having them deep down in something like packer/azure/scripts like Azure is now, but I suspect that's more of a matter of preference.

This sounds good to me. I'm leaning towards having script since it's easier to modify than having to open a PR in test-infra and also makes it work in other environments beside prow.

detiber:
> > Do you think it's better to have every test inside of a script (so we can modify the test independent of Prow), or is it better to try and modify `PATH` inside of the Prow job definition?
> 
> @detiber Do you have an opinion on this one? Trying to decide if its better to fix this in a script or in the Prow job definition. If the answer is a script, I'm more of a fan of having a dedicated `ci` folder for all CI scripts/jobs, rather than having them deep down in something like `packer/azure/scripts` like Azure is now, but I suspect that's more of a matter of preference.

hmm, I'm not sure I have enough context off the top of my head, but m y general take would be:
- modify in prow if it is something very specific to prow, and would not affect users running locally
- modify in the script if it also needs to be modified to run locally (but defaults should work for the general use case and not just prow)

codenrhoden:
Thanks both of you for the feedback. The script does seem simpler, so I've updated this PR and opened https://github.com/kubernetes/test-infra/pull/21743. Once that is merged in, I can test this.

images/capi/scripts/ci-packer-validate.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
+
+export PATH=${PWD}/.local/bin:$PATH
+export PATH=${PYTHON_BIN_DIR:-"/root/.local/bin"}:$PATH
```
detiber, line 29: Could this use `$HOME/` or `~/` here instead of `/root/`? That should make it more generally usable outside of prow.

codenrhoden: Good call, that makes sense. I had copied this straight for the Azure CI script. Will change.

codenrhoden:
/test packer-validate

This passes now. No comments on the GCP one -- looks like it just came in and it's still marked as optional so I'm ignoring it for now.

images/capi/scripts/ci-packer-validate.sh
codenrhoden: For posterity, this has to be $HOME. I tried `~/` and it didn't work. :)

CecileRobertMichon:
@codenrhoden the gcp jobs comes from https://github.com/kubernetes/test-infra/pull/21414/files which just merged... It shouldn't be running on every PR though, I missed that it had both `always_run: false` and `run_if_changed` set

codenrhoden:
Since it's marked as `required: false` I think this will still merge with an `/lgtm` (if one were given). I *think*.

CecileRobertMichon (commented):
/lgtm

we should probably move all Azure/GCP/other scripts to the script/ directory too as a follow up

codenrhoden:
> we should probably move all Azure/GCP/other scripts to the script/ directory too as a follow up

I was thinking the same. I'm willing to take that on. I was thinking of doing it in three steps:

1.  create a symlink from new location to old location (instead of outright moving first)
2. updating Prow jobs to point to new locations (which are symlinks to old)
3. replacing symlinks with the actual files (via `git mv`).

I think this way will avoid any interruption in tests passing.

PR #548, opened 2021-03-16
Updates packer patch version.

voor:

Avoids this error with building AMI images:

```
Right version of binary present
packer build -var-file="/home/imagebuilder/packer/config/kubernetes.json"  -var-file="/home/imagebuilder/packer/config/cni.json"  -var-file="/home/imagebuilder/packer/config/containerd.json"  -var-file="/home/imagebuilder/packer/config/ansible-args.json"  -var-file="/home/imagebuilder/packer/config/goss-args.json"  -var-file="/home/imagebuilder/packer/config/common.json"  -color=true -var-file="/home/imagebuilder/packer/ami/ubuntu-1804.json" -var-file="/tmp/build.json"  -var-file="/tmp/aws_settings.json"  -var-file="/tmp/additional_settings.json"  packer/ami/packer.json
Error: Failed to prepare build: "ubuntu-18.04"
1 error occurred:
	* unknown configuration key: '"launch_block_device_mappings[0].throughput"'
```

**Additional context**

### DIFF
--- a/images/capi/hack/ensure-packer.sh
+++ b/images/capi/hack/ensure-packer.sh
@@ -20,7 +20,7 @@ set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-_version="1.6.0"
+_version="1.6.6"
 
 # Change directories to the parent directory of the one in which this
 # script is located.


codenrhoden:
Thanks Robert. Good catch, and  unfortunately this means that the container published as `v0.1.7` doesn't work for AWS out of the box. I'll tag the next version to address this issue once this is in. This is something that #423 would catch... will keep pushing on that one.

/lgtm

codenrhoden:
/test pull-azure-vhds

I don't see anything in the logs about why this failed. It said everything passed. 🤔

codenrhoden:
@jsturtevant @CecileRobertMichon 

Any idea what is causing these errors in the Windows node builds?

```
    vhd-windows-2004: fatal: [default]: FAILED! => {"changed": true, "cmd": "# https://github.com/microsoft/azure-pipelines-image-generation/pull/1259/files\n$service = Get-Service \"Windefend\"\n$service.WaitForStatus(\"Running\",\"00:5:00\")\nUpdate-MpSignature", "delta": "0:00:15.324964", "end": "2021-03-16 05:57:32.287716", "msg": "non-zero return code", "rc": 1, "start": "2021-03-16 05:57:16.962752", "stderr": "Update-MpSignature : The remote procedure call failed. \r\nAt line:4 char:1\r\n+ Update-MpSignature\r\n+ ~~~~~~~~~~~~~~~~~~\r\n    + CategoryInfo          : NotSpecified: (MSFT_MpSignature:ROOT\\Microsoft\\...SFT_MpSignature) [Update-MpSignature], \r\n    CimException\r\n    + FullyQualifiedErrorId : HRESULT 0x800706be,Update-MpSignature", "stderr_lines": ["Update-MpSignature : The remote procedure call failed. ", "At line:4 char:1", "+ Update-MpSignature", "+ ~~~~~~~~~~~~~~~~~~", "    + CategoryInfo          : NotSpecified: (MSFT_MpSignature:ROOT\\Microsoft\\...SFT_MpSignature) [Update-MpSignature], ", "    CimException", "    + FullyQualifiedErrorId : HRESULT 0x800706be,Update-MpSignature"], "stdout": "", "stdout_lines": []}
```

jsturtevant:
It is the update of the windows defender signatures... I thought the wait for status fixed this (saw it early in the dev cycle) but I guess something else is happening. I'll create an issue to track fixing it

CecileRobertMichon:
I missed this... @codenrhoden @voor there is a known issue in the Packer 1.6.2+ Azure builder. @alexeldeib has a PR open to fix it https://github.com/hashicorp/packer/pull/10713 and it looks like it just got reviewed today by Packer maintainers. Might have to revert this or introduce some sort of "if Azure use older version" logic ☹️

codenrhoden:
@CecileRobertMichon Hmm, I see what you mean. I was not aware of this, I'm sorry for the trouble here. I'm definitely torn on what to do. It seems that with the current state of things, there is no Packer version we can use that works for both Azure and AWS.

As it stands, the published `v0.1.7` container image does not work for AWS (but does for Azure). The change here will make it work for AWS, but now breaks Azure.

How would you feel if I tagged `v0.1.8` with this change, and in the release notes we document that that if you are using Azure, you need to use `v0.1.7` until a new Packer version is released with the fix? We don't push any `latest` tags, so people have to choose their version anyways.

From a CI perspective, are we okay for now with Packer v1.6.6 on Azure since the entire RG get's cleaned up?

CecileRobertMichon:
@codenrhoden that action plan sounds good to me. I just checked and you're right that the entire RG gets cleaned up for CI so we shouldn't be affected.

PR #557, opened 2021-03-21
[OVAs] add calico to list of excluded nics

MaxRink:

This adds calco ifaces to the list of ifaces not reported by vmware-tools as for https://github.com/kubernetes/cloud-provider-vsphere/blob/master/docs/book/known_issues.md

@codenrhoden


MaxRink:
@codenrhoden i have no idea why the azure builds are failing, i havent touched that at all
looing at the log output the vhd-windows-2019 target is the culprit

```
[0;32m    vhd-windows-2019: fatal: [default]: FAILED! => {"changed": false, "exists": false, "msg": "Service 'docker' is not installed, need to set 'path' to create a new service"}[0m
```

PR #563, opened 2021-03-24
Adding Windows Goss support

### LINKED ISSUE #532: Add Windows support for goss
Verification of Windows images using same goss tooling used on the [linux side](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss.yaml).

Verification of the windows builds using goss.  

**Additional context**

The work for enabling this was done and recently merged:
https://github.com/YaleUniversity/packer-provisioner-goss/pull/39
https://github.com/aelsabbahy/goss/pull/633

Note not all features of goss work on windows but can use the commands to do alot of the checking for services and files.

/kind feature
/sig windows

fyi @perithompson

perithompson:
This PR adds the Goss components needed to run Goss checks on windows images built for Azure, AWS and VSphere(OVA), and is based on groundwork done in Goss by @jsturtevant 

Which issue(s) this PR fixes: Fixes #532

**Additional context**
Due to a number of fixes in the packer goss provisioned and the Goss CLI this requires an update to the latest Goss and Packer versions.  The intention here is to provide a starting point to carry on adding more checks are more functionality is added and also to stay aligned to the checks done on Linux nodes.  There are some gaps in feature parity for Goss that we need to work out but I think this lays the groundwork for more additions

/sig windows
/cc @jsturtevant

### DIFF
--- a/images/capi/hack/generate-goss-specs.py
+++ b/images/capi/hack/generate-goss-specs.py
@@ -24,16 +24,16 @@ import sys
 root_path = os.path.abspath(os.path.join(sys.argv[0], '..', '..'))
 
 # Define what OS's are supported on which providers
-builds = {'amazon': ['amazon linux', 'centos', 'flatcar', 'ubuntu'],
-          'azure':  ['centos', 'ubuntu'],
-          'ova': ['centos', 'photon', 'rhel', 'ubuntu']}
+builds = {'amazon': ['amazon linux', 'centos', 'flatcar', 'ubuntu', 'windows'],
+          'azure':  ['centos', 'ubuntu', 'windows'],
+          'ova': ['centos', 'photon', 'rhel', 'ubuntu', 'windows']}
 
-
-def generate_goss(provider, system, versions, dryrun=False, save=False):
-
+def generate_goss(provider, system, versions, runtime, dryrun=False, save=False):
     cmd = ['goss', '-g', 'packer/goss/goss.yaml', '--vars', 'packer/goss/goss-vars.yaml']
     vars = {'OS': system, 'PROVIDER': provider,
             'containerd_version': versions['containerd'],
+            'docker_ee_version': versions['docker'],
+            'distribution_version': versions['os'],
             'kubernetes_version': versions['k8s'],
             'kubernetes_deb_version': versions['k8s_deb'],
             'kubernetes_rpm_version': versions['k8s_rpm'],
@@ -41,11 +41,14 @@ def generate_goss(provider, system, versions, dryrun=False, save=False):
             'kubernetes_cni_version': versions['cni'],
             'kubernetes_cni_deb_version': versions['cni_deb'],
             'kubernetes_cni_rpm_version': versions['cni_rpm'],
-            'kubernetes_cni_source_type': 'pkg'}
+            'kubernetes_cni_source_type': 'pkg',
+            'runtime': runtime,
+            'pause_image': versions['pause']}
+
 
     # Build command
     cmd.extend(['--vars-inline', json.dumps(vars), 'render'])
-    print('generating os: %s, provider: %s' % (system, provider))
+    print('\nGenerating os: %s, provider: %s, runtime: %s' % (system, provider, runtime))
     print(cmd)
 
     # Run command with output going to file
@@ -73,14 +76,14 @@ def main():
         description='Generates GOSS specs. By default, generates all '
                     'possible specs to stdout.',
         usage='%(prog)s [-h] [--provider {amazon,azure,ova}] '
-              '[--os {al2,centos,flatcar,photon,rhel,ubuntu}]')
+              '[--os {al2,centos,flatcar,photon,rhel,ubuntu,windows}]')
     parser.add_argument('--provider',
                         choices=['amazon', 'azure', 'ova'],
                         action='append',
                         default=None,
                         help='One provider. Can be used multiple times')
     parser.add_argument('--os',
-                        choices=['al2', 'centos', 'flatcar', 'photon', 'rhel', 'ubuntu'],
+                        choices=['al2', 'centos', 'flatcar', 'photon', 'rhel', 'ubuntu', 'windows'],
                         action='append',
                         default=None,
                         help='One OS. Can be used multiple times')
@@ -107,6 +110,12 @@ def main():
     containerd = read_json_file(os.path.join(root_path, 'packer', 'config', 'containerd.json'))
     versions['containerd'] = containerd['containerd_version']
 
+    docker = read_json_file(os.path.join(root_path, 'packer', 'config', 'windows', 'docker.json'))
+    versions['docker'] = docker['docker_ee_version']
+
+    common = read_json_file(os.path.join(root_path, 'packer', 'config', 'common.json'))
+    versions['pause'] = common['pause_image']
+
     providers = builds.keys()
     if args.provider is not None:
         providers = args.provider
@@ -120,12 +129,20 @@ def main():
                 oss.append(o)
         oss = list(set(oss))
     oss = [sub.replace('al2', 'amazon linux') for sub in oss]
-
     # Generate spec for each valid permutation
     for provider, system in itertools.product(providers, oss):
         if system in builds[provider]:
-            generate_goss(provider, system, versions, args.dry_run, args.write)
-
+            if system == 'windows':
+                runtimes = ["docker_ee","containerd"]
+                os_versions = ["2019", "2004"]
+            else: 
+                runtimes = ["containerd"]
+                os_versions = [""]
+            for runtime in runtimes:
+                for version in os_versions:
+                    versions["os"] = version
+                    generate_goss(provider, system, versions, runtime, args.dry_run, args.write)
+            
 
 if __name__ == '__main__':
     main()
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -92,6 +92,36 @@
       "type": "windows-restart"
     },
     {
+      "arch": "{{user `goss_arch`}}",
+      "format": "{{user `goss_format`}}",
+      "format_options": "{{user `goss_format_options`}}",
+      "goss_file": "{{user `goss_entry_file`}}",
+      "inspect": "{{user `goss_inspect_mode`}}",
+      "target_os": "Windows",
+      "tests": [
+        "{{user `goss_tests_dir`}}"
+      ],
+      "type": "goss",
+      "url": "{{user `goss_url`}}",
+      "use_sudo": false,
+      "vars_env": {
+        "GOSS_MAX_CONCURRENT": "5",
+        "GOSS_USE_ALPHA": "1"
+      },
+      "vars_file": "{{user `goss_vars_file`}}",
+      "vars_inline": {
+        "OS": "{{user `distribution` | lower}}",
+        "PROVIDER": "ami",
+        "containerd_version": "{{user `containerd_version`}}",
+        "distribution_version": "{{user `distribution_version`}}",
+        "docker_ee_version": "{{user `docker_ee_version`}}",
+        "kubernetes_version": "{{user `kubernetes_semver`}}",
+        "pause_image": "{{user `pause_image`}}",
+        "runtime": "{{user `runtime`}}"
+      },
+      "version": "{{user `goss_version`}}"
+    },
+    {
       "elevated_password": "{{.WinRMPassword}}",
       "elevated_user": "Administrator",
       "script": "packer/ami/scripts/sysprep_prerequisites.ps1",
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -1,4 +1,5 @@
 command:
+{{ if ne .Vars.OS "windows" }} # Linux Only
   containerd --version | awk -F' ' '{print substr($3,2); }':
     exit-status: 0
     stdout: []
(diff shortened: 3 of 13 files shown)


perithompson:
@jsturtevant I thought we could iterate on this together, thank you for getting the hard bits of the goss components and the help you gave me yesterday, this passed on my first run today so I thought it was worth starting a very much WIP PR 😄

images/capi/packer/goss/goss-windows.yaml
```diff
@@ -0,0 +1,16 @@
...
+  kubeadm.exe version -o json | powershell -Command "($input | convertfrom-json).clientVersion.gitVersion":
+    exit-status: 0
+    stdout: [{{ .Vars.kubernetes_version }}]
+    stderr: []
+    timeout: 0
+  kubelet.exe --version | powershell -command "$input".Split(" ")[1]: 
+    exit-status: 0
+    stdout: [{{ .Vars.kubernetes_version }}]
+    stderr: []
+    timeout: 0
```
jsturtevant, line 16: There is a ton of things happening in the existing goss components.  It would be preferable to re-use as much as we can there, though we have to have distinct commands for the windows side due to locations and the need to use powershell. The other part is goss only supports the `command` part of [goss](https://github.com/aelsabbahy/goss/blob/master/docs/platform-feature-parity.md#matrix---testsassertions).  

The question I have is should these be backed into the existing files like https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-command.yaml which would make sure everything is together but add some noise to the file? I am leaning towards yes for the same reason we ended up consolidating some of the windows config files into the linux side: when making an update it is easier to see them and include them.

jsturtevant:
Nice to see this passing:

```
[0;32m    vhd-windows-2019:     "summary": {[0m
[0;32m    vhd-windows-2019:         "failed-count": 0,[0m
[0;32m    vhd-windows-2019:         "summary-line": "Count: 3, Failed: 0, Duration: 2.503s",[0m
[0;32m    vhd-windows-2019:         "test-count": 3,[0m
[0;32m    vhd-windows-2019:         "total-duration": 2502808900[0m
[0;32m    vhd-windows-2019:     }[0m
[0;32m    vhd-windows-2019: }[0m
[1;32m==> vhd-windows-2019: Goss validate ran successfully[0m
[1;32m==> vhd-windows-2019: 
```

images/capi/packer/goss/goss-windows.yaml
perithompson: yeah I agree, I'm trying to work out how much we can reuse but I'm not sure there is much that will just overlap,  We can use an if statement to keep things in the same file but without making it harder to understand for whats already there is going to be tough, I'm thinking for minimum checks we need to check
- binary versions
- cloudbase_metadata
- installed features
- hyper v
- windows updates ( although I think this is going to be HARD! if not impossible)
- services installed and in expected state?
any thoughts?

I adding services to goss for windows would help with overlap of somethings but as most are commands atm that could be tricky?

perithompson:
I've raised https://github.com/YaleUniversity/packer-provisioner-goss/pull/49 to deal with the fact I couldn't get the inline-vars to work with windows, it seems to be something to do with unmarshalling Json on the windows node, there still seems to be a problem with containerd on the path for these tests to pass and I'm not sure if there is a way to have a dynamic %systemdrive% using non "command" tests but starting to get there for a first pass, once the Provisioner Pull is merged. @jsturtevant if you can have a look at the style that would be good particularly the workaround for lack of services in windows

``` yaml
goss-vars.yaml
  common-windows-service:
    cloudbase-init:
      expected:
      - Manual
      - Stopped
    kubelet:
      expected: 
      - Manual 
      - Stopped
      - "/RequiredServices.+:.+containerd/"
    sshd:
      expected:
      - Automatic
      - Running

---
goss-services.yaml
{{ if eq .Vars.OS "windows"}} # Windows
# Workaround until windows services are added to goss
command:
{{range $name, $vers := index .Vars .Vars.OS "common-windows-service"}}
  "Windows Service - {{ $name }}":
    exec: powershell -command "(Get-Service {{ $name }} | select *)"
    exit-status: 0
    stdout: {{range $vers.expected}}
    - {{.}}
    {{end}}
{{end}}
{{range $name, $vers := index .Vars .Vars.OS .Vars.PROVIDER "windows-service"}}
  "Windows Service - {{ $name }}":
    exec: powershell -command "(Get-Service {{ $name }} | select *)"
    exit-status: 0
    stdout: {{range $vers.expected}}
    - {{.}}
    {{end}}
{{end}}
{{end}}
```


also @codenrhoden is there someone who would be interested in checking how I've integrated windows to the existing goss files to see if it looks ok before I go any further?

perithompson:
@jsturtevant Added some timeouts and also removed the non-interactive no profile flag that was slowing the commands down and causing them to fail, I suspect it runs everything concurrently and changing the powershell flags may speed this up?  I think this is good for a base to work on and ready for review go if the tests passed. /cc @codenrhoden

images/capi/packer/ami/packer-windows.json
```diff
@@ -94,6 +94,38 @@
...
+      "type": "goss",
+      "url": "{{user `goss_url`}}",
+      "use_sudo": false,
+      "vars_env": {
+        "GOSS_USE_ALPHA": "1"
+      },
+      "vars_file": "{{user `goss_vars_file`}}",
+      "vars_inline": {
+        "ARCH": "amd64",
+        "OS": "{{user `distribution` | lower}}",
```
kkeshavamurthy, line 116: Is this supposed to be `distribution` or `distro_name`?

images/capi/packer/ami/packer-windows.json
```diff
@@ -94,6 +94,38 @@
...
+      "inspect": "{{user `goss_inspect_mode`}}",
+      "target_os": "Windows",
+      "tests": [
+        "{{user `goss_tests_dir`}}"
+      ],
+      "type": "goss",
+      "url": "{{user `goss_url`}}",
+      "use_sudo": false,
+      "vars_env": {
+        "GOSS_USE_ALPHA": "1"
```
kkeshavamurthy, line 111: Just curious, what does `GOSS_USE_ALPHA` enable?

perithompson: It's required for windows support, you can see it here 😄  [aelsabbahy/goss/docs/platform-feature-parity.md](https://github.com/aelsabbahy/goss/blob/master/docs/platform-feature-parity.md)

images/capi/packer/goss/goss-command.yaml
```diff
@@ -69,4 +71,79 @@ command:
   {{range $key, $val := $vers}}
     {{$key}}: {{$val}}
   {{end}}
+{{end}} 
+
+{{ if eq .Vars.OS "windows" }} # Windows
+  kubectl version:
+    exit-status: 1
```
kkeshavamurthy, line 78: `exit-status: 1` ?

images/capi/packer/ami/packer-windows.json
perithompson: ah probably, ill check that out, azure uses distribution and everything else uses distro_name, need to get that consistent I think, I think you're right that aim uses distro_name

images/capi/packer/goss/goss-command.yaml
perithompson: annoyingly yes.... i think there is probably a way around it but when I run that command I get exit status 1 even though i get a result 🤯 @jsturtevant  I meant to ask you about this

perithompson: I think it's because there is no valid api-server when I run kubectl even though it's a local command

images/capi/packer/config/windows/goss-args.json
```diff
@@ -0,0 +1,11 @@
+{
```
codenrhoden, line 1: I don't see a need for this file since it is identical to `images/capi/packer/config/goss-args.json`. Just point to that one instead. Less duplication!

images/capi/packer/goss/goss-command.yaml
codenrhoden: Does the WIndows binary not have the `--client=true` flag? That's how you get it to not ask/show the version of the API server and leave it as a "local only" command.

images/capi/packer/config/windows/goss-args.json
perithompson: oh oops, yes that was the original pass, must have not removed it!

images/capi/packer/goss/goss-command.yaml
perithompson: I thought i tried that and it didn't make any difference, i will try though, i may not have done it right

jsturtevant: I was thinking the same about `--client` flag.  There are also some strange behavior in powershell if you read the error codes wrong.  If the client flag doesn't work we can look into how goss handles the return codes.

images/capi/packer/ami/packer-windows.json
perithompson: Nope, seems that AMIs and Azure use distribution, I copied the windows OS in this file from the packer.json in the AMI provider as well, maybe this is something we should get an issue to tidy up?

images/capi/packer/goss/goss-command.yaml
perithompson: nope it was just me, not sure what i was doing wrong before but that's cleaned it up! thanks all!

perithompson:
Resolved all the comments, and the tests look good too 😄

codenrhoden:
Nice work @perithompson. I'd like to get https://github.com/kubernetes-sigs/image-builder/pull/585 in before merging this one, just to make sure that adding in the Windows tests doesn't break the templates for the Linux tests. That new CI check makes sure that we can generate/populate the GOSS templates for each provider/OS we build. Take a look at it and see what you think.

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -175,3 +181,108 @@ photon:
...
+        exists: true
+        filetype: file
+        contains:
+        - "!/logging_serial_port=COM1,115200,N,8/"
+        - "metadata_services=cloudbaseinit.metadata.services.ec2service.EC2Service"
+        - "cloudbaseinit.plugins.common.ephemeraldisk.EphemeralDiskPlugin"
+        - "cloudbaseinit.plugins.common.mtu.MTUPlugin"
+        - "cloudbaseinit.plugins.common.sethostname.SetHostNamePlugin"
+  
+
```
jsturtevant, line 282: nit: lots of extra white space

jsturtevant (commented):
Thanks for adding this!  lgtm other than one minor formatting issue

perithompson:
@codenrhoden I've added the windows templating test support but the prow output seems to jumble everything together so it looks like Linux is checking windows and vice version?  seems good if you run it on your workstation 👍

perithompson:
any further comments or does this look good now @codenrhoden @kkeshavamurthy @jsturtevant

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -177,3 +183,114 @@ ubuntu:
       cloud-guest-utils:
       cloud-initramfs-copymods:
       cloud-initramfs-dyn-netconf:
+photon:
```
codenrhoden, line 186: This `photon` section is a duplicate (check out line 123 in your file).

images/capi/hack/generate-goss-specs.py
```diff
@@ -24,28 +24,36 @@
...
-
-
-def generate_goss(provider, system, versions, dryrun=False, save=False):
-
+builds = {'amazon': ['amazon linux', 'centos', 'flatcar', 'ubuntu', 'windows-2019', 'windows-2004'],
+          'azure':  ['centos', 'ubuntu', 'windows-2019', 'windows-2004'],
+          'ova': ['centos', 'photon', 'rhel', 'ubuntu', 'windows-2019', 'windows-2004']}
+
+def generate_goss(provider, system, versions, runtime, dryrun=False, save=False):
+    if 'windows' in system:
```
codenrhoden, line 32: I think there's a general simplification that can be made here... Just like we don't distinguish between Ubuntu 18.04 and Ubuntu 20.04 (there's nothing release specific in `goss-args`), I don't see anything release specific for Windows. I think in the `builds` map, you can just add `windows1`. THen this can become `if system == 'windows'`, and additional simplifications can be made.

images/capi/hack/generate-goss-specs.py
```diff
@@ -73,14 +81,14 @@ def main():
...
-              '[--os {al2,centos,flatcar,photon,rhel,ubuntu}]')
+              '[--os {al2,centos,flatcar,photon,rhel,ubuntu,windows}]')
     parser.add_argument('--provider',
                         choices=['amazon', 'azure', 'ova'],
                         action='append',
                         default=None,
                         help='One provider. Can be used multiple times')
     parser.add_argument('--os',
-                        choices=['al2', 'centos', 'flatcar', 'photon', 'rhel', 'ubuntu'],
+                        choices=['al2', 'centos', 'flatcar', 'photon', 'rhel', 'ubuntu', 'windows-2019', 'windows-2004'],
```
codenrhoden, line 91: Similarly, this can just add `'windows'`. If there is a good reason *not* to take that approach, this list should match the `usage` description above, and right now it doesn't.

codenrhoden (changes requested):
Great work @perithompson! I'm glad to see this get added, and it seems like you had no issues adding in the windows stuff to that Python script. Just a couple minor things I saw.

images/capi/hack/generate-goss-specs.py
perithompson: I only have one `windows version` check which is 

``` yaml
{{ if eq .Vars.distribution_version "2019" }}
  Windows build version is high enough:
    exit-status: 0
    exec: powershell -command "(Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion' -Name UBR).UBR -ge 1817"
    stdout:
    - "True"
    timeout: 30000
{{end}}
```

I'll have a think how i could rework this to a single windows but I did try before, it just made the code look a bit messy

perithompson: Ok ive reworked this, what do you think @codenrhoden @jsturtevant

jsturtevant: lgtm but i have less experience in the goss area

perithompson:
@codenrhoden anything else needed on this?

codenrhoden:
/approve
This looks good to me. I was able to look at the test output and see that it is generating all the expected permutations. nice work!
@kkeshavamurthy care to take a look for an LGTM?
I would like to see if this can be improved in the future in CI to log the output a bit better, such that each GOSS output immediately follows the command instead of all being jumbled together...

kkeshavamurthy:
Thanks for working on this @perithompson 
/lgtm

PR #564, opened 2021-03-25
Fixed logo image URLs used in Azure Marketplace

mboersma:

When creating the recent set of Kubernetes patch releases, this image-builder repository couldn't be synced because Azure's credential scanner flagged these verbose image URLs as containing credentials. After working around that locally, the submission still failed because the expected logo sizes have changed (there is no "wide" version any more, and the other three got bigger).

@CecileRobertMichon and I created correctly sized versions of the "turtles all the way down" logo and stored them in the same account used to build the Azure reference images. This should solve both problems.


nader-ziada:
I don't I have approval in this repo,

/lgtm
/approve

PR #566, opened 2021-03-26
Include Windows hns fix for WS 2019 - 2c+ (Feb 2021 feature update)

jsturtevant:
We found a bug in HNS Windows component when multiple LB policies are included for same endpoint then HNS hangs.  The only solution is to restart HNS.

This sets a registry key that causes HNS to throw an error instead of hanging.

**Additional context**

This was first reported in aks and aks-engine: https://github.com/Azure/aks-engine/pull/4315

/sig windows
/cc @perithompson

### DIFF
--- a/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
@@ -104,6 +104,16 @@
     start_mode: auto
     state: started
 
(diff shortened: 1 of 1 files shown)


perithompson:
I didnt see the original issue but def need to add this after reviewing the issue! Thanks @jsturtevant 

/lgtm

perithompson (approved):

Issue #575, opened 2021-04-05
Image builder on Windows Evaluation ISOs, version selection + programdata/ssh/sshd_config setup


jayunit100:
It seems like there is an issue wherein the windows evaluation ISO has 4 different choices as for the OS:

1) There are four different versions of windows server 2019 ~ so one has to be selected ~ which ones the `sane` default (and should that be put in autoattend.xml ? 

- Windows Server 2019 Standard, 
- Windows Server 2019 Standard (Desktop Experience), 
- Windows Server 2019 Datacenter, 
- Windows Server 2019 Datacenter (Desktop Experience)

2) When i pick one of these, it looks like the `ssh` setup step fails.  Is there a diference between the  way 
-  windows server Eval versions vs
-  windows server MSDN versions 
(thats i think what @perithompson  usually uses) , work in term of the default ssh key? 

[...]

PR #581, opened 2021-04-06
Improve unattend behaviour in certain failure situations and add unattend instructions

jayunit100:
Add unattend notes for windows image builder eval

### DIFF
--- a/images/capi/ansible/windows/roles/kubernetes/tasks/url.yml
+++ b/images/capi/ansible/windows/roles/kubernetes/tasks/url.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- name: Get kubernetes binaries
+- name: Download kubernetes binaries
   win_get_url:
     url: "{{ kubernetes_base_url }}/{{ item }}.exe"
     dest: '{{ kubernetes_install_path }}\'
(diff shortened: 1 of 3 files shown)


jayunit100:
Part of an iterative improvement to #575 ...  this way theres some guidance on what to do for product keys?

codenrhoden:
A couple of questions...

Is the hard coded product key a dummy key or is it real? meaning, something you wouldn't want checked in. :)

I haven't looked closely at the windows builds, but is the autoattend.xml file location configurable? Meaning can you tell image-builder to point to a different one? I ask because the notes say "you will probably want to modify this file". In  general, we never want a user to modify a file that is part of the git repo -- that produces a dirty repo and will show up in build metadata (may not be true for windows builds at this time, I'm not sure). Anything that needs to be changed should be done via build parameters.

And finally, this should be squashed to a single commit.

jayunit100:
if i remove the *users will want to modify* comment, are we all good? 

was hoping to just get this in as a first iteration on making the configuration of autoattend.xml more consumable for mortals :)

jayunit100:
I added the time zone stuff
also added an updated to the `Download Kubernetes binaries` step so its named identically to the linux equivalent step

codenrhoden:
> if i remove the _users will want to modify_ comment, are we all good?
> 
If it is the current state of things, better to document actual expectations rather than an ideal that doesn't exist yet.

/lgtm
/approve

PR #582, opened 2021-04-06
Add burrito server and README

### LINKED ISSUE #571: imgpkg / bundle  / http facade awareness for image builds

For windows nodes, you need to build an OVA from scratch - and can't borrow OVAs from elsewhere.    Thus anyone using CAPI w/ windows need to use image builder to create their images, for their nodes.

This means that the UX around how binaries are pulled into image builder needs to be dumbed down a little, since its not just cloud providers building image builder into their internal workflows - but rather - end users wanting to get a cluster up quickly.

*problem* :when we build windows images, we pull artifacts from a URL, we force users to *host* their windows kubelet/kubeproxy/containerd/hcsshim/... from an endpoint.  USers may not have a good way to host this stuff or may not have an internal process that allows for this.  

- ... One really clean and scalable way to host files as docker images is https://carvel.dev/imgpkg/, which suggests an automatic mechanism for creating a hosted package of imagbuilder windows binaries.... 
- another option would be to make imagebuilder capable of reading files from a local disk somehow, using something like scp to put them onto an image

However, imgpkg doesnt "serve" the files on an endpoint, but rather, directly copies them as a folder to a destination.

So, i  would like a way to consume from files, rather then a url.... 

1) Can we make it so that an `imgpkg` aware hosting command in image builder, i.e. one which runs:

```
imgpkg pull -i index.docker.io/periismyfriend/windowz:v1 --output /tmp/my-windows-exes
python3 -m http.server
```

for users, that way we can give folks a very clean image building experience, wherein they only need to provide a docker registry of a `imgpkg` repo as the input to image builder.  This would mean users dont need to host there own windows executables on a port... instead, the VM that is building the image could just pull down its bits from the imagebuilder node.

2) IF we're not able to bundle imgpkg awareness, can we modify the way we invoke imgbuilder so that it copys executables from its local disk to the ova somehow , rather then doing a pull from a URL ? 

Make a third artifact that connects imgpkg artifacts with imagebuilder execution, i.e. a init container which ran image builder in a pod, pulled the artifacts into a PV/hostpath , and invoked image builder on an endpoint .  this however would only work in k8s 

**Additional context**

/kind feature

jayunit100:
This mostly  allows you to take an imgpkg bundle, find the artifacts in it, and host them on an endpoint .  It has no external dependencies, just go run it.    It'll reverse engineer the 'base_url" fields by finding the root directory, thus it allows any arbitrary packaging of binaries to be surfaced to output an image builder compatible input  file... 

Output:

```
jayunit100@kcp-1:~/SOURCE/image-builder/images/capi/hack$ go run serve_artifacts.go 
kubernetes_base_url: http://127.0.0.1:8080/my/k8s
cloudbase_init_url: http://127.0.0.1:8080/**none**
wins_url: http://127.0.0.1:8080/**none**
nssm_url: http://127.0.0.1:8080/**none**
```
assuming @perithompson will have fun with this tomorrow :) :) :) 
Fixes #571

### DIFF
--- a/images/capi/hack/README.md
+++ b/images/capi/hack/README.md
@@ -1,0 +1,21 @@
+# Hack
+
+This directory has a collection of scripts that dont belong elsewhere and may go away some day.
+
+## serve-artifacts.go
+
+This script finds all relevant k8s artifacts (i.e. like the kubelet.exe) in a directory , serves them on an endpoint.
+
+It prints out the values corresponding to this endpoint so you can use it in input to image builder, i.e.
+
+
+```
+jayunit100@kcp-1:~/SOURCE/image-builder/images/capi/hack$ go run serve_artifacts.go 
+kubernetes_base_url: http://127.0.0.1:8080/my/k8s
+cloudbase_init_url: http://127.0.0.1:8080/**none**
+wins_url: http://127.0.0.1:8080/**none**
+nssm_url: http://127.0.0.1:8080/**none**
+``` 
+
+It has no dependencies, and this is intentional - it is just an example script for surfacing artifacts to image builder,
+to be adopted by vendors as needed
--- a/images/capi/hack/serve_artifacts.go
+++ b/images/capi/hack/serve_artifacts.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"path/filepath"
+	"fmt"
+	"strings"
+)
+
+var (
+	dir     = flag.String("b", "", "base directory where kubelet, kube-proxy, containerd, and other binaries live")
+	port    = flag.String("p", "8080", "port to serve files on")
+	verbose = flag.Bool("v", false, "verbose logging")
+	ip = flag.String("ip", "127.0.0.1", "ip address to print for url default")
+)
+
+// This will be printed out as a YAML file that can serve as the input to the image
+// builder process...
+
+func main() {
+	artifacts := map[string]string {
+	   "kubelet.exe": "kubernetes_base_url",
+	   //"kubernetes_base_url": "https://storage.googleapis.com/kubernetes-release/release/v1.19.2/bin/windows/amd64",
+	   "CloudbaseInit*": "cloudbase_init_url",  
+	   // "cloudbase_init_url": "https://github.com/cloudbase/cloudbase-init/releases/download/1.1.2/CloudbaseInitSetup_1_1_2_x64.msi",
+	   "wins.exe": "wins_url", //: "https://github.com/rancher/wins/releases/download/v0.0.4/wins.exe",
+	   "nssm.exe": "nssm_url",
+	   // "nssm_url": "https://azurek8scishared.blob.core.windows.net/nssm/nssm.exe",
+	   //"additional_debug_files": "https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/hack/DebugWindowsNode.ps1",
+	}
+
+	flag.Parse()
+
+	if _, err := strconv.Atoi(*port); err != nil {
+		log.Fatal("port provided must be a valid integer")
+	}
+
+	curr, err := os.Getwd()
+	if err != nil {
+		log.Fatal("os.Getwd(): ", err)
+	}
+
+	if len(*dir) > 0 {
+		if _, err := os.Stat(*dir); err != nil {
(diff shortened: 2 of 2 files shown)


nikhita:
creating a (dummy) approve comment to re-trigger CI, the approval mechanism was broken for a bit - kubernetes/test-infra#21687

/approve

perithompson:
Is your idea to copy this to the host while it is building and do something with it? The host running packer builder wouldn't always be available to the host being built so just trying to get an idea of where you'd run this, if you run in k8s the host wouldn't know the service IP to talk to the pod, would it?

jayunit100:
This PR isn't proscriptive as far as where it is run, but rather, it provided a canonical impl to reduce bikeshedding for vendors to borrow .

Possible uses:

1) run it directly on your image builder node 

2) run it in a side car pod that w or wo go runtime (compile it first) 

3) run it as part of a burrito that self hosts its own binaries , and then run THaT container as a pod in a cluster on hostNetwork

perithompson:
I'm having to use something similar to this already so would be good, to get something in to do the same for someone else, @codenrhoden have you got any thoughts on this?

jayunit100:
ive finished testing this w/ a dummy image builder, works for me.

ravisantoshgudimetla:
> I'm having to use something similar to this already so would be good, to get something in to do the same for someone else, @codenrhoden have you got any thoughts on this?

+1. Having run image builder in the past, I agree with @perithompson's comments. Unless someone else has a strong objection to merge this, I find this useful.

jayunit100:
thanks folks, sounds like consensus - can we merge + iterate on this ?

codenrhoden:
/lgtm

Going to wait a bit and see if any other comments come in.

images/capi/hack/README.md
```diff
@@ -0,0 +1,22 @@
+# Hack
+
+This directory has a collection of scripts that dont belong elsewhere and may go away some day.
+
+## serve-artifacts.go
+
+This script finds all relevant k8s artifacts (i.e. like the kubelet.exe) in a directory , serves them on an endpoint.
+
+It print out the YAML definitions corresponding to this endpoint so you can use it in input to the
```
jsturtevant, line 9:
```suggestion
It prints out the YAML definitions corresponding to this endpoint so you can use it in input to image builder, i.e.
```

images/capi/hack/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+This directory has a collection of scripts that dont belong elsewhere and may go away some day.
+
+## serve-artifacts.go
+
+This script finds all relevant k8s artifacts (i.e. like the kubelet.exe) in a directory , serves them on an endpoint.
+
+It print out the YAML definitions corresponding to this endpoint so you can use it in input to the 
+
+image builder, i.e.
```
jsturtevant, line 11:
```suggestion

```

images/capi/hack/serve_artifacts.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	dir     = flag.String("b", "", "base directory where kubelet, kube-proxy, containerd, and other binaries live")
+	port    = flag.String("p", "8080", "port to serve files on")
+	verbose = flag.Bool("v", false, "verbose logging")
+	ip = flag.String("ip", "127.0.0.1", "ip address to print for url default")
+)
+
+// This will be printed out as a YAML file that can serve as the input to the image
```
jsturtevant, line 21: These yamls would be the inputs to the ansible?  

In the case of using the make file I think it would need to be json?  It would be slick to be able to run this then use `PACKER_VAR_FILES=localserver.json make image-name`

jsturtevant (commented):
thanks, this is useful for local testing!

images/capi/hack/serve_artifacts.go
jayunit100: +1 ill make a follow on, actually we dont literally print YAML, so i updated the comments

jayunit100: made a follow on for this here https://github.com/kubernetes-sigs/image-builder/issues/593

codenrhoden:
Thanks for writing the follow-up issue, Jay. And thanks for calling out that it is more useful to have JSON output that can be consumed as-is James.

I'm ready to merge this, but I don't think it needs to be 6 commits. Please squash.

codenrhoden:
@jayunit100 Once this gets squashed, I'll merge it in.

PR #583, opened 2021-04-08
Change kubelet service state

SanikaGawhane:
Using state: restarted caused flaky behavior in kubelet and containerd service.

/cc @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/main.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/main.yml
@@ -42,7 +42,6 @@
     name: kubelet
     daemon_reload: yes
     enabled: True
-    state: restarted
 
 - name: Create the Kubernetes version file
   template:


images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -143,7 +143,7 @@
     name: containerd
     daemon_reload: yes
     enabled: True
-    state: restarted
+    state: enabled
```
detiber, line 146: I don't think enabled is sufficient here, since later tasks that would attempt to use containerd to pre-pull images would fail if containerd is not started.

codenrhoden:
@SanikaGawhane Were you able to recreate/witness the flakiness around this? I'm going solely off of memory here, but I think we had seen `kubelet` be flaky, not `containerd`.

It's also worth noting that `enabled` is not a valid option for `state`: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/systemd_module.html#parameter-state
If we truly don't care about the `state`, and instead just want to make sure it's `enabled` (which it is, and as @detiber pointed out, we need this one running), then I think the correct thing to do is to remove the `state` line entirely. It's not required.

SanikaGawhane:
@codenrhoden I've witnessed flakiness in kubelet service, not containerd though. I think we can only edit kubelet service state here.
I agree to @detiber's and your suggestion of removing `state` altogether. Thank you.

codenrhoden:
Yeah, my take here is that we don't want to be starting (via "restarted") the `kubelet` service at this stage of things, only ensure that it is enabled. We have seen flakiness in tests where restarting the service and checking to see if it is running is flaky. Normally, we are relying on additional files and config to be put in place (via cloud-init) from CAPI providers, so I think even checking that the service starts here is a false sense of security. *if* it is successful it doesn't indicate at all that it will be successful on the next boot with a different config in place.

@detiber Do you have thoughts on this one? I am in favor of this change.

detiber:
> Yeah, my take here is that we don't want to be starting (via "restarted") the `kubelet` service at this stage of things, only ensure that it is enabled. We have seen flakiness in tests where restarting the service and checking to see if it is running is flaky. Normally, we are relying on additional files and config to be put in place (via cloud-init) from CAPI providers, so I think even checking that the service starts here is a false sense of security. _if_ it is successful it doesn't indicate at all that it will be successful on the next boot with a different config in place.

+1, I highly suspect that the flakiness is related to the fact that the service is crashlooping at this point, since there isn't a valid configuration in place yet.

PR #584, opened 2021-04-08
Clean up and *use* the ansible.cfg file

codenrhoden:
The `ansible.cfg` file that is present in `images/capi/ansible` is not used
in any way because the Ansible commands won't pick it up in that
location. Since Ansible is run with a PWD of `images/cap`i, put it there
instead.

Additionally, the .cfg file mistakenly had 'default' instead of
'defaults', meaning that all entries were being ignored. We no longer
need to specify a search path for filter plugins since we haven't used
any in a while, and the setting for retry files is already the default
in Ansible.

**Additional context**
This makes what is being sought in #531 a *lot* easier (I tested it, it works, but wanted to leave that PR separate).

/hold

With this change in place, SSH Pipeline **will** be used. This is a change that can significantly speed up builds. I've got the `/hold` in place because we at least need to make sure that every OS we currently build works with this change. If any OS has `requiretty` set by default in `/etc/sudoers`, it won't work. This is becoming less common, but I have no idea who sets it and who doesn't at this point.

/assign @kkeshavamurthy @detiber @CecileRobertMichon 
if anyone does get a chance to take this for a spin, mind putting in a note what OS/provider combintation you tried to make sure the SSH pipelining worked?

### DIFF
--- a/images/capi/.dockerignore
+++ b/images/capi/.dockerignore
@@ -3,6 +3,7 @@
 
 # Exceptions
 !ansible
+!ansible.cfg
 !cloudinit
 !hack
 !packer
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -28,6 +28,7 @@ USER imagebuilder
 WORKDIR /home/imagebuilder/
 
 COPY --chown=imagebuilder:imagebuilder ansible ansible/
+COPY --chown=imagebuilder:imagebuilder ansible.cfg ansible.cfg
 COPY --chown=imagebuilder:imagebuilder cloudinit cloudinit/
 COPY --chown=imagebuilder:imagebuilder hack hack/
 COPY --chown=imagebuilder:imagebuilder packer packer/
--- a/images/capi/ansible.cfg
+++ b/images/capi/ansible.cfg
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-[default]
-remote_tmp = /tmp/.ansible/
-filter_plugins = ./filter_plugins
-retry_files_enabled = False
+[defaults]
+remote_tmp = /tmp/.ansible
 
 [ssh_connection]
-pipelining = True
+pipelining = False
--- a/images/capi/ansible/playbook.retry
+++ b/images/capi/ansible/playbook.retry
@@ -1,1 +1,0 @@
-default
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -75,9 +75,6 @@
   ],
   "provisioners": [
     {
-      "ansible_env_vars": [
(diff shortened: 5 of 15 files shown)


codenrhoden:
### Tested
OVAs
- [ ] Photon 3
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04

AMIs
- [ ] Amazon 2
- [ ] CentOS 7
- [ ] Flatcar
- [ ] Ubuntu  18.04
- [ ] Ubuntu 20.04
- [ ] Windows 2004
- [ ] Windows 2019

Azure
- [ ] CentOS 7
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] Windows 2004
- [ ] Windows 2019

I don't think that testing Openstack/RAW separate from OVAs is necessary since they start with the same ISO they will have the same default `/etc/sudoers` settings. Main question is whether we also need to test Digital Ocean or GCE. How likely is it that those have a different setting in their source template OS within `/etc/sudoers`? I have no idea.

codenrhoden:
I'm also fairly confident this doesn't affect Windows, since the issue is about TTY and `sudo`....

codenrhoden:
Both CentOS and AL2 on Amazon failed with:

```
    centos-7: fatal: [default]: UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo /tmp/.ansible/ `\"&& mkdir \"` echo /tmp/.ansible/ansible-tmp-1617918862.6195111-29036-271485260635057 `\" && echo ansible-tmp-1617918862.6195111-29036-271485260635057=\"` echo /tmp/.ansible/ansible-tmp-1617918862.6195111-29036-271485260635057 `\" ), exited with result 1", "unreachable": true}
```

This is from the the `remote_tmp = /tmp/.ansible/` line in the config. The default is `~/.ansible/tmp`, and since this config file was *not* being used before, I see no reason to change it. I'm going to remove that line.

codenrhoden:
Wow, this opened a can of worms on my end. Starting to piece together all effects that are happening, and i think this will turn into a meaningiful cleanup. But it was more impactful than I suspected. Stay tuned.

codenrhoden:
After playing with this for a bit, I figured out that enabling SSH pipelining is likely to take more work. So, in the interest of time, this just places the `ansible.cfg` file in the correct place and maintains all the settings that we were _already_ running. So this does not introduce any change other than having the `ansible.cfg` file be read, which will make things like #531 a lot easier.

/hold cancel
/assign @kkeshavamurthy

kkeshavamurthy:
Thanks for fixing this @codenrhoden . 
/lgtm

PR #586, opened 2021-04-10
Allow CI tests to be called from images/capi/scripts dir

codenrhoden:
All CI scripts are being relocated into images/capi/scripts. This commit creates symlinks to those that can be symlink'd easily, and for the Azure CI test, creates an intermediate script file that calls the original. Once CI jobs are updated to reference new locations, files will be moved for real.

**Additional context**

### DIFF
--- a/images/capi/scripts/ci-azure-e2e.sh
+++ b/images/capi/scripts/ci-azure-e2e.sh
@@ -16,28 +16,7 @@
 
 ###############################################################################
 
-set -o errexit
-set -o nounset
-set -o pipefail
-
-[[ -n ${DEBUG:-} ]] && set -o xtrace
-
 CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
-cleanup() {
-    returnCode="$?"
-    exit "${returnCode}"
-}
-
-trap cleanup EXIT
-
-json_files=$(find . -type f -name "*.json" | sort -u)
-for f in ${json_files}
-do
-  if ! diff <(jq -S . ${f}) ${f} >> /dev/null; then
-    echo "json files are not sorted!! Please sort them with \"make json-sort\" in \"images/capi\" before commit"
-    echo "Unsorted file: ${f}"
-    exit 1
-  fi
-done
+packer/azure/scripts/ci-azure-e2e.sh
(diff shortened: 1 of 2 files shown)


images/capi/hack/ci-json-sort.sh
```diff
@@ -22,7 +22,7 @@ set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/../../..
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
```
codenrhoden, line 25: I'm not sure why this was doing `../../..` before, as that puts it at the repo root. This is meant to just look in `images/capi`, and that's why I changed this. Otherwise this picked up unsorted files in the docs.

images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+###############################################################################
+
+CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+cd "${CAPI_ROOT}" || exit 1
+
+packer/azure/scripts/ci-azure-e2e.sh
```
codenrhoden, line 22: I tried to do this file as a symlink, but was unable get passed issues with `dirname` if I did. If you called the script through the symlink, it would pick up the location of that file instead of file being pointed to. I even tried modifying the source script to use `readlink`, but no luck. I figured this was just easier -- not worth the time to come up with an elegant solution when this file is going away in the next round of commits anyways.

codenrhoden:
/assign @CecileRobertMichon 

once tis goes in, Prow can be flipped over to use the new locations with https://github.com/kubernetes/test-infra/pull/21753

images/capi/scripts/ci-azure-e2e.sh
CecileRobertMichon: can you please update the PR description & title to reflect that?

codenrhoden: Updated!

PR #590, opened 2021-04-12
Relocate CI scripts

codenrhoden:

Move existing CI scripts into `images/capi/scripts`.

**Additional context**
The GitHub diff makes it look like the old files are deleted instead of being moved. Not entirely sure why. I purposely have this as two separate commits (1st deletes the links/pointers, 2nd is a `git mv`) so that git history for these files is preserved. If I do a `git blame` on them after these commits, it looks correct to me.

In hindsight, I think the better way to do this would have been to `git mv` the files to begin with and then create symlinks from the old location to the new, rather than the other way around. Then the second commit would just delete the links. Lesson learned.

/hold
^ waiting for https://github.com/kubernetes/test-infra/pull/21753
/assign @CecileRobertMichon


CecileRobertMichon:
/retest

https://github.com/kubernetes/test-infra/pull/21753 merged

PR #591, opened 2021-04-14
Custom role names comma delimeter

RageZBla:

This PR is required because, we found out during our testing that second ansible roles is being ignored when using `custom_role_names` delimited by spaces. I think this most likely due to the way packer is passing the variable values to ansible. 

Anyway it is probably easier to use commas as delimiters.

### DIFF
--- a/docs/book/src/capi/capi.md
+++ b/docs/book/src/capi/capi.md
@@ -48,7 +48,7 @@ Several variables can be used to customize the image build.
 | Variable | Description | Default |
 |----------|-------------|---------|
 | `custom_role` | If set to `"true"`, this will cause `image-builder` to run a custom Ansible role right before the `sysprep` role to allow for further customization. | `"false"` |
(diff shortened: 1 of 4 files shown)


kkeshavamurthy:
I've run into this before as well. Thanks for the PR @RageZBla .

RageZBla:
I signed it

RageZBla:
@kkeshavamurthy thanks for the super quick review 👍

codenrhoden:
I need to go back and look at some of the history around this one. At one point, things bounced back and forth between comma delimited and space delimited, but I can't remember why we ultimately landed on space delimited.

I do know that if we are going to make a change, we need to be consistent across *everything*, and this change would only change it for one item, which to me makes things more confusing.

Adding to this problem is that when I look at the docs, there is already one item (`no_proxy`) that is comma delimited, but others are space delimited. Ugh.

I do wonder if the problem you are hitting can be fixed with correct escaping -- I know that I've run multiple ansible roles many times without an issue, but you do need to escape like in [this example](https://image-builder.sigs.k8s.io/capi/capi.html#passing-in-extra-packages-to-the-image).

kkeshavamurthy:
You are right @codenrhoden. The space needs to be escaped properly and it works.

PR #592, opened 2021-04-14
WIP: Added ability to set timezone in auto unattend

### LINKED ISSUE #587: (OVA/Windows) IP address not reported by VMware Tools in vSphere due to timezone mismatch

- Using a vSphere server that is **not** on "Pacific Standard Time" 
- Follow the usual steps to build a Windows Server 2019 node on vSphere i.e. `make build-node-ova-vsphere-windows-2019`
- Packer waits at the following message, while the temporary VM awaits instruction: 
```
==> vsphere: Set boot order temporary...
==> vsphere: Power on VM...
==> vsphere: Waiting for IP...
```
- vSphere console indicates the "VMware Tools is not installed on this virtual machine." 

- The temporary VM communicates the IP address to vSphere
- Packer continues with the following steps over WinRM

After much investigation, it was determined that if the temporary VM and vSphere are on different timezones, this causes VMCI to fail to start on Windows.

The fix is to set the <TimeZone /> element in autounattend.xml to match the vSphere build environment timezone. Ref: https://communities.vmware.com/t5/Virtual-Machine-Guest-OS-and-VM/Unattended-VMtools-failing-on-Windows-Enterprise-1909/m-p/2822266/highlight/true#M48815 

Suggested fix in image-builder:
- Parameterize the `<TimeZone>` element within `packer/ova/windows-*/autounattend.xml` with new packer var TBD
- Add a default of "Pacific Standard Time" to avoid a breaking change and allow this to be overridden (e.g. with "GMT Standard Time" for those of us building in the UK)

(PR incoming for the above)

/kind bug
/sig windows
/assign

perithompson:
This patch fixes an issue where the IP address is not picked up by vmware tools

Which issue(s) this PR fixes: Fixes #587 

**Additional context**
@jsturtevant @stuartpreston @codenrhoden I am aware this is verging on a hack but I wanted to put this up here to see if anyone had any better suggestions that work for replacing XML in the unattend before packer runs the installer


stuartpreston:
Thanks @perithompson - this is much better than my hack involving sed and doesn't appear to introduce any new dependencies to image builder. Short of putting in a bigger system to manage more of these files before upload I think this is a defendable hack for now! Will let others comment on the actual implementation.

perithompson:
Cheers @stuartpreston, I agree, I've tried to at least make this extensible so that we can add product key or other things to it at a later time, I'm just concerned that its a little messy so hoping @codenrhoden or @jsturtevant has some ideas or at least think this is an ok approach to?

images/capi/hack/windows-ova-unattend.py
```diff
@@ -0,0 +1,82 @@
...
+import json
+
+def set_xmlstring(root, location, key, value):
+  setting = root.find(location)
+  setting.find(key).text = value
+  return setting
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Uploads an OVA created from a Packer build")
```
jsturtevant, line 30: This looks like it just modifies an ova script?  Is this copy-pasta?

images/capi/Makefile
```diff
@@ -300,6 +302,8 @@ $(NODE_OVA_ESX_BUILD_TARGETS): deps-ova
 
 .PHONY: $(NODE_OVA_VSPHERE_BUILD_TARGETS)
 $(NODE_OVA_VSPHERE_BUILD_TARGETS): deps-ova
+	$(if $(findstring windows,$@),packer build $(PACKER_WINDOWS_NODE_FLAGS) -var-file="packer/ova/packer-common.json" -var-file="$(abspath packer/ova/$(subst build-node-ova-vsphere-,,$@).json)" -only=file $(ABSOLUTE_PACKER_VAR_FILES) packer/ova/packer-windows.json,)
```
jsturtevant, line 305: What does this line do?

jsturtevant (commented):
If I understand correctly, the idea is to patch the unattended file?  I would guess there are other fields that might want to be edited in the unattended eventually?  Doesn't look like there are existing tools for this type modification the answerfiles on the file.  Did you consider allowing folks to provide there own answer files through configuration?

images/capi/hack/windows-ova-unattend.py
perithompson: Yeah just a copy and paste, was trying to be consistent but that is to consistent 🤣

images/capi/Makefile
perithompson: It writes a JSON file with interpolated variables.... And skips if not windows.... Not ideal obviously would need lots of commenting here!

perithompson:
I think there are a few fields that would be good to replace but all of this is a bit of a challenge, I considered that for sure, maybe I could lump all of the scripts into one file to make it easier to sub these out but I haven't thought of a good way to handle that either 🙁

codenrhoden:
My main thought here is for the build scenario, an `autounattend` file is required, correct? If that's the case, I think it is reasonable to expect that this file might need several customizations for a given environment, and thus you might want to pass your own. The version in the repo could be pointed to by default, and serve as a decent template, but I'd rather be able to pass a variable that points to my own version of the file to override it rather than having to add code every time a new use case is uncovered for a different entry to be modified.

Thoughts on that?

perithompson:
I quite like the idea of providing my own but there is some commands that need to be run, perhaps I should replace the steps with just one pressed script then its easier to bring your own... @stuartpreston any thoughts?

perithompson:
We are still seeing this bug around the timezone in vsphere so I would like to get this closer to merging, do we think we could add this functionality and then perhaps I can make a new PR to allow users to swap out the autoattend if they would like to?
@jsturtevant @codenrhoden if we can agree that I can merge this kind of change I will fix up this PR ready to merge

codenrhoden:
I feel like this is a bit hacky still, but I at least understand what it is doing.
If this gets you unblocked. I'm okay with it. @jsturtevant Anything to add?

images/capi/hack/windows-ova-unattend.py
```diff
@@ -0,0 +1,82 @@
...
+      print("windows-ova-unattend: Updating %s ..." % args.unattend_file)
+      unattend.write(args.unattend_file)
+    
+    if modified == 0:
+      print("windows-ova-unattend: skipping...")
+
+if __name__ == "__main__":
+    main()
+
+
```
jsturtevant, line 82: nit: extra line

images/capi/Makefile
jsturtevant: I see so this generated a json file, which is then used by the python script to update the values, then packer is used to build the image.

Can you add a comment here and  the other place above?

jsturtevant (commented):
Agree its a bit different but I don't know of another way. 

Only request is to leave a pointer in the Makefile for anyone reading why packer build is called twice.

perithompson:
found an error after my last rebase
/hold

images/capi/Makefile
```diff
@@ -110,7 +110,6 @@ IMAGE_NAME ?= cluster-node-image-builder
 CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= dev
 ARCH ?= amd64
-BASE_IMAGE ?= docker.io/library/ubuntu:focal
```
codenrhoden, line 113: I don't think this change (and the corresponding docker build stuff further down) were intended. I'm guessing also an issue after your rebase. We definitely want to keep there here. :)

codenrhoden (changes requested):
I think once the docker build stuff gets revertedm this is ready to go? Sounds like there is general agreement to take it in.

codenrhoden:
Somehow this had fallen off my radar. Apologies @perithompson.

I see the Dockerfile changes have been taken out, so I think this is ready now? Just want to double-check one more time since it's been a few weeks. I also know there is that other fix that depends on this.

perithompson:
no problem thanks @codenrhoden, yep its good to go from my side of things

PR #603, opened 2021-05-04
Use Azure STORAGE_ACCOUNT_NAME env if set during init-vhd

kkeshavamurthy:
`STORAGE_ACCOUNT_NAME` was hardcoded in the init-vhd.sh. 
Make it configurable if `STORAGE_ACCOUNT_NAME` env var is set. 

**Additional context**


images/capi/packer/azure/scripts/init-vhd.sh
```diff
@@ -2,20 +2,22 @@
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
+export RESOURCE_GROUP_NAME="${RESOURCE_GROUP_NAME:-cluster-api-images}"
```
devigned, line 5: Can the storage account changes be moved back under the `echo "create storage account` statement?

devigned (changes requested):
Seems like a reasonable change. I'd just like to see the code flow to be returned to it's previous flow.

images/capi/packer/azure/scripts/init-vhd.sh
kkeshavamurthy: I'll make the change. 
I thought it's general practice to have variables defined and top and then have the commands.

devigned:
/lgtm

Thank you for addressing the feedback. You are probably right about the env var declarations. Perhaps, that's something we should pursue in a refactor PR.

@nader-ziada wdyt?

nader-ziada:
looks goods, but can we also add a note about this new env variable in the azure doc here https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/azure.md#prerequisites-for-azure

kkeshavamurthy:
> looks goods, but can we also add a note about this new env variable in the azure doc here https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/azure.md#prerequisites-for-azure

@nader-ziada There's no new variable. Default value of `RESOURCE_GROUP_NAME` is overridden if it's set in the user environment.

nader-ziada:
> > looks goods, but can we also add a note about this new env variable in the azure doc here https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/azure.md#prerequisites-for-azure
> 
> @nader-ziada There's no new variable. Default value of `RESOURCE_GROUP_NAME` is overridden if it's set in the user environment.

aren't you now using the value of `STORAGE_ACCOUNT_NAME` if its set as an env var?

codenrhoden:
Looks like maybe the `/approve` here didn't go through because of the docs change as well. Nader's should have been enough on the Azure front... weird. I feel like the bot usually lists things out separately if different OWNER approver configs are needed...

/approve

PR #604, opened 2021-05-06
Fix apt-daily services and add GOSS tests

SanikaGawhane:
Remove ansible tasks of disabling `apt-daily` and `apt-daily-upgrade`. These are static services and cannot be enabled/disabled (it has no provisions for enabling in the [Install] unit file section). 
```
root@ip-172-31-77-217:/home/ubuntu# systemctl list-unit-files | grep 'apt'
apt-daily-upgrade.service                      static
apt-daily.service                              static
apt-daily-upgrade.timer                        disabled
apt-daily.timer                                disabled
```
Found an old, similar issue in GOSS - https://github.com/aelsabbahy/goss/issues/266

Adding GOSS tests to check if the daily timers are disabled will ensure that other ones will never run (since 
static services can only be run as a dependency of others - in this case the daily timers that we *do* disable).

/cc @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -73,9 +73,7 @@
     state: stopped
     enabled: false
   loop:
-  - apt-daily
   - apt-daily.timer
-  - apt-daily-upgrade
   - apt-daily-upgrade.timer
 
 - name: Get installed packages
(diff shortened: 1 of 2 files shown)


codenrhoden:
Thanks for the fix @SanikaGawhane !
Looks like only need to disable (and test for) the `timer` services, because the `static` services are dependencies of the timers and won't run without the timers enabled. TIL in systemd land...

/lgtm
/approve

I think with this flaky test fixed, we can/should enable the GOSS tests as enforcing.

PR #534, opened 2021-03-04
Added additional windows download tasks

perithompson:
What this PR does / why we need it:

Allows the download of additional files and executables in windows image builder.

Which issue(s) this PR fixes: 
Fixes #518 

**Additional context**
Add any other context for the reviewers

/sig windows

PR #607, opened 2021-05-13
Add variable to set custom list of windows kubernetes binaries to download

perithompson:
When using contained, currently we need to pull kube-proxy down onto the host as well, rather than make this an additional download it would be good to allow the user to choose additional binaries to pull into the kubernetes base path. 

**Additional context**
This would be done by specifying `"kubernetes_bin_list": "kubeadm,kubelet,kubectl,kube-proxy"` 

If the option is not set then just kubeadm, kubelet and kubectl will be used

Once contained can run cni in a pod this would no longer be needed but for the foreseeable future there may be a requirement to run kube-proxy as a service


perithompson:
@jayunit100

codenrhoden:
@perithompson From what you are describing, it sounds to me like if `runtime==containerd`, that things will *not* workout without also installing `kube-proxy`. If that is true, I think that logic should be built directly into the Ansible role, rather than changing the list of `kubernetes-bins` as a user (e.g. variables passed by Packer).

Put another way, it doesn't make sense to me that if a user sets `runtime` to `containerd`, the would also have to set a second variable with the magic list of kubernetes bins to make it work. Setting `containerd` should just take care of it. Do you think that is possible?

perithompson:
the hard coded way would be 
```yaml
- name: Download kube-proxy binary
  win_get_url:
    url: "{{ kubernetes_base_url }}/kube-proxy.exe"
    dest: '{{ kubernetes_install_path }}\'
  when: runtime == "contained" 
  ```

codenrhoden:
> the hard coded way would be
> 
> ```yaml
> - name: Download kube-proxy binary
>   win_get_url:
>     url: "{{ kubernetes_base_url }}/kube-proxy.exe"
>     dest: '{{ kubernetes_install_path }}\'
>   when: runtime == "contained" 
> ```

If you don't want to hardcode a seperate task like this, you could "extend" the `kubernetes-bins` list conditionally based on `runtime` with logic like this one: https://github.com/kubernetes-sigs/image-builder/blob/5c9ccea21d1a32b0bfbcd2bff8bdbe17e6fbce2d/images/capi/ansible/roles/kubernetes/defaults/main.yml#L35.

It's still kind of hardcoding it, but instead it would be adding `kube-proxy` to the list conditionally, no matter what the original list is set to.

jsturtevant:
Could you use https://github.com/kubernetes-sigs/image-builder/pull/534 to pull kube-proxy?  Since this is temporary (once hostprocess lands we wouldn't want kube-proxy pulled)  it seems wrong to hard code it now

perithompson:
> Could you use #534 to pull kube-proxy? Since this is temporary (once hostprocess lands we wouldn't want kube-proxy pulled) it seems wrong to hard code it now

probably, as they all came from the same source to the same location just seemed easier to include in this step, rather than have to move it at run time, although I suppose we need to setup the service anyway so its only saving one line. fair points, thanks both, ill focus on #534

PR #611, opened 2021-05-14
add owners file for windows

jayunit100:


codenrhoden (commented):
I like the idea of having a seperate OWNERS file for windows related roles here. I think maybe creating a new `image-builder-windows-maintainers` alias in `OWNERS_ALIASES` might be a good approach? We've tried to avoid naming individuals any place but at the top level. I would then put @jsturtevant and @perithompson  in that alias.

/cc @CecileRobertMichon @jsturtevant

CecileRobertMichon:
> I like the idea of having a seperate OWNERS file for windows related roles here. I think maybe creating a new image-builder-windows-maintainers alias in OWNERS_ALIASES might be a good approach? We've tried to avoid naming individuals any place but at the top level. I would then put @jsturtevant and @perithompson in that alias.

+1

codenrhoden:
sounds like there is some consensus here, @jayunit100, if you can find the time to make the suggested change. Thanks for this!

jayunit100:
im going to close this so the img builder folks can implement it ,   not really my wheelhouse to define OWNERS files :)

PR #624, opened 2021-06-02
locked ansible base version

perithompson:
An issue has been identified in [ansible-base 2.10.10](https://github.com/ansible/ansible/issues/74887) that causes win_templates to fail to copy to a windows host.  As there is currently no way to lock the ansible base version I have added a downgrade to lock this until we are happy that the issue has been resolved in ansible

### DIFF
--- a/images/capi/hack/ensure-ansible.sh
+++ b/images/capi/hack/ensure-ansible.sh
@@ -32,5 +32,6 @@ if command -v ansible >/dev/null 2>&1; then exit 0; fi
 
 ensure_py3
 pip3 install --user "ansible==${_version}"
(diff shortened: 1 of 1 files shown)


perithompson:
@jsturtevant @codenrhoden  got there eventually, the problem with the templates was the version of ansible base

codenrhoden:
Thanks for tracking that down, @perithompson. I get why you are setting this explicitly in your other PR now...
This seems like a temporary workaround until we can set a new Ansible version (> 2.10.10) that has this problem fixed.
/assign @jsturtevant 
^ PTAL
/lgtm

jsturtevant:
/lgtm 
/approve

Thanks for tracking this down @perithompson!

PR #635, opened 2021-06-20
Add more user configurable goss provisioner options

kkeshavamurthy:
Adds the following user configurable goss provisioner options 

```
  "goss_download_path": "",
  "goss_remote_folder": "",
  "goss_remote_path": "",
  "goss_skip_install": "",
```
It's useful in using a pre-downloaded goss binary to run tests in air gapped environments. 

**Additional context**

### DIFF
--- a/docs/book/src/capi/goss/goss.md
+++ b/docs/book/src/capi/goss/goss.md
@@ -25,9 +25,13 @@ Supported arguments are passed through file: `packer/config/goss-args.json`
 ```json
 {
   "goss_arch": "amd64",
+  "goss_download_path": "",
   "goss_entry_file": "goss/goss.yaml",
   "goss_format": "json",
   "goss_inspect_mode": "true",
+  "goss_remote_folder": "",
+  "goss_remote_path": "",
+  "goss_skip_install": "",
   "goss_tests_dir": "packer/goss",
   "goss_url": "",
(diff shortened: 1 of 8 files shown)


docs/book/src/capi/goss/goss.md
```diff
@@ -25,9 +25,13 @@ Supported arguments are passed through file: `packer/config/goss-args.json`
 ```json
 {
   "goss_arch": "amd64",
+  "goss_download_path": "",
   "goss_entry_file": "goss/goss.yaml",
   "goss_format": "json",
   "goss_inspect_mode": "true",
+  "goss_remote_folder": "",
+  "goss_remote_path": "",
+  "goss_skip_install": "",
```
codenrhoden, line 34: Looks like the code and examples show this as a `bool`, so should this be `"goss_skip_install": false,` instead?
https://github.com/YaleUniversity/packer-provisioner-goss/blob/054242a2771427641a2563893ff9b4cc25b4836c/packer-provisioner-goss.go#L38

codenrhoden: Or at least, to be consistent with everything else, let's not rely on the default parsing. I think we should set it to `"false"`.

kkeshavamurthy: Agreed. Done.

PR #636, opened 2021-06-21
Fix for multiple lines returned for Azure OS_DISK_URI

mboersma:

`make builder-azure-vhd-$os-$version` recently began returning more than one URL into the packer.out file, causing build failures when this code looks for `OSDiskURI:`. This change takes only the first match, which is the correct URI.

Which issue(s) this PR fixes: 

N/A

**Additional context**:

cc: @CecileRobertMichon @devigned


mboersma:
/retest

Looks like a failure to clean up a Windows vm.

PR #639, opened 2021-06-30
Consistent cpu & ram capacity for all vsphere builders

kkeshavamurthy:
Currently CPU & Ram are set to 1 & 2G for `vsphere-iso` & `vsphere-iso-base` builders. 
Since /tmp was bumped to 5Gb in PR #629 , downloading data into /tmp eats up the ram and slows down the build VM to a standstill. 
So bumping all `vsphere-*` builder CPU to 4 & RAM to 16Gb. 

**Additional context**

/assign @codenrhoden


codenrhoden:
Hmm, we say this is for "all vSphere builders", but this skips vsphere-clone and Windows.

Right now it is indeed a mess, and looks like this:

| | vsphere-iso | vsphere-iso (windows) | vsphere-iso-base | vsphere-clone | vmware-iso | vmware-iso-base | vmware-iso (windows) |
| -- | -- | -- | -- | -- | -- | -- | -- |
| CPU | 1 | 4 | 1 | 4 | 1 | 1 | 4 |
| RAM | 2048 | 8192 | 2048 | 16384 | 2048 | 2048 | 4096 |

And on top of that, everything that has a "cores" setting has it set to "1", except for the the vphsere-clone builder which has it set to "12", which I *think* means that we are setting 4 sockets with 12 cores each for a 48 core machine. o_0

I kind of feel like 16GB might be high on the RAM side for most builders, especially if doing local. I totally get that it might be nice to set it that high if available, though. My recommendation is this:

Add vars in packer-common so that we can standardize across all builders, *and* allow them to be overridden by users.
Then, set the default CPUs to 4 and the default RAM to 8GB for *all* builders, including Windows.
I would also go ahead and set `cpu_cores` to one for the vsphere-clone builder.

Thoughts?

/cc @perithompson just for input on the Windows side. My suggestion would bump the default ram on the WIndows vmware-iso builder from 4 to 8GB, and that would be the only change.

kkeshavamurthy:
SGTM

perithompson:
@codenrhoden Agreed, we can bump it up to 8, I only left it lower to save people with smaller systems but I don't really think many people would try to build windows on a workstation with that little ram anyway!

images/capi/packer/ova/packer-common.json
```diff
@@ -11,6 +13,7 @@
   "guestinfo_datasource_slug": "https://raw.githubusercontent.com/vmware/cloud-init-vmware-guestinfo",
   "headless": "true",
   "insecure_connection": "false",
+  "memory": "8196",
```
codenrhoden, line 16: I think you want `8192` here. :)

codenrhoden: After that, LGTM

PR #641, opened 2021-07-01
capg: add options to set a service account or disable the default one

cpanato:
As part of the investigations why the nightly job is not working properly after we drop some GCP permissions, see https://github.com/kubernetes/k8s.io/pull/2147

We are adding two more options in the GCE builder to allow to disable the default service account and another to use a specific service account.

In this PR we are using `service_account_email` to bound with the service account we setup for the gcp project.

/assign @spiffxp @ameukam @dims @codenrhoden 

**Additional context**

### DIFF
--- a/images/capi/packer/gce/ci/nightly/overwrite-1-18.json
+++ b/images/capi/packer/gce/ci/nightly/overwrite-1-18.json
@@ -3,5 +3,6 @@
   "kubernetes_deb_version": "1.18.18-00",
   "kubernetes_rpm_version": "1.18.18-0",
   "kubernetes_semver": "v1.18.18",
-  "kubernetes_series": "v1.18"
+  "kubernetes_series": "v1.18",
+  "service_account_email": "gcb-builder-cluster-api-gcp@k8s-staging-cluster-api-gcp.iam.gserviceaccount.com"
 }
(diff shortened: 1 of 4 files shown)


spiffxp:
/lgtm
/hold
I'll leave it to you to /unhold, I'm not sure if the pull-azure-vhds failure is something you care about

codenrhoden:
/lgtm
/hold cancel

The Azure issue right now is a flake. Some changes went into Prow yesterday that should make it *not* required for this PR, as it does not touch anything that would result in a change in Azure.

cpanato:
@codenrhoden I will try to retest this, but if that fails again can we override the test? thanks!

/test pull-azure-vhds

codenrhoden:
> @codenrhoden I will try to retest this, but if that fails again can we override the test? thanks!

@cpanato yeah, let's do it. I actually think it was one of your comments I saw one time that showed me the `override` command! I've never tried to use it. :)

/override pull-azure-vhds

codenrhoden:
Well, I guess that command doesn't work. The docs say you need to be a repo admin, and I am for this repo. :/

cpanato:
@codenrhoden maybe we can make this job not required?

cpanato:
works now :)

PR #645, opened 2021-07-05
cleanup netplan files, cleanup journald

MaxRink:

Removes netplan files generated at install, if present
Cleans up journald as best as possible

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -109,6 +109,24 @@
     path: "{{ item.path }}"
   loop: "{{ temp_files.files }}"
 
+- name: Find netplan files
+  find:
+    depth: 1
+    file_type: any
+    paths:
+    - /lib/netplan
+    - /etc/netplan
+    - /run/netplan
+    pattern: '*.yaml'
+  register: netplan_files
+
+- name: Delete netplan files
+  file:
+    state: absent
+    path: "{{ item.path }}"
+  loop: "{{ netplan_files.files }}"
+  when: netplan_files.files is defined and (netplan_files.files|length>0)
+
 - name: Find SSH host keys
   find:
     path: /etc/ssh
@@ -167,3 +185,14 @@
   loop:
   - { path: /root/.bash_history }
   - { path: "/home/{{ ansible_env.SUDO_USER }}/.bash_history" }
+
+- name: Rotate journalctl to archive logs
+  shell:
+    cmd: |
+      journalctl --rotate
+  when: not ( ansible_os_family == "RedHat" and ansible_distribution_major_version|int <= 7 )
+
+- name: Remove archived journalctl logs
+  shell:
+    cmd: |
+      journalctl -m --vacuum-time=1s


images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -109,6 +109,24 @@
...
+- name: Find netplan files
+  find:
+    depth: 1
+    file_type: any
+    paths:
+    - /lib/netplan
+    - /etc/netplan
+    - /run/netplan
+    pattern: '*.yaml'
+  register: netplan_files
```
kkeshavamurthy, line 121: If the default network config is deleted, how will the VM obtain IP addresses etc?

MaxRink: Via cloud-init, every provider ive worked with so far sets network config in cloud init.
Not removing the defautls from build can lead to weird behaviour when new netplan config and old config merge tho, thst why im dropping it here

MaxRink: we should propably clarify that with other folks tho

kkeshavamurthy: Sometimes logging into the standalone image is useful in debugging some issues.

MaxRink: then you can still provde network config via cloud init
you already need to do that with users anyway, as the default users are all locked

randomvariable: > @randomvariable i guess CAPA also gets its network data from amazon, right?

Yes, that's right. It will set up link-local address, then ping 169.254.169.254 and  get its networking info

randomvariable: > then you can still provde network config via cloud init
> you already need to do that with users anyway, as the default users are all locked

Yeah, I agree. Really should not leave any networking configuration behind. It's going to default to DHCP anyway.

>  In the absence of network configuration in any of the above sources , Cloud-init will write out a network configuration that will issue a DHCP request on a “first” network interface.

From https://cloudinit.readthedocs.io/en/latest/topics/network-config.html

bavarianbidi: CAOP also got the network config via `cloud init`.

below the generated `50-cloud-init.yaml` which got created on instance boot and the `01-netcfg.yaml` which got generated during image-build process.
```bash
ls -ltrah /etc/netplan/
total 20K
-rw-r--r--  1 root root  193 Jul  2 10:30 01-netcfg.yaml
drwxr-xr-x  2 root root   54 Jul  2 11:30 .
-rw-r--r--  1 root root  485 Jul  6 03:40 50-cloud-init.yaml
drwxr-xr-x 91 root root 8.0K Jul  6 03:40 ..
```

```bash
cat /etc/netplan/01-netcfg.yaml 
# This file describes the network interfaces available on your system
# For more information, see netplan(5).
network:
  version: 2
  renderer: networkd
  ethernets:
    ens4:
      dhcp4: yes
```

but `ens4` doesn't exists:
```bash
ip a s ens4
Device "ens4" does not exist.
```

where `ens192` exists:

```bash
cat /etc/netplan/50-cloud-init.yaml 
# This file is generated from information provided by the datasource.  Changes
# to it will not persist across an instance reboot.  To disable cloud-init's
# network configuration capabilities, write a file
# /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with the following:
# network: {config: disabled}
network:
    version: 2
    ethernets:
        ens192:
            dhcp4: true
            match:
                macaddress: fa:16:3e:40:8c:2a
            set-name: ens192
```

```bash
ip a s ens192
2: ens192: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether fa:16:3e:40:8c:2a brd ff:ff:ff:ff:ff:ff
    altname enp11s0
    inet 10.6.0.166/24 brd 10.6.0.255 scope global ens192
       valid_lft forever preferred_lft forever
    inet6 fe80::f816:3eff:fe40:8c2a/64 scope link 
       valid_lft forever preferred_lft forever
```

MaxRink: so the only thing still in question would be CAPZ, right?
@CecileRobertMichon do you happen to know how azure handles this?

randomvariable: There's an example IMDS, and it's pretty much the same as AWS here https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=linux

randomvariable:
`journalctl --rotate` is returning non-0 exit in the Azure test.

randomvariable:
Can confirm
``` shell
➜ docker run -it centos:7 /bin/bash
Unable to find image 'centos:7' locally
7: Pulling from library/centos
2d473b07cdd5: Pull complete
Digest: sha256:0f4ec88e21daf75124b8a9e5ca03c37a5e937e0e108a255d890492430789b60e
Status: Downloaded newer image for centos:7
[root@7a3c8fd84af1 /]# yum install systemd
Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
 * base: mirrors.vinters.com
 * extras: uk.mirrors.clouvider.net
 * updates: uk.mirrors.clouvider.net
base                                                                                                                     | 3.6 kB  00:00:00
extras                                                                                                                   | 2.9 kB  00:00:00
updates                                                                                                                  | 2.9 kB  00:00:00
(1/4): base/7/x86_64/group_gz                                                                                            | 153 kB  00:00:00
(2/4): extras/7/x86_64/primary_db                                                                                        | 242 kB  00:00:00
(3/4): base/7/x86_64/primary_db                                                                                          | 6.1 MB  00:00:03
(4/4): updates/7/x86_64/primary_db                                                                                       | 8.8 MB  00:00:03
Resolving Dependencies
--> Running transaction check
---> Package systemd.x86_64 0:219-78.el7 will be updated
---> Package systemd.x86_64 0:219-78.el7_9.3 will be an update
--> Processing Dependency: systemd-libs = 219-78.el7_9.3 for package: systemd-219-78.el7_9.3.x86_64
--> Running transaction check
---> Package systemd-libs.x86_64 0:219-78.el7 will be updated
---> Package systemd-libs.x86_64 0:219-78.el7_9.3 will be an update
--> Finished Dependency Resolution

Dependencies Resolved

================================================================================================================================================
 Package                             Arch                          Version                                 Repository                      Size
================================================================================================================================================
Updating:
 systemd                             x86_64                        219-78.el7_9.3                          updates                        5.1 M
Updating for dependencies:
 systemd-libs                        x86_64                        219-78.el7_9.3                          updates                        418 k

Transaction Summary
================================================================================================================================================
Upgrade  1 Package (+1 Dependent package)

Total download size: 5.5 M
Is this ok [y/d/N]: y
Downloading packages:
Delta RPMs disabled because /usr/bin/applydeltarpm not installed.
warning: /var/cache/yum/x86_64/7/updates/packages/systemd-libs-219-78.el7_9.3.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID f4a80eb5: NOKEY
Public key for systemd-libs-219-78.el7_9.3.x86_64.rpm is not installed
(1/2): systemd-libs-219-78.el7_9.3.x86_64.rpm                                                                            | 418 kB  00:00:00
(2/2): systemd-219-78.el7_9.3.x86_64.rpm                                                                                 | 5.1 MB  00:00:01
------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                           3.6 MB/s | 5.5 MB  00:00:01
Retrieving key from file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
Importing GPG key 0xF4A80EB5:
 Userid     : "CentOS-7 Key (CentOS 7 Official Signing Key) <security@centos.org>"
 Fingerprint: 6341 ab27 53d7 8a78 a7c2 7bb1 24c6 a8a7 f4a8 0eb5
 Package    : centos-release-7-9.2009.0.el7.centos.x86_64 (@CentOS)
 From       : /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
Is this ok [y/N]: y
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Updating   : systemd-libs-219-78.el7_9.3.x86_64                                                                                           1/4
  Updating   : systemd-219-78.el7_9.3.x86_64                                                                                                2/4
Failed to get D-Bus connection: Operation not permitted
  Cleanup    : systemd-219-78.el7.x86_64                                                                                                    3/4
  Cleanup    : systemd-libs-219-78.el7.x86_64                                                                                               4/4
  Verifying  : systemd-219-78.el7_9.3.x86_64                                                                                                1/4
  Verifying  : systemd-libs-219-78.el7_9.3.x86_64                                                                                           2/4
  Verifying  : systemd-libs-219-78.el7.x86_64                                                                                               3/4
  Verifying  : systemd-219-78.el7.x86_64                                                                                                    4/4

Updated:
  systemd.x86_64 0:219-78.el7_9.3

Dependency Updated:
  systemd-libs.x86_64 0:219-78.el7_9.3

Complete!
[root@7a3c8fd84af1 /]# journalctl --help
journalctl [OPTIONS...] [MATCHES...]

Query the journal.

Flags:
     --system              Show the system journal
     --user                Show the user journal for the current user
  -M --machine=CONTAINER   Operate on local container
  -S --since=DATE          Show entries not older than the specified date
  -U --until=DATE          Show entries not newer than the specified date
  -c --cursor=CURSOR       Show entries starting at the specified cursor
     --after-cursor=CURSOR Show entries after the specified cursor
     --show-cursor         Print the cursor after all the entries
  -b --boot[=ID]           Show current boot or the specified boot
     --list-boots          Show terse information about recorded boots
  -k --dmesg               Show kernel message log from the current boot
  -u --unit=UNIT           Show logs from the specified unit
  -t --identifier=STRING   Show entries with the specified syslog identifier
  -p --priority=RANGE      Show entries with the specified priority
  -e --pager-end           Immediately jump to the end in the pager
  -f --follow              Follow the journal
  -n --lines[=INTEGER]     Number of journal entries to show
     --no-tail             Show all lines, even in follow mode
  -r --reverse             Show the newest entries first
  -o --output=STRING       Change journal output mode (short, short-iso,
                                   short-precise, short-monotonic, verbose,
                                   export, json, json-pretty, json-sse, cat)
     --utc                 Express time in Coordinated Universal Time (UTC)
  -x --catalog             Add message explanations where available
     --no-full             Ellipsize fields
  -a --all                 Show all fields, including long and unprintable
  -q --quiet               Do not show privilege warning
     --no-pager            Do not pipe output into a pager
  -m --merge               Show entries from all available journals
  -D --directory=PATH      Show journal files from directory
     --file=PATH           Show journal file
     --root=ROOT           Operate on catalog files underneath the root ROOT
     --interval=TIME       Time interval for changing the FSS sealing key
     --verify-key=KEY      Specify FSS verification key
     --force               Override of the FSS key pair with --setup-keys

Commands:
  -h --help                Show this help text
     --version             Show package version
  -F --field=FIELD         List all values that a specified field takes
     --new-id128           Generate a new 128-bit ID
     --disk-usage          Show total disk usage of all journal files
     --vacuum-size=BYTES   Reduce disk usage below specified size
     --vacuum-time=TIME    Remove journal files older than specified date
     --flush               Flush all journal data from /run into /var
     --header              Show journal header information
     --list-catalog        Show all message IDs in the catalog
     --dump-catalog        Show entries in the message catalog
     --update-catalog      Update the message catalog database
     --setup-keys          Generate a new FSS key pair
     --verify              Verify journal file consistency
[root@7a3c8fd84af1 /]# journalctl --rotate
journalctl: unrecognized option '--rotate'
```

MaxRink:
Oh yeah, RHEL before v8 isnt persisting journald anyway, so i guess we can just skip if for those versions

MaxRink:
I guess 
`when: not ( ansible_os_family == "RedHat" and ansible_lsb.major_release|int <= 7 )`
should do the trick, any edge-case im not thinking about?

randomvariable:
that should be fine

MaxRink:
had to switch to release version instead of lsb, as that is also not supported in centos :D
But i think it should be solved now

randomvariable:
/lgtm cancel

let's squash these commits before merging.

MaxRink:
@randomvariable done

randomvariable:
OVA timed out, not sure what happened with vhds.

randomvariable:
otherwise,

/lgtm

PR #647, opened 2021-07-07
Skip ovf/vmdk export in vsphere-iso-base builder

kkeshavamurthy:
The idea of a `vsphere-iso-base` builder is to create a base vm template on the vCenter and build images on top of that using the `vsphere-clone` builder. In this process exporting the ovf & vmdk from the vC to the build VM seems like an unnecessary step. So this PR removes that. 

**Additional context**


codenrhoden:
/lgtm
Thanks for this, @kkeshavamurthy !

PR #623, opened 2021-06-01
Added windows build fixes and optimisations

perithompson:
What this PR does / why we need it:
Adding fixes for cloudbase template failures, kubernetes binary download retries and openSSH failed

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #595,#621 

**Additional context**
Add any other context for the reviewers

PR #648, opened 2021-07-08
[OVA] [Ubuntu] Support disabling IPv6 module during boot

kkeshavamurthy:
Some networks provide IPv6 addresses but the network seems to be very flaky. 
Due to the flakiness I often see installation fail in this [step](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg#L101) when it tries to pull packages. Apt seems to prefer to pull packages using the IPv6 targets when both v4 & v6 configs are present. 
The current behavior will be the default and provide a way for the user to turn off IPv6 module during installation. 
This is done only during boot and IPv6 is not disabled in the actual built image. 

**Additional context**


images/capi/packer/ova/ubuntu-1804.json
```diff
@@ -1,6 +1,7 @@
...
+  "boot_command_prefix": "<enter><wait><f6><wait><esc><wait><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs><bs>/install/vmlinuz auto console-setup/ask_detect=false console-setup/layoutcode=us debconf/frontend=noninteractive debian-installer=en_US fb=false initrd=/install/initrd.gz kbd-chooser/method=us keyboard-configuration/layout=USA keyboard-configuration/variant=USA locale=en_US netcfg/get_domain=local netcfg/get_hostname=localhost grub-installer/bootdev=/dev/sda ipv6.disable={{ user `boot_disable_ipv6` }} preseed/file=",
```
codenrhoden, line 2: where is `boot_disable_ipv6` defined? Should it go in `packer/ova/packer-common.json` with a default value of `"false"`?

kkeshavamurthy: it's defined in the `ubuntu-*.json` file. Since this is specific to ubuntu only, I did not add in `packer-common`

codenrhoden: wait, was there the whole time and I just missed it? Wow...

kkeshavamurthy:
@perithompson @jsturtevant 
Any idea why some of the windows goss tests and builds are so flaky? Here are a couple of errors that I've seen often. 
Does increasing the `kubelet --version` timeout help or is there another underlying issue? 


```
 ==> vhd-windows-2019: Timeout waiting for machine to restart. 
````

```
     vhd-windows-2004:             "summary-line": "kubelet --version: exit-status: Error: Command execution timed out (10s)", 
```

perithompson:
I'm hoping https://github.com/kubernetes-sigs/image-builder/pull/623 will help, Goss on windows is pretty resource hungry and I think the commands just time out...

kkeshavamurthy:
Maybe we should bump the VM size in ci?

PR #655, opened 2021-07-16
capg: make capg node images available to use in another accounts

cpanato:
Now the job is building again (https://prow.k8s.io/?job=periodic-image-builder-gcp-all-nightly) and we resolve all issues regarding permissions

The next step is make those images available to use in CI to accelerate the testing time for CAPG (because in the existing jobs we build the image from scratch all the time)

GCP reference to make the images public: https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly

Slack 🧵  https://kubernetes.slack.com/archives/CCK68P2Q2/p1626358090270500

/assign @dims @spiffxp @codenrhoden 

**Additional context**

### DIFF
--- a/images/capi/scripts/ci-gce-nightly.sh
+++ b/images/capi/scripts/ci-gce-nightly.sh
@@ -56,3 +56,7 @@ echo "Displaying the generated image information"
 filter="name~cluster-api-ubuntu-*"
 gcloud compute images list --project "$GCP_PROJECT" \
   --no-standard-images --filter="${filter}"
+
+echo "Making images public to use in CI"
+(gcloud compute images list --project "$GCP_PROJECT" --no-standard-images --filter="${filter}" --format="value(name[])" | \
+awk '{print "gcloud compute images add-iam-policy-binding --project '"$GCP_PROJECT"' " $1 " --member='"'allAuthenticatedUsers'"' --role='"'roles/compute.imageUser'"' \n"}' | bash)


cpanato:
it will need `'compute.images.setIamPolicy'` not sure if this is in place

cpanato:
looks like it have the permission we are using `roles/compute.instanceAdmin.v1` and this role have `compute.images.*` permission

ref: https://cloud.google.com/iam/docs/understanding-roles

dims:
let's give it a shot.

/approve
/lgtm

images/capi/scripts/ci-gce-nightly.sh
```diff
@@ -56,3 +56,7 @@ echo "Displaying the generated image information"
 filter="name~cluster-api-ubuntu-*"
 gcloud compute images list --project "$GCP_PROJECT" \
   --no-standard-images --filter="${filter}"
+
+echo "Making images public to use in CI"
+(gcloud compute images list --project "$GCP_PROJECT" --no-standard-images --filter="${filter}" --format="value(name[])" | \
+awk '{print "gcloud compute images add-iam-policy-binding --project '"$GCP_PROJECT"' " $1 " --member='"'allAuthenticatedUsers'"' --role='"'roles/compute.imageUser'"' \n"}' | bash)
```
spiffxp, line 62: I suspect this needs a k8s.io PR to grant privileges to do this. Not sure the existing roles cover this

spiffxp (approved):
/approve
/lgtm
/hold
I think this will break if merged before permissions are sorted out

images/capi/scripts/ci-gce-nightly.sh
cpanato: We have this role `roles/compute.instanceAdmin.v1` set for the service account https://github.com/cpanato/k8s.io/blob/main/infra/gcp/ensure-staging-storage.sh#L361

and based on this documentation: https://cloud.google.com/iam/docs/understanding-roles the `roles/compute.instanceAdmin.v1` have the permissions to `compute.images.*`, the start it means all under the `.images` which is the one we need here.

spiffxp: /hold cancel
Thanks for looking it up @cpanato

PR #657, opened 2021-07-17
Use 1.7 ansible galaxy package for Windows

jsturtevant:

Ansible windows package broke win_templates in https://github.com/ansible/ansible/issues/74887 and has been [fixed](https://github.com/ansible-collections/ansible.windows/issues/212#issuecomment-823524227) in the package which is not avaliable via standard 2.10.11 ansible package.

This ensures we have the latest windows package with the fixes

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes https://github.com/kubernetes-sigs/image-builder/pull/624

**Additional context**

/sig windows
/cc @perithompson @mboersma

### DIFF
--- a/images/capi/hack/ensure-ansible.sh
+++ b/images/capi/hack/ensure-ansible.sh
@@ -32,6 +32,6 @@ if command -v ansible >/dev/null 2>&1; then exit 0; fi
 
 ensure_py3
 pip3 install --user "ansible==${_version}"
-pip3 install --user "ansible-base==2.10.9" --force-reinstall
 ensure_py3_bin ansible
 ensure_py3_bin ansible-playbook
+ansible-galaxy collection install ansible.windows:==1.7.0


CecileRobertMichon:
@codenrhoden @kkeshavamurthy OVA test is failing with

> --> vsphere: ServerFaultCode: Cannot complete login due to an incorrect user name or password.

Same failure is happening in other PRs too

perithompson:
@jsturtevant are you able to fix up your commits?  looks like one of them does nothing but a space? otherwise lgtm

codenrhoden:
/test pull-container-image-build

I'm confused when I see the title of this PR saying it's installing Ansible 2.10, but the code is installing ansible.windows 1.7.

Ansible has so many versions floating around now. =) I get what's happening, but still.

I think at some point we should explore moving to Ansible 4.x (with ansible-core 2.11.x), and maybe taking the ansible-core route with only installing the modules we need to reduce sizes.

mboersma:
> I think at some point we should explore moving to Ansible 4.x 

FWIW, I [fixed a related problem](https://github.com/deis/docker-go-dev/pull/307) that @jsturtevant and I were working on by upgrading from ansible 2.10.x to 4.2.0. It builds at least the Azure images without any problems, so there's one data point in favor.

CecileRobertMichon:
/lgtm

@jsturtevant please fix the commits and ping me when it's ready for approve

PR #553, opened 2021-03-17
Add retries while fetching Windows Defender signatures

jsturtevant:
What this PR does / why we need it:
Update Windows Defender signatures can fail when reaching out to MS servers to grab the signatures.  Testing to see if retries helps mitigate the failures

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #549 

**Additional context**
Add any other context for the reviewers

I've had a hard time reproducing on my sub.  Will run the test a few times before merging to see that it resolves flakes

/cc @perithompson

PR #659, opened 2021-07-20
bump containerd to 1.5.4

MaxRink:
https://github.com/containerd/containerd/releases/tag/v1.5.4
Tl;dr: fixes a CVE

**Additional context**

### DIFF
--- a/images/capi/packer/config/containerd.json
+++ b/images/capi/packer/config/containerd.json
@@ -1,7 +1,7 @@
 {
   "containerd_additional_settings": null,
   "containerd_cri_socket": "/var/run/containerd/containerd.sock",
-  "containerd_sha256": "32a9bf1b7ab2adbd9d2a16b17bf1aa6e61592938655adfb5114c40d527aa9be7",
-  "containerd_sha256_windows": "9ad0323b1ffc717795b27708ede5a69b653bfa44edb73c7891aa06ccc2f4c809",
-  "containerd_version": "1.5.3"
+  "containerd_sha256": "591e4e087ea2f5007e6c64deb382df58d419b7b6922eab45a1923d843d57615f",
+  "containerd_sha256_windows": "64a84f45d1eb21ec8298905e2f5d9014811c9100f71d589b9887bb37891845f5",
+  "containerd_version": "1.5.4"
 }


MaxRink:
Hmm, weird. looking at the logs containerd installs correctly on windows but the kubelet fails afterwards.

CecileRobertMichon:
@jsturtevant @perithompson any ideas?

perithompson:
It's like packer timed out on the final reboot before sysprep, and only on 2019, if it was containerd I'd expect it to effect both versions? @jsturtevant do we get any machine logs? To see if there was an error on reboot that stopped the host from booting?

jsturtevant:
I am not seeing any errors for kubelet starting and the ansible tasks are completing successfully for both 2019 and 2004. The  failures seem to have timeouts during the final restart:

```
--> vhd-windows-2019: Timeout waiting for machine to restart.
```

This restart time hasn't been an issue before but could vary depending on the Windows updates being applied.  In aks-engine  we have the timeouts set to 10m instead of the default 5m (https://www.packer.io/docs/provisioners/windows-restart).  

We can either add a restart timeout here or in a different PR.

CecileRobertMichon:
@jsturtevant any ideas why it would start failing on this particular PR? seems like it's still passing on other PRs, eg. https://github.com/kubernetes-sigs/image-builder/pull/645

jsturtevant:
Unfortunately no, the timeout isn't even consistent in this PR, it failed for 2004 the first round and 2019 the second.  @marosset do you have any ideas on why it take longer from time to time? My understanding is the subset of Core Windows packages could be slightly different causing a slower update.  We've bumped aks-engine to 10 mins per restart for this reason.

marosset:
> Unfortunately no, the timeout isn't even consistent in this PR, it failed for 2004 the first round and 2019 the second. @marosset do you have any ideas on why it take longer from time to time? My understanding is the subset of Core Windows packages could be slightly different causing a slower update. We've bumped aks-engine to 10 mins per restart for this reason.

I don't and trying to root-cause delays like that can be very time consuming :(

jsturtevant:
@MaxRink https://github.com/kubernetes-sigs/image-builder/pull/660 merged to help resolve the timeout. could you re-base and try again?  Thanks!

MaxRink:
done

jsturtevant:
This error we have seen before, and thought was resolved in https://github.com/kubernetes-sigs/image-builder/pull/553. I will look into this some more.

```
ASK [systemprep : Update Windows Defender signatures] *************************[0m
[0;32m    vhd-windows-2019: fatal: [default]: FAILED! => {"changed": true, "cmd": "$service = Get-Service \"Windefend
```

/retest

CecileRobertMichon (commented):
/approve

jsturtevant:
/retest

This timeout is due to goss running in parallel for windows.  We have started to see it after enabling goss to fail. Spoke with @perithompson and we should lower the concurrency here for now and address in up stream goss.

PR #660, opened 2021-07-21
Increase Restart timeout for Windows

jsturtevant:

#659 is seeing builds timeout for Windows (https://github.com/kubernetes-sigs/image-builder/pull/659#issuecomment-884357758)

**Additional context**

I didn't update the restart timeout for OVA as there was custom logic.  @perithompson do you think we should increase it there too?

/sig windows

### DIFF
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -91,6 +91,7 @@
       "user": "Administrator"
     },
     {
+      "restart_timeout": "10m",
       "type": "windows-restart"
     },
     {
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
/lgtm

pending @perithompson's answer to
> I didn't update the restart timeout for OVA as there were custom logic. @perithompson do you think we should increase it there too?

jsturtevant:
Goss time-outed:

```
kubelet --version: exit-status: Error: Command execution timed out (10s)
```

@perithompson  Is this related to some of the other timeouts we've seen with goss?

/retest

perithompson:
so far its been ok, ill need to do a bit more work to test if the logic can be the same? but I can raise a PR for that as another PR?
/lgtm

PR #662, opened 2021-07-22
Add a script to delete unused Azure image storage

mboersma:

In the process of building reference images for CAPZ users, builds fail or are stopped for other reasons. In some cases these leave orphaned storage accounts, and a number of these have piled up.

This script deletes storage accounts in a specified Azure subscription and resource group which are not referenced by the marketplace offer or offers specified, assuming they are older than 30 days and match the expected naming pattern. By default, it only prints what it would do, so it is safe to use in a "dry run" fashion.

N/A

**Additional context**

I have used a nearly identical version of this script with @CecileRobertMichon and @devigned to clean up a large number of unreferenced large accounts.

This seems to me the most logical home for such a script, even though this is arguably an Azure internal detail.

### DIFF
--- a/images/capi/packer/azure/scripts/delete-unused-storage.sh
+++ b/images/capi/packer/azure/scripts/delete-unused-storage.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


devigned:
/lgtm

Thank you for adding the script here. I'm sure others with image build automation with thank you.

CecileRobertMichon (commented):
/lgtm
/approve

PR #670, opened 2021-07-30
capi: avoid specifying containerd version for Flatcar

dongsupark:
Containerd version in Flatcar could be different from the default containerd version for other distros.
However, it is difficult to update the version in the code every time when Flatcar stable gets released.
Instead, delete the Flatcar-specific version, and take a different approach, specifying a user-defined containerd json config.

Now that Flatcar-specific versions and sha256 variables are gone, in `hack/image-build-flatcar.sh`, we need to provide containerd.json that includes Flatcar-specific version to the make command.

Users are also allowed to pass their own Flatcar version and sha256. e.g.

```
FLATCAR_CONTAINERD_VERSION=1.5.4 hack/image-build-flatcar.sh
```

### DIFF
--- a/images/capi/hack/image-build-flatcar.sh
+++ b/images/capi/hack/image-build-flatcar.sh
@@ -6,12 +6,20 @@ export VAGRANT_VAGRANTFILE=${VAGRANT_VAGRANTFILE:-/tmp/Vagrantfile.builder-flatc
 export VAGRANT_SSH_PRIVATE_KEY=${VAGRANT_SSH_PRIVATE_KEY:-/tmp/vagrant-insecure-key}
 export VAGRANT_SSH_PUBLIC_KEY=${VAGRANT_SSH_PUBLIC_KEY:-/tmp/vagrant-insecure-key.pub}
 
+CONTAINERD_TMPDIR="$(mktemp -d)"
+CONTAINERD_CHECKSUM_FILE="${CONTAINERD_TMPDIR}/containerd-sha256sum"
+PACKER_VAR_FILES_CONTAINERD="${CONTAINERD_TMPDIR}/containerd.json"
+
 usage() {
     echo "Usage: $0 [<channel>] [<version>]"
     echo "          <channel> is one of: edge alpha beta stable (defaults to"
     echo "                      stable)"
     echo "          <version> release version to use (defaults to the latest"
     echo "                      release available for <channel>)"
+    echo ""
+    echo "To specify Flatcar-specific containerd version and/or its sha256:"
+    echo ""
+    echo "  FLATCAR_CONTAINERD_VERSION=1.5.4 FLATCAR_CONTAINERD_SHA256=abcd... ./hack/image-build-flatcar.sh"
 }
 # --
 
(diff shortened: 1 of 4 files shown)


codenrhoden (commented):
Hi @dongsupark,

That's awesome news!

These values are now the same as what is in `images/capi/packer/config/containerd.json`. Is there an expectation that in the future Flatcar will need to maintain a different default than the other providers? If not, the lines being edited here can be deleted instead of changing them, and Flatcar will just pick up the "global" default that is already being set.  These flatcar-specific values were essentially overrides that are no longer needed if I am understanding correctly.

And, it is still the case that a user could provide custom values at build time if they ever need to point to a different version.

dongsupark:
@codenrhoden Thanks for the comment.

> These values are now the same as what is in images/capi/packer/config/containerd.json. Is there an expectation that in the future Flatcar will need to maintain a different default than the other providers?

Yes, containerd version of Flatcar, specifically Stable channel, can be different from upstream containerd releases.
Main reason is that it takes around 2~3 months until one major version (like 2905) could be promoted from Beta to Stable. Especially big jumps like containerd 1.4 to 1.5 could require multiple weeks for the promotion.
When some security issues or bug fixes are available, we try to backport the fixes to Stable as soon as possible. In that case, the gap might not be that long.

> If not, the lines being edited here can be deleted instead of changing them, and Flatcar will just pick up the "global" default that is already being set. These flatcar-specific values were essentially overrides that are no longer needed if I am understanding correctly.

Yes, ideally I would like to do so, instead of updating the versions in the image-builder on every Flatcar release.
Fully understand your concerns.

> And, it is still the case that a user could provide custom values at build time if they ever need to point to a different version.

Not sure how a user could easily do that.
As far as I understand, user would need to specify both `containerd_version` and `containerd_sha256` in `packer/config/containerd.json` or similar for that purpose.
So if we have a new containerd version x.y.z, we need to get 
https://github.com/containerd/containerd/releases/download/vx.y.z/cri-containerd-cni-x.y.z-linux-amd64.tar.gz.sha256sum to be used for `containerd_sha256`, right?
Or am I missing something?

codenrhoden:
> > And, it is still the case that a user could provide custom values at build time if they ever need to point to a different version.
> 
> Not sure how a user could easily do that.
> As far as I understand, user would need to specify both `containerd_version` and `containerd_sha256` in `packer/config/containerd.json` or similar for that purpose.
> So if we have a new containerd version x.y.z, we need to get
> https://github.com/containerd/containerd/releases/download/vx.y.z/cri-containerd-cni-x.y.z-linux-amd64.tar.gz.sha256sum to be used for `containerd_sha256`, right?
> Or am I missing something?

Yes, a user would still need to lookup the correct values in order to pass them along.

For example, if I was building flatcar and I wanted to install the *old* containerd (1.4.4), I would do it this way:

```
$ cat containerd.json
{
  "containerd_sha256": "96641849cb78a0a119223a427dfdc1ade88412ef791a14193212c8c8e29d447b",
  "containerd_version": "1.4.4"
}
$ PACKER_VAR_FILES=containerd.json make build-ami-flatcar
```


From your comments, it sounds like you would like to keep the containerd version locked in place for Flatcar a bit more, as a bump to the default is more likely to break Flatcar. It does get bumped pretty frequently, usually within a few days of a new containerd release. So maybe keeping things the way you have them is the better approach.

dongsupark:
@codenrhoden Thanks for the comment.
As you suggested, I took a different approach.
Removed Flatcar-specific containerd versions and sha256.
Instead made `hack/image-build-flatcar.sh` specify a custom `containerd.json` including a containerd version and its sha256.

codenrhoden:
/lgtm

I like your approach! It looks like if you don't specify custom containerd values, it will just use what is defined in the project as the default, but if you do set those FLATCAR env vars, it will put the needed values into a temporary file and pass it into Packer.

I'll see if anyone else has any comments, but LGTM.

PR #671, opened 2021-07-30
Add option to enable .local lookups via DNS in Photon

randomvariable:

Lot of folk using Photon are still using .local domains in contravention of the RFCs. This adds an option to enable resolution of .local via DNS rather than MulticastDNS.

As this is technically a leak, it's disabled by default.

**Additional context**

### DIFF
--- a/docs/book/src/capi/providers/vsphere.md
+++ b/docs/book/src/capi/providers/vsphere.md
@@ -4,16 +4,16 @@
 
 The images may be built using one of the following hypervisors:
 
-| OS | Builder | Build target |
-|----|---------|--------------|
-| Linux | VMware Workstation (vmware-iso) | build-node-ova-local-<OS> |
-| macOS | VMware Fusion (vmware-iso)| build-node-ova-local-<OS> |
-| ESXi | ESXi | build-node-ova-esx-<OS> |
-| vSphere | vSphere >= 6.5 | build-node-ova-vsphere-<OS> |
-| vSphere | vSphere >= 6.5 | build-node-ova-vsphere-base-<OS> |
-| vSphere Clone | vSphere >= 6.5 | build-node-ova-vsphere-clone-<OS> |
-| Linux | VMware Workstation (vmware-vmx) | build-node-ova-local-vmx-<OS> |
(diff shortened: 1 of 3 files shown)


images/capi/hack/ensure-ansible.sh
```diff
@@ -35,3 +35,4 @@ pip3 install --user "ansible==${_version}"
 ensure_py3_bin ansible
 ensure_py3_bin ansible-playbook
 ansible-galaxy collection install ansible.windows:==1.7.0
+ansible-galaxy collection install community.general:==3.4.0
```
codenrhoden, line 38: While I get why this is here, it seems unrelated to this PR?

I think using the newer syntax (e.g. `community.general.ini_file`) is a laudable goal, and an upgrade we need to do, I'd rather not see it mixed in piecemeal. I find it harder to read (but that may just be me, happy to hear other's opinions). For this particular PR, I'd rather see the changes made just `ini_file` until we make a bigger, across the board bump in Ansible. Thoughts?

codenrhoden (changes requested):
Hi @randomvariable,

Thanks for looking into this and coming up with a way for users to use `.local`. I agree that it shouldn't be enabled by default. I do think we should add this variable to the table at https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/capi.md#customization to document it's use? Or maybe since this is Photon specific, maybe it belongs in https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/vsphere.md instead? We don't really have a table there for provider-specific variables, but we don't have one for AWS, so there is an example: https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/providers/aws.md#common-aws-options.

images/capi/hack/ensure-ansible.sh
randomvariable: Does `ini_file` still work? The Ansible docs have removed mention of it.

randomvariable: Reverted to the 2.9 docs version

codenrhoden:
Thanks for the docs updates!

/lgtm
/approve

PR #674, opened 2021-08-04
Bump k8s version to v1.20.9

CecileRobertMichon:
What this PR does / why we need it: Now that K8s v1.22.x has been released, we can bump the default K8s version to the latest in the 1.20.x series, keeping with the n-2 approach.

**Additional context**


codenrhoden:
/lgtm

Thanks!

PR #677, opened 2021-08-16
provide a way for users to bring their own containerd config

kkeshavamurthy:
A way for users to use their own containerd config file. 

`containerd_config_file` points to the default config file that's being used currently. Users can set the var to point it to their custom config to use. 

**Additional context**


codenrhoden (approved):
I think doing the copy using `template` here is a good thing -- it makes using the existing template easy (no changes), but it also means that if you provide your own file, that file could be static (pre-filled) or could in fact be a Jinja template.

I think it's worth getting input from @MaxRink on this, to see if it this meets the needs he was going for in #474. I think it does, but want to double check.

MaxRink:
@codenrhoden kinda :D 
Containerd 1.5.0 and up also have a directory stucture that gets ued for registry-specific config ( https://github.com/containerd/containerd/releases/tag/v1.5.0 ) which isnt covered by this.

kkeshavamurthy:
@codenrhoden Should we merge this PR and create a new issue with @MaxRink 's suggestion?

codenrhoden:
> @codenrhoden Should we merge this PR and create a new issue with @MaxRink 's suggestion?

that sounds reasonable to me. I think looking at what would be the better way to support containerd's directory structure for configuration is something we should definitely look into.

/lgtm

PR #679, opened 2021-08-18
Allow specifying kubernetes version

mdbooth:
Bumps the default kubernetes version to current latest: 1.22
Another version can be specified with:

  make &lt;target&gt; KUBERNETES_VERSION=1.18

**Additional context**

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -111,6 +111,7 @@ CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= dev
 ARCH ?= amd64
 BASE_IMAGE ?= docker.io/library/ubuntu:focal
+KUBERNETES_VERSION ?= 1.22
 
 ## --------------------------------------
 ## Packer flags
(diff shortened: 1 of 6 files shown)


jsturtevant:
@mdbooth thanks for the PR! we typically keep the k8s version n-2. see https://github.com/kubernetes-sigs/image-builder/pull/674 as an example

We provide the overridding ability via passing in additional values via files as outlined in https://github.com/kubernetes-sigs/image-builder/blob/master/docs/book/src/capi/capi.md#examples

There is also the ability to override variables via packer flags: `PACKER_FLAGS="-var=kubernetes_semver=1.20.2"`

I realize this isn't as simple as what you have but I don't know if adding this opens the door for overriding more variables and adds maintenance burden.  @codenrhoden might have thoughts?

Improving the docs to make the options we do have more visible is definitely something we could improve. See https://github.com/kubernetes-sigs/image-builder/pull/439#issuecomment-765677898 for more conversation there.

mdbooth:
The GCE folks are already maintaining this data, btw, which is where I took them from: https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi/packer/gce/ci/nightly

It feels like a generally useful thing.

PR #681, opened 2021-08-20
Support building Azure images using custom networks

kkeshavamurthy:
When building in VPN enabled Azure accounts, the build VM needs to be hooked up to custom networks to access internal networks. This PR adds a way for users to do that. 

**Additional context**


images/capi/packer/azure/packer-windows.json
```diff
@@ -185,7 +189,11 @@
     "manifest_output": "manifest.json",
     "nssm_url": null,
     "prepull": null,
+    "private_virtual_network_with_public_ip": "",
```
codenrhoden, line 192: So from the packer docs, it says this is:
> private_virtual_network_with_public_ip (bool) - This value allows you to set a virtual_network_name and obtain a public IP. If this value is not set and virtual_network_name is defined Packer is only allowed to be executed from a host on the same subnet / virtual network.

I take that to mean it's a boolean that toggles whether the VM will get a public IP in addition to being on the private network. Is that how this works? So I'm guessing you need to set this if you still want to be able to access the public internet?

I'd like to see a description of this added to the Azure docs.

cc @CecileRobertMichon

kkeshavamurthy: Yes, that' correct. I didn't add it to the docs since these are straight up packer vars. Let me know if you still want them, I'll add it.

codenrhoden:
Thanks for the docs update, @kkeshavamurthy 

/lgtm
/approve

PR #682, opened 2021-08-24
Adding Windows Server 2022 support for vSphere

stuartpreston:

This PR adds Windows Server 2022 support to image builder for vSphere. It's pretty much a copy of the Windows Server 2019 autounattend.xml file with the KMS client key switched out for the latest one that matches the edition (Standard Server Core), plus an example configuration.

### DIFF
--- a/images/capi/packer/ova/windows-2022.json
+++ b/images/capi/packer/ova/windows-2022.json
@@ -1,0 +1,16 @@
+{
+  "build_name": "windows-2022",
+  "disk_controller_type": "pvscsi",
+  "distro_arch": "amd64",
+  "distro_name": "windows",
+  "distro_version": "2022",
+  "guest_os_type": "Windows2019Server-64",
+  "iso_checksum": "none",
+  "local_guest_os_type": "windows9srv-64",
+  "os_display_name": "Windows Server 2022",
+  "os_iso_path": "[datastore] ISO/en-us_windows_server_2022_x64_dvd_620d7eac.iso",
+  "os_iso_url": "file:/path/en-us_windows_server_2022_x64_dvd_620d7eac.iso",
+  "vmtools_iso_path": "[datastore] ISO/vmtools/windows-11.1.5.iso",
+  "vmtools_iso_url": "file:/path/VMware-tools-windows-11.1.5-16724464.iso",
+  "vsphere_guest_os_type": "windows9Server64Guest"
+}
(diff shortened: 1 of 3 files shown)


images/capi/packer/ova/windows-2022.json
```diff
@@ -0,0 +1,16 @@
+{
+  "build_name": "windows-2022",
+  "disk_controller_type": "pvscsi",
+  "distro_arch": "amd64",
+  "distro_name": "windows",
+  "distro_version": "2022",
+  "guest_os_type": "Windows2019Server-64",
```
jayunit100, line 7: shud b 2022 ? just hacking around to get this working and caught it?

stuartpreston: No, I left this deliberately as 2022 is not recognized as a guest OS type yet in the VM tools or vSphere.

images/capi/packer/ova/windows/windows-2022/autounattend.xml
```diff
@@ -0,0 +1,250 @@
...
+                            <Value>Windows Server 2022 SERVERSTANDARDCORE</Value>
+                        </MetaData>
+                    </InstallFrom>
+                </OSImage>
+            </ImageInstall>
+            <UserData>
+                <AcceptEula>true</AcceptEula>
+                <FullName>Administrator</FullName>
+                <Organization>Organization</Organization>
+                <ProductKey>
```
jayunit100, line 82: i still feel like id like to parameterize this at some point, but doesnt need to be in this PR :), for me i manually commented it out to test

jayunit100:
just tested, appears to work for me with https://www.microsoft.com/en-us/evalcenter/evaluate-windows-server-2022 

thank you sir !

images/capi/packer/ova/windows-2022.json
jsturtevant: that will be confusing in the future 😉 this is when json having comments would be useful but makes sense

jsturtevant:
/ok-to-test

lgtm but @perithompson has more vsphere experience

jayunit100:
@perithompson i think tide will do that for us now.. 
merge merge merge !

perithompson:
/assign @cecilerobertmichon
Please can you take a look, I guess we need to do the same for the other providers @jsturtevant

codenrhoden:
FWIW, these commits did not get squashed. I'm not sure if there is some prow plugin we need to enable/change to make tha that happen or what, but in the future I'll be more of a stickler for squashing (I also thing having a plugin do it is weird, because then you don't have control over what the overall commit message looks like -- it probably just appends them all, but that's usually not what you want).

I also don't see `/tide` even mentioned on https://prow.k8s.io/command-help. @jayunit100 is this something you've seen in action?

PR #690, opened 2021-09-10
block traffic to 168.63.129.16 port 80 for cve-2021-27075

devigned:
What this PR does / why we need it:

This PR explicitly ensures that TCP traffic bound for the reserved Azure IP 168.63.129.16 via TCP on port 80 is dropped to remediate https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-27075. 

Needs: #691 

/cc @CecileRobertMichon

PR #694, opened 2021-09-14
add Windows firewall rule to block 168.63.129.16:80 for cve-2021-27075

devigned:

This PR explicitly ensures that TCP traffic bound for the reserved Azure IP 168.63.129.16 via TCP on port 80 is dropped to remediate https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-27075 on Windows nodes.

**Additional context**
Related to: #690

### DIFF
--- a/images/capi/ansible/windows/node_windows.yml
+++ b/images/capi/ansible/windows/node_windows.yml
@@ -70,6 +70,8 @@
         name: cloudbase-init
       when: install_cloudbase_init
     - include_role:
+        name: providers
+    - include_role:
         name: runtimes
     - include_role:
         name: kubernetes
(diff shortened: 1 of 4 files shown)


devigned:
/hold

I'm adding a goss verifier.

codenrhoden:
/lgtm

Makes me realize we didn't add a GOSS verification for this on Linux, but since this PR is focused on Windows that seems out of scope.

devigned:
> /lgtm
> 
> Makes me realize we didn't add a GOSS verification for this on Linux, but since this PR is focused on Windows that seems out of scope.

I’m happy to open another pr to add the Linux test. Def should have added one.

PR #696, opened 2021-09-17
Checking if ssh directory exists before creating it

knabben:

If the .ssh directory already exists in the image, a warning is raised, checking the directory before creating

**Additional context**

### DIFF
--- a/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
@@ -111,8 +111,7 @@
     type: string
 
 - name: Create SSH program data folder
-  win_shell: mkdir "$env:ProgramData\ssh"
-  ignore_errors: yes
+  win_shell: If (-Not (Test-Path -Path "$env:ProgramData\ssh")) { mkdir "$env:ProgramData\ssh" }
 
 - name: Enable ssh login without a password
   win_shell: Add-Content -Path "$env:ProgramData\ssh\sshd_config" -Value "PasswordAuthentication no`nPubkeyAuthentication yes"


knabben:
```
PS C:\ProgramData> rmdir .ssh
 PS C:\ProgramData> ls .ssh
 ls : Cannot find path 'C:\ProgramData\.ssh' because it does not exist.
 PS C:\ProgramData> If (-Not (Test-Path -Path C:\ProgramData\.ssh)) {mkdir c:\ProgramData\.ssh } 
Directory: C:\ProgramData
----                                                                  d-----         9/17/2021  11:35 AM                .ssh

PS C:\ProgramData> If (-Not (Test-Path -Path C:\ProgramData\.ssh)) {mkdir c:\ProgramData\.ssh } 
PS C:\ProgramData>
```

codenrhoden:
/assign @perithompson @jsturtevant 

This makes me wonder how things are setup "out of the box" when installing from ISO -- since we are always doing a clean install (we don't run roles against pre-existing servers), I feel like the directory is either always already there or it isn't.

perithompson:
I think it is strange that the folder doesn't get added when using evaluation media but does when adding the feature on a licensed ISO @jsturtevant any thoughts on why?

images/capi/ansible/windows/roles/systemprep/tasks/main.yml
```diff
@@ -111,7 +111,7 @@
     type: string
 
 - name: Create SSH program data folder
-  win_shell: mkdir "$env:ProgramData\ssh"
+  win_shell: If (-Not (Test-Path -Path "$env:ProgramData\ssh")) { mkdir "$env:ProgramData\ssh" }
   ignore_errors: yes
```
perithompson, line 115: id remove the ignore errors now because if it cant create then it should fail really

jsturtevant:
> I think it is strange that the folder doesn't get added when using evaluation media but does when adding the feature on a licensed ISO @jsturtevant any thoughts on why?

I don't know, haven't tried with the evaluation media but this seems like a reasonable approach.

/lgtm
/approve

PR #697, opened 2021-09-17
Bump Ubuntu 18.04 ISO to currently available ISO

detiber:

Current builds reference the `ubuntu-18.04.5-server-amd64.iso`, which is no longer available and has been replaced with `ubuntu-18.04-6-server-amd64.iso`


codenrhoden:
/lgtm

It's so weird that they pull the older ISO for `amd64`, but not others like `s390x` or `ppc64el`.

Issue #688, opened 2021-09-03
goss is failing regularly on Azure gen2 images


jsturtevant:
**What steps did you take and what happened:**
[A clear and concise description on how to REPRODUCE the bug.]

We have seen this flake on ubuntu 1804 and 2004 multiple times.  As an example:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/652/pull-azure-sigs/1433463530325020672

```
==> Some builds didn't complete successfully and had errors:
--> sig-ubuntu-1804-gen2: Error running Goss: goss non-zero exit status
```

If looks like the issue is with the kublet check:

```
"summary-line": "Service: kubelet: running:\nExpected\n    \u003cbool\u003e: true\nto equal\n    \u003cbool\u003e: false",[0m
```

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi), [kube-deploy/imagebuilder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/kube-deploy/imagebuilder), [konfigadm](https://github.com/kubernetes-sigs/image-builder/tree/master/images/konfigadm)):

Additional info for Image Builder for Cluster API related issues:

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`):
- Packer Version: 1.
- Packer Provider: azure sigs
- Ansible Version: 
- Cluster-api version (if using):
- Kubernetes version: (use `kubectl version`):

/kind bug
[One or more /area label. See https://github.com/kubernetes-sigs/cluster-api/labels?q=area for the list of labels]

jepio:
I think its ubuntu images in general, not gen2, and I hit the same on my PR #701. I believe its related to `kubelet.service` being started and being in a restart-loop, so occasionally its running when goss looks. I've added a commit that ensures the service is stopped (but stays enabled), so that the test should not fail any more.

PR #701, opened 2021-09-23
Pothos+jepio/flatcar fixes

jepio:
This PR adds a bunch of tweaks to the Flatcar Container Linux image build for CAPI with the following goals:
* better align behavior of Flatcar images with other distributions
* fix issues found while developing CAPI provider support (Azure/vSphere)

We need it so that users can easily use produced Flatcar images with CAPI.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): no open issues

**Additional context**

Here's a walkthrough of the changes:
* we install containerd in the same way as other distros now - by unpacking upstream tarballs instead of relying on a mixture of Flatcar and upstream binaries
* remove /etc/machine-id and /etc/hostname during build, and create /boot/flatcar/first_boot to ensure ignition and systemd treat the next provisioning as the first boot again
* disable OS updates by default. While automatic OS updates are a nice feature of Flatcar, in the context of CAPI they aren't so useful as we expect users to roll out new images on their own schedule (or explicitly enable OS updates again)
*  ensure packer user is deprovisioned so that on provisioning new admin user is created (typically capi). For Azure this is done through waagent, QEMU relied on a default ignition file in the OEM partition.
* mask docker to ensure containerd is used by capi. Previously docker was automatically discovered and so dockershim ended up being used.
* improve compatibility with kubernets by making /usr/libexec work for flex volumes and ensuring /opt/bin is in PATH when `kubeadm join` is executed.
* a weird issue with the installed /etc/hosts that prevented port-forwarding to etcd from working (but only when containerd is CRI)
* we made it easier to use the latest Flatcar release by adding support for `make FLATCAR_VERSION=current`


jepio:
/cc @dongsupark 
/cc @whites11 
@whites11: this includes some fixes that I noticed were necessary when testing the Flatcar Azure SIG work that you submitted.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -52,31 +47,19 @@
...
-    dest: /etc/systemd/system/containerd.service
-    src: etc/systemd/system/containerd-flatcar.service
-    mode: 0600
+      - --absolute-names
+      - --transform
+      - 's@usr@opt@'
+      - --transform
+      - 's@sbin@bin@'
+      - --transform
+      - 's@opt/local@opt@'
```
dongsupark, line 62: The `--transform` option is really great idea. :+1:

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -196,3 +197,17 @@
...
     cmd: |
       journalctl -m --vacuum-time=1s
+
+- name: Ensure ignition runs on next boot
+  file:
+    state: touch
+    path: /boot/flatcar/first_boot
+    owner: root
+    group: root
+  when: ansible_os_family == "Flatcar"
```
dongsupark, line 207: That looks good.
Question, is it then possible to delete [`touch /boot/flatcar/first_boot`](https://github.com/kubernetes-sigs/image-builder/blob/5f3d1d6998a29ac1de63f0ae914bcb266a242078/images/capi/packer/files/bootstrap-flatcar.sh#L43) in `packer/files/bootstrap-flatcar.sh`?

dongsupark (commented):
Thanks for the improvements PR with detailed commit messages.
I think the changes are mostly going into the right direction.

Comments below:

images/capi/Makefile
```diff
@@ -215,7 +216,10 @@ WINDOWS_VERSIONS		:=	windows-2019 windows-2004
 
 # Set Flatcar Container Linux channel and version if not supplied
 FLATCAR_CHANNEL ?= stable
-FLATCAR_VERSION ?= 2765.2.6
+FLATCAR_VERSION ?= 2905.2.3
+ifeq ($(FLATCAR_VERSION),current)
+FLATCAR_VERSION := $(shell hack/image-grok-latest-flatcar-version.sh $(FLATCAR_CHANNEL))
+endif
```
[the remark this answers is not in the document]
jepio: Added both items to `README-flatcar.md`

images/capi/ansible/roles/containerd/tasks/main.yml
jepio: All @pothos's idea 👍

images/capi/ansible/roles/sysprep/tasks/main.yml
jepio: Yes, did that just note.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -21,11 +21,6 @@
 - import_tasks: photon.yml
   when: ansible_os_family == "VMware Photon OS"
 
-- name: Fail if containerd version does not match for Flatcar
-  shell: containerd --version | cut -d\  -f 3
-  register: command_result
-  failed_when: "ansible_os_family == 'Flatcar' and containerd_version not in command_result.stdout"
-
```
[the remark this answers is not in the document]
jepio: Previously there was a mixture of downloaded shims and OS containerd, and it made sense to check that versions match. Now that both shims and containerd come from the same upstream tarball and are placed in `/opt/bin`, the check felt redundant. We also install the systemd file (https://github.com/kubernetes-sigs/image-builder/pull/701/files#diff-5f3a94dda85ffcc60388bcf6dce435f0e7f554e1d72c17d4665abca68e90b5a1) that ensures the `/opt/bin` paths are used for containerd via ansible. So I don't feel like we gain anything by checking again.

t-lo:
@dongsupark @codenrhoden anything left to do for us to un-block merging?

dongsupark:
As I mentioned in the [comment](https://github.com/kubernetes-sigs/image-builder/pull/701#discussion_r715638337), ideally I would like to have an ack from @moshloop .
Not sure how active he is in these days, but anyway we should wait at least for 1 week.
If there is no reply until Friday this week, then I can happily approve this PR.

jepio:
The CI fail was this flake: https://github.com/kubernetes-sigs/image-builder/issues/688. I've pushed a fix (https://github.com/kubernetes-sigs/image-builder/pull/701/commits/88e5e034e292c0248f5d3c2c331c1d65dab9cced)
/retest.

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
```diff
@@ -16,3 +16,37 @@
   set_fact:
     last_log_mode:   "0644"
     machine_id_mode: "0444"
+
+- name: Invalidate the machine-id file for systemd to reevaluate presets
+  file:
+    path: /etc/machine-id
+    state: absent
```
codenrhoden, line 23: Is there something about the "generic" version of this that runs for all OS's that doesn't work for Flatcar?
https://github.com/kubernetes-sigs/image-builder/blob/1c55f366d663baa944423a0734c6fc68c1103c4c/images/capi/ansible/roles/sysprep/tasks/main.yml#L33-L42

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -40,6 +40,7 @@
   loop:
   - { path: /etc/machine-id, state: absent, mode: "{{ machine_id_mode }}" }
   - { path: /etc/machine-id, state: touch,  mode: "{{ machine_id_mode }}" }
+  when: ansible_os_family != "Flatcar"
```
codenrhoden, line 43: Is it the `touch` that makes this not work for Flatcar? Any idea why?

images/capi/packer/azure/packer.json
```diff
@@ -175,6 +175,17 @@
...
       },
       "version": "{{user `goss_version`}}"
+    },
+    {
+      "environment_vars": [
+        "BUILD_NAME={{user `build_name`}}"
+      ],
+      "inline": [
+        "if [ $BUILD_NAME != \"flatcar\" ]; then exit 0; fi",
+        "sudo bash -c \"/usr/share/oem/python/bin/python /usr/share/oem/bin/waagent -force -deprovision+user && sync\""
```
codenrhoden, line 185: Can I get some detail on what this command is doing? It seems weird to add a new command that only runs when *not* flatcar in a PR that is about improving Flatcar.

images/capi/ansible/roles/containerd/tasks/main.yml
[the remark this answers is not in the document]
codenrhoden: @dongsupark Are you satisfied with this explanation?

images/capi/ansible/roles/sysprep/files/etc/hosts
```diff
@@ -1,2 +1,2 @@
+127.0.0.1      localhost localhost.local
```
codenrhoden, line 1: Is this about the port forwarding to etcd?

images/capi/packer/azure/packer.json
codenrhoden: Is this meant to be ` == Flatcar` instead? I feel like it must be, as the other OS's don't have `/usr/share/oem/python` (that I'm aware of)... But, if that's true, it seems like Azure builds should fail at this step. 🤔

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -37,11 +37,12 @@
     group: root
     mode: 0644
 
-- name: Start and Enable kubelet service
+- name: Enable kubelet service
   systemd:
     name: kubelet
     daemon_reload: yes
     enabled: True
+    state: stopped
```
codenrhoden, line 45: I'm definitely confused by this -- it shouldn't be necessary.
We tried to fix the flake that is brought up in https://github.com/kubernetes-sigs/image-builder/issues/688 with https://github.com/kubernetes-sigs/image-builder/pull/583. Specifically, we don't need to care what the state is (it could be running or not), we just need to ensure it is enabled. If we are really seeing more flakes because the service is maybe running, maybe not, then asking Ansible to ensure it is stopped seems like it would just create more flakes. What am I missing?

codenrhoden: Oh wait, I see. The flake is in GOSS, at which point again, the service may or may not be running.

So I guess the question is do we take this change here and explicitly stop the service, or do we fix our GOSS test to test what we really care about -- that the service is enabled (and not test if it is running or not).

codenrhoden: Our GOSS test explicitly checks that the service is enabled and stopped, so this seems like the right thing to do.

https://github.com/kubernetes-sigs/image-builder/blob/1c55f366d663baa944423a0734c6fc68c1103c4c/images/capi/packer/goss/goss-service.yaml#L11

I would prefer if this was in a separate PR, rather than buried with 17 other commits since it's not a Flatcar specific fix, but at the end of the day the commit would be the same and it *is* a separate commit, so thanks for this fix!

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
jepio: An empty `/etc/machine-id` leads to `/etc/machine-id` being regenerated, but is not enough to trigger first-boot semantics of systemd (https://www.freedesktop.org/software/systemd/man/machine-id.html#First%20Boot%20Semantics). Only when `/etc/machine-id` is missing is the next boot considered "first", and we need that to be the case on the next ignition run when the image-build images are provisioned again. One of the things that was broken was that systemd presets (https://www.freedesktop.org/software/systemd/man/systemd.preset.html) were not being (re)evaluated. Flatcar (or rather: ignition) uses those to enable services.

images/capi/ansible/roles/sysprep/tasks/main.yml
jepio: Yes, we want the file to be missing and not just empty (see the comment above). This gives first-boot semantics and causes the next boot to have all "ignition" steps properly applied.

images/capi/packer/azure/packer.json
jepio: The logic is "if not Flatcar then exit", for Flatcar the next line is also evaluated and the command run. See the CI log https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/701/pull-azure-sigs/1445004864366055424/build-log.txt, search for `flatcar ']`.

As to what the command does: Azure docs specify that it's the correct way to deprovision a VM to turn it into an image (https://docs.microsoft.com/en-us/azure/virtual-machines/linux/build-image-with-packer). It removes the provisioning user account and ensures the next time waagent runs it performs provisoning steps. We observed that without it, on next boot `waagent` (Azure linux agent) sees `/var/lib/waagent/provisioned` and doesn't copy over new provisioning data (`ovf-env.xml`) and instead re-applied old provisioning data (from the packer build), resulting in the packer SSH public key being added to the admin user again and old hostname being re-applied.

I think the interaction between cloud-init and waagent is different, this does not appear to be an issue on Ubuntu but with ignition it is.

images/capi/ansible/roles/sysprep/files/etc/hosts
jepio: Yes, i added more context in the commit message https://github.com/kubernetes-sigs/image-builder/pull/701/commits/2effdf64dce2b0c7d846c03438faf5fd659b77d9: containerd can't resolve localhost to the ipv4 address if the order is swapped
```
containerd[1003]: E0922 13:09:39.190414    1003 httpstream.go:257] error forwarding port 2379 to pod b0b39725d5e18ff487dd4f9e9f04d6f84dd33e6c13d2a7a77cbb33694efc924a, uid : failed to execute portforward in network namespace "host": failed to connect to localhost:2379 inside namespace "b0b39725d5e18ff487dd4f9e9f04d6f84dd33e6c13d2a7a77cbb33694efc924a", IPv4: dial tcp4: lookup localhost: No address associated with hostname IPv6 dial tcp6 [::1]:2379: connect: connection refused
```
This wasn't an issue before because the Flatcar image-builder images were accidentally relying on OS provided docker which does not seem to have the same issue.

images/capi/ansible/roles/kubernetes/tasks/main.yml
jepio: You got it: it could be running or not, so better explicitly stop it. I don't know goss, so i don't know if it can be told to not check the `running` state.

I'm happy keeping the commit here, so that we don't hit the flake again, and going forward no other PR will either :)

images/capi/ansible/roles/containerd/tasks/main.yml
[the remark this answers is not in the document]
dongsupark: >  Are you satisfied with this explanation?

@codenrhoden Yes, I am.

images/capi/packer/azure/packer.json
codenrhoden: > The logic is "if not Flatcar then exit", for Flatcar the next line is also evaluated and the command run. See the CI log https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/701/pull-azure-sigs/1445004864366055424/build-log.txt, search for `flatcar ']`.

🤦  Yeah, I should have seen that.

> As to what the command does: Azure docs specify that it's the correct way to deprovision a VM to turn it into an image (https://docs.microsoft.com/en-us/azure/virtual-machines/linux/build-image-with-packer). It removes the provisioning user account and ensures the next time waagent runs it performs provisoning steps. We observed that without it, on next boot `waagent` (Azure linux agent) sees `/var/lib/waagent/provisioned` and doesn't copy over new provisioning data (`ovf-env.xml`) and instead re-applied old provisioning data (from the packer build), resulting in the packer SSH public key being added to the admin user again and old hostname being re-applied.
> 
> I think the interaction between cloud-init and waagent is different, this does not appear to be an issue on Ubuntu but with ignition it is.

Interesting. Similar to the systemd first-boot symantics, I'm wondering whether we should look at this for non-Flatcar (non-Ignition) systems as well. While cloud-init may be side-stepping the issue, it'd be worth checking whether `waagent` in Azure is re-applying old data because we don't de-provision the agent. Seems like a second item to be captured into an issue and explored seperate. 
/cc @jsturtevant @CecileRobertMichon

codenrhoden:
/lgtm
This seems ready to go to me. I'll let this go until EOD to see if anyone else sees something, but nice work @jepio!

PR #713, opened 2021-10-06
Added a WS2022 VHD and SIG target

nick5616:
Adds Windows Server 2022 as a VHD and SIG target. This will allow us to add Windows Server 2022 to the Azure marketplace


jsturtevant:
ubuntu seems to be having trouble with package:

```
build-8tafnb0q/cryptography/setup.py\", line 14, in <module>\n        from setuptools_rust import RustExtension\n    ModuleNotFoundError: No module named 'setuptools_rust'\n 
```

Will try one more time incase it is a temporary issue.
/retest

PR #719, opened 2021-10-12
Allow subset of users to have access to Azure wireserver for Windows

jsturtevant:

#694 resulted in Azure VM's not coming online when provisioned.  This is due to the GuestAgent and cloudbase-init needing access to the wireserver.  

This configured the block rule to apply to everyone but LocalSystem and users in the local Administrator group which includes GuestAgent and cloudbase-init.

**Additional context**

This does mean that if you have an application installed on the VM that is not in the administrator group then you will not be able to access the wireserver. 

`ContainerAdministrator` is considered a part of the Administrator group and therefore containers running as `ContainerAdministrator` would be given access to the wireserver.  We have two options:
- use cni specific acl (https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/1672/commits/dc241572c96d13f2ea2e51b35494a8a0aaf8e543#diff-d0c3cd272af154ea5bc8ae6f19ef250f1ea9c2f6d78d92d7fdc3b3d61ad68f2a)
- create a group and add cloudbase-init too it to remove `containeradministrator`'s access 

**update**: went with option two, adding a group.  The blocks access to the wireserver for `conatineradministrator` and allows for adding permissions for other users/apps.

### DIFF
--- a/images/capi/ansible/windows/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/windows/roles/providers/tasks/azure.yml
@@ -10,9 +10,39 @@
 # limitations under the License.
 ---
 
+- name: Create Azure wireserver access group
+  ansible.windows.win_group:
+    name: WireServerAccess
+    description: Controls access to the Azure WireServer
+
+# AzureGuestAgent and Cloudbase-init need access to wireserver otherwise VM doesn't boot
+# So we give the users access via the firewall security filters
+# https://docs.microsoft.com/en-us/powershell/module/netsecurity/set-netfirewallsecurityfilter
+# 
+# Permissions set on the Firewall rule:
+#   S-1-1-0 is Everyone. We mark this as Allow (A) to ensure the Block is enforced for all users other than on the exception list.
+#   S-1-5-18 is LocalSystem used by AzureGuestAgent.  We mark this as Deny (D) to add to Block exception list. 
+#   We also add the newly created group WireServerAccess to the block exception list and add Cloudbase-init user later.
+#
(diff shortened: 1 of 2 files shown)


marosset:
Nice work!
The new changes (for the changes to create a wire server user group and grant access) LGTM

devigned:
> /retest
> 
> @devigned looks like there might possible flake with the iptables lock
> 
> ```
> �[0;32m    sig-ubuntu-2004-gen2: TASK [providers : Block traffic to 168.63.129.16 port 80 for cve-2021-27075] ***�[0m
> �[0;32m    sig-ubuntu-2004-gen2: fatal: [default]: FAILED! => {"changed": false, "cmd": "/usr/sbin/iptables -t filter -A FORWARD -p tcp -d 168.63.129.16 -j DROP --destination-port 80 -m comment --comment 'block traffic to 168.63.129.16 for cve-2021-27075'", "msg": "Another app is currently holding the xtables lock. Perhaps you want to use the -w option?", "rc": 4, "stderr": "Another app is currently holding the xtables lock. Perhaps you want to use the -w option?\n", "stderr_lines": ["Another app is currently holding the xtables lock. Perhaps you want to use the -w option?"], "stdout": "", "stdout_lines": []}�[0m
> �[0;32m    sig-ubuntu-2004-gen2:�[0m
> ```

I've updated to configure 60s of wait time before the iptables mutation will fail. See: https://github.com/kubernetes-sigs/image-builder/pull/717/files#diff-adaa5bbfc20ce5e21aed6ea1e95dfca1d060eb4cf11622555d5242007ec02798R41

PR #720, opened 2021-10-14
Added ability to configure the packer builder VM to trust custom ca certs on Red Hat

SayaliAsh:
Co-authored-by: Peri Thompson <perit@vmware.com>

When using image builder, an early image builder step is to install and update packages from a yum repo. If this environment has a private yum repo that requires a custom cert, it's not possible. This PR adds the ability to provide a customer root cert for private repositories.

**Additional context**
If this variable is provided it copies the certificate to the target and CA trust is updated. This only works on Red Hat currently.


codenrhoden:
I think the request that @randomvariable made here is valid. Let's make it more flexible by being able to take multiple certs.

SayaliAsh:
> I think the request that @randomvariable made here is valid. Let's make it more flexible by being able to take multiple certs.

Thank you for the suggestion. I am working on including this change.

codenrhoden:
@SayaliAsh Do you still think you'll be able to make the requested change?

SayaliAsh:
> @SayaliAsh Do you still think you'll be able to make the requested change?

yes. sorry this slipped from my head. I am on wedding leaves until Feb, I will finish that right away once I am back.

SOrry for the delay

codenrhoden:
No worries at all, @SayaliAsh ! I just was doing some clean up and wanted to check if this can stay open. Which it can. Enjoy the time away!

PR #724, opened 2021-10-22
kernel values to support --protect-kernel-defaults

xtreme-nitin-ravindran:
Kernel setting to support `--protect-kernel-defaults=true` 

Which issue(s) this PR fixes: 
Fixes k/k issue [#66241](https://github.com/kubernetes/kubernetes/issues/66241)

**Additional context**
`--protect-kernel-defaults` setting is required by both CIS and STIG kubernetes benchmarks
- CIS  rule 4.2.6
- STIG rule V-242434

### DIFF
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -51,6 +51,9 @@
     - { param: net.ipv6.conf.all.forwarding, val: 1 }
     - { param: net.ipv6.conf.all.disable_ipv6, val: 0 }
     - { param: net.ipv4.tcp_congestion_control, val: bbr }
+    - { param: vm.overcommit_memory, val: 1 }
+    - { param: kernel.panic, val: 10 }
+    - { param: kernel.panic_on_oops, val: 1 }
 
 - name: Disable swap memory
   shell: |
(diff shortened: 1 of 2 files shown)


codenrhoden:
If we are making this change common to all Linux kernels (and looks like we are/should), please add these same variables to the GOSS tests, here: https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-kernel-params.yaml

codenrhoden:
/lgtm
/assign @CecileRobertMichon 

@randomvariable Naadir, sounds like you are good with this on the AWS side
@CecileRobertMichon It's worth pointing this out to you as it would change some Linux kernel params for all providers, so I definitely want you to chime in from the Azure side.

images/capi/ansible/roles/node/tasks/main.yml
```diff
@@ -51,6 +51,9 @@
     - { param: net.ipv6.conf.all.forwarding, val: 1 }
     - { param: net.ipv6.conf.all.disable_ipv6, val: 0 }
     - { param: net.ipv4.tcp_congestion_control, val: bbr }
+    - { param: vm.overcommit_memory, val: 1 }
+    - { param: kernel.panic, val: 10 }
+    - { param: kernel.panic_on_oops, val: 1 }
```
jackfrancis, line 56: @alexeldeib this lgtm for a general purpose k8s node OS foundation, wdyt?

codenrhoden:
@CecileRobertMichon @jsturtevant any concerns with this one?

PR #725, opened 2021-10-26
Update Flatcar-related URLs

johananl:

We have some Flatcar-related URLs which point to older Flatcar servers which may not always be up to date, for example when querying for release metadata. This fixes the URLs to point to the current, authoritative servers.

**Additional context**
None


codenrhoden:
/approve
/assign @dongsupark 
An `/lgtm` from @dongsupark should merge this in

PR #729, opened 2021-11-01
Added virtual floppy disk feature to centos-7

spagno:
The PR add the virtual floppy disk support to Centos7 to be used to load the kickstart file instead of using http. It's useful if you want to use image-builder as container

**Additional context**
If you try to use image-builder as container (for example following this doc: https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.4/vmware-tanzu-kubernetes-grid-14/GUID-build-images-index.html#build ), for Centos7 (and other distros like RHEL or Photon) it doesn't work because the created VM tries to connect to packer to http://containerIP:randomPort to obtain the kickstart file. It would work only if the container runs with the "--network=host" option. Ubuntu already uses the virtual floppy disk, why we can use it also for centos?


codenrhoden:
/assign @kkeshavamurthy 
^ Kiran, we were just talking about this last week, and you had pointed out this wasn't working at the time you added for Ubuntu. Can you check this out and see if it works now?
/ok-to-test

@spagno If this is working, let's please add `centos-7` to the list of OVAs that get a test build, by adding it at https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/scripts/ci-ova.sh#L24. This is the feature that has kept it from there being there.

codenrhoden:
Okay, with Kiran's comment [here](https://github.com/kubernetes-sigs/image-builder/pull/730#issuecomment-956517504) I think this is good to go. I'm just going to wait and see if @spagno can add the centos test into the CI script before merge. :)

Huge thanks, @spagno!

kkeshavamurthy:
The floppy side of things work fine. 
But the Centos OVA building is broken I think. @codenrhoden do you know if Centos-7 was building successfully before? 
Because, after the base image is installed, packer waits to get the IP of the VM and times out. I checked the build VM and found that `open-vm-tools` was not installed. I think a repo from where the packages need to be installed is needed here.

codenrhoden:
> @codenrhoden do you know if Centos-7 was building successfully before? Because, after the base image is installed, packer waits to get the IP of the VM and times out. I checked the build VM and found that `open-vm-tools` was not installed. I think a repo from where the packages need to be installed needs to be provided in the ks file (similar to rockylinux).

It was working at one point, for sure. :) But there's a few different variations, depending on whether you are doing a base image first, whether you are using vCenter or a local hypervisor, etc. It's entirely possible something slipped through where CentOS doesn't build on vCenter (but works fine on Fusion) because of the open-vm-tools thing.

We do explicitly install it in the kickstart file: https://github.com/kubernetes-sigs/image-builder/blob/81e0169cd40dffd0afaa9055b7d952d0a5b0a8cf/images/capi/packer/ova/linux/centos/http/7/ks.cfg#L56,
However since that kickstart is installing from `cdrom` only, if that package isn't in the repo that's on the DVD, I think it just skips it and continues. Maybe something changed there between an older 7.x version and where we are at now, or perhaps it just was missed entirely.

kkeshavamurthy:
`However since that kickstart is installing from cdrom only, if that package isn't in the repo that's on the DVD, I think it just skips it and continues. Maybe something changed there between an older 7.x version and where we are at now, or perhaps it just was missed entirely.`
+1
Also you are correct, `open-vm-tools` is not needed for Fusion builder, just vCenter. 

I'm happy to track that issue separately and get this PR in.

codenrhoden:
yeah, if CentOS isn't working on vCenter, we should fix that separately and add it to the CI setup. So let's merge this one in.

Thanks for fixing this, @spagno! A much simpler setup and better experience for all. 🎉 

/approve

spagno:
I also found why open-vm-tools is not installed (and it cause the process remains stuck in "waiting for IP"
the problem is that open-vm-tools is inside the "Update" repo and with the --ignoremissing in the default kickstart we don't have any error. 

I already prepared a different ks which fix that, i wanted to open a dedicated pr for this

regards

PR #730, opened 2021-11-01
Updated to rhel7.9 and added virtual floppy disk support

spagno:
The PR add the virtual floppy disk support to RHEL7 to be used to load the kickstart file instead of using http. It's useful if you want to use image-builder as container.

The PR also updates the RHEL version from 7.7 to 7.9

Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #

Additional context
If you try to use image-builder as container (for example following this doc: https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.4/vmware-tanzu-kubernetes-grid-14/GUID-build-images-index.html#build ), for RHEL7 (and other distros like Centos or Photon) it doesn't work because the created VM tries to connect to packer to http://containerIP:randomPort to obtain the kickstart file. It would work only if the container runs with the "--network=host" option. Ubuntu already uses the virtual floppy disk, why we can use it also for RHEL?


kkeshavamurthy:
I tried this and it works fine. 
Thanks for working on this @spagno 
/ok-to-test

PR #731, opened 2021-11-01
fixed kickstart file for centos-7 and rhel-7 and added centos-7 to ci…

spagno:
…-ova.sh script

This PR will fix the installation of open-vm-tools for centos and RHEL.
open-vm-tools package is in "Update" Repo and with the default --ignoremissing we don't see any error like "open-vm-tools package not found"

**Additional context**
Without open-vm-tools installed, if you try to build centos/rhel OVA the process will remain stuck in "Waiting for IP address"

### DIFF
--- a/images/capi/packer/ova/linux/centos/http/7/ks.cfg
+++ b/images/capi/packer/ova/linux/centos/http/7/ks.cfg
@@ -53,7 +53,6 @@ part / --grow --asprimary --fstype=ext4 --label=slash
 openssh-server
 sed
 sudo
-open-vm-tools
 
 # Remove unnecessary firmware
 -*-firmware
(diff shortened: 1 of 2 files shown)


kkeshavamurthy:
Confirmed that centos-7 base OVA builds correctly on vCenter.

kkeshavamurthy:
Seems like Ubuntu-2004 flaked. Centos-7 passed 👍 
/test pull-ova-all

codenrhoden:
With this change, I feel like we should get rid of `--ignoremissing`, because if something *is* missing, it's definitely a problem. Thoughts?

spagno:
if you remove --ignoremissing and there's something which actually missing, the process will stuck waiting for an user input

codenrhoden:
> if you remove --ignoremissing and there's something which actually missing, the process will stuck waiting for an user input

Ah, thanks for the reminder. Been a while since I read the docs there, but just did again. :)

/approve

PR #732, opened 2021-11-01
Add networkd-dispatcher script to access NTP servers provided by DHCP

### LINKED ISSUE #676: DHCP Option 42 doesn't working on Ubuntu 20.04

Chrony will only have its time source configured via DHCP using dhclient, for which there is a script in /etc/dhcp/dhclient-exit-hooks.d/chrony 

The dhclient script for DHCP on Ubuntu is incompatible with chrony and only supports timesyncd.

Attempting to run dhclient manually gets:

```shell
root@wl-104-md-0-5ccdb4fffd-lv64j:/usr/lib# dhclient
RTNETLINK answers: File exists
Failed to try-restart systemd-timesyncd.service: Unit systemd-timesyncd.service is masked.
```

but chrony does get the time source. Note that the time source gets appended to the default ubuntu list, which might be not ideal.

The issue is that Ubuntu 20.04 uses systemd-networkd, programmed by netplan to configure networking. systemd includes its own DHCP client, which means that the dhclient scripts are no longer used. This would mean that NTP DHCP support would only be supported via systemd-timesyncd and not chrony.

This runs counter us standardising on chrony because we wanted consistent cloud-init support.

There's potentially 2 options:

1. Get chrony working somehow with systemd-networkd.
2. Revert to systemd-timesyncd, but verify cloud-init works

**Environment:**

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`): Ubuntu 20.04
- Packer Version:
- Packer Provider:
- Ansible Version: 

/kind bug

SanikaGawhane:
Chrony is unable to get time source from DHCP provided NTP servers.
To fix this, we use [networkd-dispatcher](https://gitlab.com/craftyguy/networkd-dispatcher) to add a new dispatcher script that parses DHCP provided NTP servers from the networkctl status output and add them to chrony.

This fix is for Ubuntu-2004 and Photon-3 images. `networkd-dispatcher` is already packaged for Ubuntu-2004 but needs to be installed from source for Photon.
 
Fixes #676

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -57,11 +57,13 @@ common_photon_rpms:
 - audit
 - conntrack-tools
 - chrony
+- dbus-python3
 - distrib-compat
 - ebtables
 - jq
 - net-tools
 - openssl-c_rehash
+- python3-pygobject
 - python-netifaces
 - python3-pip
 - python-requests
--- a/images/capi/ansible/roles/providers/files/usr/libexec/chrony-helper
+++ b/images/capi/ansible/roles/providers/files/usr/libexec/chrony-helper
@@ -1,0 +1,251 @@
+#!/bin/bash
+# This script configures running chronyd to use NTP servers obtained from
+# DHCP and _ntp._udp DNS SRV records. Files with servers from DHCP are managed
+# externally (e.g. by a dhclient script). Files with servers from DNS SRV
+# records are updated here using the dig utility. The script can also list
+# and set static sources in the chronyd configuration file.
+
+chronyc=/usr/bin/chronyc
+chrony_conf=/etc/chrony.conf
+chrony_service=chronyd.service
+helper_dir=/var/run/chrony-helper
+added_servers_file=$helper_dir/added_servers
+
+network_sysconfig_file=/etc/sysconfig/network
+dhclient_servers_files=/var/lib/dhclient/chrony.servers.*
+dnssrv_servers_files=$helper_dir/dnssrv@*
+dnssrv_timer_prefix=chrony-dnssrv@
+
+chrony_command() {
+    $chronyc -a -n -m "$1"
+}
+
+is_running() {
+    chrony_command "tracking" &> /dev/null
+}
+
+is_update_needed() {
+    for file in $dhclient_servers_files $dnssrv_servers_files \
+            $added_servers_file; do
+        [ -e "$file" ] && return 0
+    done
+    return 1
+}
+
+update_daemon() {
+    local all_servers_with_args all_servers added_servers
+
+    if ! is_running; then
+        rm -f $added_servers_file
+        return 0
+    fi
+
+    all_servers_with_args=$(
+        cat $dhclient_servers_files $dnssrv_servers_files 2> /dev/null)
+
+    all_servers=$(
+        echo "$all_servers_with_args" |
+            while read server serverargs; do
+                echo "$server"
+            done | sort -u)
+    added_servers=$( (
+        cat $added_servers_file 2> /dev/null
+        echo "$all_servers_with_args" |
+            while read server serverargs; do
+                [ -z "$server" ] && continue
+                chrony_command "add server $server $serverargs" &> /dev/null &&
+                    echo "$server"
+            done) | sort -u)
+
+    comm -23 <(echo -n "$added_servers") <(echo -n "$all_servers") |
+        while read server; do
+            chrony_command "delete $server" &> /dev/null
+        done
+
+    added_servers=$(comm -12 <(echo -n "$added_servers") <(echo -n "$all_servers"))
+
+    [ -n "$added_servers" ] && echo "$added_servers" > $added_servers_file ||
+        rm -f $added_servers_file
+}
+
+get_dnssrv_servers() {
+    local name=$1
+
+    if ! command -v dig &> /dev/null; then
+        echo "Missing dig (DNS lookup utility)" >&2
+        return 1
+    fi
+
+    (
+        . $network_sysconfig_file &> /dev/null
+
+        output=$(dig "$name" srv +short +ndots=2 +search 2> /dev/null)
+        [ $? -ne 0 ] && return 0
+
+        echo "$output" | while read prio weight port target; do
+            server=${target%.}
+            [ -z "$server" ] && continue
+            echo "$server port $port ${NTPSERVERARGS:-iburst}"
+        done
+    )
+}
+
+check_dnssrv_name() {
+    local name=$1
+
+    if [ -z "$name" ]; then
+        echo "No DNS SRV name specified" >&2
+        return 1
+    fi
+
+    if [ "${name:0:9}" != _ntp._udp ]; then
+        echo "DNS SRV name $name doesn't start with _ntp._udp" >&2
+        return 1
+    fi
+}
+
+update_dnssrv_servers() {
+    local name=$1
+    local srv_file=$helper_dir/dnssrv@$name servers
+
+    check_dnssrv_name "$name" || return 1
+
+    servers=$(get_dnssrv_servers "$name")
+    [ -n "$servers" ] && echo "$servers" > "$srv_file" || rm -f "$srv_file"
+}
+
+set_dnssrv_timer() {
+    local state=$1 name=$2
+    local srv_file=$helper_dir/dnssrv@$name servers
+    local timer=$dnssrv_timer_prefix$(systemd-escape "$name").timer
+
+    check_dnssrv_name "$name" || return 1
+
+    if [ "$state" = enable ]; then
+        systemctl enable "$timer"
+        systemctl start "$timer"
+    elif [ "$state" = disable ]; then
+        systemctl stop "$timer"
+        systemctl disable "$timer"
+        rm -f "$srv_file"
+    fi
+}
+
+list_dnssrv_timers() {
+    systemctl --all --full -t timer list-units | grep "^$dnssrv_timer_prefix" | \
+            sed "s|^$dnssrv_timer_prefix\(.*\)\.timer.*|\1|" |
+        while read -r name; do
+            systemd-escape --unescape "$name"
+        done
+}
+
+prepare_helper_dir() {
+    mkdir -p $helper_dir
+    exec 100> $helper_dir/lock
+    if ! flock -w 20 100; then
+        echo "Failed to lock $helper_dir" >&2
+        return 1
+    fi
+}
+
+is_source_line() {
+    local pattern="^[ \t]*(server|pool|peer|refclock)[ \t]+[^ \t]+"
+    [[ "$1" =~ $pattern ]]
+}
+
+list_static_sources() {
+    while read line; do
+        is_source_line "$line" && echo "$line" || :
+    done < $chrony_conf
+}
+
+set_static_sources() {
+    local new_config tmp_conf
+
+    new_config=$(
+        sources=$(
+            while read line; do
+                is_source_line "$line" && echo "$line"
+            done)
+
+        while read line; do
+            if ! is_source_line "$line"; then
+                echo "$line"
+                continue
+            fi
+
+            tmp_sources=$(
+                local removed=0
+
+                echo "$sources" | while read line2; do
+                    [ "$removed" -ne 0 -o "$line" != "$line2" ] && \
+                        echo "$line2" || removed=1
+                done)
+
+            [ "$sources" == "$tmp_sources" ] && continue
+            sources=$tmp_sources
+            echo "$line"
+        done < $chrony_conf
+
+        echo "$sources"
+    )
+
+    tmp_conf=${chrony_conf}.tmp
+
+    cp -a $chrony_conf $tmp_conf &&
+        echo "$new_config" > $tmp_conf &&
+        mv $tmp_conf $chrony_conf || return 1
+
+    systemctl try-restart $chrony_service
+}
+
+print_help() {
+    echo "Usage: $0 COMMAND"
+    echo
+    echo "Commands:"
+    echo "	update-daemon"
+    echo "	update-dnssrv-servers NAME"
+    echo "	enable-dnssrv NAME"
+    echo "	disable-dnssrv NAME"
+    echo "	list-dnssrv"
+    echo "	list-static-sources"
+    echo "	set-static-sources < sources.list"
+    echo "	is-running"
+    echo "	command CHRONYC-COMMAND"
+}
+
+case "$1" in
+    update-daemon|add-dhclient-servers|remove-dhclient-servers)
+        is_update_needed || exit 0
+        prepare_helper_dir && update_daemon
+        ;;
+    update-dnssrv-servers)
+        prepare_helper_dir && update_dnssrv_servers "$2" && update_daemon
+        ;;
+    enable-dnssrv)
+        set_dnssrv_timer enable "$2"
+        ;;
+    disable-dnssrv)
+        set_dnssrv_timer disable "$2" && prepare_helper_dir && update_daemon
+        ;;
+    list-dnssrv)
+        list_dnssrv_timers
+        ;;
+    list-static-sources)
+        list_static_sources
+        ;;
+    set-static-sources)
+        set_static_sources
+        ;;
+    is-running)
+        is_running
+        ;;
+    command|forced-command)
+        chrony_command "$2"
+        ;;
+    *)
+        print_help
+        exit 2
+esac
+
+exit $?
--- a/images/capi/ansible/roles/providers/tasks/main.yml
+++ b/images/capi/ansible/roles/providers/tasks/main.yml
@@ -33,6 +33,16 @@
   when: packer_builder_type is search('qemu') and 
     build_target is search('raw')
 
+- include_tasks: vmware-photon.yml
+  when: (packer_builder_type is search('vmware') or
+    packer_builder_type is search('vsphere')) and
+    ansible_os_family == "VMware Photon OS"
+
+- include_tasks: vmware-ubuntu.yml
+  when: (packer_builder_type is search('vmware') or
+    packer_builder_type is search('vsphere')) and
+    ansible_distribution == "Ubuntu"
+
 # Create a boot order configuration
 # b/w containerd and cloud final, cloud config services
 
--- a/images/capi/ansible/roles/providers/tasks/vmware-photon.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware-photon.yml
@@ -1,0 +1,70 @@
+# Copyright 2021 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+- name: Install networkd-dispatcher service (Download from source)
+  unarchive:
+    src: "{{ networkd_dispatcher_download_url }}"
+    dest: /tmp
+    remote_src: yes
+
+- name: Create needed directories
+  file:
+    path: "{{ item.dir }}"
+    state: directory
+  loop:
+    - { dir: /etc/conf.d }
+    - { dir: /etc/networkd-dispatcher/carrier.d }
+    - { dir: /etc/networkd-dispatcher/configured.d }
+    - { dir: /etc/networkd-dispatcher/configuring.d }
+    - { dir: /etc/networkd-dispatcher/degraded.d }
+    - { dir: /etc/networkd-dispatcher/dormant.d }
+    - { dir: /etc/networkd-dispatcher/no-carrier.d }
+    - { dir: /etc/networkd-dispatcher/off.d }
+    - { dir: /etc/networkd-dispatcher/routable.d }
+
+- name: Install networkd-dispatcher service (Move files)
+  command: mv "{{ item.src }}" "{{ item.dest }}"
+  loop:
+    - { src: /tmp/networkd-dispatcher-2.1/networkd-dispatcher, dest: /usr/bin }
+    - { src: /tmp/networkd-dispatcher-2.1/networkd-dispatcher.service, dest: /etc/systemd/system }
+    - { src: /tmp/networkd-dispatcher-2.1/networkd-dispatcher.conf, dest: /etc/conf.d }
+
+- name: Install networkd-dispatcher service (Run networkd-dispatcher)
+  systemd:
+    name: networkd-dispatcher
+    state: started
+    enabled: yes
+
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: a+x
+  vars:
+    server_dir: "/var/lib/dhclient"
+    chrony_helper_dir: "/usr/libexec"
+  loop:
+    - { src: files/etc/networkd-dispatcher/routable.d/20-chrony.j2, dest: /etc/networkd-dispatcher/routable.d/20-chrony }
+    - { src: files/etc/networkd-dispatcher/off.d/20-chrony.j2, dest: /etc/networkd-dispatcher/off.d/20-chrony }
+    - { src: files/etc/networkd-dispatcher/no-carrier.d/20-chrony.j2, dest: /etc/networkd-dispatcher/no-carrier.d/20-chrony }
+
+- name: Copy chrony-helper script
+  copy:
+    src: files/usr/libexec/chrony-helper
+    dest: /usr/libexec/chrony-helper
+    owner: root
+    group: root
+    mode: a+x
--- a/images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
@@ -1,0 +1,27 @@
+# Copyright 2021 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: a+x
+  vars:
+    server_dir: "/var/lib/dhcp"
+    chrony_helper_dir: "/usr/lib/chrony"
+  loop:
+    - { src: files/etc/networkd-dispatcher/routable.d/20-chrony.j2, dest: /etc/networkd-dispatcher/routable.d/20-chrony }
+    - { src: files/etc/networkd-dispatcher/off.d/20-chrony.j2, dest: /etc/networkd-dispatcher/off.d/20-chrony }
+    - { src: files/etc/networkd-dispatcher/no-carrier.d/20-chrony.j2, dest: /etc/networkd-dispatcher/no-carrier.d/20-chrony }
--- a/images/capi/ansible/roles/providers/defaults/main.yml
+++ b/images/capi/ansible/roles/providers/defaults/main.yml
@@ -13,3 +13,4 @@
 # limitations under the License.
 ---
 cloud_init_deb_download_url: "https://launchpad.net/ubuntu/+source/cloud-init/21.1-19-gbad84ad4-0ubuntu1~20.04.2/+build/21434629/+files/cloud-init_21.1-19-gbad84ad4-0ubuntu1~20.04.2_all.deb"
+networkd_dispatcher_download_url: "https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.bz2"
--- a/images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/no-carrier.d/20-chrony.j2
+++ b/images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/no-carrier.d/20-chrony.j2
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# This is a networkd-dispatcher script for chronyd to handle its NTP
+# sources. It sets the NTP sources online or offline when a network
+# interface is configured or removed. On DHCP change, chrony will
+# update its NTP sources passed from DHCP options.
+
+export LC_ALL=C
+
+DHCP_SERVER_FILE={{ server_dir }}/chrony.servers.$IFACE
+
+clear_servers_from_dhcp() {
+    if [ -f "$DHCP_SERVER_FILE" ]; then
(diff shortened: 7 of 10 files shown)


images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -82,3 +84,4 @@ sysctl_conf_file: "{{ '/etc/sysctl.d/99-sysctl.conf' if ansible_os_family == 'VM
 pause_image: "k8s.gcr.io/pause:3.6"
 containerd_additional_settings: null
 leak_local_mdns_to_dns: false
+networkd_dispatcher: "https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.bz2"
```
kkeshavamurthy, line 87: Would it be better to rename it to something like `networkd_dispatcher_download_url`?

images/capi/ansible/roles/node/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,47 @@
...
+
+if [ "$STATE" = "routable" ]; then
+    add_servers_from_dhcp
+elif [ "$STATE" = "off" ]; then
+    clear_servers_from_dhcp
+fi
+
+if [ "$STATE" = "routable" ] || [ "$STATE" = "off" ]; then
+    chronyc onoffline > /dev/null 2>&1
+fi
```
kkeshavamurthy, line 45: Can you pls add a comment why this is needed.

images/capi/ansible/roles/node/tasks/ubuntu.yml
```diff
@@ -0,0 +1,34 @@
...
+    dest: /etc/networkd-dispatcher/routable.d/20-chrony
+    owner: root
+    group: root
+    mode: a+x
+- name: Point server_dir in chrony-helper to DHCP provided NTP servers
+  lineinfile:
+    path: /usr/lib/chrony/chrony-helper
+    regexp: 'dhclient_servers_files=/var/lib/dhclient/chrony.servers.*'
+    line: 'dhclient_servers_files=$CHRONY_SERVER_DIR/chrony.servers.*'
+    backrefs: yes
```
kkeshavamurthy, line 28: Can we re-use the helper script added for photon?

images/capi/ansible/roles/node/tasks/ubuntu.yml
```diff
@@ -0,0 +1,35 @@
...
+    group: root
+    mode: a+x
+- name: Copy chrony-helper script
+  copy:
+    src: files/usr/lib/chrony/chrony-helper
+    dest: /usr/lib/chrony/chrony-helper
+    owner: root
+    group: root
+    mode: a+x
+- name: Restart systemd-networkd service
```
codenrhoden, line 30: I doubt this part is needed -- what we are doing here doens't really matter until the image is booted, so why do we restart this service?

images/capi/ansible/roles/node/tasks/main.yml
```diff
@@ -18,6 +18,9 @@
 - import_tasks: amazonLinux2.yml
   when: ansible_distribution == "Amazon"
 
+- import_tasks: ubuntu.yml
```
codenrhoden, line 21: I think we only want to do these changes for OVAs, right? If so, I would expect changes to go in `roles/providers`, not `roles/node`. As it is, this will apply to all Ubuntu images, including Azure and AWS.

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -35,3 +35,46 @@
     sysctl_set: yes
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
+
+- name: Install networkd-dispatcher service (Download from source)
+  unarchive:
+    src: "{{ networkd_dispatcher_download_url }}"
+    dest: /tmp
+    remote_src: yes
+- name: Install networkd-dispatcher service (Create directory for networkd-dispatcher.conf script)
```
codenrhoden, line 44: nit: we've been consistent about having a newline between each Ansible task.

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -35,3 +35,46 @@
...
+    group: root
+    mode: a+x
+- name: Copy chrony-helper script
+  copy:
+    src: files/usr/lib/chrony/chrony-helper
+    dest: /usr/lib/chrony/chrony-helper
+    owner: root
+    group: root
+    mode: a+x
+- name: Restart systemd-networkd service
```
codenrhoden, line 75: I don't think we need to restart the service during this installation phase.

images/capi/ansible/roles/node/tasks/photon.yml
```diff
@@ -35,3 +35,46 @@
     sysctl_set: yes
     reload: yes
     sysctl_file: "{{ sysctl_conf_file }}"
+
+- name: Install networkd-dispatcher service (Download from source)
```
codenrhoden, line 39: While Photon is currently only in the OVA builder, I think this logic should be in the vmware-specific provider role instead.

It would be totally fine to put this in a `vmware-photon.yml` file and import that from `vmware.yml` though.

images/capi/ansible/roles/node/tasks/ubuntu.yml
```diff
@@ -0,0 +1,35 @@
+# Copyright 2020 The Kubernetes Authors.
```
codenrhoden, line 1: as a new file, this would be (c) 2021.

I think this should be in a vmware-ubunut.yml file in the providers role.

codenrhoden (commented):
I'm going to need to some time to dig into this `chrony-helper` thing, because it confuses me and I wasn't expecting it to be involved (until you gave me a heads up yesterday).

What changes to it are needed?

On Ubuntu, I see that it is installed with the DEB, and I see that it is referenced by the systemd unit file.

On Photon, it is not included in the RPM, so it would be totally new.

I was expecting the dispatcher scripts to do everything with `chronyc`, so anything you can help with there to explain with the `chrony-helper` script is needed (and for Ubuntu, why the standard one doesn't work` would be really helpful.

images/capi/ansible/roles/node/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,49 @@
...
+export LC_ALL=C
+
+# chrony_server_dir is read by `chrony-helper`
+export CHRONY_SERVER_DIR=/var/lib/chrony-servers
+DHCP_SERVER_FILE=$CHRONY_SERVER_DIR/chrony.servers.$IFACE
+
+mkdir -p "$CHRONY_SERVER_DIR"
+
+# Read DHCP provided NTP servers
+networkctl status | grep -A3 'NTP' | sed 's/NTP: //' > $CHRONY_SERVER_DIR/temp-ntp-servers
```
codenrhoden, line 17: One other question.  Why did we end up running this command directly and using `grep` rather than parsing `$json` and using `jq`?

images/capi/ansible/roles/node/tasks/ubuntu.yml
SanikaGawhane: Makes sense. Removing this step.

images/capi/ansible/roles/node/files/etc/networkd-dispatcher/routable.d/20-chrony
SanikaGawhane: In the networkd-dispatcher provided `$json`, NTP is inside a top-level key `Search D`. The part of the output of `echo $json | jq .` that has NTP info looks something like this - 
```
  "Search D": [
    "ains: <something>.vmware.com",
    "vmware.com",
    "NTP: 10.128.152.81",
    "10.188.26.119",
    "10.166.1.120",
    "10.84.55.42"
  ]
  ```
  Looking for `NTP` returns only the IP listed as a value of key `NTP` and ignores the rest of the servers because of the way it is structured. 
  Hence, had to revert to running the command directly and grep-ing from it.

Comment (codenrhoden) [reply images/capi/ansible/roles/node/files/etc/networkd-dispatcher/routable.d/20-chrony:17]: Hmm, this example helps, but it makes me see that the solution here isn't generic enough yet.  I think you've used `-A3` because the infra you are working on happens to provide 4 NTP servers (which is a lot).  What if it was only 1 or 2?

It's really disappointing that `$json` didn't instead report something like:

```json
"NTP": [ "ip1", "ip2", "..."],
```

Guess we aren't that lucky. I'll have to look and see if there is an additional thing we can do here.

images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -57,11 +57,14 @@ common_photon_rpms:
 - audit
 - conntrack-tools
 - chrony
+- dbus-python3
 - distrib-compat
 - ebtables
 - jq
 - net-tools
 - openssl-c_rehash
+- open-vm-tools
```
codenrhoden, line 66: Please rebase when you can - this PR doesn't need to add `open-vm-tools` back in here.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
```
codenrhoden, line 1: Okay, i got a chance to play with this script, and I actually don't run into any issues with `$json` not having a clean `NTP` field. In my home setup, I see this:

```json
    "NTP": [
        "216.229.0.49",
        "216.239.35.0"
    ],
```

So, I'd like to offer an alternative approach to this script, so that you can test it and see if it works. Will put it at the bottom, and put some explanations of things attached to this one.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,49 @@
...
+
+# This is a networkd-dispatcher script for chronyd to handle its NTP
+# sources. It sets the NTP sources online or offline when a network
+# interface is configured or removed. On DHCP change, chrony will
+# update its NTP sources passed from DHCP options.
+
+export LC_ALL=C
+
+# chrony_server_dir is read by `chrony-helper`
+export CHRONY_SERVER_DIR=/var/lib/chrony-servers
```
codenrhoden, line 11: I finally understand what's going on with `chrony-helper`. I do think we need to use it, but I think (for Ubuntu at least) we can use it unmodified. I think your changes here were because you were trying to avoid writing files to `/var/lib/dhcp` because that was remnant of `dhclient`? Even if it is, that directory exists out of the box on Ubuntu, so I don't think we need to modify things. The script I'll paste here works with an unmodified `chrony-helper`.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,49 @@
...
+    fi
+}
+
+if [ "$STATE" = "routable" ]; then
+    add_servers_from_dhcp
+elif [ "$STATE" = "off" ]; then
+    clear_servers_from_dhcp
+fi
+
+if [ "$STATE" = "routable" ] || [ "$STATE" = "off" ]; then
```
codenrhoden, line 43: From my testing, I didn't see any instances of this script running with `$STATE` set to anything other than `routable`, because this script lives in the `routable.d` directory. If `$STATE == off`, then it would run scripts in `off.d` instead, so I don't think this logic is necessary.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/routable.d/20-chrony
```diff
@@ -0,0 +1,49 @@
...
+    # The onoffline command tells chronyd to switch all sources to
+    # the online (routable) or offline (off) status according to the current network configuration.
+    chronyc onoffline > /dev/null 2>&1
+fi
+
+exit 0
```
codenrhoden, line 49: Okay, here is what I tested with... Please try it out on your end:

`routable.d/20-chrony`
```sh
#!/bin/bash

# This is a networkd-dispatcher script for chronyd to handle its NTP
# sources. It sets the NTP sources online or offline when a network
# interface is configured or removed. On DHCP change, chrony will
# update its NTP sources passed from DHCP options.

export LC_ALL=C

DHCP_SERVER_FILE=/var/lib/dhcp/chrony.servers.$IFACE

add_servers_from_dhcp() {
    if [ -f "$DHCP_SERVER_FILE" ]; then
        rm -f "$DHCP_SERVER_FILE"
    fi
    echo "$json" | jq -r 'select(.NTP !=null) .NTP[]' >> $DHCP_SERVER_FILE
    /usr/lib/chrony/chrony-helper update-daemon || :
}

if [ "$STATE" = "routable" ]; then
    add_servers_from_dhcp
    # The onoffline command tells chronyd to switch all sources to
    # the online (routable) or offline (off) status according to the current network configuration.
    chronyc onoffline > /dev/null 2>&1
fi

exit 0
```

I also added two scripts in `off.d` and `no-carrier.d` to remove NTP servers received via DHCP. 

`off.d/20-chrony`
```sh
#!/bin/bash

# This is a networkd-dispatcher script for chronyd to handle its NTP
# sources. It sets the NTP sources online or offline when a network
# interface is configured or removed. On DHCP change, chrony will
# update its NTP sources passed from DHCP options.

export LC_ALL=C

DHCP_SERVER_FILE=/var/lib/dhcp/chrony.servers.$IFACE

clear_servers_from_dhcp() {
    if [ -f "$DHCP_SERVER_FILE" ]; then
        rm -f "$DHCP_SERVER_FILE"
        /usr/lib/chrony/chrony-helper update-daemon || :
    fi
}

if [ "$STATE" = "off" ]; then
    clear_servers_from_dhcp
    # The onoffline command tells chronyd to switch all sources to
    # the online (routable) or offline (off) status according to the current network configuration.
    chronyc onoffline > /dev/null 2>&1
fi

exit 0
```

`no-carrier.d/20-chrony`
```sh
#!/bin/bash

# This is a networkd-dispatcher script for chronyd to handle its NTP
# sources. It sets the NTP sources online or offline when a network
# interface is configured or removed. On DHCP change, chrony will
# update its NTP sources passed from DHCP options.

export LC_ALL=C

DHCP_SERVER_FILE=/var/lib/dhcp/chrony.servers.$IFACE

clear_servers_from_dhcp() {
    if [ -f "$DHCP_SERVER_FILE" ]; then
        rm -f "$DHCP_SERVER_FILE"
        /usr/lib/chrony/chrony-helper update-daemon || :
    fi
}

if [ "$STATE" = "no-carrier" ]; then
    clear_servers_from_dhcp
    # The onoffline command tells chronyd to switch all sources to
    # the online (routable) or offline (off) status according to the current network configuration.
    chronyc onoffline > /dev/null 2>&1
fi

exit 0
```

I tested the `no-carrier` state by pulling the NIC from a running VM. I didn't come in over a serial console, so I wasn't able to test the `off` state...

I think checking the `$STATE` variable in these scripts at all is probably redundant, but am not positive about it. I always want to look a bit more if its really necessary to run `chrony-helper` followed by `chronyc onoffline` or not.

images/capi/ansible/roles/providers/files/usr/lib/chrony/chrony-helper
```diff
@@ -0,0 +1,251 @@
+#!/bin/bash
```
codenrhoden, line 1: So, after digging a bit more, I get why this is here. Since photon doesn't package this wrapper, we will need to carry it in order to get support for multiple chrony conf files. I still need to check if CentOS packages this file, and if it does, we should probably use that as a base for Photon rather than the Debian version.

The dispatcher scripts I've pasted above all work with an unmodified version of this script on Ubuntu. I don't think we need to change it.

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+
+---
+
+- name: Install networkd-dispatcher service (Download from source)
+  unarchive:
+    src: "{{ networkd_dispatcher_download_url }}"
+    dest: /tmp
+    remote_src: yes
+- name: Install networkd-dispatcher service (Create directory for networkd-dispatcher.conf script)
```
codenrhoden, line 22: nit: still would like to see newlines between each of these stanzas

images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
```diff
@@ -0,0 +1,29 @@
...
+
+---
+- name: Copy 20-chrony script to add DHCP provided NTP servers
+  copy:
+    src: files/etc/networkd-dispatcher/routable.d/20-chrony
+    dest: /etc/networkd-dispatcher/routable.d/20-chrony
+    owner: root
+    group: root
+    mode: a+x
+- name: Copy chrony-helper script
```
codenrhoden, line 23: If the scripts I've offered work, this step isn't needed for Ubuntu.

codenrhoden (commented):
Please take the scripts I'm suggested for a testdrive and see if they work in your environment. I believe they work on Ubuntu without any modifcation to `chrony-helper`, and also will work with any number of NTP servers present within the DHCP options. But I've only tested one env so far...

I need to check on CentOS and Photon still.

images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -57,11 +57,14 @@ common_photon_rpms:
...
+- python3-gobject
```
codenrhoden, line 67: Does `python3-gobject` exist? I don't see it in https://github.com/vmware/photon/tree/3.0/SPECS

Using Photon, I get this:

```
# tdnf install python3-gobject
python3-gobject package not found or not installed
Error(1011) : No matching packages
```

There is this: https://github.com/vmware/photon/blob/3.0/SPECS/pygobject

What are the actual dependencies here on Photon?

codenrhoden: The `networkd-dispatcher` script is using Python3, but doing `tdnf install pygobject` just makes `from gi.repository import GLib as glib` work on Python 2.  So I'm still trying to sort this out on Photon 3.

codenrhoden:
Okay, so I was able to get this to work on Photon as well with the following changes....

Don't install `python3-gobject`, install `python3-pygobject`.
Install the `chrony-helper` script from CentOS (with I pulled out of `chrony-3.4-1.el7.x86_64.rpm` to `/usr/libexec/chrony-helper`.

Change all references in the `networkd-dispatcher` scripts from `/var/lib/dhcp` to `/var/lib/dhclient`.

So here's what I think you should do...

Make the `networkd-dispatcher` scripts Jinja templates, and on Ubuntu, use `/var/lib/dhcp`, but on Red Hat/Photon families use `/var/lib/dhclient`

The `chrony-helper` script should be the EL version, not debian, as it's not packaged in CentOS or Photon. The defaults out of networkd-dispatcher reference this file at `/usr/libexec/chrony-helper`, not at `/usr/lib/chrony/chrony-helper`, so let's keep that instead.

In general, we need OS-specific handling of the following:

- `chrony-helper` (only installed on Photon and CentOS)
- All the `networkd-dispatcher` scripts (switch between `/var/lib/dhcp` and `/var/lib/dhclient` depending on Ubuntu vs CentOS/Photon)

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/routable.d/20-chrony
SanikaGawhane: Tested on my setup that these scripts work as expected for Ubuntu and Photon.

codenrhoden:
This is failing in the Ansible YAML:

```
The offending line appears to be:\n\n  template:\n    src: {{ src }}\n          ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}v
```

You should be able to verify that this builds successfully locally, so do please try to build both Ubuntu and Photon OVAs to verify the builds are successful. 

I'll take a look at the other changes here next.

images/capi/ansible/roles/providers/tasks/main.yml
```diff
@@ -33,6 +33,16 @@
   when: packer_builder_type is search('qemu') and 
     build_target is search('raw')
 
+- include_tasks: vmware-photon.yml
+  when: (packer_builder_type is search('vmware') or
+        packer_builder_type is search('vsphere')) and
```
codenrhoden, line 38: nit: let's have this indentation match the rest of the file.

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,62 @@
...
+    remote_src: yes
+
+- name: Install networkd-dispatcher service (Create directory for networkd-dispatcher.conf script)
+  file:
+    path: /etc/conf.d
+    state: directory
+
+- name: Make directory structure for drop-in files
+  command: mkdir -p /etc/networkd-dispatcher/{routable,dormant,no-carrier,off,carrier,degraded,configuring,configured}.d
+- name: Install networkd-dispatcher service (Move files)
```
codenrhoden, line 30: nit: newline above this line

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: {{ src }}
+    dest: {{ dest }}
+    mode: a+x
+  vars:
+    server_dir: "/var/lib/dhclient"
+    chrony_helper_dir: "/usr/libexec/"
+  loop:
+    - { src: files/etc/networkd-dispatcher/routable.d/20-chrony.j2, dest: /etc/networkd-dispatcher/routable.d/20-chrony }
```
codenrhoden, line 52: no need for this three to have different numbers (e.g. 20, 21, 22). They can all be 20. (and the number isn't important for us anyways, but 20 is fine).

images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: {{ src }}
+    dest: {{ dest }}
+    mode: a+x
+  vars:
+    server_dir: "/var/lib/dhcp"
+    chrony_helper_dir: "/usr/lib/chrony"
+  loop:
+    - { src: files/etc/networkd-dispatcher/routable.d/20-chrony.j2, dest: /etc/networkd-dispatcher/routable.d/20-chrony }
```
codenrhoden, line 25: same here, just use 20.

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,62 @@
...
+    enabled: yes
+
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: {{ src }}
+    dest: {{ dest }}
+    mode: a+x
+  vars:
+    server_dir: "/var/lib/dhclient"
+    chrony_helper_dir: "/usr/libexec/"
```
codenrhoden, line 50: In the ubuntu file down below, you don't have a trailing slash for this path. I think we can drop it here too to be consistent.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/no-carrier.d/22-chrony.j2
```diff
@@ -0,0 +1,26 @@
...
+# update its NTP sources passed from DHCP options.
+
+export LC_ALL=C
+
+DHCP_SERVER_FILE={{ server_dir }}/chrony.servers.$IFACE
+
+clear_servers_from_dhcp() {
+    if [ -f "$DHCP_SERVER_FILE" ]; then
+        rm -f "$DHCP_SERVER_FILE"
+        /{{ chrony_helper_dir }}/chrony-helper update-daemon || :
```
codenrhoden, line 15: I think it's weird that the `{{ server_dir }}` above doesn't have a beginning `/`, but this one does. Looking at the Ansible role, both vars are given as full paths, so I don't think the leading `/` is necessary here.

images/capi/ansible/roles/providers/files/etc/networkd-dispatcher/off.d/21-chrony.j2
```diff
@@ -0,0 +1,26 @@
...
+# update its NTP sources passed from DHCP options.
+
+export LC_ALL=C
+
+DHCP_SERVER_FILE={{ server_dir }}/chrony.servers.$IFACE
+
+clear_servers_from_dhcp() {
+    if [ -f "$DHCP_SERVER_FILE" ]; then
+        rm -f "$DHCP_SERVER_FILE"
+        /{{ chrony_helper_dir }}/chrony-helper update-daemon || :
```
codenrhoden, line 15: no need for the leading slash.

codenrhoden (commented):
Great improvements! Just a few little things.

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,62 @@
...
+
+- name: Install networkd-dispatcher service (Run networkd-dispatcher)
+  systemd:
+    name: networkd-dispatcher
+    state: started
+    enabled: yes
+
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: {{ src }}
```
codenrhoden, line 45: this needs to be `src: "{{ src }}"`
That's what's failing in CI. So, both this line and the one below it. Same in the Ubuntu file.

codenrhoden:
@SanikaGawhane once you address the existing comments, please go ahead and remove "WIP" from the title as well.

codenrhoden:
I'll keep an eye on the tests, but right now this is looking good to me. :)

Can you remove "WIP" from the title?

codenrhoden:
Looks like tests are passing ! I'm going to build these and test them in my home setup as well. But I think you have it!

images/capi/ansible/roles/providers/files/usr/libexec/chrony-helper
```diff
@@ -0,0 +1,251 @@
...
+# and set static sources in the chronyd configuration file.
+
+chronyc=/usr/bin/chronyc
+chrony_conf=/etc/chrony.conf
+chrony_service=chronyd.service
+helper_dir=/var/run/chrony-helper
+added_servers_file=$helper_dir/added_servers
+
+network_sysconfig_file=/etc/sysconfig/network
+dhclient_servers_files=$CHRONY_SERVER_DIR/chrony.servers.*
```
codenrhoden, line 15: I didn't look closely at this file, as I thought it was being copied in unmodified. Looks like `/var/lib/dhclient` got replaced with `$CHRONY_SERVER_DIR` here. That would make sense if we were putting this script on multiple distros with multiple paths, but we are not. This is only going in for Photon, and this needs to be hardcoded to `/var/lib/dhclient`. As it is, this script doesn't work on Photon.

This env var isn't set by anything, and this script is not a template, so if you build Photon and look at this file, it ends up with `$CHRONY_SERVER_DIR` in literal text, which I don't think is what you expected.

images/capi/ansible/roles/providers/tasks/vmware-photon.yml
```diff
@@ -0,0 +1,63 @@
...
+    dest: /tmp
+    remote_src: yes
+
+- name: Install networkd-dispatcher service (Create directory for networkd-dispatcher.conf script)
+  file:
+    path: /etc/conf.d
+    state: directory
+
+- name: Make directory structure for drop-in files
+  command: mkdir -p /etc/networkd-dispatcher/{routable,dormant,no-carrier,off,carrier,degraded,configuring,configured}.d
```
codenrhoden, line 29: `command` does not execute in a shall, so bash-isms like `{...}` don't get expanded the way you would think. When I ran this and built Photon, none of these directories got created. Let's instead use the more pure Ansible way, and combine this step with the previous one, and do this instead:

```yaml
- name: Create needed directories
  file:
    path: "{{ item.dir }}"
    state: directory
  loop:
    - { dir: /etc/conf.d }
    - { dir: /etc/networkd-dispatcher/carrier.d }
    - { dir: /etc/networkd-dispatcher/configured.d }
    - { dir: /etc/networkd-dispatcher/configuring.d }
    - { dir: /etc/networkd-dispatcher/degraded.d }
    - { dir: /etc/networkd-dispatcher/dormant.d }
    - { dir: /etc/networkd-dispatcher/no-carrier.d }
    - { dir: /etc/networkd-dispatcher/off.d }
    - { dir: /etc/networkd-dispatcher/routable.d }
```

codenrhoden (commented):
Hi @SanikaGawhane,

When I built this over the weekend, neither build targets (Ubuntu 20.04 or Photon-3) built successfully. Both had an issue with unfound `networkd_dispatcher_download_url` variable.

Once I fixed that, Ubuntu 20.04 was tested successfully, but Photon-3 didn't work on the NTP side. I've detailed the fixes necessary in this review.

I'm going to peak at our Ubuntu test to see why it didn't fail. It seems like it should have to be. Photon 3 isn't part of the CI tests, so it wouldn't have been caught, but Ubuntu should have been. I must stress again how important it is to build and test images with the exact content of your PR.

codenrhoden:
@SanikaGawhane 

> When I built this over the weekend, never build targets (Ubuntu 20.04 or Photon-3) built successfully. Both had an issue with the the unfound networkd_dispatcher_download_url variable.

Sorry, I misspoke here. the `networkd-dispatcher_download_url` variable is only used for Photon, since that's the only one where we install from source, so that's why Ubuntu still passed CI. So Ubuntu was working fine (the build issue I had at first was a local issue), it was just Photon that had some issues.

It does make me wonder, though.... I don't see any place in here where `networkd-dispatcher` is explicitly installed as a package in Ubuntu. Any idea how/where that is happening?

codenrhoden:
> It does make me wonder, though.... I don't see any place in here where `networkd-dispatcher` is explicitly installed as a package in Ubuntu. Any idea how/where that is happening?

I had no idea, `networkd-dispatcher` is a dependency for both systemd and chrony:

```
# apt-cache rdepends --installed networkd-dispatcher
networkd-dispatcher
Reverse Depends:
  chrony
  systemd
  systemd
  systemd
  chrony
  systemd
```

For posterity, it's worth noting that this PR also installs these scripts on Ubuntu 18.04, and I've confirmed from older builds that I have that Ubuntu 18.04 also has `networkd-dispatcher` installed by default. However, it's an older version (1.7), and there is no update available. While testing this, Sanika and I found that the default version on Ubuntu 20.04 (2.0) was buggy, but the latest update (2.1) was working fine. This makes me doubt this fix works as-is on Ubuntu 18.04. Unfortunately, when I tried to build and test this on my own, I found that a brand new build of Ubuntu 18.04 in my home setup failed to get an IP address. Not sure what is going on there, but I'm going to have to come back to that at another time.

images/capi/ansible/roles/providers/files/usr/libexec/chrony-helper
```diff
@@ -0,0 +1,251 @@
...
+# and set static sources in the chronyd configuration file.
+
+chronyc=/usr/bin/chronyc
+chrony_conf=/etc/chrony.conf
+chrony_service=chronyd.service
+helper_dir=/var/run/chrony-helper
+added_servers_file=$helper_dir/added_servers
+
+network_sysconfig_file=/etc/sysconfig/network
+dhclient_servers_files={{ chrony_helper_dir }}/chrony.servers.*
```
codenrhoden, line 15: @SanikaGawhane, this does not need to be a variable. This will now evaluate to `/usr/libexec`, which is also the wrong place. Set it to `/var/lib/dhclient`. That's it. That's the only value it ever needs to be.

codenrhoden:
I built/tested both Ubuntu 20.04 and Photon 3 OVAs with this change, and the both picked up DHCP Option 42 provided NTP servers on bootup. 🎉 

I think many will be happy to see this addition.
/approve

PR #734, opened 2021-11-08
remove python2-pip from photon GOSS pkg test

codenrhoden:
A recent change to GOSS package test structure caused python2-pip to be
an expected package on Photon. Photon does not install/use python2-pip,
so this test is failing. This patch removes the test for that package.

**Additional context**
I noticed GOSS tests for Photon were failing over the weekend. This PR fixes the problem that introduced last week. @kkeshavamurthy this is a good example of something sneaking in since Photon is in the list of OVA CI targets. I know we can't really fix that yet, just something to be aware of.

/assign @kkeshavamurthy

### DIFF
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -152,7 +152,6 @@ photon:
       open-vm-tools:
       cloud-init:
       cloud-utils:
-      python2-pip:
       python3-netifaces:
 rockylinux:
   common-package: *common_rpms


codenrhoden:
@jsturtevant have you ever seen this issue when trying to install the Azure CLI?

```
ModuleNotFoundError: No module named 'setuptools_rust'
```

Happening in the Ubuntu 20.04 build, a couple times in a row now. Not sure if it's a flake or not.

kkeshavamurthy:
My bad. Thanks for fixing it @codenrhoden 
/lgtm

Issue #589, opened 2021-04-12
Improve UX for PR tests


CecileRobertMichon:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

It's currently hard to tell which test failed and why when looking at the logs for azure PR tests, eg. https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/585/pull-azure-vhds/1381615892579749888

I usually end up searching for "failed" or "error" in the logs to know which packer buid failed and why it failed. It would be great to have some sort of overall summary either using junit (not sure how feasible that is) or via a high level summary at the end. 

**Describe alternatives you've considered**

**Additional context**

/kind feature

codenrhoden:
/remove-lifecycle rotten
I'll address the latest comments on #736 so we can get this one closed out.

codenrhoden:
I think this is addressed for Azure in #736, and in vSphere with #872.

/close

PR #736, opened 2021-11-17
Separate Azure CI logs

codenrhoden:
During CI, log each target to its own file so that that file can be
displayed separately later. This makes the Prow logs look cleaner and
easier to parse.

**Additional context**

### DIFF
--- a/images/capi/scripts/ci-azure-e2e.sh
+++ b/images/capi/scripts/ci-azure-e2e.sh
@@ -28,6 +28,23 @@ set -o pipefail
 CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
+export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
+mkdir -p "${ARTIFACTS}/azure-sigs" "${ARTIFACTS}/azure-vhds"
+
+# Get list of Azure target names from common file
+source azure_targets.sh
+
+# Convert single line entries into arrays
+IFS=' ' read -r -a VHD_TARGETS <<< "${VHD_TARGETS}"
+IFS=' ' read -r -a SIG_TARGETS <<< "${SIG_TARGETS}"
+IFS=' ' read -r -a SIG_GEN2_TARGETS <<< "${SIG_GEN2_TARGETS}"
+
+# Append the "gen2" targets to the original SIG list
+for element in "${SIG_GEN2_TARGETS[@]}"
+do
+    SIG_TARGETS+=("${element}-gen2")
+done
+
 # shellcheck source=parse-prow-creds.sh
 source "packer/azure/scripts/parse-prow-creds.sh"
 
@@ -65,8 +82,34 @@ make deps-azure
 # Disable them for CI runs so don't run into timeouts
 export PACKER_VAR_FILES="packer/azure/scripts/disable-windows-prepull.json scripts/ci-disable-goss-inspect.json"
 
+declare -A PIDS
 if [[ "${AZURE_BUILD_FORMAT:-vhd}" == "sig" ]]; then
-    make -j build-azure-sigs
+    for target in ${SIG_TARGETS[@]};
+    do
+        make build-azure-sig-${target} > ${ARTIFACTS}/azure-sigs/${target}.log 2>&1 &
+        PIDS["sig-${target}"]=$!
+    done
 else
-    make -j build-azure-vhds
+    for target in ${VHD_TARGETS[@]};
+    do
+        make build-azure-vhd-${target} > ${ARTIFACTS}/azure-vhds/${target}.log 2>&1 &
+        PIDS["vhd-${target}"]=$!
+    done
+fi
+
+# need to unset errexit so that failed child tasks don't cause script to exit
+set +o errexit
+exit_err=false
+for target in "${!PIDS[@]}"; do
+  wait ${PIDS[$target]}
+  if [[ $? -ne 0 ]]; then
+    exit_err=true
+    echo "${target}: FAILED. See logs in the artifacts folder."
+  else
+    echo "${target}: SUCCESS"
+  fi
+done
+
+if [[ "${exit_err}" = true ]]; then
+  exit 1
 fi
--- a/images/capi/.dockerignore
+++ b/images/capi/.dockerignore
@@ -8,3 +8,4 @@
 !hack
 !packer
 !Makefile
+!azure_targets.sh
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -34,6 +34,7 @@ COPY --chown=imagebuilder:imagebuilder cloudinit cloudinit/
 COPY --chown=imagebuilder:imagebuilder hack hack/
 COPY --chown=imagebuilder:imagebuilder packer packer/
 COPY --chown=imagebuilder:imagebuilder Makefile Makefile
+COPY --chown=imagebuilder:imagebuilder azure_targets.sh azure_targets.sh
 
 ENV PATH="/home/imagebuilder/.local/bin:${PATH}"
 ENV PACKER_ARGS ''
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -260,9 +260,15 @@ HAPROXY_OVA_VSPHERE_BUILD_NAMES		:=	$(addprefix haproxy-ova-vsphere-,$(PHOTON_VE
 
 AMI_BUILD_NAMES			   ?= ami-centos-7 ami-ubuntu-1804 ami-ubuntu-2004 ami-amazon-2 ami-flatcar ami-windows-2019 ami-windows-2004
 GCE_BUILD_NAMES			   ?= gce-ubuntu-1804 gce-ubuntu-2004
-AZURE_BUILD_VHD_NAMES	   ?= azure-vhd-ubuntu-1804 azure-vhd-ubuntu-2004 azure-vhd-centos-7 azure-vhd-windows-2019 azure-vhd-windows-2019-containerd azure-vhd-windows-2022-containerd
-AZURE_BUILD_SIG_NAMES	   ?= azure-sig-ubuntu-1804 azure-sig-ubuntu-2004 azure-sig-centos-7 azure-sig-windows-2019 azure-sig-windows-2019-containerd azure-sig-flatcar azure-sig-windows-2022-containerd
-AZURE_BUILD_SIG_GEN2_NAMES ?= azure-sig-ubuntu-1804-gen2 azure-sig-ubuntu-2004-gen2 azure-sig-centos-7-gen2
+
+# Make needs these lists to be space delimited, no quotes
+VHD_TARGETS := $(shell grep VHD_TARGETS azure_targets.sh | sed 's/VHD_TARGETS=//' | tr -d \")
+SIG_TARGETS := $(shell grep SIG_TARGETS azure_targets.sh | sed 's/SIG_TARGETS=//' | tr -d \")
+SIG_GEN2_TARGETS := $(shell grep SIG_GEN2_TARGETS azure_targets.sh | sed 's/SIG_GEN2_TARGETS=//' | tr -d \")
+AZURE_BUILD_VHD_NAMES	   ?= $(addprefix azure-vhd-,$(VHD_TARGETS))
+AZURE_BUILD_SIG_NAMES	   ?= $(addprefix azure-sig-,$(SIG_TARGETS))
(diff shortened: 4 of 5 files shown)


codenrhoden:
/hold
Well I broke something. 😆  Will look into it later.

jsturtevant:
fixes https://github.com/kubernetes-sigs/image-builder/issues/589

jsturtevant:
hmm something doens't look like it ran correctly: 

```
ERROR: Please run 'az login' to setup account.
```

codenrhoden:
Ha, Thanks for linking that @jsturtevant . I'll have to take it more seriously now. :)

I was just playing around, trying to do the same thing we do in the OVA tests. Right now the script is missing the `wait`, which is why it's not working, but in looking at that, I'm worried that the OVA tests aren't properly grabbing the exit code from all the backgrounded processes. As it is, I think it's only grabbing the very last one in the list...

images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -28,6 +28,10 @@ set -o pipefail
 CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
+VHD_TARGETS=("ubuntu-1804")
+#VHD_TARGETS=("ubuntu-1804" "ubuntu-2004" "centos-7" "windows-2019" "windows-2019-containerd" "windows-2022-containerd")
+SIG_TARGETS=("ubuntu-1804" "ubuntu-2004" "centos-7" "windows-2019" "windows-2019-containerd" "windows-2022-containerd" "flatcar" "ubuntu-1804-gen2" "ubuntu-2004-gen2" "centos-7-gen2")
```
jsturtevant, line 33: my concern with this would be maintaining this list with the one in the make file

jsturtevant:
No worries. Let me know when it is further along. This one has been on my backlog for awhile.  It will be a big improvement

codenrhoden:
Well, shoot. I'm going to have to take a different approach. It turns out that a background task spawned by a shell can't use `wait` to get status from child tasks that **it** has spawned. 

IF the parent/child relationship is spawned as A->B->C, B can't `wait` on C, as C is considered a child of A, not B. Learned something new today.

Our jobs are run using `runner.sh` in kubekins, and that is done as a background task. I'll take a closer look at `runner.sh` and see if it is doing anything we need -- maybe the answer is to just run the script directly instead of using runner.sh

codenrhoden:
Well shoot, I tried one more thing with some test scripts, and I still can't figure out what's going. If any bash experts can see where I am going wrong, here is what I tried...

child.sh
```sh
#!/bin/bash

echo "child script is pid $$"
for i in {1..5}; do
  ./test.sh &
  echo "spawned background process with pid: $!"
  PIDS+=($!)
done

jobs -l
wait

for pid in ${PIDS[@]}; do
  echo "checking status of pid: ${pid}"
  wait ${pid}
  exit_code=$?
  if [[ ${exit_code} -ne 0 ]]; then
    echo "pid: ${pid} failed with status ${exit_code}"
    exit ${exit_code}
  fi
done
```

test.sh
```sh
#!/bin/bash

echo "$$ parent is ${PPID}"
sleep 5
exit 0
```

When I run this, here is what I see:

```term
$ ./child.sh
child script is pid 72528
spawned background process with pid: 72529
spawned background process with pid: 72530
spawned background process with pid: 72531
spawned background process with pid: 72532
spawned background process with pid: 72533
[1]  72529 Running                 ./test.sh &
[2]  72530 Running                 ./test.sh &
[3]  72531 Running                 ./test.sh &
[4]- 72532 Running                 ./test.sh &
[5]+ 72533 Running                 ./test.sh &
72529 parent is 72528
72530 parent is 72528
72531 parent is 72528
72532 parent is 72528
72533 parent is 72528
checking status of pid: 72529
./child.sh: line 16: wait: pid 72529 is not a child of this shell
pid: 72529 failed with status 127
```

It is very strange to me that it fails saying that pid 72529 is not a child of the shell, because well, it **is**. :) I even captured a view with `pstree` while this was running:

```terminal
 |     \-+= 08827 trhoden -bash
 |       \-+= 72528 trhoden /bin/bash ./child.sh
 |         |-+- 72529 trhoden /bin/bash ./test.sh
 |         | \--- 72535 trhoden sleep 5
 |         |-+- 72530 trhoden /bin/bash ./test.sh
 |         | \--- 72534 trhoden sleep 5
 |         |-+- 72531 trhoden /bin/bash ./test.sh
 |         | \--- 72536 trhoden sleep 5
 |         |-+- 72532 trhoden /bin/bash ./test.sh
 |         | \--- 72537 trhoden sleep 5
 |         \-+- 72533 trhoden /bin/bash ./test.sh
 |           \--- 72538 trhoden sleep 5
```

🤷‍♂️ 

These scripts were used to recreate the problem I'm seeing in the CI jobs launched by Prow, and I was running Bash 5.0.18 on macOS. At least the behavior is consistent.

codenrhoden:
I had the problem pointed out to me... it was because I was using `wait` twice. Once without a pid, and one with. Can't do that. :)

codenrhoden:
Hey, I think this is working now!
The SIG suite had a failure in it, and it got picked up correctly: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/736/pull-azure-sigs/1468266032136196096

re-running to get tests to pass because it's from a flake, but that log at least shows the new logic works.
/test pull-azure-sigs

@jsturtevant This is ready for a real review, I believe.

images/capi/scripts/ci-azure-e2e.sh
codenrhoden: yeah, I hear you. There isn't really a way to do `make -j <all>` and get the logs separated out, so I'm not sure what to do about that.

codenrhoden:
I've never seen this hit the 1h timeout before:

```
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 1h0m0s timeout","severity":"error","time":"2021-12-07T18:57:12Z"}
```

/test pull-azure-sigs

codenrhoden:
Most recent failure is strange:

```
sig-flatcar: fatal: [default]: FAILED! => {"msg": "The conditional check 'ansible_memory_mb.swap.total != 0' failed. The error was: error while evaluating conditional (ansible_memory_mb.swap.total != 0): 'ansible_memory_mb' is undefined
```

If that was really undefined all the time, the tests would never pass on any other PRs... 🤔 

/test pull-azure-sigs

codenrhoden:
Still looking like flakes. This time no issues with flatcar, but two other targets failed, one due to a yum failure, the other from a pip failure.

/test pull-azure-sigs

images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -54,6 +57,25 @@ export JOB_NAME="${JOB_NAME:-"image-builder-e2e"}"
 export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
 
 cleanup() {
+    if [[ "${AZURE_BUILD_FORMAT:-vhd}" == "sig" ]]; then
```
jsturtevant, line 60: Is there a reason to cat these file together here?  Could we just upload them to the storage as is and print out which one failed in the main log?

images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -66,7 +88,29 @@ make deps-azure
...
+    do
+        make build-azure-vhd-${target} > ${target}-${TIMESTAMP}.log 2>&1 &
+        PIDS+=($!)
+    done
+fi
+
+# need to unset errexit so that failed child tasks don't cause script to exit
+set +o errexit
+exit_err=false
+for pid in "${PIDS[@]}"; do
```
jsturtevant, line 107: I generally understand what is happening here but not sure if there are edge cases that could be encountered.  Do we have anyone that might be able to review this?

codenrhoden:
/test pull-azure-sigs
/hold cancel

@CecileRobertMichon can you take a peak at this one?

images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -66,7 +88,29 @@ make deps-azure
 export PACKER_VAR_FILES="packer/azure/scripts/disable-windows-prepull.json scripts/ci-disable-goss-inspect.json"
 
 if [[ "${AZURE_BUILD_FORMAT:-vhd}" == "sig" ]]; then
-    make -j build-azure-sigs
+    for target in ${SIG_TARGETS[@]};
+    do
+        make build-azure-sig-${target} > ${target}-${TIMESTAMP}.log 2>&1 &
```
CecileRobertMichon, line 93: I'm not seeing this file getting uploaded in prow artifacts, where is the file saved in the job?

images/capi/scripts/ci-azure-e2e.sh
CecileRobertMichon: we _could_ move the list of VHD_TARGETS and SIG_TARGETS to a separate .sh file and reuse that both in the makefile and in this script

images/capi/scripts/ci-azure-e2e.sh
CecileRobertMichon: I guess this echoes my comment below, right now the job output is very hard to read. It might be better to upload each file to prow artifacts and then just print the name of the jobs with either "succeeded" or "failed".

then this PR would fix #589 (at least for Azure)

CecileRobertMichon: At the beginning of the script:

```
export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
mkdir -p "${ARTIFACTS}/azure-sigs" "${ARTIFACTS}/azure-vhds"
```

And then just write the files under `${ARTIFACTS}/azure-sigs` or `${ARTIFACTS}/azure-vhds`.

CecileRobertMichon:
This has been one of my pain points for a while, excited to see it merge. Thank you for doing this @codenrhoden. 

Added some comments with some ideas for a couple of improvements.

images/capi/scripts/ci-azure-e2e.sh
jsturtevant: I like this approach as it will be easier to find which file has an error.  What do you think @codenrhoden?

images/capi/scripts/ci-azure-e2e.sh
codenrhoden: Hi @CecileRobertMichon, sorry for the super slow response here. The individual files are not saved, but rather concatenated into `stoud` so it shows up in the log automatically. See lines 61-78. I think looking at a test output example would show what it looks like. For example, this [build-log.txt](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/736/pull-azure-sigs/1479583732280070144).  So it is still one big log file, rather than separate ones, that is just more organized. A step in the right direction?

codenrhoden:
I got this in a state where tests are passing, after incorporating suggestions. Just pushed a new commit that will cause a failure on purpose, just to make sure that failures are caught and listed out correctly.

kkeshavamurthy:
/test pull-azure-vhds
Thanks Travis, logs are definitely cleaner/easier to look at. 
Is it possible to post a link to the logfiles in the main log output?

codenrhoden:
I purposely made a change to make CentOS fail (made the base image undiscoverable), and here are the runs for that:
[VHDs](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/736/pull-azure-vhds/1496932795677675520)
[SIGs](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/736/pull-azure-sigs/1496932795719618560)

Looks like the Ubuntu 2004 build experienced a flake at the same time. Logs are separated, and a summary is available at the bottom of the job output. I think this is good to go! @jsturtevant @CecileRobertMichon  Please take another look. And thank you both for your help getting this in a state much better than where it started.

@kkeshavamurthy The links to the logs are pretty long/involved, so I'd rather not try to generate them ourselves. But when you click on a job result, you just need to click "artifacts" near the top right and drill in from there to find that right log. Not quite as easy as a link, but not too bad either. :)

jsturtevant:
Nice! I see the following in the failed build logs which makes it pretty easy to figure out which one failed.

```
vhd-centos-7: FAILED. See logs.
vhd-ubuntu-1804: FAILED. See logs.
```

kkeshavamurthy:
Seems like everyone is happy with the improvements. Thanks @codenrhoden 
/lgtm

CecileRobertMichon:
This makes the output much easier to process, thank you!

PR #739, opened 2021-11-22
set defaults for the ansible variables

mkumatag:

Set defaults for the variable used in the ansible roles, this will help the user to use the ansible independently without the packer tool.

- Made `common_raw_rpms` and `common_raw_photon_rpms` as array instead of a string variable
- defined `packer_builder_type` and `build_target`
- defined `epel_rpm_gpg_key` which was missing
- `ansible_env.SUDO_USER` fails if user logins as root(this shell environment variable does not exist if logged as root), hence adding a default block to use `ansible_user_id` to avoid failure

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/providers/defaults/main.yml
+++ b/images/capi/ansible/roles/providers/defaults/main.yml
@@ -14,3 +14,5 @@
 ---
 cloud_init_deb_download_url: "https://launchpad.net/ubuntu/+source/cloud-init/21.1-19-gbad84ad4-0ubuntu1~20.04.2/+build/21434629/+files/cloud-init_21.1-19-gbad84ad4-0ubuntu1~20.04.2_all.deb"
 networkd_dispatcher_download_url: "https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.bz2"
+packer_builder_type: ""
+build_target: ""
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -84,13 +84,13 @@ common_virt_debs:
 common_virt_photon_rpms:
 - open-vm-tools
 
-common_raw_rpms:
+common_raw_rpms: []
 
 common_raw_debs:
 - linux-cloud-tools-generic
 - linux-tools-generic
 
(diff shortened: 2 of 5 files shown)


images/capi/ansible/roles/providers/defaults/main.yml
```diff
@@ -13,4 +13,6 @@
 # limitations under the License.
 ---
 cloud_init_deb_download_url: "https://launchpad.net/ubuntu/+source/cloud-init/21.1-19-gbad84ad4-0ubuntu1~20.04.2/+build/21434629/+files/cloud-init_21.1-19-gbad84ad4-0ubuntu1~20.04.2_all.deb"
-networkd_dispatcher_download_url: "https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.bz2"
\ No newline at end of file
+networkd_dispatcher_download_url: "https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.bz2"
+packer_builder_type: ""
```
codenrhoden, line 17: `packer_builder_type` is also used by Windows, so this should also be defined in `images/capi/ansible/windows/roles/providers/defaults/main.yml` (that file would need to be added).

images/capi/ansible/roles/providers/defaults/main.yml
```diff
@@ -13,4 +13,6 @@
...
+build_target: ""
```
codenrhoden, line 18: Should we actually define a useful default value here? The Ansible is always called with this value set to `"virt"` or `"raw"`, I feel like we should set this to `"virt"` here to match. In the vast majority of cases, that's what a user will want.

One additional question -- `build_target` is also used in the node role, not just the providers role. So I think this would have to be defined in `images/capi/ansible/roles/node/defaults/main.yml` as well.

mkumatag: make sence, lemme set set value.

mkumatag:
@codenrhoden fixed the review comments, ptal latest commit

codenrhoden:
/lgtm

Thanks for this improvement, @mkumatag! I'll see if anyone else wants to take a peek at this, but if not I'll merge it in soon.

PR #740, opened 2021-11-22
Switch base image from focal to latest

PushkarJ:
- Current base image has many vulnerabilities that are fixed in newer release of ubuntu.
- Bumping this to latest will allow applying fixes to these vulnerabilities

**Additional context**
Slack thread: https://kubernetes.slack.com/archives/C01E0Q35A8J/p1636653385019700

/cc @CecileRobertMichon 

/sig security cluster-lifecycle

### DIFF
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE=docker.io/library/ubuntu:focal
+ARG BASE_IMAGE=docker.io/library/ubuntu:latest
 FROM $BASE_IMAGE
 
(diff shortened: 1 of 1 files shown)


codenrhoden:
Thanks @PushkarJ, especially for the Slack link for context.

As I understand it, as of right now, this change would have no effect, since `focal` is already the most recent LTS. `focal` itself is a mutable tag, as it will get moved as newer base images are built for Focal/20.04. So it's already all about when you build the image.

I still like the change, as once a new LTS release is made (i.e. from Focal to Jammy) it will pick that up automatically.

/lgtm

Going to wait and see if anyone else has input before merging.

CecileRobertMichon (commented):
/lgtm

Issue #741, opened 2021-11-24
Windows: Waiting for IP


jayunit100:
**What steps did you take and what happened:**

Tried to run image-builder like so:

```
docker run -it --rm --mount type=bind,source=$(pwd)/windows.json,target=/windows.json
--mount type=bind,source=$(pwd)/autounattend.xml,target=/home/imagebuilder/packer/ova/windows/windows-2019/autounattend.xml 
-e PACKER_VAR_FILES="/windows.json" 
-e IB_OVFTOOL=1 -e IB_OVFTOOL_ARGS='--skipManifestCheck' 
-e PACKER_FLAGS='-force -on-error=ask' 
-t projects-stg.registry.vmware.com/tkg/image-builder:v0.1.10_vmware.1 build-node-ova-vsphere-windows-2019
```

# Logs

```hack/ensure-ansible-windows.sh                                                                                                                                                                                                           [84/1457
hack/ensure-packer.sh
hack/ensure-goss.sh
Right version of binary present
hack/ensure-ovftool.sh
packer build -var-file="/home/imagebuilder/packer/config/kubernetes.json"  -var-file="/home/imagebuilder/packer/config/windows/kubernetes.json"  -var-file="/home/imagebuilder/packer/config/containerd.json"  -var-file="/home/imagebuilder/packe
r/config/windows/containerd.json"  -var-file="/home/imagebuilder/packer/config/windows/docker.json"  -var-file="/home/imagebuilder/packer/config/windows/ansible-args-windows.json"  -var-file="/home/imagebuilder/packer/config/windows/common.js
on"  -var-file="/home/imagebuilder/packer/config/windows/cloudbase-init.json"  -var-file="/home/imagebuilder/packer/config/goss-args.json"  -var-file="/home/imagebuilder/packer/config/additional_components.json"  -force -on-error=ask -color=t
rue  -var-file="packer/ova/packer-common.json" -var-file="/home/imagebuilder/packer/ova/windows-2019.json" -var-file="packer/ova/vsphere.json" -except=esx -except=local -only=vsphere-iso -var-file="/windows.json"  -only=vsphere packer/ova/pac
ker-windows.json
2021/11/24 11:33:40 [INFO] Packer version: 1.7.2 [go1.16.3 linux amd64]
2021/11/24 11:33:40 [TRACE] discovering plugins in /home/imagebuilder/.local/bin
[...]

PR #742, opened 2021-11-24
Move VMware tools installation to specialize phase of Windows install

stuartpreston:
Signed-off-by: Stuart Preston <spreston@vmware.com>

This PR moves the VMware guest tools into the Specialize configuration pass, meaning they will be installed prior to any first run scripts in the OOBE pass. The guest tools have a higher chance of deployment success when in its own phase and with a reboot immediately afterwards. #741 

This change also solves a previous issue being tracked where the TimeZone was required to be set to match the ESXi host otherwise the guest tools would not install correctly. #587 

Finally the PR removes the requirement to ship extra files on the boot floppy.

Fixes: #741 
Fixes: #587

### DIFF
--- a/images/capi/packer/ova/packer-windows.json
+++ b/images/capi/packer/ova/packer-windows.json
@@ -67,7 +67,6 @@
         "./packer/ova/windows/disable-network-discovery.cmd",
         "./packer/ova/windows/disable-winrm.ps1",
         "./packer/ova/windows/enable-winrm.ps1",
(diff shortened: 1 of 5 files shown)


codenrhoden:
/sig windows

This one is still looking for more Windows expertise to review

/assign @perithompson 
/assign @jsturtevant

perithompson:
Thanks @stuartpreston this is a better way to achieve the install! 
/lgtm

PR #748, opened 2021-12-01
Azure: archive storage accounts used for reference images

mboersma:

Azure has a limit on the number of storage accounts in use in a region, and we have sometimes hit that limit while building reference images for CAPZ. This PR enhances our periodic cleanup script to archive "keeper" build artifacts into a single archive storage account. This will keep us from hitting the limit as we continue to add reference images for new Kubernetes versions.

N/A

**Additional context**

I've run this by hand several times without the final step of deleting the old SA. I think it's sane, but I'm not clear on how to poll that a batch `start-copy` operation has actually completed. In practice, it seems to complete very quickly (a couple seconds).

Also I need to verify that moving the location of the SA assets will in no way affect the existing Partner Center offer.

cc: @devigned @CecileRobertMichon

### DIFF
--- a/images/capi/packer/azure/scripts/delete-unused-storage.sh
+++ b/images/capi/packer/azure/scripts/delete-unused-storage.sh
@@ -29,6 +29,7 @@ PUBLISHER=${PUBLISHER:-cncf-upstream}
 OFFERS=${OFFERS:-capi capi-windows}
 PREFIX=${PREFIX:-capi}
 LONG_PREFIX=${LONG_PREFIX:-${PREFIX}[0-9]{10\}}
+ARCHIVE_STORAGE_ACCOUNT=${ARCHIVE_STORAGE_ACCOUNT:-${PREFIX}archive}
 
 which pub &> /dev/null || (echo "Please install pub from https://github.com/devigned/pub/releases" && exit 1)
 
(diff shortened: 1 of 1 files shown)


images/capi/packer/azure/scripts/delete-unused-storage.sh
```diff
@@ -46,7 +48,12 @@ for name in ${OFFERS}; do
 done
 NOW=$(date +%s)
 
+# ensure the existence of the archive storage account
+echo "Creating storage account ${ARCHIVE_STORAGE_ACCOUNT}..."
+az storage account create -g "${RESOURCE_GROUP}" -n "${ARCHIVE_STORAGE_ACCOUNT}" --sku Standard_LRS
```
CecileRobertMichon, line 53: is this actually setting the tier of the blobs to archive? https://docs.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview

codenrhoden:
@mboersma looks like this PR is waiting on a question to you.

mboersma:
I'm sorry for not updating this. I still have some questions about whether this is a safe backup strategy overall, so perhaps I opened the PR too soon.

I'll re-open once I've reviewed this internally and have more confidence.

PR #749, opened 2021-12-02
support specifying remote_folder for packer provisioner

kkeshavamurthy:
Azure `packer.json` has a shell provisioner block after the ansible roles are run. This shell provisioner runs the scripts from `/tmp` by default on the build VM. Sometimes permissions to `/tmp` can restricted after customizing the node image. This change adds a way to run the scripts from a different folder. 

**Additional context**

### DIFF
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -185,6 +185,7 @@
         "sudo bash -c \"/usr/share/oem/python/bin/python /usr/share/oem/bin/waagent -force -deprovision+user && sync\""
       ],
       "inline_shebang": "/bin/bash -x",
+      "remote_folder": "{{user `provisioner_remote_folder`}}",
       "type": "shell"
     }
   ],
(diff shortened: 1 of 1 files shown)


devigned (commented):
The change lgtm. However, it would be nice to understand a scenario when /tmp is restricted. Can you expand on that?

codenrhoden:
> The change lgtm. However, it would be nice to understand a scenario when /tmp is restricted. Can you expand on that?

@kkeshavamurthy Is there anything you can link to on this one?

@devigned I know that Kiran has been exploring implementing enhanced node security hardening, specifically around CIS and STIG. I bet there's a specific requirement that he can link to.

kkeshavamurthy:
@devigned From the CIS benchmark docs
[CIS Docs](https://downloads.cisecurity.org/#/) (Free registration might be needed)

```
1.1.5 Ensure noexec option set on /tmp partition (Automated) 
Profile Applicability: 
  Level 1 - Server 
  Level 1 - Workstation 
Description: 
The noexec mount option specifies that the filesystem cannot contain executable binaries. 
Rationale: 
Since the /tmp filesystem is only intended for temporary file storage, set this option to 
ensure that users cannot run executable binaries from /tmp.
```

devigned:
lgtm! Thank you for the providing the context, @kkeshavamurthy.

codenrhoden:
/approve

Will wait to see if anyone else wants to `/lgtm`

PR #751, opened 2021-12-03
Make debug_tools configurable

hxietkg:
debug_tools is always true, and not configurable, we need an option to disable it.

**Additional context**


images/capi/packer/ova/packer-windows.json
```diff
@@ -264,6 +265,7 @@
     "windows_service_manager": null,
     "windows_updates_categories": null,
     "windows_updates_kbs": null,
-    "wins_url": "https://github.com/rancher/wins/releases/download/v{{user `wins_version`}}/wins.exe"
+    "wins_url": "https://github.com/rancher/wins/releases/download/v{{user `wins_version`}}/wins.exe",
+    "debug_tools": "true"
```
jsturtevant, line 269: This just needs to be sorted alphabetically.  We should add it to the https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/config/windows/common.json file to keep as default for all providers. With the new field being passed through the Azure configs didn't get the debug files on the CI builds.

jsturtevant (changes requested):
We should add the `debug_tools=true` it to the https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/config/windows/common.json file to keep as default for all providers. With the new field being passed through the Azure configs didn't get the debug files on the CI builds.

images/capi/packer/ova/packer-windows.json
hxietkg: done

PR #752, opened 2021-12-03
Fix OVA CI script to catch background errors

codenrhoden:
The OVA CI script was using `wait` with no arguments to wait until all
background processes finished before proceeding. From `help wait` in
BASH:

  - If ID is not given, waits for all currently active child processes,
    and the return status is zero.

I know I've seen the script fail when a background `make` failed,
so I am confused on why that would happen if `wait` was always exiting
with a 0 return code. Regardless, I ran experiments where I forced the
first `make` target to fail, and saw that the script exited with 0,
showing that tests passed when they didnt. The observed behavior was
`wait` returning the exit code of the last checked child process.

This patch changes the behavior to explicitly check the return code
for each child process, and exits non-zero if any of them are
non-zero.

**Additional context**

/assign @kkeshavamurthy 

I know the CI tests are failing right now due to auth issues, but figured I'd go ahead and open this anyways.
I stumbled across this behavior while working on https://github.com/kubernetes-sigs/image-builder/pull/736

### DIFF
--- a/images/capi/scripts/ci-ova.sh
+++ b/images/capi/scripts/ci-ova.sh
@@ -103,8 +103,22 @@ cat << EOF > ci-${target}.json
 "build_version": "capv-ci-${target}-${TIMESTAMP}"
 }
 EOF
-  PACKER_VAR_FILES="ci-${target}.json" make -j build-node-ova-vsphere-${target} > ${target}-${TIMESTAMP}.log 2>&1 &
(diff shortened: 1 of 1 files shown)


codenrhoden:
@kkeshavamurthy I believe this is good to go

kkeshavamurthy:
Thanks for fixing it @codenrhoden 
/lgtm

PR #754, opened 2021-12-07
Add support for OCI Images

shyamradhakrishnan:
This PR adds support for Oracle Cloud Infrastructure(OCI) support in image-builder project. We are starting off with support to Ubuntu 18.04/20.04 and Oracle Linux 8 image.

**Additional context**
Uses OCI packer support similar to other providers.

### DIFF
--- a/docs/book/src/capi/providers/oci.md
+++ b/docs/book/src/capi/providers/oci.md
@@ -1,0 +1,63 @@
+# Building Images for Oracle Cloud Infrastructure (OCI)
+
+## Prerequisites
+
+- An OCI account
+- [The OCI plugin for Packer supports three options for authentication](https://www.packer.io/docs/builders/oracle/oci#authentication)
+  . You may use any of these options when building the Cluster API images.
+
+## Building Images
+
+The build [prerequisites](../capi.md#prerequisites) for using `image-builder` for
+building OCI images are managed by running:
+
+```bash
+make deps-oci
+```
+
+From the `images/capi` directory, run `make build-do-<OS>` where `<OS>` is
+the desired operating system. The available choices are listed via `make help`.
+
+### Configuration
+
+In addition to the configuration found in `images/capi/packer/config`, the `oci`
+directory includes several JSON files that define the default configuration for
+the different operating systems.
+
+| File | Description |
+|------|-------------|
+| `oracle-linux-8.json` | The settings for the Oracle Linux 8 image |
+| `ubuntu-1804.json` | The settings for the Ubuntu 18.04 image |
+| `ubuntu-2004.json` | The settings for the Ubuntu 20.04 image |
+
+#### Common options
+
+This table lists several common options that a user must set via
+`PACKER_VAR_FILES` to customize their build behavior.  This is not an exhaustive
+list, and greater explanation can be found in the
+[Packer documentation for the OCI builder](https://www.packer.io/docs/builders/oracle/oci#required-configuration-parameters).
+
+| Variable | Description | Default | Mandatory |
+|----------|-------------|---------|---------|
+| `compartment_ocid` | The OCID of the compartment that the instance will run in. |  | Yes |
+| `subnet_ocid` |  The name of the subnet within which a new instance is launched and provisioned. |  | Yes |
+| `availability_domain` | The name of the Availability Domain within which a new instance is launched and provisioned. The names of the Availability Domains have a prefix that is specific to your tenancy. |  | Yes |
+| `shape` | An OCI region. Overrides value provided by the OCI config file if present. This cannot be used along with the use_instance_principals key. | `VM.Standard.E4.Flex` | No |
+
+#### Steps to create Packer VAR file
+
+Create a file with the following contents and name it as `oci.json`
+
+```json
+{
+  "compartment_ocid": "Fill compartment OCID here",
+  "subnet_ocid": "Fill Subnet OCID here",
+  "availability_domain": "Fill Availbility Domain here"
+}
+```
+
+#### Example make command with Packer VAR file
+
+```bash
+PACKER_VAR_FILES=oci.json make build-oci-oracle-linux-8
+```
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -48,7 +48,7 @@ version: ## Display version of image-builder
 
 .PHONY: deps
 deps: ## Installs/checks all dependencies
-deps: deps-ami deps-azure deps-do deps-gce deps-ova deps-qemu deps-raw
+deps: deps-ami deps-azure deps-do deps-gce deps-ova deps-qemu deps-raw deps-oci
 
 .PHONY: deps-ami
 deps-ami: ## Installs/checks dependencies for AMI builds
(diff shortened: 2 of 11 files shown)


docs/book/src/capi/providers/oci.md
```diff
@@ -0,0 +1,63 @@
+# Building Images for Oracle Cloud Infrastructure (OCI)
```
CecileRobertMichon, line 1: the page needs to be added to https://github.com/kubernetes-sigs/image-builder/blob/f3b6852ae8efd0b0eb9d4293925b4b9e9dae18e4/docs/book/src/SUMMARY.md to show up in the book TOC

shyamradhakrishnan: Thanks @CecileRobertMichon , fixed that

CecileRobertMichon:
@shyamradhakrishnan since we don't have any PR tests in place that run this right now, are you able to share the output of a local run so we know this works / for the record?

shyamradhakrishnan:
[oracle-linux-8.txt](https://github.com/kubernetes-sigs/image-builder/files/7687273/oracle-linux-8.txt)
[ubuntu-18.04.txt](https://github.com/kubernetes-sigs/image-builder/files/7687274/ubuntu-18.04.txt)
[ubuntu-20.04.txt](https://github.com/kubernetes-sigs/image-builder/files/7687275/ubuntu-20.04.txt)
Sure @CecileRobertMichon I have uploaded the make output of the images. I hope this is what you wanted us to share, if not my apologies, can you please explain a bit more on what test you want us to run?

images/capi/ansible/roles/providers/tasks/oci.yml
```diff
@@ -0,0 +1,34 @@
...
+    state: absent
+    regexp: "-A FORWARD -j REJECT --reject-with icmp-host-prohibited"
+  when: ansible_distribution == "Ubuntu"
+
+- name: Disable firewalld service
+  systemd:
+    name: firewalld
+    state: stopped
+    enabled: false
+  when: ansible_distribution == "OracleLinux"
```
kkeshavamurthy, line 34: Should this be added in the node role rather than provider?

images/capi/ansible/roles/providers/tasks/oci.yml
```diff
@@ -0,0 +1,34 @@
...
+    state: absent
+    regexp: "-A INPUT -j REJECT --reject-with icmp-host-prohibited"
+  when: ansible_distribution == "Ubuntu"
+
+- name: Remove the default input reject all iptable rule
+  lineinfile:
+    path: /etc/iptables/rules.v4
+    state: absent
+    regexp: "-A FORWARD -j REJECT --reject-with icmp-host-prohibited"
+  when: ansible_distribution == "Ubuntu"
```
kkeshavamurthy, line 27: Just curious, this is not done for other Ubuntu images. What's the reason for doing it for the Oracle Ubuntu images?

images/capi/packer/oci/packer.json
```diff
@@ -0,0 +1,86 @@
...
+    },
+    {
+      "ansible_env_vars": [
+        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'"
+      ],
+      "extra_arguments": [
+        "--extra-vars",
+        "{{user `ansible_common_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_extra_vars`}}"
```
kkeshavamurthy, line 40: Having `ansible_usr_vars` might be useful here. Helps to pass more vars when needed. Refer to other packer.json files.

images/capi/ansible/roles/providers/tasks/oci.yml
shyamradhakrishnan: This is a specific step required only for OCI provided images. I dont think we should pollute the node role with this. I hope it is fine, we have similar for other providers as well - https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/providers/tasks/aws.yml#L61 . Please suggest if I am missing something here.

images/capi/ansible/roles/providers/tasks/oci.yml
shyamradhakrishnan: OCI provided ubuntu images has strict rules by default, which almost blocks all ingress and egress traffic. We tried this in other providers and they dont have the strict rules by default. If we dont do this, none of the communication (apiserver LB to node, etcd to etcd, node port etc) wont work. As I said, this seem to be a quirk only in the ubuntu base image provided by OCI.

images/capi/packer/oci/packer.json
shyamradhakrishnan: Fixed, Thanks @kkeshavamurthy

CecileRobertMichon:
pull-ova-all has a known issue (#775) but pull-azure-sigs should succeed

images/capi/Makefile
```diff
@@ -564,6 +581,11 @@ build-raw-ubuntu-2004: ## Builds Ubuntu 20.04 RAW image
 build-raw-ubuntu-2004-efi: ## Builds Ubuntu 20.04 RAW image that EFI boots
 build-raw-all: $(RAW_BUILD_TARGETS) ## Builds all RAW images
 
+build-oci-ubuntu-1804: ## Builds the OCI ubuntu-1804 image
+build-oci-ubuntu-2004: ## Builds the OCI ubuntu-2004 image
+build-oci-oracle-linux-8: ## Builds the OCI Oracle Linux 8.x image
+build-oci-all: $(OCI_BUILD_TARGETS) ## Builds all OCI image
```
codenrhoden, line 587: can we also document the OCI 'validate' targets below? There should be a `validate` version of each of these.

codenrhoden: ... down around line 655, add the `validate-oci-*` targets, and make sure the `-all` target lists `$(OCI_VALIDATE_TARGETS` as a dependency. That will make sure the validate CI job actually runs the validate script against OCI.

images/capi/packer/oci/packer.json
```diff
@@ -0,0 +1,88 @@
...
+      "ansible_env_vars": [
+        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'"
+      ],
+      "extra_arguments": [
+        "--extra-vars",
+        "{{user `ansible_common_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_extra_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_user_vars`}}"
```
codenrhoden, line 42: I see this was added at @kkeshavamurthy's request. It's good thing to add, but I think we need to define the var in this file (I think the `validate` script may have picked up it's missing once it starts running). See https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ami/packer.json#L148 as an example. I'd just add it right after the ones below.

codenrhoden (changes requested):
Thanks for adding this @shyamradhakrishnan !

My main comment is to get the `validate` target setup in the Makefile. That will make sure the  `pull-packer-validate` job is running against OCI as well, which right now it is not.

images/capi/Makefile
shyamradhakrishnan: Fixed

images/capi/packer/oci/packer.json
shyamradhakrishnan: Fixed

shyamradhakrishnan:
@codenrhoden I have added the validate targets to CI job and made the changes you suggested. The only extra I had to do was to create a private key as OCI packer validates if the private key in the auth config is valid. Hope this is OK.

CecileRobertMichon (commented):
/lgtm

PR #756, opened 2021-12-10
photon: disable apparmor in kernel

codenrhoden:
Previous attempts to disable AppArmor on Photon were not sufficient. So
long as AppArmor is enabled in the kernel, kubelet and containerD are
loading an apparmor profile that gets enforced and does not work on
Photon at this time.

This patch appends `apparmor=0` to the kernel boot command line to
disable.

**Additional context**
I made sure the regex was idempotent. A negative lookahead assertion is used to only match if the change isn't already present.

/assign @kkeshavamurthy 

I also moved the previous block of code disabling the service from the `setup` role to the `node` role, as `setup` is more of a generic library for installing and bootstrapping, and the changes here are specific to K8s, so I think it is better place in the `node` role.

### DIFF
--- a/images/capi/ansible/roles/node/tasks/photon.yml
+++ b/images/capi/ansible/roles/node/tasks/photon.yml
@@ -20,6 +20,7 @@
     option: Domains
     value: "~local"
   when: leak_local_mdns_to_dns
+
 - name: Leak mDNS to DNS (dhcp) (enable .local domain lookups)
   ini_file:
     path: /etc/systemd/network/99-dhcp-en.network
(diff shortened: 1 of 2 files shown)


codenrhoden:
relevant line in question in `/boot/photon.cfg`:
```
photon_cmdline=init=/lib/systemd/systemd ro loglevel=3 quiet no-vmw-sta cgroup.memory=nokmem apparmor=0
```

kkeshavamurthy:
Thanks for the PR @codenrhoden 
/lgtm

PR #761, opened 2021-12-14
Fix failure observed during adding additional registry images

### LINKED ISSUE #760: Adding additional registry images fails
While attempting to bake additional registry images by setting `load_additional_components` and `additional_registry_images` to true, and specifying the image(s) against `additional_registry_images_list`, encountered image build failure, with the following error:

```
qemu: failed: [default] (item=docker.io/calico/cni:v3.20.1) => {"ansible_loop_var": "item", "changed": true, "cmd": "CONTAINERD_NAMESPACE=\"k8s.io\" /usr/bin/ctr --address=/var/run/containerd/containerd.sock images pull docker.io/calico/cni:v3.20.1", "delta": "0:00:00.004665", "end": "2021-12-13 05:53:20.155166", "item": "docker.io/calico/cni:v3.20.1", "msg": "non-zero return code", "rc": 127, "start": "2021-12-13 05:53:20.150501", "stderr": "/bin/sh: /usr/bin/ctr: No such file or directory", "stderr_lines": ["/bin/sh: /usr/bin/ctr: No such file or directory"], "stdout": "", "stdout_lines": []}
```

It is expected that the ctr binary is already present at the required location, i.e., `/usr/bin` and the image is successfully pulled

The issue happens as the symlink of `/usr/local/bin/ctr` is created at `/usr/bin/` only when `~/images/capi/ansible/roles/kubernetes/tasks/url.yml` is imported, which happens when the  `kubernetes_source_type` and `kubernetes_cni_source_type` are `http`

**Environment:**

Project: CAPI

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`): observed across different OS - Centos 7, Ubuntu 20.04, Centos 8 (On a PR that has been created)
- Packer Version: 1.7.4
- Packer Provider: (tested with) Qemu (probably observed against others too)
- Ansible Version: 2.11.5
- Cluster-api version (if using): N/A
- Kubernetes version: (use `kubectl version`): v1.21.1 

/kind bug

Pearl1594:

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #760 

**Additional context**
Moves the task of creating symlink for `/usr/local//bin/ctr` so as to make it available when images are to be pulled in registry.yml

Post this modification, the image build successfully completes, snippet of the log pertaining to pulling of the specified image from the registry:
```
...
qemu:  ___________________________________________________________
    qemu: / TASK [load_additional_components : Pull additional images \
    qemu: \ from registry]                                            /
    qemu:  -----------------------------------------------------------
    qemu:         \   ^__^
    qemu:          \  (oo)\_______
    qemu:             (__)\       )\/\
    qemu:                 ||----w |
    qemu:                 ||     ||
    qemu:
    qemu: changed: [default] => (item=docker.io/calico/cni:v3.20.1)
...
```

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/main.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/main.yml
@@ -50,6 +50,19 @@
     src: etc/kubernetes-version
     mode: 0644
 
+- name: Symlink cri-tools
+  file:
+    src:   "/usr/local/bin/{{ item }}"
+    dest:  "/usr/bin/{{ item }}"
+    mode: 0777
+    state: link
+    force: yes
+  loop:
+  - ctr
+  - crictl
+  - critest
+  when: ansible_os_family != "Flatcar"
+
 # TODO: This section will be deprecated once https://github.com/containerd/cri/issues/1131 is fixed. It is used to support ECR with containerd.
 - name: Check if Kubernetes container registry is using Amazon ECR
   set_fact:
--- a/images/capi/ansible/roles/kubernetes/tasks/url.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/url.yml
@@ -12,19 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- name: Symlink cri-tools
-  file:
-    src:   "/usr/local/bin/{{ item }}"
-    dest:  "/usr/bin/{{ item }}"
-    mode: 0777
-    state: link
-    force: yes
-  loop:
-  - ctr
-  - crictl
-  - critest
-  when: ansible_os_family != "Flatcar"
-
 - name: Create CNI directory
   file:
     state: directory


jsturtevant:
might not be needed? https://github.com/kubernetes-sigs/image-builder/issues/760#issuecomment-993938561  

seems like a reasonable change to me though. @kkeshavamurthy what are you thoughts?

kkeshavamurthy:
I'm fine with the change. Does not hurt. 
Also `sysusr_prefix` could probably be updated to `/usr/local` by default.

codenrhoden:
It took me a few minutes to understand what was happening here. :) Now that I do, I have one question...

Since the point is to make sure that `ctr` is available at `{sysusr_prefix}/ctr`, would it make more sense to have the symlink task actually link to `{sysusr_prefix}` instead of the hardcoded `/usr/bin`? I'm not sure what this would mean for flatcar which sets this to `/opt`.

just trying to make this a bit less fragile.

@kkeshavamurthy

kkeshavamurthy:
The task is not run for flatcar. So, should be fine I guess.

kkeshavamurthy:
@Pearl1594 any more thoughts how you'd like to proceed with this?

Pearl1594:
Sorry for the delay in getting back. @kkeshavamurthy I believe it's been established that since this task isn't run for flatcar, the initial concern raised isn't an issue per se. So do you think any changes are required?  Or could we move ahead with this fix?

rohityadavcloud (approved):

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -50,6 +50,19 @@
     src: etc/kubernetes-version
     mode: 0644
 
+- name: Symlink cri-tools
```
sidharthsurana, line 53: This introduces regression since the code in url.yml expects the ctr bin to be present at `/usr/bin/ctr` but now this change moves this symlink to a later point.

This task needs to happen prior to the include of url.yml to work properly.

sidharthsurana:
This seems to bring in a regression when building photon ova. In building that I run into the same error where ansible complaints about missing file at `/usr/bin/ctr`

PR #763, opened 2021-12-14
fix ova-ci script to use vCenter IP directly

kkeshavamurthy:
Due to a CVE, VMC locked down certain access to vCenter on AWS. 
Adding the IP of the vCenter should force the connection to go thru the VPN and unblock the CI. 
In the long run, we should update the prow secrets. 

**Additional context**

/assign @codenrhoden

### DIFF
--- a/images/capi/scripts/ci-ova.sh
+++ b/images/capi/scripts/ci-ova.sh
@@ -42,7 +42,7 @@ cleanup_build_vm() {
   chmod +x govc
   mv govc /usr/local/bin/govc
 
-  export GOVC_URL="${VSPHERE_SERVER}"
+  export GOVC_URL="10.2.224.4"
   export GOVC_USERNAME="${VSPHERE_USERNAME}"
   export GOVC_PASSWORD="${VSPHERE_PASSWORD}"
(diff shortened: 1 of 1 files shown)


images/capi/scripts/ci-ova.sh
```diff
@@ -69,7 +69,7 @@ fi
 
 cat << EOF > packer/ova/vsphere.json
 {
-    "vcenter_server":"${VSPHERE_SERVER}",
+    "vcenter_server":"https://10.2.224.4",
```
codenrhoden, line 72: this one shouldn't have the `https://` in front of it. That's what I take from the log.

jsturtevant:
lgtm, ova job passed. Is there a issue to track updating the PROW secrets?

Is that a static IP address or will it change frequently?

kkeshavamurthy:
I think capv folks are looking into updating the prow secrets.

codenrhoden:
> I think capv folks are looking into updating the prow secrets.

yeah, this is something I will track with @yastij, as the secret is a CAPV one.

Thanks for updating this, @kkeshavamurthy 

/lgtm
/approve

PR #778, opened 2022-01-11
Fix target argument in Makefile substitution command

abhay-krishna:
This PR fixes the Make substitution command to derive the value of `RAW_CLEAN_TARGETS` from `RAW_BUILD_TARGETS` instead of self-referencing itself.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -438,7 +438,7 @@ QEMU_CLEAN_TARGETS := $(subst build-,clean-,$(QEMU_BUILD_TARGETS))
 $(QEMU_CLEAN_TARGETS):
 	rm -fr output/$(subst clean-qemu-,,$@)-kube*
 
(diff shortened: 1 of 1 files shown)


abhay-krishna:
@kkeshavamurthy Any idea why the tests are failing? Are they flakes?

kkeshavamurthy:
Expect ova tests to fail. Known issue. 
I think azure tests might be a flake. They worked on my other open PR. 

/test pull-azure-vhds
/test pull-azure-sigs

abhay-krishna:
> Expect ova tests to fail. Known issue. I think azure tests might be a flake. They worked on my other open PR.
> 
> /test pull-azure-vhds /test pull-azure-sigs

Thanks for the info. In case of consistent known failures, is the practice to force-merge or add a `/override` comment for the presubmit context to pass for auto-merge via tide?

abhay-krishna:
@kkeshavamurthy Alright, it looks like the Azure presubmits are passing now.

abhay-krishna:
Could I get an update on this? Thanks!

Issue #766, opened 2021-12-15
Azure CI is flaking when installing azure cli


jsturtevant:
**What steps did you take and what happened:**
[A clear and concise description on how to REPRODUCE the bug.]

This time it was the ubuntu error  
```
0;32m    sig-ubuntu-1804-gen2: fatal: [default]: FAILED! => {"changed": false, "cmd": ["/usr/bin/pip3", "install", "azure-cli"], "msg": "stdout: Collecting azure-cli\n  Downloading https://files.pythonhosted.org/packages/ba/5a
If you are seeing an error here please try the following to\n            s
uccessfully install cryptography:\n    \n           
 Upgrade to the latest pip and try again. This will fix errors for most\n    
```

Doesn't seem to occur every time.

See https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/762/pull-azure-sigs/1471216005047914496 for full logs

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Looks like it is this line: 

https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/providers/tasks/debian.yml#L17

**Environment:**

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi), [kube-deploy/imagebuilder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/kube-deploy/imagebuilder), [konfigadm](https://github.com/kubernetes-sigs/image-builder/tree/master/images/konfigadm)):

Additional info for Image Builder for Cluster API related issues:

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`):
- Packer Version:
- Packer Provider:
- Ansible Version: 
- Cluster-api version (if using):
- Kubernetes version: (use `kubectl version`):

/kind bug
[One or more /area label. See https://github.com/kubernetes-sigs/cluster-api/labels?q=area for the list of labels]

jsturtevant:
This looks like it is an issue with the pip version on the host: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html

>  Although it executes using the Ansible Python interpreter, the pip module shells out to run the actual pip command, so it can use any pip version you specify with executable. By default, it uses the pip version for the Ansible Python interpreter. For example, pip3 on python 3, and pip2 or pip on python 2.
[...]

PR #781, opened 2022-01-17
[ansible][windows] Add `ExtendVolumesPlugin` cloubase-init plugin

ionutbalutoiu:
### What this PR does / why we need it:

The official capi Windows marketplace image from the `cncf-upstream:capi`
are generated without this plugin, and the images have a 30GB system
root partition disk. Using a flavor with a bigger OS disk will not extend the
root system partition, unless manually extended.

The [ExtendVolumesPlugin](https://cloudbase-init.readthedocs.io/en/latest/plugins.html#cloudbaseinit.plugins.windows.extendvolumes.ExtendVolumesPlugin) plugin is required to extend the root system
partition to the maximum size, by the cloudbase-init.

### DIFF
--- a/images/capi/ansible/windows/roles/cloudbase-init/templates/cloudbase-init.conf
+++ b/images/capi/ansible/windows/roles/cloudbase-init/templates/cloudbase-init.conf
@@ -33,5 +33,6 @@ netbios_host_name_compatibility={{ netbios_host_name_compatibility }}
 metadata_services={{ cloudbase_metadata_services }}
 plugins=cloudbaseinit.plugins.windows.createuser.CreateUserPlugin, 
         cloudbaseinit.plugins.common.setuserpassword.SetUserPasswordPlugin, 
+        cloudbaseinit.plugins.windows.extendvolumes.ExtendVolumesPlugin,
         cloudbaseinit.plugins.common.userdata.UserDataPlugin,
         {{ cloudbase_plugins }}


jsturtevant:
Was there a failure happening do to this?

Will this plugin work across all the providers (ami, azure, ova) or is provider specific? If it works across providers this should be a fine place to put it otherwise we can add it to the providers that require it.

/ok-to-test

jsturtevant:
The ova failure is known is being worked on.  It is not blocking

ionutbalutoiu:
> Was there a failure happening do to this?
> 
> 
> 
> Will this plugin work across all the providers (ami, azure, ova) or is provider specific? If it works across providers this should be a fine place to put it otherwise we can add it to the packer configuration for providers that require it.
> 
> 
> 
> /ok-to-test

Hello, James!

We started using the upstream capi Windows images in the K8s Flannel CNI CAPZ-based CI.

In the CI, we pre-pull all the container images for the conformance tests. Due to the root partition disk being 30GB, the disk will fill up quickly, leading to "out of space" errors.

Currently, we have a "preKubeadm" command to extend the partition to maximum size. So, nothing is blocking.

However, I feel that anyone using the public capi Windows images with bigger workloads, will be required to add the same "preKubeadm" command. Otherwise, they'll hit the same issue.

In my opinion, the cloudbase-init plugin usage is a more elegant solution.

Also, the plugin works on any provider.

jsturtevant:
flake was https://github.com/kubernetes-sigs/image-builder/issues/766
/retest

PR #799, opened 2022-02-10
Add option to compress capi/qemu images

apricote:

Adds two options to the capi qemu packer configuration:

- [discard:unmap](https://www.packer.io/plugins/builders/qemu#disk_discard)
- [Compression](https://www.packer.io/plugins/builders/qemu#disk_compression)

`disk_compression` option is set to the packer default, while `disk_discard` was set to `unmap`, this only breaks in really old versions of qemu (~2014).

When both are set, the resulting qcow2 image is reduced in size by ~50%, see this comparison from the current `master` branch build (e2738129b35fec82d55a3172aeb9c07bae8c856f, kube v1.20.10):

```json
{
  "disk_compression": "true",
  "disk_discard": "unmap"
}
```

- Without Options (current): `3.7G` (baseline)
- With Options:  `1.6G` (- 57%)


kkeshavamurthy:
/ok-to-test

Any reason why both the options should not be turned ON by default?
I do not work with qemu/raw images to know :)

apricote:
/assign @mdbooth

Assigning you as per https://github.com/kubernetes-sigs/image-builder/pull/799#issuecomment-1035013841

mdbooth:
/lgtm

Agree that compression should not be a default option due to the runtime cost. However, I think discard=unmap could just be a default. The only reason I can think of not to do this would be for support of extremely old versions of qemu, but I think they'd have to be older than 2014 so I think we can move on now.

apricote:
> Agree that compression should not be a default option due to the runtime cost. However, I think discard=unmap could just be a default. The only reason I can think of not to do this would be for support of extremely old versions of qemu, but I think they'd have to be older than 2014 so I think we can move on now.

Sounds good to me @mdbooth. If no one objects, I will push an update soon

PR #813, opened 2022-02-22
Modify permission of SSH folder

### LINKED ISSUE #815: Win32-OpenSSH is not abled to be connected

image-builder is supporting `ssh_source_url` to install external ssh server for windows image build, but it does not work for windows server 2019 because it does not set the  correct permission on the ssh folder `C:\Program Files\OpenSSH`, so we are not able to ssh to the windows node.

SSH works.

**Environment:**

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`): windows server 2019
- Packer Version:
- Packer Provider:
- Ansible Version: 

/kind bug

hxietkg:
SSH is installed in folder 'C:\Program Files\OpenSSH', according to
wiki from Win32-OpenSSH:
   https://github.com/PowerShell/Win32-OpenSSH/wiki/Install-Win32-OpenSSH
so we need the folder has correct permission set.

SSH server does not work

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #815

**Additional context**

### DIFF
--- a/images/capi/ansible/windows/roles/systemprep/tasks/ssh-archive.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/ssh-archive.yml
@@ -20,6 +20,36 @@
     - '{{ programfiles.stdout | trim }}\OpenSSH'
     - '{{ alluserprofile.stdout | trim }}\ssh'
 
+# Win32-OpenSSH requires SYSTEM and Administrator groups having Write
+# permissions on directory 'C:\Program Files\OpenSSH', authenticated
+# users having only Read and Execute permissions on it, see:
+#   https://github.com/PowerShell/Win32-OpenSSH/wiki/Install-Win32-OpenSSH
+#
+#   "Make sure binary location has the Write permissions to just to SYSTEM,
+#   Administrator groups. Authenticated users should and only have Read and
+#   Execute."
+#
+# Folder 'C:\Program Files\OpenSSH' inherits users and permissions from its
+# parent folder when it is created, by default, SYSTEM and Administrator
+# already have Write permissions on it, the only exception is the inherited
+# user BUILTIN\Users has ReadAndExecute permission but only authenticated
+# users are allowed to have such permission, this prevent us from connecting
+# to the sshd server, just remove it.
+- name: Disable inheritance of OpenSSH directory
+  win_acl_inheritance:
+    path: '{{ programfiles.stdout | trim }}\OpenSSH'
+    state: absent
+    reorganize: yes
+- name: Remove permission for Users
+  win_acl:
+    path: '{{ programfiles.stdout | trim }}\OpenSSH'
+    user: BUILTIN\Users
+    rights: ReadAndExecute,Synchronize
+    type: allow
+    state: absent
+    inherit: 'None'
+    propagation: 'None'
+
 - name: Download OpenSSH Archive
   win_get_url:
     url: '{{ ssh_source_url }}'
@@ -41,4 +71,3 @@
     Get-ChildItem -Path "{{ tempdir.stdout | trim }}\OpenSSH-Win64\*" -Recurse | Move-Item -Destination "{{ programfiles.stdout | trim }}\OpenSSH"
     Get-ChildItem -Path "{{ programfiles.stdout | trim }}\OpenSSH" | Unblock-File
     & 'C:\Program Files\OpenSSH\install-sshd.ps1'
-    
--- a/images/capi/hack/generate-goss-specs.py
+++ b/images/capi/hack/generate-goss-specs.py
@@ -43,7 +43,8 @@ def generate_goss(provider, system, versions, runtime, dryrun=False, save=False)
             'kubernetes_cni_rpm_version': versions['cni_rpm'],
             'kubernetes_cni_source_type': 'pkg',
             'runtime': runtime,
-            'pause_image': versions['pause']}
+            'pause_image': versions['pause'],
+            'ssh_source_url': versions['ssh_url']}
 
 
     # Build command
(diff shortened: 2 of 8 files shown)


hxietkg:
@jayunit100 @perithompson

jayunit100:
thanks hongsheng
 can you file a bug that this fixes ?

codenrhoden:
/assign @jsturtevant 

Just a drive-by comment here - if the permission are wrong, why would this step be optional? If they are sometimes wrong (maybe it depends on OS versoin? OpenSSH version?), wouldn't we still want the task to always ensure the perms are correct? I wasn't expecting to see a new boolean flag that controls whether or not to change it. Seems like an always things to me (but again, no first-hand experience).

hxietkg:
> /assign @jsturtevant
> 
> Just a drive-by comment here - if the permission are wrong, why would this step be optional? If they are sometimes wrong (maybe it depends on OS versoin? OpenSSH version?), wouldn't we still want the task to always ensure the perms are correct? I wasn't expecting to see a new boolean flag that controls whether or not to change it. Seems like an always things to me (but again, no first-hand experience).

Thanks for your comment, image-builder supports many version of windows, e.g. 2004, 2019 and 2022, the version I am using is 2019, I did run test on the other versions, so I am not sure whether it works for them, can we keep the boolean flag optional until all the versions are verified?

images/capi/ansible/windows/roles/systemprep/tasks/ssh-archive.yml
```diff
@@ -20,6 +20,18 @@
...
 
+- name: Modify permission of OpenSSH directory
+  win_shell: |
+    $FilePath = "{{ programfiles.stdout | trim }}\OpenSSH"
+    $acl = Get-Acl -Path $FilePath
+    $acl.SetAccessRuleProtection($True, $True)
+    Set-Acl -Path $FilePath -AclObject $acl -Confirm:$false
+    $acl = Get-Acl -Path $FilePath
+    $ace = New-Object System.Security.AccessControl.FileSystemAccessRule("BUILTIN\Users", "ReadAndExecute, Synchronize", "None", "None", "Allow")
+    $acl.RemoveAccessRule($ace)
```
jsturtevant, line 31: Can we build up the requirements here instead of adding all and removing some? It would be easier to avoid a permission slipping by that was already assigned.

From the install notes it looks like it should be:

```
(Make sure binary location has the Write permissions to just to SYSTEM, Administrator groups. Authenticated users should and only have Read and Execute.)
```

jsturtevant (commented):
I agree with @codenrhoden that this should be something that is enabled across all the versions. It is unlikely that behavior has changed. @perithompson has more experience here.

I would also I like to see a goss test add that verifies we have the proper permissions set here: https://github.com/kubernetes-sigs/image-builder/blob/4f73b99cf951941cbfef1b3cbfbbcd89247833d2/images/capi/packer/goss/goss-command.yaml#L70

jayunit100:
talking w/ @hxietkg this issue comes up specifically in an ipv6 environment...

images/capi/ansible/windows/roles/systemprep/tasks/ssh-archive.yml
hxietkg: Folder 'OpenSSH' inherits users and permissions from its parent folder when it is created, so it has always has correct permissions for SYSTEM and Administrators groups, the only exception is BUILTIN/Users, it has wrong permission, let's move it, I updated the comments in the code, please have a look.

hxietkg:
@jsturtevant, @perithompson, @jayunit100, can you please take a look at this PR? thanks!

perithompson:
thanks @hxietkg! good to get more tests in there!
/lgtm

images/capi/packer/goss/goss-command.yaml
```diff
@@ -175,4 +175,23 @@ command:
     stderr: []
     timeout: 30000
 {{end}}
+
+  Check permission of OpenSSH directory for SYSTEM:
+    exec: powershell -command "!(Test-Path 'C:\Program Files\OpenSSH') -or ((Get-Acl 'C:\Program Files\OpenSSH').Access | Where-Object{$_.IdentityReference -eq 'NT AUTHORITY\SYSTEM' -and $_.FileSystemRights -eq 'FullControl'}) -ne $null"
```
jsturtevant, line 180: This works since the folder doesn't exist if installed via the feature.  @perithompson Is there a way with the goss setup to only run this if `ssh_source_url` is set?

hxietkg: > 
I think there is a way, but we need to add ssh_source_url to a config file, I'll try that.

jsturtevant: awesome thanks!

images/capi/hack/generate-goss-specs.py
```diff
@@ -113,6 +115,9 @@ def main():
     docker = read_json_file(os.path.join(root_path, 'packer', 'config', 'windows', 'docker.json'))
     versions['docker'] = docker['docker_ee_version']
 
+    wincommon = read_json_file(os.path.join(root_path, 'packer', 'config', 'windows', 'common.json'))
+    versions['ssh_url'] = wincommon['ssh_source_url']
```
jsturtevant, line 119: looking at this now, it looks like there isn't a better way to do this as there isn't a section for loading arbitrary values.  

This overload the `versions` array to get this working, @codenrhoden is this ok or should we be using a separate array here?  Is this the way to accomplish running validations on various types of setups?

jsturtevant:
Overall this looks good, I am just wondering on the best approach to feed that value to goss.  @codenrhoden could you take a look?

codenrhoden:
> Overall this looks good, I am just wondering on the best approach to feed that value to goss. @codenrhoden could you take a look?

I see what you are saying, @jsturtevant. Right now that does appear to be the best way. I don't see a better one without re-writing it with a method for load arbitrary values.

/lgtm

And sorry I was slow to respond on this one. In my head this was "a Windows PR", so I wasn't looking that closely. Thanks for pinging me again @knabben.

PR #821, opened 2022-03-08
Add support for Arm based images in OCI

shyamradhakrishnan:
This PR has the necessary changes to build Arm based images on Oracle Cloud Infrastructure.

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/node/meta/main.yml
+++ b/images/capi/ansible/roles/node/meta/main.yml
@@ -21,6 +21,12 @@ dependencies:
 
   - role: setup
     vars:
+      rpms: "{{ common_rpms }}"
+      debs: "{{ common_debs }}"
+    when: packer_builder_type == "oracle-oci" and ansible_architecture == "aarch64"
+
+  - role: setup
+    vars:
       rpms: "{{ ( common_photon_rpms + lookup('vars', 'common_' + build_target + '_photon_rpms') ) if ansible_os_family == 'VMware Photon OS' else ( ( common_rpms + rh7_rpms + lookup('vars', 'common_' + build_target + '_rpms') ) if (ansible_os_family == 'RedHat' and ansible_distribution_major_version == '7') else ( common_rpms + rh8_rpms + lookup('vars', 'common_' + build_target + '_rpms') ) ) }}"
       debs: "{{ common_debs +  lookup('vars', 'common_' + build_target + '_debs') }}"
-    when: ansible_distribution != "Amazon"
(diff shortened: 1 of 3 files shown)


images/capi/ansible/roles/node/meta/main.yml
```diff
@@ -19,8 +19,14 @@ dependencies:
       debs: "{{ common_debs }}"
     when: ansible_distribution == "Amazon"
 
+  - role: setup
+    vars:
+      rpms: "{{ common_rpms }}"
+      debs: "{{ common_debs }}"
+    when: packer_builder_type == "oracle-oci" and ansible_architecture == "aarch64"
```
shyamradhakrishnan, line 26: Arm does not have the common virt rpms and debs

images/capi/ansible/roles/setup/tasks/debian.yml
```diff
@@ -17,6 +17,7 @@
     src: etc/apt/sources.list.j2
     dest: /etc/apt/sources.list
     mode: 0644
+  when: packer_builder_type != "oracle-oci"
```
shyamradhakrishnan, line 20: OCI Base images has the required apt sources list embedded inside the image. IF we copy the templated file, Arm based build are failing as the sources list is not correct for Arm

codenrhoden:
overall the changes look fine to me.
/lgtm

I suspect there is some overall refactoring to be done to properly allow a choice of CPU arch -- x86 is pretty much hardcoded across the project for all providers. So I think other providers would run into similar issues around trying to reach `common_virt` packages that don't exist on ARM.

images/capi/ansible/roles/setup/tasks/debian.yml
jsturtevant: This seems like a good bit of information to have in a comment so it is not lost in future

codenrhoden:
/approve
@shyamradhakrishnan has mentioned in Slack that he's got a follow-on PR in progress where he can add the requested comment.

jsturtevant:
azure flake was https://github.com/kubernetes-sigs/image-builder/issues/766.  

ova failed on another PR today to (#825). Not sure if there is something flaky or if it is configuration for ova builds:

```
vsphere-clone: fatal: [default]: UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory.
```

/retest

images/capi/ansible/roles/setup/tasks/debian.yml
shyamradhakrishnan: Added comment

PR #823, opened 2022-03-10
Install nfs-common package for Azure image

lzhecheng:
Signed-off-by: Zhecheng Li <zhechengli@microsoft.com>

Azure image installs netbase and nfs-common package.

**Additional context**

related: https://github.com/kubernetes/test-infra/issues/24957

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -54,3 +54,14 @@
     state: restored
     path: /etc/iptables/rules.v4
   when: ansible_os_family == "Debian"
+
+- name: Install netbase and nfs-common
+  apt:
+    name: "{{ packages }}"
+    state: present
+    force_apt_get: yes
+  vars:
+    packages:
+    - netbase
+    - nfs-common
+  when: ansible_os_family == "Debian"


jsturtevant:
My understanding this is required for some of the features in the Azure cloudprovider, specifically for some e2e tests around functionality.  We discussed it briefly but maybe @lzhecheng can expand upon what features this enabled in the cloudprovider.

jsturtevant:
As a stop gap we are adding it to the pre-kubeadm commands in capz: https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2158.

CecileRobertMichon:
@codenrhoden my understanding is that it's required to pass full k8s conformance on Azure (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2158#issuecomment-1064670635). I guess it depends if we want to be able to say image-builder as producing k8s conformant images or not. If so I would lean towards including it.

@lzhecheng can you please clarify if users would require this package to be installed outside of conformance tests?

lzhecheng:
@CecileRobertMichon thank you for the explanation!

@codenrhoden so far I only see it needed for some conformance tests: "should access to two volumes with the same volume mode and retain data across pod recreation on the same node". I think we need to include nfs-common into image or update those tests to skip on nodes without the package.

codenrhoden:
> I guess it depends if we want to be able to say image-builder as producing k8s conformant images or not. If so I would lean towards including it.

@CecileRobertMichon I think that's definitely what we want to say, so I agree. Thanks for the discussion!

/hold cancel
/lgtm

kkeshavamurthy:
@lzhecheng pls rebase. #831 should fix the failing ova tests

lzhecheng:
@kkeshavamurthy I have updated the PR. Thank you!

CecileRobertMichon (commented):
/lgtm

PR #825, opened 2022-03-11
fix(windows): custom role arguments passing

mweibel:
`custom_role` needs to be passed to ansible to make the `include_role`
directive work at all. `custom_role_names` needs to be quoted because
the arguments are space separated, otherwise only the first role gets
passed.


mweibel:
/sig windows

(not sure that works)

jsturtevant:
> custom_role_names needs to be quoted because
the arguments are space separated, otherwise only the first role gets
passed.

Does this need to be fixed for Linux as well? https://github.com/kubernetes-sigs/image-builder/blob/4577f24d377fe6113ac42aa9471dd267e8753714/images/capi/packer/config/ansible-args.json#L2 cc: @codenrhoden

mweibel:
> > custom_role_names needs to be quoted because
> > the arguments are space separated, otherwise only the first role gets
> > passed.
> 
> Does this need to be fixed for Linux as well?
> 
> https://github.com/kubernetes-sigs/image-builder/blob/4577f24d377fe6113ac42aa9471dd267e8753714/images/capi/packer/config/ansible-args.json#L2
> cc: @codenrhoden

good question, I wondered as well. I believe it can't hurt and therefore I added the change for linux, too.

codenrhoden:
/assign @kkeshavamurthy 
Kiran, can you check out if this works for you on the Linux side? I know you've done a lot with custom roles. We've always taken the approach of passing in quoted strings.

As an example, we called this out in docs here: https://image-builder.sigs.k8s.io/capi/capi.html#passing-in-extra-packages-to-the-image

It's really easy to miss, though, and i've made this mistake a few times. This may be a much better solution. :)

mweibel:
> As an example, we called this out in docs here: https://image-builder.sigs.k8s.io/capi/capi.html#passing-in-extra-packages-to-the-image

Oh, I **did** miss this, hence the PR ;)

jsturtevant:
I think this approach is cleaner.  @kkeshavamurthy what are your thoughts?

kkeshavamurthy:
Hello, sorry been busy the last few days. I'll take this for a run and let you know. Sorry for the delay.

kkeshavamurthy:
Thanks for doing this @mweibel . This is definitely better looking. 
/lgtm

PR #826, opened 2022-03-11
feat: allow overriding volume_type and throughput

mweibel:
- Upgrades windows to gp3 volume type (default, can be configured) and accordingly specify the throughput. 
- Allows to specify builder_instance_type as variable

### DIFF
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -20,7 +20,8 @@
           "delete_on_termination": true,
           "device_name": "{{ user `root_device_name` }}",
           "volume_size": "{{ user `volume_size` }}",
-          "volume_type": "gp2"
+          "volume_type": "{{ user `volume_type` }}",
+          "throughput": "{{ user `throughput` }}"
         }
       ],
       "name": "{{user `build_name`}}",
@@ -187,6 +188,8 @@
     "subnet_id": "",
     "temporary_security_group_source_cidrs": "",
     "volume_size": "40",
+    "volume_type": "gp2",
+    "throughput": "",
     "vpc_id": "",
     "windows_service_manager": null,
     "windows_updates_kbs": null,
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -13,15 +13,15 @@
       "disable_stop_instance": true,
       "encrypt_boot": "{{user `encrypted`}}",
       "iam_instance_profile": "{{user `iam_instance_profile`}}",
-      "instance_type": "t3.large",
+      "instance_type": "{{user `builder_instance_type`}}",
       "kms_key_id": "{{user `kms_key_id`}}",
       "launch_block_device_mappings": [
         {
           "delete_on_termination": true,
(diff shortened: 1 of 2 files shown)


images/capi/packer/ami/packer-windows.json
```diff
@@ -13,14 +13,15 @@
...
-      "instance_type": "t3.large",
+      "instance_type": "{{user `instance_type`}}",
       "kms_key_id": "{{user `kms_key_id`}}",
       "launch_block_device_mappings": [
         {
           "delete_on_termination": true,
           "device_name": "{{ user `root_device_name` }}",
           "volume_size": "{{ user `volume_size` }}",
-          "volume_type": "gp2"
+          "volume_type": "{{ user `volume_type` }}",
```
codenrhoden, line 23: this is one that is not parameterized on Linux:
https://github.com/kubernetes-sigs/image-builder/blob/7b6c9f93f45dcf28559130fb068aade0e44d5447/images/capi/packer/ami/packer.json#L22

I feel like something was there that made it required to be `gp3`, but now I can't remember... If that can be consistent with Linux (e.g. make it a variable in Linux too), that would be the ideal outcome.

images/capi/packer/ami/packer-windows.json
```diff
@@ -13,14 +13,15 @@
       "disable_stop_instance": true,
       "encrypt_boot": "{{user `encrypted`}}",
       "iam_instance_profile": "{{user `iam_instance_profile`}}",
-      "instance_type": "t3.large",
+      "instance_type": "{{user `instance_type`}}",
```
codenrhoden, line 16: could we make this `instance_builder_type` to match the Linux builds? `https://github.com/kubernetes-sigs/image-builder/blob/7b6c9f93f45dcf28559130fb068aade0e44d5447/images/capi/packer/ami/packer.json#L13

codenrhoden (changes requested):
Hi @mweibel, thanks for the enhancements here!

It looks like all the changes are to improve the Windows OS builder. There are a couple of things here that have equivalent abilities in the Linux builder, and I'd mostly like to see them be consistent with each other.

/ok-to-test

images/capi/packer/ami/packer-windows.json
mweibel: done :+1:

images/capi/packer/ami/packer-windows.json
mweibel: done :+1: I also switched the default on windows to `gp3` to keep that consistent across the operating systems (switching to gp3 was actually the reason for this PR ;)). Hope this is fine.

mweibel:
I changed the requested things. The change for gp2 to gp3 for windows is something which could result in higher costs when building the image, I believe. Does that need to be pointed out in some way specifically?

mweibel:
/sig windows

is that too, I believe :)

codenrhoden (approved):
Thanks for making the requested changes!

> I changed the requested things. The change for gp2 to gp3 for windows is something which could result in higher costs when building the image, I believe. Does that need to be pointed out in some way specifically?

It's not something we've done before... But I will be sure to point it out the next set of release notes.

/approve

PR #827, opened 2022-03-14
Make Flatcar build process more generic

johananl:

This PR modifies the Flatcar build process to make it as generic as possible and removes "special", Flatcar-specific cases. Following this change set Flatcar no longer uses its own set of GOSS validations but rather uses the same set as the other distros.

Summary of changes:

- Make GOSS validations execute for Flatcar.
- Ensure GOSS validations pass for Flatcar.

**Additional context**

To test this PR I executed `PACKER_VAR_FILES=/tmp/packer-vars-aws.json FLATCAR_VERSION=3033.2.2 make build-ami-flatcar` with the following `packer-vars-aws.json` file:

```json
{
    "kubernetes_deb_version": "1.20.10-00",
    "kubernetes_rpm_version": "1.20.10-0",
    "kubernetes_semver": "v1.20.10",
    "kubernetes_series": null,
    "goss_inspect_mode": "false",
    "ami_regions": "eu-central-1",
    "aws_region": "eu-central-1"
}
```

### DIFF
--- a/images/capi/ansible/roles/setup/tasks/flatcar.yml
+++ b/images/capi/ansible/roles/setup/tasks/flatcar.yml
@@ -57,3 +57,8 @@
     owner: root
     group: root
     mode: "0755"
+
+- name: Enable systemd-timesyncd unit
+  systemd:
+    enabled: yes
+    name: systemd-timesyncd.service
--- a/images/capi/packer/ami/flatcar.json
+++ b/images/capi/packer/ami/flatcar.json
@@ -4,7 +4,6 @@
   "ansible_extra_vars": "ansible_python_interpreter=/opt/bin/python",
   "build_name": "flatcar-{{env `FLATCAR_CHANNEL`}}",
   "crictl_source_type": "http",
-  "goss_entry_file": "goss/goss-flatcar.yaml",
   "kubernetes_cni_source_type": "http",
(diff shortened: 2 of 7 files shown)


jepio (approved):
LGTM

johananl:
The CI is failing because of a pip error which seems to be solvable by upgrading pip according to the error message. The strange thing is - I'm not getting this error when running the build process from my laptop against Azure. Maybe the difference is because the CI is using a different Python package mirror than I am?

jsturtevant:
> The CI is failing because of a pip error which seems to be solvable by upgrading pip according to the error message. The strange thing is - I'm not getting this error when running the build process from my laptop against Azure. Maybe the difference is because the CI is using a different Python package mirror than I am?

flake is tracked here: https://github.com/kubernetes-sigs/image-builder/issues/766

/retest
/lgtm

/assign @codenrhoden @kkeshavamurthy

jsturtevant:
That was a flake in packer itself when cleanup the vms.  The VHD build process passed.
/retest

johananl:
Great, CI is green now. Thanks @jsturtevant!

codenrhoden:
I love that some of the Flatcar specific logic is no longer needed. Nice work!

/test pull-ova-all

PR #829, opened 2022-03-14
Fix QEMU Flatcar build

### LINKED ISSUE #828: QEMU Flatcar build is broken

The QEMU Flatcar build is broken.

To reproduce, run the following:

```
cd images/capi
FLATCAR_VERSION=3033.2.2 make build-qemu-flatcar
```

Output:

```
flatcar: output will be in this color.

==> flatcar: Retrieving ISO
==> flatcar: Trying https://stable.release.flatcar-linux.net/amd64-usr/3033.2.2/flatcar_production_iso_image.iso
==> flatcar: Trying https://stable.release.flatcar-linux.net/amd64-usr/3033.2.2/flatcar_production_iso_image.iso?checksum=md5%3Abd514676b122fed3231192d79ee93e0e
==> flatcar: https://stable.release.flatcar-linux.net/amd64-usr/3033.2.2/flatcar_production_iso_image.iso?checksum=md5%3Abd514676b122fed3231192d79ee93e0e => /home/jliebermann/microsoft/cluster-api/image-builder/images/capi/packer_cache/125fcbf7de539e2b5e51edb9618e2cd751d8c9c3.iso
==> flatcar: Error finding "./packer/qemu/linux/flatcar/http/": stat ./packer/qemu/linux/flatcar/http/: no such file or directory
==> flatcar: Deleting output directory...
Build 'flatcar' errored after 438 milliseconds 147 microseconds: Error finding "./packer/qemu/linux/flatcar/http/": stat ./packer/qemu/linux/flatcar/http/: no such file or directory

==> Wait completed after 438 milliseconds 190 microseconds

==> Some builds didn't complete successfully and had errors:
--> flatcar: Error finding "./packer/qemu/linux/flatcar/http/": stat ./packer/qemu/linux/flatcar/http/: no such file or directory

==> Builds finished but no artifacts were created.
make: *** [Makefile:414: build-qemu-flatcar] Error 1
```

I expected the build to work.

None

/kind bug
 
/assign

johananl:

The Flatcar QEMU build is currently broken. This PR makes the build pass.

TODOs:

- [x] Remove temporary commits
- [x] Squash commits

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #828

**Additional context**

To test this PR, run the following:

```
cd images/capi
FLATCAR_VERSION=3033.2.2 make build-qemu-flatcar
```


johananl:
Strange. Installing `azure-cli` is [failing](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/829/pull-azure-vhds/1503737531273842688/artifacts/azure-vhds/ubuntu-1804.log) but locally things seem to work.

codenrhoden (commented):
It's been a while since I looked at the `qemu` folder in here, and poking around, I see what you are looking to accomplish, and I have some alternative suggestions (depends on how much you are willing to refactor).

At first when I looked and saw the removal of the HTTP stuff, I though "well that will break all the other builders that are expecting a kickstart/preseed file"... However, then I noticed that this was in a builder specifically for flatcar. So, I wonder two things

- Why does Flatcar have a separate builder?
- Does Flatcar use a preseed file?

It looks pretty clear that Flatcar does not need a preseed file. That will be useful in a bit...

Going line by line through the two builders, it looks like the only difference between the two builders is the flatcar builder literally just adds `"name": "flatcar"`. However, all builders should have `"name" defined, and it should be set to `{{user `build_name`}}`, which is defined in the individual <OS>.json file. So that's the first thing I would change/add, and that will eliminate the needed for defining a second builder (and the post-processor and provisioner logic that says "only flatcar" will still work). A couple of examples:
[AWS](https://github.com/kubernetes-sigs/image-builder/blob/e1ccea7bdd8a9834d2833c7c4ced2a05c4036ced/images/capi/packer/ami/packer.json#L25)
[Azure](https://github.com/kubernetes-sigs/image-builder/blob/e1ccea7bdd8a9834d2833c7c4ced2a05c4036ced/images/capi/packer/azure/packer.json#L23)

From there, I would still leave

```
      "boot_command": [
        "{{user `boot_command_prefix`}}",
        "http://{{ .HTTPIP }}:{{ .HTTPPort }}",
      ],
```

in place, except that it needs to be changed to:

```
    {
      "boot_command": [
        "{{user `boot_command_prefix`}}",
        "{{user `boot_media_path`}}",
      ],
```

That way `boot_media_path` can be set for those OS's that need it.

Flatcar can set both `boot_media_path` and `boot_command_suffix` to `""` and they will be empty, getting the result you want.

What's weird to me is that I see `boot_media_path` defined in both `qemu-centos-7.json` and `qemu-rockylinux-8.json`, but they are both unused because that var isn't referenced in packer.json. I think there was a good bit of copy/paste happening here. The `ova` folder has [a good reference](https://github.com/kubernetes-sigs/image-builder/blob/e1ccea7bdd8a9834d2833c7c4ced2a05c4036ced/images/capi/packer/ova/packer-node.json#L125-L129) of how this can be used, which shows that some OS's pull a preseed/kickstart [from an HTTP endpoint](https://github.com/kubernetes-sigs/image-builder/blob/e1ccea7bdd8a9834d2833c7c4ced2a05c4036ced/images/capi/packer/ova/photon-3.json#L2-L4), others from [a mounted floppy disk](https://github.com/kubernetes-sigs/image-builder/blob/e1ccea7bdd8a9834d2833c7c4ced2a05c4036ced/images/capi/packer/ova/ubuntu-2004.json#L2-L5). And in your case (Flatcar+QEMU) you don't need to pull one at all, so both media path and the suffix can be empty, and you can set `boot_command_prefix` to the way that you are current proposing to set `boot_command`.

this will make all the builders look the same, and also make this qemu folder be structures more like AWS, Azure, and OVA.

I knot that's a lot of text, but the changes are pretty straightforward. If they don't make sense, let me know and I will make a PR that changes the `qemu` folder to bring it in line with the other providers first, and then you can rebase your PR on top of those changes to make your fixes.

Thanks for looking at this @johananl !

johananl:
Superb feedback @codenrhoden. Thanks!

Your suggestion regarding `boot_command` makes a lot of sense to me.

AFAICT the motivation for the separate Flatcar builder is being able to run provisioners and post-processors only for Flatcar builds, and whether we still need *that* I'm not sure. I think there is a good chance this divergence from the generic build process has historical reasons which may well be no longer relevant.

Removing the dedicated build makes a lot of sense to me as it continues what we've started in https://github.com/kubernetes-sigs/image-builder/pull/827. I'll try to figure out if we still need "special treatment" in terms of provisioners and post-processors and will open a new PR with the relevant cleanups.

codenrhoden:
> Superb feedback @codenrhoden. Thanks!

Glad it was well received!

> AFAICT the motivation for the separate Flatcar builder is being able to run provisioners and post-processors only for Flatcar builds, and whether we still need _that_ I'm not sure. I think there is a good chance this divergence from the generic build process has historical reasons which may well be no longer relevant.

I agree. And I don't have any problem with those special provisioners staying. They work by setting that `"only"` section to the `"name"` that is defined in the builder. So for "flatcar", it's literally: https://github.com/kubernetes-sigs/image-builder/blob/a09b089b1b344d75275cf741cafdbd877050f660/images/capi/packer/qemu/packer.json#L53

However, the only diff I see between builder [1](https://github.com/kubernetes-sigs/image-builder/blob/a09b089b1b344d75275cf741cafdbd877050f660/images/capi/packer/qemu/packer.json#L3-L32) and [2](https://github.com/kubernetes-sigs/image-builder/blob/a09b089b1b344d75275cf741cafdbd877050f660/images/capi/packer/qemu/packer.json#L33-L63) is that `"name": "flatcar"` line. If that's really the case, the two can be combined. Instead the remaining builder should have a line that says ```"name": "{{user `build_name`}}"```.

Then, each of the OS-specific `.json` files should have something like `"build_name": "flatcar"` or `"build_name": "ubuntu-2004"`. Does that make sense? This change can be made and the flatcar specific post-processors and provisions will still only run for flatcar.

codenrhoden:
>  This change can be made and the flatcar specific post-processors and provisions will still only run for flatcar.

Oh wait, I thought about this a bit more and I'm going to have to eat my words. I remember working on this for the OVA builder, and Packer only behaves this way when `"name"` is defined directly, not as a `{{user ...}}` variable. I remember being really disappointed about this. 

So for now I think what you have looks good. I'd just ask that you squash the commits.

johananl:
@codenrhoden following your latest comments, I looked into the duplicate builder matter again. It seems to me that we don't need it after all because in the Flatcar-specific provisioner we check the build name in the provisioner itself, which means it will be triggered for all distros when we remove the `only` directive **but won't actually execute**. As for the dedicated Vagrant post-processor, I'm not sure why we would need it for Flatcar only, i.e. if this post-processor is useful we may want to use it for all distros and if it's useless we should remove it.

So, I've removed the dedicated builder. This allowed me to greatly simplify the Makefile and in the current state of the PR we treat Flatcar as yet another distro along with the rest of the QEMU targets.

Flatcar builds fine this way. I still need to ensure I haven't broken any of the other providers and double check the Vagrant post-processor matter. Will update.

johananl:
> > I still need to ensure I haven't broken any of the other providers and double check the Vagrant post-processor matter.
> 
> I've manually built all providers except RHEL 8 which seems to be broken on `master`. I think we should be good to merge.
> 
> Regarding the CI failure, looks like we've been bitten by [ansible/ansible#77537](https://github.com/ansible/ansible/issues/77537) 😢 (which is unrelated to this change).

https://github.com/kubernetes-sigs/image-builder/pull/863 works around the issue.

johananl:
The CI failure looks like a DNS-related flake.

/test pull-azure-sigs

johananl:
>I'd just ask that you squash the commits.

Done!

Also, with #863 merged the CI issue should be fixed.

/unhold

codenrhoden:
This looks like a great set of clean ups, @johananl !

Thanks for fixing all this.

/lgtm
/approve

PR #830, opened 2022-03-14
Add Azure pipelines for FoodX

gnadaban:


gnadaban:
Mistakenly opened, please disregard.

PR #834, opened 2022-03-17
flatcar Azure gen2 images

jepio:
* add support for building Flatcar Azure Gen2 SIG images
* fix handling for FLATCAR_VERSION=current; previously specifying it as make argument had different behavior than passing it in the environment
* fix appending local packer directory to path

**Additional context**

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -231,7 +231,7 @@ WINDOWS_VERSIONS		:=	windows-2019 windows-2004 windows-2022
 FLATCAR_CHANNEL ?= stable
 FLATCAR_VERSION ?= 2905.2.3
 ifeq ($(FLATCAR_VERSION),current)
-FLATCAR_VERSION := $(shell hack/image-grok-latest-flatcar-version.sh $(FLATCAR_CHANNEL))
+override FLATCAR_VERSION := $(shell hack/image-grok-latest-flatcar-version.sh $(FLATCAR_CHANNEL))
 endif
 
 export FLATCAR_CHANNEL FLATCAR_VERSION
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -469,6 +469,7 @@ build-azure-vhd-windows-2022-containerd: ## Builds for Windows Server 2022 with 
 build-azure-vhd-windows-2004: ## Builds for Windows Server 2004 SAC
 build-azure-sig-centos-7-gen2: ## Builds CentOS Gen2 managed image in Shared Image Gallery
 build-azure-sig-flatcar: ## Builds Flatcar Azure managed image in Shared Image Gallery
+build-azure-sig-flatcar-gen2: ## Builds Flatcar Azure Gen2 managed image in Shared Image Gallery
 build-azure-sig-ubuntu-1804-gen2: ## Builds Ubuntu 18.04 Gen2 managed image in Shared Image Gallery
 build-azure-sig-ubuntu-2004-gen2: ## Builds Ubuntu 20.04 Gen2 managed image in Shared Image Gallery
 build-azure-vhds: $(AZURE_BUILD_VHD_TARGETS) ## Builds all Azure VHDs
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -22,7 +22,7 @@ SHELL := /usr/bin/env bash
 
 # This option is for running docker manifest command
 export DOCKER_CLI_EXPERIMENTAL := enabled
-export PATH := $(PATH):$(PWD)/.local/bin
+export PATH := $(PATH):$(CURDIR)/.local/bin
 
 export IB_VERSION ?= $(shell git describe --dirty)
 
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -229,7 +229,7 @@ WINDOWS_VERSIONS		:=	windows-2019 windows-2004 windows-2022
 
 # Set Flatcar Container Linux channel and version if not supplied
 FLATCAR_CHANNEL ?= stable
-FLATCAR_VERSION ?= 2905.2.3
+FLATCAR_VERSION ?= 3033.2.4
 ifeq ($(FLATCAR_VERSION),current)
 override FLATCAR_VERSION := $(shell hack/image-grok-latest-flatcar-version.sh $(FLATCAR_CHANNEL))
 endif
--- a/images/capi/packer/azure/packer.json
+++ b/images/capi/packer/azure/packer.json
@@ -182,7 +182,7 @@
         "BUILD_NAME={{user `build_name`}}"
       ],
       "inline": [
-        "if [ $BUILD_NAME != \"flatcar\" ]; then exit 0; fi",
+        "if [[ $BUILD_NAME != \"flatcar\"* ]]; then exit 0; fi",
         "sudo bash -c \"/usr/share/oem/python/bin/python /usr/share/oem/bin/waagent -force -deprovision+user && sync\""
       ],
       "inline_shebang": "/bin/bash -x",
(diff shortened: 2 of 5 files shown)


images/capi/Makefile
```diff
@@ -22,7 +22,7 @@ SHELL := /usr/bin/env bash
 
 # This option is for running docker manifest command
 export DOCKER_CLI_EXPERIMENTAL := enabled
-export PATH := $(PATH):$(PWD)/.local/bin
+export PATH := $(PATH):$(CURDIR)/.local/bin
```
kkeshavamurthy, line 25: TIL the difference between CURDIR & PWD :)

jepio: Me too :)

jepio:
Centos test got a 403 accessing EPEL rpm, Flatcar test failed to due not having terms accepted for stable-gen2 (didn't know this is tied to a specific sku and not the product). Added a commit that automatically accepts those terms.

invidian (approved):

images/capi/packer/azure/scripts/init-sig.sh
```diff
@@ -98,3 +98,15 @@ az sig image-definition create \
...
+   --resource-group ${RESOURCE_GROUP_NAME} \
+   --gallery-name ${GALLERY_NAME} \
+   --gallery-image-definition capi-flatcar-${FLATCAR_CHANNEL}-${FLATCAR_VERSION}-gen2 \
+   --publisher capz \
+   --offer capz-demo \
+   --sku flatcar-${FLATCAR_CHANNEL}-${FLATCAR_VERSION}-gen2 \
+   --hyper-v-generation V2 \
+   --os-type Linux
+
+az vm image terms accept --urn kinvolk:flatcar-container-linux-free:${FLATCAR_CHANNEL}:latest
```
CecileRobertMichon, line 111: not sure it's a good idea to run this in this script which is common to all SIG definitions and accept license terms on behalf of the user. Perhaps we could split it into a specific flatcar setup script so that it only gets run if the user is trying to create a flatcar image, that way it's at least only run when the user has expressed intent to use the flatcar source images and we don't end up accepting terms on every subscription that runs this, regardless of their need for flatcar. The other option is to not run this in a script and add it as a one-time setup instruction in docs but I can see why that is less attractive.

invidian: Maybe there is a way to check if terms has been accepted and bail when flatcar target is built? This way we fail early and tell user what to do to get through.

CecileRobertMichon: the terms for the gen2 image have now been accepted for the sponsorship subscription used by prow jobs so tests should pass without this

jepio: Dropped this commit.

jepio: I don't see a way to check if terms have been accepted and the error message from the VM create is clear on what needs to be done and provides links to Azure docs.

FWIW: we would all prefer that the user not have to agree to any terms to use any free Flatcar images, but this is not something that the Azure marketplace allows us to express.

invidian:
CI is failing because of #861

invidian:
CI failure seems legit. Does building work for you locally @jepio ?

jepio:
Worked locally with `FLATCAR_VERSION=current`, I'll retry with 2905.2.3 and report back.

invidian:
For me building locally fails the same as on CI. Perhaps we should bump the default Flatcar version to latest then.

jepio:
Oops, forgot that the first release to support Azure Gen2 was [2983.0.0](https://github.com/flatcar-linux/coreos-overlay/commit/cf1258472343e57c5d2ed865ab006dab3e733eba). Will bump to a more recent release, 3033.2.4, which will be the base of the coming flatcar lts channel.

jepio:
Looks like it's snowing.
/retest

invidian:
/retest

VHD eventually passed

images/capi/packer/azure/scripts/init-sig.sh
```diff
@@ -98,3 +98,12 @@ az sig image-definition create \
    --sku centos-7.7-gen2 \
    --hyper-v-generation V2 \
    --os-type Linux
+az sig image-definition create \
```
jsturtevant, line 101: This will conflict with https://github.com/kubernetes-sigs/image-builder/pull/867.  I would suggest we get #867 in first?

invidian: I'm fine with rebasing either, shouldn't be too hard :)

jepio: Either rebased this PR

jsturtevant: Thanks!

jepio:
This is in mergeable shape, if I could get some lgtm's

codenrhoden:
/lgtm
@jsturtevant @kkeshavamurthy @johananl any final comments before merging?

images/capi/packer/azure/packer.json
```diff
@@ -182,7 +182,7 @@
         "BUILD_NAME={{user `build_name`}}"
       ],
       "inline": [
-        "if [ $BUILD_NAME != \"flatcar\" ]; then exit 0; fi",
+        "if [[ $BUILD_NAME != \"flatcar\"* ]]; then exit 0; fi",
```
johananl, line 185: We could use the format `if [[ \"${BUILD_NAME}\" == *\"flatcar\"* ]]` which we already use in other builds:

https://github.com/kubernetes-sigs/image-builder/blob/ddd0f293f79a70dc31577364bb76ecf0cce53c9d/images/capi/packer/ami/packer.json#L91

https://github.com/kubernetes-sigs/image-builder/blob/ddd0f293f79a70dc31577364bb76ecf0cce53c9d/images/capi/packer/qemu/packer.json#L49

images/capi/Makefile
```diff
@@ -469,6 +469,7 @@ build-azure-vhd-windows-2022-containerd: ## Builds for Windows Server 2022 with
 build-azure-vhd-windows-2004: ## Builds for Windows Server 2004 SAC
 build-azure-sig-centos-7-gen2: ## Builds CentOS Gen2 managed image in Shared Image Gallery
 build-azure-sig-flatcar: ## Builds Flatcar Azure managed image in Shared Image Gallery
+build-azure-sig-flatcar-gen2: ## Builds Flatcar Azure Gen2 managed image in Shared Image Gallery
```
johananl, line 472: I'd consider adding the equivalent `validate-*` target so that `make validate-azure-sig-flatcar-gen2` works. However, we're already inconsistent in that not all build targets have matching validate targets... The same applies to `build-azure-sig-flatcar`, too.

johananl (approved):
LGTM! Nice work @jepio.

I added two non-blocking nits which IMO can be done in follow up PRs.

codenrhoden:
@jepio please capture the needed follow-up tasks in an issue so they are not forgotten.
@johananl nice catch about the validate target. I try to make sure those don't get out of sync, but it sounds like they already are.

/approve

johananl:
>@johananl nice catch about the validate target. I try to make sure those don't get out of sync, but it sounds like they already are.

Maybe there is a way to define all platforms and distros once and enforce that for every combination there be a build target and a corresponding validate target.

invidian:
>Maybe there is a way to define all platforms and distros once and enforce that for every combination there be a build target and a corresponding validate target.

If the validation tests are fast and stable, maybe they could be executed after the build target, to make sure they are defined and are actually passing.

jepio:
The validate targets actually exist for all images, the seperate `validate-` targets are only used by `make help`. Will capture this in an issue though.

images/capi/Makefile
jepio: Sorry for getting this out of sync: `make validate-azure-sig-flatcar-gen2` actually works already, but `make help` does not show the line without a dedicated `validate-xxxx: ## Builds xxx` entry.

PR #837, opened 2022-03-18
Make al-2 images immutable by disabling security updates on boot.

kkeshavamurthy:
1. al-2 images run security updates on first boot of the base ami. This task sometimes is slow and tasks [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/setup/tasks/redhat.yml#L29) fail to obtain the yum lock in 60s and the image build fails. Since we update all packages in ansible, disabling this on first boot thru `user_data` in `packer.json` should not affect anything. 

2. According to this [FAQ](https://aws.amazon.com/amazon-linux-ami/faqs/), al-2 images always pull security updates on every boot making the image different from build time. To make the images immutable, `repo_upgrade: none` is added to the cloud.cfg. 

Thanks @codenrhoden for the pointers. 

**Additional context**
/assign @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -106,3 +106,4 @@ pause_image: "k8s.gcr.io/pause:3.6"
 containerd_additional_settings: null
 leak_local_mdns_to_dns: false
 build_target: "virt"
+cloud_cfg_file: "/etc/cloud/cloud.cfg"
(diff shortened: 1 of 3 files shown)


codenrhoden:
/lgtm

@sedefsavas Just wanted to make you aware of this one -- Turns out Amazon Linux 2 AMIs weren't immutable as they are supposed to be for CAPI. On boot up, it was possible that they were installing security updates (nominally a good thing, but not what we are supposed to do).

PR #839, opened 2022-03-22
fix(AWS): add SSH key to windows images

mweibel:
Without this change, connecting via SSH to windows machines on AWS is not possible because there is no key in `authorized_keys`.

### DIFF
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -167,7 +167,7 @@
     "cloudbase_init_url": "https://github.com/cloudbase/cloudbase-init/releases/download/{{user `cloudbase_init_version`}}/CloudbaseInitSetup_{{user `cloudbase_init_version` | replace_all `.` `_` }}_x64.msi",
     "cloudbase_metadata_services": "cloudbaseinit.metadata.services.ec2service.EC2Service",
     "cloudbase_metadata_services_unattend": "cloudbaseinit.metadata.services.base.EmptyMetadataService",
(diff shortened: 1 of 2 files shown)


jsturtevant:
Good catch, I guess AWS's metadata service supports setting ssh keys for Windows?  

If you are using this with Capi, another alternative is to specify the ssh key in the user data: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d120f2d9a35df48ca2aaefeacfb65fb99cd2599c/templates/cluster-template-windows-containerd.yaml#L310-L314  This is how we do it in Azure as the metadata service there doesn't support sshkeys.

Could you add a check to goss configuration like OVA's: https://github.com/kubernetes-sigs/image-builder/blob/f245056d646f4d6c8be62c65fb64e464f6306756/images/capi/packer/goss/goss-vars.yaml#L390-L391

here:
https://github.com/kubernetes-sigs/image-builder/blob/f245056d646f4d6c8be62c65fb64e464f6306756/images/capi/packer/goss/goss-vars.yaml#L404-L405

mweibel:
> If you are using this with Capi, another alternative is to specify the ssh key in the user data: https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/d120f2d9a35df48ca2aaefeacfb65fb99cd2599c/templates/cluster-template-windows-containerd.yaml#L310-L314 This is how we do it in Azure as the metadata service there doesn't support sshkeys.

thanks for this - wasn't aware of it, even though I scanned through the azure capi templates a couple of times already ;) 

Thanks for the review. I added the requested change to goss :+1:

PR #843, opened 2022-03-24
fix: windows goss wrong provider on AWS

mweibel:
Goss for windows builds on AWS didn't work previous this change because the goss configuration expects `amazon`, not `ami`.

### DIFF
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -120,7 +120,7 @@
       "vars_file": "{{user `goss_vars_file`}}",
       "vars_inline": {
         "OS": "{{user `distribution` | lower}}",
-        "PROVIDER": "ami",
+        "PROVIDER": "amazon",
         "containerd_version": "{{user `containerd_version`}}",
         "distribution_version": "{{user `distribution_version`}}",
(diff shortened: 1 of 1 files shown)


jsturtevant:
@codenrhoden do we have a CI job that tests ami's?

/ok-to-test

jsturtevant:
This matches https://github.com/kubernetes-sigs/image-builder/blob/f245056d646f4d6c8be62c65fb64e464f6306756/images/capi/hack/generate-goss-specs.py#L26-L28

jsturtevant:
As well as Linux config https://github.com/kubernetes-sigs/image-builder/blob/83b29e1b9fa57c3d7ce9c6534851b6e8aa148828/images/capi/packer/ami/packer.json#L130

/lgtm 
/approve

codenrhoden:
/test pull-goss-populate

> @codenrhoden do we have a CI job that tests ami's?

@jsturtevant Unfortunately, no. We've long been blocked on getting access to AWS credentials through prow. Credentials for image-builder were created a long while back (like over a year), but then we got blocked on something. I can't remember exactly what. cc @kkeshavamurthy  An attempt was made with https://github.com/kubernetes-sigs/image-builder/pull/423, but that never got finished.

PR #850, opened 2022-04-01
Add ability to supply global pip config file

codenrhoden:
This patch creates a new variable to Packer and Ansible, `pip_conf_file`,
which allows the user to specify a global config file for pip, the
Python installer. This file is copied into /etc/pip.conf, and can be
used for many purposes, including defining a PyPI mirror. This is useful
when building in internet-restricted environments that do not have
access to public package repositories.

This serves as the Python version of handing non-public repos like we
already do for RPM and DEB repositories.

**Additional context**
/assign @kkeshavamurthy


kkeshavamurthy:
Thanks for adding this @codenrhoden 
/lgtm

PR #851, opened 2022-04-06
Fix tdnf install in firstboot role

codenrhoden:
The `tdnf` command in the `firstboot` role is happening before any custom
repos are put in place. This means that if you happen to define a
private or internal mirror to use instead of a public one, the `firstboot`
role ignores this and uses the public repo anyways.

The logic in the `setup` role (meant to be used as a dependency) already
accounts for this, and needs to be used instead.

**Additional context**
/assign @kkeshavamurthy

### DIFF
--- a/images/capi/ansible/roles/firstboot/meta/main.yml
+++ b/images/capi/ansible/roles/firstboot/meta/main.yml
@@ -1,4 +1,4 @@
-# Copyright 2020 The Kubernetes Authors.
+# Copyright 2022 The Kubernetes Authors.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


images/capi/ansible/roles/firstboot/tasks/photon.yml
```diff
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Perform a tdnf distro-sync
-  command: tdnf distro-sync -y --refresh
```
codenrhoden, line 16: this is no longer needed, as it is done automatically in the `setup` role (which properly handles custom repos).

kkeshavamurthy:
/lgtm
Thanks for fixing this @codenrhoden

PR #863, opened 2022-04-14
Fix building Flatcar images for Flatcar stable releases after 3139.2.0

### LINKED ISSUE #861: image_version is not being passed to resulting CAPI SIG images on Azure
Discovered while investigating https://github.com/flatcar-linux/Flatcar/issues/719 and https://github.com/kubernetes-sigs/image-builder/issues/859. Turns out `image_version` parameter is only passed to the first builder in https://github.com/kubernetes-sigs/image-builder/blob/a09b089b1b344d75275cf741cafdbd877050f660/images/capi/packer/azure/packer.json#L2, which makes resulting image always use latest version of SKU, even if different one is specified.

Specified image version to be used as a base image.

image-builder version a09b089b1

CC @jepio

/kind bug

### LINKED ISSUE #862: Building Flatcar images fails due to os_family Ansible mismatch

Because of https://github.com/kubernetes-sigs/image-builder/issues/861 and the regression in recent Flatcar stable release https://github.com/flatcar-linux/Flatcar/issues/719, building SIG Flatcar images is right now failing.

Ci to succeed.

/kind bug

invidian:

Fixes for building Flatcar Azure SIG images related to recent regression. See commit messages for more details.

Closes #862
Closes #861

### DIFF
--- a/images/capi/ansible/python.yml
+++ b/images/capi/ansible/python.yml
@@ -13,6 +13,8 @@
 # limitations under the License.
 ---
 - hosts: all
+  # Gathering facts requires Python to be available, so it's a chicken and egg
+  # problem as this playbook installs Python.
   gather_facts: no
   become: yes
 
--- a/images/capi/ansible/roles/python/tasks/main.yml
+++ b/images/capi/ansible/roles/python/tasks/main.yml
@@ -17,5 +17,6 @@
   register: distrib_id
 
 - include_tasks: flatcar.yml
+  # We can't use ansible_os_family fact here for consistency, as facts gathering
+  # is disabled in the playbook which includes this role. See playbook for more details.
   when: distrib_id.stdout_lines[0] is search("Flatcar")
-
--- a/images/capi/ansible/roles/setup/tasks/flatcar.yml
+++ b/images/capi/ansible/roles/setup/tasks/flatcar.yml
@@ -12,16 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- name: Fetch /etc/os-release
-  raw: cat /etc/os-release
-  register: os_release
-  changed_when: false
-  # This command should always run, even in check mode
-  check_mode: false
-  environment: {}
-
 - include_tasks: bootstrap-flatcar.yml
-  when: '"Flatcar" in os_release.stdout'
 
 - name: Create /opt/libexec overlay directories
   file:
--- a/images/capi/ansible/roles/setup/tasks/main.yml
+++ b/images/capi/ansible/roles/setup/tasks/main.yml
@@ -16,7 +16,10 @@
   when: ansible_os_family == "Debian"
 
 - import_tasks: flatcar.yml
-  when: ansible_os_family == "Flatcar"
+  # This task overrides ansible_os_family to "Flatcar" as a workaround for
+  # regression between Flatcar and Ansible, so rest of the code can use just
+  # "Flatcar" for comparison, which is the correct value.
+  when: ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]
 
 - import_tasks: redhat.yml
   when: ansible_os_family == "RedHat"
--- a/images/capi/ansible/roles/setup/tasks/bootstrap-flatcar.yml
+++ b/images/capi/ansible/roles/setup/tasks/bootstrap-flatcar.yml
@@ -21,3 +21,14 @@
     ansible_python_interpreter: "{{ bin_dir }}/python"
   tags:
     - facts
+
+# Some tasks are not compatible with Flatcar, so to centralize and deduplicate the logic of checking
+# if we run on Flatcar, we define it here.
+#
+# This is required until https://github.com/ansible/ansible/issues/77537 is fixed and used.
+- name: Override Flatcar's OS family
+  set_fact:
+    ansible_os_family: Flatcar
+  when: ansible_os_family == "Flatcar Container Linux by Kinvolk"
+  tags:
+    - facts
--- a/images/capi/packer/azure/flatcar.json
+++ b/images/capi/packer/azure/flatcar.json
@@ -8,6 +8,7 @@
   "image_offer": "flatcar-container-linux-free",
   "image_publisher": "kinvolk",
   "image_sku": "{{env `FLATCAR_CHANNEL`}}",
+  "image_version": "{{env `FLATCAR_VERSION` }}",
   "kubernetes_cni_source_type": "http",
   "kubernetes_source_type": "http",
   "plan_image_offer": "{{user `image_offer`}}",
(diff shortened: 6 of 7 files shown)


invidian:
Updated to include more generic fix. Let's see if that fixes the CI.

images/capi/ansible/roles/python/tasks/main.yml
```diff
@@ -17,5 +17,4 @@
   register: distrib_id
 
 - include_tasks: flatcar.yml
-  when: distrib_id.stdout_lines[0] is search("Flatcar")
-
+  when: ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]
```
invidian, line 20: We still end up with couple of `ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]` and many previously used `ansible_os_family == "Flatcar"`.

Ideally, we would be able to unify it and this will happen, when Ansible is fixed and updates, however, it may take a while.

Alternative path would be to use `set_fact` at some higher level in execution hierarchy, e.g. in the playbook before roles are applied. It could override either `ansible_os_family` or introduce some custom fact like `is_flatcar`, which should be easier to use, however inconsistent with other existing uses of `ansible_os_family`.

To summarize, I think proposed set of changes is not ideal, it's "long-term temporary", but it should do the job and it minimizes the changes needed now and in the future.

invidian:
Ubuntu error on CI seems to be a fluke, but Flatcar is something I need to look into.

Ankitasw:
hi @invidian, any more changes left in this PR, as this change works fine while building CAPA AMI images for flatcar OS. Any idea for the timeline of new release with this fix?

images/capi/ansible/python.yml
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 ---
 - hosts: all
-  gather_facts: no
```
invidian, line 16: > hi @invidian, any more changes left in this PR, as this change works fine while building CAPA AMI images for flatcar OS. Any idea for the timeline of new release with this fix?

So this change does not work without Python installed and this playbook does exactly that, so I have to re-evaluate how to make Flatcar detection here, as it seems it cannot be consistent with other roles. `ansible_os_family` doesn't seem available, while `distrib_id` definitely is and `os_release` may probably be available as well.

invidian:
The CI is passing now, but I'd appreciate more opinions on introduced changes. @jepio or @johananl maybe one of you could have a look please?

images/capi/ansible/roles/setup/tasks/flatcar.yml
```diff
@@ -21,7 +21,7 @@
   environment: {}
 
 - include_tasks: bootstrap-flatcar.yml
-  when: '"Flatcar" in os_release.stdout'
+  when: ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]
```
jepio, line 24: This check could be removed since we are already in `flatcar.yml`. That also removes the need to check /etc/os-release here.

images/capi/ansible/roles/setup/tasks/main.yml
```diff
@@ -16,7 +16,7 @@
   when: ansible_os_family == "Debian"
 
 - import_tasks: flatcar.yml
-  when: ansible_os_family == "Flatcar"
+  when: ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]
```
jepio, line 19: This could use the comment about flatcar.yml being the one that overrides `ansible_os_family` to make it uniform everywhere else.

images/capi/ansible/python.yml
jepio: Interesting thing is this (python.yml) playbook is only ever used by the azure builder. All other builders directly call `bootstrap-flatcar.sh` as a script.

images/capi/ansible/roles/setup/tasks/flatcar.yml
invidian: Done, thanks!

invidian:
CI failures seems random and unrelated to the PR content :cry: a

images/capi/ansible/python.yml
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 ---
 - hosts: all
+  # Gathering facts require Python to be available, so it's a chicken egg
```
johananl, line 16:
```suggestion
  # Gathering facts requires Python to be available, so it's a chicken and egg
```

images/capi/ansible/python.yml
```diff
@@ -13,6 +13,8 @@
...
+  # problem, as this playbook installs Python.
```
johananl, line 17:
```suggestion
  # problem as this playbook installs Python.
```

images/capi/ansible/roles/python/tasks/main.yml
```diff
@@ -17,5 +17,6 @@
   register: distrib_id
 
 - include_tasks: flatcar.yml
+  # We can't use ansible_os_family fact here for consistency, as facts gathering
+  # is disabled in playbook including this role. See playbook for more details.
```
johananl, line 21:
```suggestion
  # is disabled in the playbook which includes this role. See playbook for more details.
```

Resolve ambiguity.

images/capi/ansible/roles/setup/tasks/flatcar.yml
```diff
@@ -21,7 +21,6 @@
   environment: {}
 
 - include_tasks: bootstrap-flatcar.yml
-  when: '"Flatcar" in os_release.stdout'
```
johananl, line 24: Can't we remove the `Fetch /etc/os-release` task, now that we don't care about its output?

images/capi/ansible/roles/setup/tasks/main.yml
```diff
@@ -16,7 +16,10 @@
   when: ansible_os_family == "Debian"
 
 - import_tasks: flatcar.yml
-  when: ansible_os_family == "Flatcar"
+  # This task will override ansible_os_family to "Flatcar" as a workaround for
```
johananl, line 19:
```suggestion
  # This task overrides ansible_os_family to "Flatcar" as a workaround for
```

Avoid future tense in docs.

images/capi/ansible/roles/setup/tasks/main.yml
```diff
@@ -16,7 +16,10 @@
...
+  # regression between Flatcar and Ansible, so rest of the code can use just
+  # "Flatcar" for comparison, which is the desired value.
```
johananl, line 21:
```suggestion
  # "Flatcar" for comparison, which is the correct value.
```

johananl (approved):
This seems to work well on my testing. Nice job overriding the fact rather than modifying all the `when` cases.

Added some non-blocking nits.

images/capi/ansible/roles/setup/tasks/flatcar.yml
invidian: Good point! Removed :)

codenrhoden:
I like the way this looks now, nice work incorporating all the feedback, @invidian. 

It appears this fixes an outstanding issue with Flatcar in the Azure sigs test, so it will be nice to have that test stabilize. I will keep an eye out if any other feedback comes in, but I'd like to see this merge today.

PR #868, opened 2022-04-26
Set transparent huge pages to madvise

SanikaGawhane:
Photon defaults to always using transparent huge pages. This is fine for desktops but is risky for servers. Some applications can lose up to 30% performance. Setting it to `madvise` can help with this.

Ubuntu 20.04 has this set to `madvise` by default. This PR is to add and make it uniform across all -EL distros. 
A corresponding GOSS test ensures this setting is applied.

/cc @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/firstboot/tasks/photon.yml
+++ b/images/capi/ansible/roles/firstboot/tasks/photon.yml
@@ -14,4 +14,11 @@
 
 # no-op task just to have something for the role to do. Right now
 # all the work happens in the setup role
-- meta: noop
+# - meta: noop
+
+- name: Set transparent huge pages to madvise
+  lineinfile:
+    path: /boot/photon.cfg
+    backrefs: yes
+    regexp: "^(?!.*transparent_hugepage=madvise)(photon_cmdline.*)"
+    line: '\1 transparent_hugepage=madvise'
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -95,3 +95,11 @@
     sysctl_set: yes
     reload: yes
   when: ansible_distribution == "Ubuntu"
+
+- name: Set transparent huge pages to madvise
+  lineinfile:
+    path: /etc/default/grub
+    backrefs: yes
+    regexp: "^(?!.*transparent_hugepage=madvise)(GRUB_CMDLINE_LINUX=.*)(\"$)"
+    line: '\1 transparent_hugepage=madvise"'
+  when: ansible_os_family == "RedHat"
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -59,6 +59,13 @@ command:
     stderr: []
     timeout: 0
 {{end}}
+{{if eq .Vars.OS "photon"}}
+  cat /sys/kernel/mm/transparent_hugepage/enabled:
+    exit-status: 0
+    stdout: ["always [madvise] never"]
+    stderr: []
+    timeout: 0
+{{end}}
 {{range $name, $vers := index .Vars .Vars.OS .Vars.PROVIDER "command"}}
   {{ $name }}:
   {{range $key, $val := $vers}}


codenrhoden (commented):
At first glance this looks good, @SanikaGawhane. I'll have to take it for a spin. Thanks for adding the GOSS test!

We should consider (probably as a follow-up) moving the logic into a `debian.yml` file within the `node` role. We already have one for AL2 and Photon. It's probably time to add a Debian/Ubuntu one now that more and more distro-specific logic is coming in.

codenrhoden:
@SanikaGawhane A few follow-up questions...

Did you find that Ubuntu 18.04 was behaving differently than 20.04? I ask because we definitely noticed that Ubuntu 20.4 had this set already, but your changes add logic to change it for 18.04. However, when I build an 18.04 image, it *also* has this set already, so I'm wondering if you saw the same, and if that change is needed for 18.04.

When I run this, the OVA test for Photon 3 fails. It took me a bit to realize why. As we know, Photon does *not* have this set by default. The change you've made does the right thing -- if you boot a Photon 3 OVA with this change, it will be set to madvice. however, the GOSS test is testing a running system before it's been rebooted. BY adding the change to the cmdline, that change doesn't take effect until after a reboot. But the photon OVA test *passed* here, so I'm really confused on what is happening.

The failures in Azure are interesting, as they all show a null stdout. I'm wondering if perhaps there is something different about the Azure VMs that they have this setting completely disabled or something. It may be worth booting up an Azure VM and poking around to see.

I'm going to re-run the tests just to if they are consistent.

/test pull-azure-vhds
/test pull-ova-all

SanikaGawhane:
@codenrhoden thank you for the review. 
I tested with an 18.04 image and, like you said, it is set to `madvise` already. So, maybe we don't need to do anything at all for Ubuntu. I'll double check to be sure.
Photon should be okay, Ubuntu has an `update grub.cfg` step to apply the changes. It will be useful if we have something similar for Photon too. Not sure what is going on with the GOSS tests.
Also looking at Azure VMs.

codenrhoden:
@SanikaGawhane Any updates on this PR? I'd like this change to come in before tagging a new release.

SanikaGawhane:
@codenrhoden - I'll have an update on this by EOD today.

SanikaGawhane:
Updates so far - 
In `pull-ova-all` test, GOSS tests passed only for Ubuntu-1804 and Ubuntu-2004 since those were set to `madvise` by default. Photon-3 and Centos tests are failing but somehow GOSS inspect mode is on for OVAs and hence it passed without failing packer.

`pull-azure-vhds` doesn't have GOSS inspect mode on and hence fails image-builds - as it should. 
I built and sshed into Ubuntu-2004 and Centos azure images and noticed a few things here - 

1. Manually cat-ing the file works fine but GOSS can't seem to find it.
```
azureuser@ubuntu-2004-sng:/$ cat /sys/kernel/mm/transparent_hugepage/enabled
always madvise [never]
```
```
[root@centos-7-sng /]# cat /sys/kernel/mm/transparent_hugepage/enabled
always madvise [never]
```
2. transparent huge page is set to `never` on both vhds. (weird for Ubuntu 2004, given we know it is set to `madvise`). Might be worth exploring if there are any Azure specific steps that set these to `never` in all images by default.

I am still trying to find if there is a way to update grub on a running system. But in general, looks like there are a lot of discrepancies among systems. Will keep digging.

SanikaGawhane:
Update - For Photon, this task can be a part of firstboot role so that it is applied during the reboot and GOSS test can pick it up.

/test pull-ova-all

images/capi/ansible/roles/firstboot/tasks/photon.yml
```diff
@@ -15,3 +15,10 @@
 # no-op task just to have something for the role to do. Right now
 # all the work happens in the setup role
 - meta: noop
```
kkeshavamurthy, line 17: `noop` is prob not needed anymore?

SanikaGawhane:
Update - Photon has this task as a part of firstboot role so that it is applied during the reboot and GOSS test picks it up.
For EL-based systems, it is run on a node, hence no corresponding GOSS test is included.
The plan is to explore and discuss and explore this as a community issue.

codenrhoden:
This looks good to me.
/lgtm

@kkeshavamurthy any objections?

kkeshavamurthy:
Thanks @SanikaGawhane 
Can you pls open up an issue for this
```
For EL-based systems, it is run on a node, hence no corresponding GOSS test is included.
The plan is to explore and discuss and explore this as a community issue.
```

/approve

PR #869, opened 2022-04-26
Remove deprecated HAProxy OVA

codenrhoden:
CAPV has not needed the HAProxy OVA for a couple years now, and we had
long since announced the HAProxy as deprecated. This commit removes the
build target and associated config for building the HAProxy OVA.

**Additional context**
I didn't rename the existing `build-node-ova-*` targets (to get rid of the word "node") because that would be a breaking change and would break a lot of existing scripts/CI, etc. 

The change to the `image-build-ova` script shoudn't break anyone because it's always called as a packer post-processor, so while it does have a breaking change (to not pass `--node` anymore), the change is contained and not really visible.

/assign @kkeshavamurthy


kkeshavamurthy:
Thanks for removing the haproxy stuff @codenrhoden 
/lgtm

Issue #854, opened 2022-04-08
Flatcar may inadvertently update during image creation


johananl:
In https://github.com/kubernetes-sigs/image-builder/pull/701/commits/c73b6e7a28b93dd03a747a8454600838e656ff5c we've disabled Flatcar updates. However, this change leaves a time gap which still allows Flatcar to download updates and reboot **during image creation**, resulting in either a broken build or an image with an unexpected version.

Following is a high-level description of the Flatcar image build process for ISO-based Packer builds (e.g. QEMU, OVA):

1. We boot Flatcar from an official ISO or base image (e.g. AMI) of a specific Flatcar release.
2. We execute `flatcar-install` while passing an Ignition file and reboot.
3. Flatcar boots from disk.
4. Ansible is executed.
5. An image is created from the provisioned machine and the machine is terminated.

Currently, **Flatcar may inadvertently update between stage 3 and 5** (inclusive). In order to prevent this, we need to disable updates in the Ignition config we pass to `flatcar-install` at **stage 2**.

>NOTE: I've confirmed this bug exists for OVA builds, however it could apply also for AWS and Azure builds: Although these builds aren't based on booting from ISO, there is still a phase where the temporary VM is running while Ansible is executing. We should double-check updates are disabled during that phase, too.

## TODO

- [ ] Ensure updates are disabled throughout the entire build process as well as the final image for all supported providers.
  - [ ] AMI - handled by #1150
  - [ ] Azure (SIG + VHD) - handled by #1150
  - [x] OVA - handled by #895
  - [x] QEMU - handled by #895
  - [x] Raw - handled by #895

/kind bug
/assign

johananl:
Fixed in https://github.com/flatcar-linux/flatcar-packer-qemu/pull/5. Once merged, we need to update the following:

https://github.com/kubernetes-sigs/image-builder/blob/2e85b15e9e75f5a194eb0c4c3f21c47918281a41/images/capi/packer/qemu/qemu-flatcar.json#L3

https://github.com/kubernetes-sigs/image-builder/blob/2e85b15e9e75f5a194eb0c4c3f21c47918281a41/images/capi/packer/raw/raw-flatcar.json#L3

johananl:
Following https://github.com/kubernetes-sigs/image-builder/pull/873#discussion_r866007976, looks like the solution to this issue depends on https://github.com/kubernetes-sigs/image-builder/issues/890.

/hold

johananl:
[...]

PR #817, opened 2022-02-24
capi: support Flatcar OVA images

johananl:
What this PR does / why we need it:
This PR adds Flatcar as one of the distros in the VMware OVA image builder.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): None

**Additional context**

TODOs:

- [x] Add Flatcar as a CI [target](https://github.com/kubernetes-sigs/image-builder/blob/142e815b24593287025b6a53ff3f2c6625d7cfac/images/capi/scripts/ci-ova.sh#L24)?
- [ ] I'm having a hard time testing the `build-node-ova-local-flatcar` target because I can't run VMware Workstation on my machine. Should this be a blocker?
- [x] Fix GOSS errors (the build passes though).
- [x] Ensure Flatcar builds OK for all supported platforms.
- [x] Add Flatcar to https://image-builder.sigs.k8s.io/capi/providers/vsphere.html#configuration.
- [x] Squash commits?

I've tested the changes by running the following:

```
PACKER_VAR_FILES=/tmp/packer-ova.json FLATCAR_VERSION=3033.2.4 SSH_PRIVATE_KEY_FILE=/tmp/packer-key SSH_PUBLIC_KEY_FILE=/tmp/packer-key.pub make build-node-ova-vsphere-flatcar
```

Sample `/tmp/packer-ova.json` contents:

```JSON
{
    "vcenter_server":"vsphere.example.com",
    "username":"foo",
    "password":"bar",
    "datacenter": "MyDC",
    "datastore":"My Datastore",
    "folder": "foo",
    "cluster": "10.1.2.3",
    "network": "VM Network",
    "kubernetes_semver": "v1.20.10",
    "convert_to_template": "true"
}
```

PR #873, opened 2022-04-29
qemu,raw: Freeze Flatcar version

johananl:

Flatcar is an auto-updating Linux distro. We want to ensure that Flatcar doesn't update **during image creation** because if that happens we could end up with an image containing an unexpected Flatcar version.

Currently the only providers which seem to be affected by this issue are QEMU and raw. This PR fixes the problem by using an Ignition bootstrap file which disables updates during image creation.

This PR is required for #854 but doesn't close the issue on its own since additional testing and possible fixes are required for the rest of the providers, too.

**Additional context**
None


images/capi/packer/qemu/qemu-flatcar.json
```diff
@@ -1,6 +1,6 @@
 {
   "ansible_extra_vars": "ansible_python_interpreter=/opt/bin/python",
-  "boot_command_prefix": "curl -sLo /tmp/ignition.json https://raw.githubusercontent.com/flatcar-linux/flatcar-packer-qemu/917f209e1afd262e71f41c65c1295f29c08cb8c6/ignition-builder.json<enter>sudo flatcar-install -d /dev/sda -C {{user `channel_name`}} -V {{user `release_version`}} -i /tmp/ignition.json<enter>sudo reboot<enter><wait{{user `install_wait`}}>",
+  "boot_command_prefix": "curl -sLo /tmp/ignition.json https://raw.githubusercontent.com/flatcar-linux/flatcar-packer-qemu/daba089c7cde0456008115a5c44f90676da6d590/ignition-builder.json<enter>sudo flatcar-install -d /dev/sda -C {{user `channel_name`}} -V {{user `release_version`}} -i /tmp/ignition.json<enter>sudo reboot<enter><wait{{user `install_wait`}}>",
```
jsturtevant, line 3: Realize that this is using same pattern prior to this update but is there a released version of the file that should be used with details around what the difference between these files are?  Or should this ignition file be included in this repository for easier tracking of changes? This is a bit opaque as to what the change really is and why.  

for instance the new file is missing the `storage":` entry. and the new one has a `{
        "name": "update-engine.service",
        "mask": true
      },` Are these intentional changes? For give my ignorance with flatcare :-)

codenrhoden: good points, @jsturtevant . Looking at this, I also wonder if that hash could be coming from something like the `FLATCAR_CHANNEL` or `FLATCAR_VERSION`.

johananl: Great points @jsturtevant. Flatcar builds require Ignition bootstrap config on some providers. While it's sometimes possible to pass Ignition data using provider-specific mechanisms such as AWS user data, this isn't always the case. For that reason we've used `curl` with an internet URL for the QEMU build (and also for the OVA build that is WIP in https://github.com/kubernetes-sigs/image-builder/pull/817).

Personally I'm not happy with this pattern, especially when the Ignition files are managed in a separate repository. I've been thinking about this problem for a while now and think that we should solve the problem using one of the following ways in order of preference:

1. Get rid of the need to pass Ignition config at build time altogether. This would be best as it gets rid of the problem bit is likely impossible.
2. Find a provider-native way (e.g. AWS user data) to pass Ignition data for *all* providers.
3. Use `curl` with an internet URL as part of the build process and host the Ignition files inside the image-builder repo.

I think that the quick and easy solution is to host the Ignition files as part of image-builder (option no. 3). I can amend this PR to include the relevant Ignition file. Then, we might want to look at option no. 2 and maybe even option no. 1 in subsequent issues/PRs.

How does that sound?

johananl: You've raised valid questions regarding the diff @jsturtevant. We aren't currently managing the Ignition files using proper releases with notes etc. These are updated ad hoc, which is far from ideal. Regarding your two specific questions:

- We've dropped `storage` because we no longer need Ignition to alter disk contents (e.g. write files). We do all the necessary changes as usual using Ansible.
- We've masked two systemd units - `update-engine.service` and `locksmithd.service` in order to prevent Flatcar from initiating an update during a build. More info in the PR's description.

I think that if we start to host these files in the image-builder repo we would no longer need to treat them as an external dependency and therefore would get rid of the need to release them separately, write release notes etc. WDYT?

johananl: I've opened https://github.com/kubernetes-sigs/image-builder/issues/890 to move the discussion out of this PR. I'm currently working on a new PR for hosting the files in image-builder.

johananl:
Following https://github.com/kubernetes-sigs/image-builder/pull/873#discussion_r866007976 I'm holding this for now.

/hold

images/capi/packer/qemu/qemu-flatcar.json
jsturtevant: think that makes sense for now. Thanks!

johananl:
Once #895 is merged, we should edit this PR to consume the Ignition files from the image-builder repo.

johananl:
> Once #895 is merged, we should edit this PR to consume the Ignition files from the image-builder repo.

Following a change of direction in #895, we might want to **close** this PR eventually since we'd no longer need to consume Ignition files over the internet at all.

johananl:
Following #895 this is now obsolete.

PR #874, opened 2022-04-29
ami: Use cloud-init only on supported distros

johananl:

Not all Linux distros support cloud-init. Flatcar Container Linux, for example, uses Ignition as a bootstrap format.

In #837, a cloud-init user data script was added **for all distros**.

- Move user_data to a variable to allow per-distro overrides.
- Set the default to the existing cloud-init script.
- Override to an empty script for Flatcar.

**Additional context**
While we're at it, the cloud-init snippet added in #837 seems to be specific for Amazon Linux. Are there no side effects on other distros? IMO we may want to consider setting the `user_data` var to `""` for all distros except Amazon Linux.

cc @kkeshavamurthy

### DIFF
--- a/images/capi/packer/ami/flatcar.json
+++ b/images/capi/packer/ami/flatcar.json
@@ -12,5 +12,6 @@
   "ssh_username": "core",
   "systemd_prefix": "/etc/systemd",
   "sysusr_prefix": "/opt",
-  "sysusrlocal_prefix": "/opt"
+  "sysusrlocal_prefix": "/opt",
+  "user_data": ""
 }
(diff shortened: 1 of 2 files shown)


codenrhoden:
@kkeshavamurthy did you have any thoughts on this comment:

> While we're at it, the cloud-init snippet added in https://github.com/kubernetes-sigs/image-builder/pull/837 seems to be specific for Amazon Linux. Are there no side effects on other distros? IMO we may want to consider setting the user_data var to "" for all distros except Amazon Linux.

kkeshavamurthy:
`#cloud-config\nrepo_upgrade: none` is a cloud-init config and should not have any undesired effects on any distro which uses cloud-init. This explicitly disables `update` on first boot of the AMI and we don't have to keep track of the default settings. 


> **Additional context**
> While we're at it, the cloud-init snippet added in #837 seems to be specific for Amazon Linux. Are there no side effects on other distros? IMO we may want to consider setting the `user_data` var to `""` for all distros except Amazon Linux.

PR #876, opened 2022-05-03
Quote multi-value fields in ansible-args.json

ThatsMrTalbot:
When multiple space separated values are passed to fields such as extra_repos the intent is they all get actioned by ansible, but due to a lack of quoting in ansible-args.json only the first value gets passed in to ansible at all.

Fixes issue where only the first value in the extra_repos parameter gets actioned by ansible. The fields changed to have quotations are extra_repos, extra_debs and extra_rpms.


kkeshavamurthy:
Thanks @ThatsMrTalbot 
/lgtm
/approve

PR #878, opened 2022-05-05
Update Azure scripts for new naming scheme

mboersma:

The Azure Marketplace offer used to host CAPZ reference images has a limit on SKUs, and we've used too many. This changes the publishing scripts according to our plan to conserve SKUs and use more versions underneath.

Which issue(s) this PR fixes:

N/A

**Additional context**

### DIFF
--- a/images/capi/packer/azure/.pipelines/create-sku.yaml
+++ b/images/capi/packer/azure/.pipelines/create-sku.yaml
@@ -1,15 +1,16 @@
 # Required pipeline variables:
 # - BUILD_POOL - Azure DevOps build pool to use
 # - CONTAINER_IMAGE - Dev container image URL to use. Should have Azure CLI, Packer and Ansible.
-# - AZURE_TENANT_ID - tenant ID
 # - AZURE_CLIENT_ID - Service principal ID
 # - AZURE_CLIENT_SECRET - Service principal secret
+# - AZURE_TENANT_ID - tenant ID
 # - KUBERNETES_VERSION - version of Kubernetes to create the sku for, e.g. `1.16.2`
-# - PUBLISHER - the name of the publisher to create the sku for
 # - OFFER - the name of the offer to create the sku for
-# - SKU_TEMPLATE_FILE - the base template file to use for the sku
 # - OS - target of build e.g. `Ubuntu/Windows`
-# - OS_VERSION - target of build e.g. `18.04/2004/2019`
+# - OS_VERSION - target of build e.g. `18.04/2004/2019/2022-containerd`
+# - PUBLISHER - the name of the publisher to create the sku for
+# - SKU_TEMPLATE_FILE - the base template file to use for the sku
+# - VM_GENERATION - VM generation to use, e.g. `gen2`
 
 jobs:
 - job: create_sku
--- a/images/capi/packer/azure/scripts/new-sku.sh
+++ b/images/capi/packer/azure/scripts/new-sku.sh
@@ -2,18 +2,20 @@
 
 OS=${OS:-"Ubuntu"}
 OS_VERSION=${OS_VERSION:-"18.04"}
+VM_GENERATION=${VM_GENERATION:-"gen1"}
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
 required_env_vars=(
-    "KUBERNETES_VERSION"
-    "SKU_TEMPLATE_FILE"
-    "AZURE_TENANT_ID"
     "AZURE_CLIENT_ID"
     "AZURE_CLIENT_SECRET"
-    "PUBLISHER"
+    "AZURE_TENANT_ID"
+    "KUBERNETES_VERSION"
     "OFFER"
     "OS"
     "OS_VERSION"
+    "PUBLISHER"
+    "SKU_TEMPLATE_FILE"
+    "VM_GENERATION"
 )
 
 for v in "${required_env_vars[@]}"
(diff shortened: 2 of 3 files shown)


mboersma:
/assign @CecileRobertMichon @jackfrancis @jsturtevant
/hold

I've tested this internally on Azure pipelines and it seems sane. But we should review the actual Marketplace offer results together before merging this PR.

images/capi/packer/azure/scripts/new-sku.sh
```diff
@@ -2,18 +2,20 @@
 
 OS=${OS:-"Ubuntu"}
 OS_VERSION=${OS_VERSION:-"18.04"}
+VM_GENERATION=${VM_GENERATION:-"gen1"}
```
jackfrancis, line 5: Is this not an appropriate time to rev to gen2 by default?

mboersma: 🤷🏻 I'm not sure what prevents us from doing that. I think the set of available SKUs would change somewhat, which could potentially impact users who had been accustomed to a certain VM type which now won't work with our gen2 reference images.

jackfrancis (approved):
/lgtm

images/capi/packer/azure/scripts/new-sku.sh
CecileRobertMichon: let's make that change separately and keep this PR scoped to the renaming. We haven't switched to gen 2 by default yet, mostly because the VM size support was limited for gen2 in the beginning.

images/capi/packer/azure/scripts/new-sku.sh
```diff
@@ -29,17 +31,9 @@ if [ ! -f "$SKU_TEMPLATE_FILE" ]; then
...
-IFS=' ' # reset to default value after usage
-
-major=${ADDR[0]}
-minor=${ADDR[1]}
-patch=${ADDR[2]}
-
 os=$(echo "$OS" | tr '[:upper:]' '[:lower:]')
 version=$(echo "$OS_VERSION" | tr '[:upper:]' '[:lower:]' | tr -d .)
-sku_id="k8s-${major}dot${minor}dot${patch}-${os}-${version}"
+sku_id="${os}-${version}-${VM_GENERATION}"
```
CecileRobertMichon, line 36: @codenrhoden FYI, not sure who is the right person to cc from the VMWare side for Azure image publishing to the Marketplace. We are making these changes to the "capi" offer structure that Microsoft publishes to Azure Marketplace to mitigate the risk of running into the hard limit of max number of plans that we recently discovered. If anyone wants to reach out to me or @mboersma, happy to provide more detail and discuss how this might affect you folks.

CecileRobertMichon:
> The Azure Marketplace offer used to host CAPZ reference images has a limit of 300 SKUs, and we've used over 200

shouldn't this be 200 and 100 ?

images/capi/packer/azure/.pipelines/create-sku.yaml
```diff
@@ -1,15 +1,16 @@
...
 # - AZURE_CLIENT_ID - Service principal ID
 # - AZURE_CLIENT_SECRET - Service principal secret
+# - AZURE_TENANT_ID - tenant ID
 # - KUBERNETES_VERSION - version of Kubernetes to create the sku for, e.g. `1.16.2`
-# - PUBLISHER - the name of the publisher to create the sku for
 # - OFFER - the name of the offer to create the sku for
-# - SKU_TEMPLATE_FILE - the base template file to use for the sku
 # - OS - target of build e.g. `Ubuntu/Windows`
-# - OS_VERSION - target of build e.g. `18.04/2004/2019`
+# - OS_VERSION - target of build e.g. `18.04/2004/2019/2022-containerd`
```
CecileRobertMichon, line 10: question for @jsturtevant are we actually going to use `2022-containerd` as os version for 2022? Is there another "flavor" of 2022?

CecileRobertMichon (commented):
/lgtm

images/capi/packer/azure/.pipelines/create-sku.yaml
jsturtevant: You are right there isn't another flavor but there techinically could be.  Right now this value is mapped through from `build-vhd.yaml` and is needed in this format to build the correct `make` target there

images/capi/packer/azure/scripts/new-sku.sh
codenrhoden: @CecileRobertMichon Both @kkeshavamurthy and @SanikaGawhane know way more than I do about how VMware is publishing images to marketplace. I'm not sure if they/we are using this scripts internally or not, but they will know, and should probably take a look at this to prepare for any changes.

CecileRobertMichon: @kkeshavamurthy thanks for confirming. Even if you do not use this script, you should be aware that you may run into the same SKU limit issue (note that deprecated SKUs _still_ count toward the limit, so this is 200 total SKUs total in the lifetime of the offer) as I've noticed the VMWare offer uses the same SKU naming schema. We recommend mitigating the risk before it's too late and you are blocked from further publishing.

Also checkout the preview [Azure Community Galleries](https://docs.microsoft.com/en-us/azure/virtual-machines/azure-compute-gallery#community) offering, we are ourselves considering switching from Marketplace publishing to using community galleries in the future. It is a the same as what is currently known as "Shared Image Gallery" (SIG) in the image-builder repo, but now supports making the images "public" so they can be consumed across subscriptions and tenants.

CecileRobertMichon (commented):
/approve
/hold cancel

images/capi/packer/azure/scripts/new-sku.sh
CecileRobertMichon: see also https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2294

images/capi/packer/azure/scripts/new-sku.sh
mboersma: See kubernetes-sigs/cluster-api-provider-azure#1003. Sounds like we need to find a new GPU SKU before we could switch to gen2 by default.

images/capi/packer/azure/scripts/new-sku.sh
kkeshavamurthy: That's good to know. I was not aware of the lifetime 200 limit. We deprecated a bunch of plans recently to speed up the publishing process. 
I did see the community gallery offering recently. It seems to be still in preview stage and hopefully GAs soon.

PR #882, opened 2022-05-10
Fix raw build target

### LINKED ISSUE #879: Raw build target is broken
The raw build target is currently broken: **Two** Packer builders get triggered at the same time and the boot command is being typed into the wrong session.

To reproduce, run `make build-raw-ubuntu-2004` for example.

I expected `make build-raw-ubuntu-2004` to work.

In https://github.com/kubernetes-sigs/image-builder/pull/829 we've cleaned up and simplified the QEMU build. Since the raw target is nearly identical to QEMU, we can probably address the raw target in a similar way.

/kind bug

johananl:
This PR fixes the _raw_ build target which is currently broken.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #879 

**Additional context**
To test this PR, run `make build-raw-all` and `make validate-raw-all`.


images/capi/packer/raw/packer.json
```diff
@@ -27,45 +27,9 @@
...
-      "boot_command": [
-        "{{user `boot_command_prefix`}}",
-        "http://{{ .HTTPIP }}:{{ .HTTPPort }}",
-        "{{user `boot_command_suffix`}}"
-      ],
-      "boot_wait": "{{user `boot_wait`}}",
-      "cpus": "{{user `cpus`}}",
-      "disk_interface": "virtio-scsi",
-      "disk_size": "{{user `disk_size`}}",
-      "firmware": "{{user `firmware`}}",
```
thebsdbox, line 42: Does removing all of this still build all of the various images? Also does it not make some of the `variables` section at the bottom redundant?

johananl: >Does removing all of this still build all of the various images?

Yes, this Flatcar-specific builder was originally added in order to be able to trigger custom provisioners and post-processors for Flatcar only, which is no longer needed.

As a matter of fact, the various images don't build **with** this builder in place since the two builders are currently run at the same time... You can verify this by running `make build-raw-ubuntu-1804` on master.

images/capi/packer/raw/packer.json
```diff
@@ -4,7 +4,7 @@
       "accelerator": "{{user `accelerator`}}",
       "boot_command": [
         "{{user `boot_command_prefix`}}",
-        "http://{{ .HTTPIP }}:{{ .HTTPPort }}",
+        "{{user `boot_media_path`}}",
         "{{user `boot_command_suffix`}}"
```
kkeshavamurthy, line 8: you prob need `boot_command_suffix` set to "" for flatcar.

johananl: AFAICT there is no default for this var so it's likely empty by default. Do you still think we should explicitly set the value to an empty string for Flatcar? If so, we likely need to do that for other distros and platforms.

johananl: From the Packer [docs](https://www.packer.io/docs/templates/legacy_json_templates/user-variables#usage):

>Even if you want a user variable to default to an empty string, it is best to explicitly define it. This explicitness helps reduce the time it takes for newcomers to understand what can be modified using variables in your template.

Looks like we don't have explicit defaults for `boot_command_prefix` and `boot_command_suffix`. If we decide to fix this we should probably do it in a separate PR because we need to fix additional providers.

johananl:
## TODO

- [ ] Update Flatcar boot command once https://github.com/kubernetes-sigs/image-builder/pull/895 is merged

/hold

johananl:
## TODO

- [ ] Fix GOSS for Flatcar

/hold

kkeshavamurthy:
@johananl can you pls squash the commits?

johananl:
> @johananl can you pls squash the commits?

Done @kkeshavamurthy.

images/capi/packer/raw/packer.json
```diff
@@ -99,7 +62,33 @@
...
+      "ansible_env_vars": [
+        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'"
+      ],
+      "extra_arguments": [
+        "--extra-vars",
+        "{{user `ansible_common_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_extra_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_user_vars`}}"
```
kkeshavamurthy, line 91: Can this block get executed before reboot is complete in the previous step?
Should something like `pause_before` or `pause_after` be used?

Issue #881, opened 2022-05-10
Ensure all targets are tested


johananl:
It seems like not all targets are tested by the CI. I assume that's the case because some targets have been broken for a long time and no one seemed to notice. Examples:

- https://github.com/kubernetes-sigs/image-builder/issues/879
- https://github.com/kubernetes-sigs/image-builder/issues/828

In addition, Flatcar (and maybe additional distros) aren't tested by the CI.

We should ensure all supported platforms are tested by the CI to avoid regressions.

codenrhoden:
This is always been a question of having the resources/infra to run tests on. vSphere tests are run on VMC (paid for by VMware), a public IaaS provider. Same as Azure (paid for by MSFT I believe). AWS tests have been stalled for well over a year for integration with boskos (accounts managed by K8s/CNCF). There is currently no infra to run qemu builders. OCI hasn't offered resources for their builder, neither has Digital Ocean.

QEMU builders would at least be easy -- all you need is a VM *anywhere* with qemu installed.

johananl:
Thanks for adding context @codenrhoden. Makes sense. So I guess running QEMU VMs on Prow itself is out of the question then? 🙂 
Anyway, in case I have bandwidth for improving the CI, I'll look into a solution for QEMU builds. IMO it would make a lot of sense to use QEMU for testing **all distros** and even consider making these tests blocking for PRs, since many distro-specific changes are provider agnostic and could be caught on QEMU.

johananl:
I happen to have some experience with running VMs inside containers on a CI system, however it's tricky and I expect to run into Prow resource constraints. I think your suggestion about spinning up a VM with QEMU makes the most sense.

invidian:
CC @kopiczko re: testing more targets on CI :)

PR #888, opened 2022-05-11
Fix regression in Flatcar provisioner

johananl:

In #883 I've accidentally introduced a regression which seems to have broken the AMI, QEMU and raw Flatcar builds (other distros weren't affected).

Turns out that removing the `BUILD_NAME` assignment inside `execute_command` was a mistake because variables defined within `environment_vars` don't affect `execute_command`, which made `if [[ \"${BUILD_NAME}\" == *\"flatcar\"* ]]` always evaluate to false.

Therefore, we first revert 4fb09d311a139aa67ded9c7cdcc3f8d35bd50c07 to fix the build.

Now, we are left with two "layers" of conditionals: Inside `execute_command` and inside `bootstrap-flatcar.sh` Even though I initially thought it'd be good to keep the `if` inside `bootstrap-flatcar.sh` as a failsafe, I no longer think so, hence the 2nd commit. See the commit message for a rationale.

**Additional context**
To test this PR, ensure the following builds pass:

```
make build-qemu-flatcar
make build-qemu-ubuntu-1804
```

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -91,7 +91,7 @@
       "environment_vars": [
         "BUILD_NAME={{user `build_name`}}"
       ],
-      "execute_command": "if [[ \"${BUILD_NAME}\" == *\"flatcar\"* ]]; then sudo {{.Vars}} -S -E bash '{{.Path}}'; fi",
+      "execute_command": "BUILD_NAME={{user `build_name`}}; if [[ \"${BUILD_NAME}\" == *\"flatcar\"* ]]; then sudo {{.Vars}} -S -E bash '{{.Path}}'; fi",
       "script": "./packer/files/bootstrap-flatcar.sh",
       "type": "shell"
     },
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -88,9 +88,6 @@
       "type": "shell"
     },
     {
-      "environment_vars": [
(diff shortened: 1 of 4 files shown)


johananl:
cc @kkeshavamurthy

johananl:
I'm still running manual tests. Will mark as ready for review when done.

johananl:
> I'm still running manual tests. Will mark as ready for review when done.

Done. Upon my testing both Flatcar and Ubuntu QEMU builds are passing now.

kkeshavamurthy:
Thanks @johananl 
Did you run any of the non-flatcar images and make sure nothing was broken, just to make sure?

johananl:
> Thanks @johananl Did you run any of the non-flatcar images and make sure nothing was broken, just to make sure?

Yes, as I wrote above I tested Ubuntu, too 🙂
Do you think we should check anything else? I could of course test all distros, however I didn't want to delay the fix further.

johananl:
As a side note, I think it would be good to rely on the CI to test all platforms regularly (see https://github.com/kubernetes-sigs/image-builder/issues/881).

kkeshavamurthy:
I missed Ubuntu :)
Yeah I had done some initial work around getting AWS creds for having CI. It got tossed around a while in k8s/test-infra and then I lost interest. I need to join the k8s infra meeting and try to get the creds again. Let me know if you are interested in looking into it further and I'll give you more context, it would be very helpful. I think 75% of the work is done.

PR #889, opened 2022-05-11
ova: Remove downgrade of cloud-init, no longer install guestinfo datasource when not needed

### LINKED ISSUE #675: Remove install of vsphere guestinfo metadata source for cloud-init after v0.21.3

https://github.com/canonical/cloud-init/pull/953 (thanks @akutz ) adds a new vSphere metadata source into cloud-init, which means we should remove the custom install in image builder at such a time that distro provided cloud-init packages include the new datasource by default. This should be v21.3 (@akutz  will confirm), but with the caveat that the datasource requires netifaces >=0.10.9 but Ubuntu 20.04 is currently 0.10.4, so we will need to review the Ubuntu 20.04 package that actually includes it.

The new datasource is backwards compatible , with key consumers being Cluster API and Cluster API Provider vSphere.

/kind feature

codenrhoden:
Commit 1:
    With cloud-init v22.1 available, no longer a need to downgrade on photon
    and Ubuntu. EL distros are still not >- 21.3, so they still need to the
    VMware Guestinfo Datasource.

Commit 2:
    A lot of distro-specific logic was in main.yml, while we already had a
    couple of vmware-distro.yml files to be included. this patch adds a
    vmware-redhat.yml file, and makes sure all vmware-distro files are
    included by vmware.yml. This refactoring makes it much easier to see
    what logic is running under what circumstances.

It may be easier to review this as separate commits (not PRs). Just a friendly reminder that that view is available.

Fixes #675

**Additional context**
/assign @kkeshavamurthy

### DIFF
--- a/images/capi/ansible/roles/providers/defaults/main.yml
+++ b/images/capi/ansible/roles/providers/defaults/main.yml
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-cloud_init_deb_download_url: "https://launchpad.net/ubuntu/+source/cloud-init/21.1-19-gbad84ad4-0ubuntu1~20.04.2/+build/21434629/+files/cloud-init_21.1-19-gbad84ad4-0ubuntu1~20.04.2_all.deb"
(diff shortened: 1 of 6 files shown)


kkeshavamurthy:
Thanks for the PR @codenrhoden 
/lgtm
/approve

PR #895, opened 2022-05-16
Add Flatcar bootstrap Ignition files

### LINKED ISSUE #890: Host bootstrap Ignition files in image-builder repo
Flatcar Container Linux uses Ignition (instead of e.g. Kickstart) as a bootstrap config format. On some platforms (e.g. QEMU, OVA) it's necessary to provide Ignition config during the image build process.

Right now we're consuming these files from https://github.com/flatcar-linux/flatcar-packer-qemu. Example:

https://github.com/kubernetes-sigs/image-builder/blob/60128fff6ce397998e58c491ac593473372c700b/images/capi/packer/qemu/qemu-flatcar.json#L3

This is far from ideal for multiple reasons, the main one being that the stability of the image-builder Flatcar builds is now dependent on an external repository which currently doesn't use proper versioning, releases etc.

In https://github.com/kubernetes-sigs/image-builder/pull/873#discussion_r866007976 we're discussing the matter. There are multiple ways to address the issue:

1. Get rid of the need to pass Ignition config at build time altogether. This would be best as it gets rid of the problem but is likely not feasible because we might need Ignition in some cases (for example to ensure we disable automatic updates at build time).
1. Find a provider-native way (e.g. AWS user data) to pass Ignition data for _all_ providers.
1. Use `curl` with an internet URL as part of the build process and host the Ignition files inside the image-builder repo.

I suggest we start by hosting the Ignition files in the image-builder repo, just like we're already doing with Kickstart used by other distros. In future iterations we may be able to switch to a provider-native way to inject Ignition data to builder VMs or even get rid of the need for bootstrap Ignition config altogether, however at least for now we'll have something stable that works.

/assign

johananl:
This PR adds a process for managing Ignition files used by Flatcar builds directly in the image-builder repo (until now these files used to be managed in an external repo - see #890 for more context).

In addition, the PR organizes the `images/capi/packer/files` directory a little better and adds a readme to help contributors figure out what the various files are used for and how to maintain them.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #890

**Additional context**
To test this PR, run the following and ensure the build exits cleanly:

```
make build-qemu-flatcar
```

I've tested QEMU on Linux (the raw target also uses Ignition but is currently broken - see https://github.com/kubernetes-sigs/image-builder/issues/879).

~Also, if someone with a macOS machine could ensure `make gen-ignition` works, that would be great 🙂~ Done - thanks @codenrhoden!


images/capi/packer/files/flatcar/ignition/bootstrap.json
```diff
@@ -0,0 +1,44 @@
...
+  "ignition": {
+    "config": {},
+    "security": {
+      "tls": {}
+    },
+    "timeouts": {},
+    "version": "2.3.0"
+  },
+  "networkd": {},
+  "passwd": {
```
CecileRobertMichon, line 11: I'm concerned that this will trigger the cred scanner on Azure pipelines and prevent us from running https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/azure/.pipelines/build-vhd.yaml

why are we checking in a password / password hash / can it be avoided?

johananl: Good point.

Packer needs a way to SSH into the temporary VM created during the build process in order to provision it using Ansible. For SSH to work, we need to have *some* auth mechanism in place.

A common practice in image-builder is to use password auth with the user `builder` and password `builder` for platforms which don't support automatic authorization of public SSH keys. AFAIK these platforms are vSphere, QEMU and raw. In order to be able to build Flatcar images for these platforms, we have to ensure Packer can SSH using `builder:builder` into the temporary VM. The obvious way to do so in the case of Flatcar is to configure `/etc/passwd` during early boot using Ignition. This is what we're doing.

The hash you're seeing is simply an MD5 hash of the string `builder`.

johananl: I'll try to think of a way to avoid having to specify the hash.

johananl: One way to get rid of the hash is to use public SSH keys for authenticating to the temporary VMs. This approach has the following downsides:

- Anyone who wants to build a Flatcar image would have to specify a both a private and a public SSH key for the build to succeed.
- Longer build time since the entire contents of the public key file have to be interactively typed over console as part of the boot command. The public key can't be known in advance and baked into the Ignition file because it's specified by the user at build time. We could resort to using a "special" insecure key pair like Vagrant [does](https://github.com/hashicorp/vagrant/tree/master/keys) but then we'd have to commit a private SSH key to the image-builder repo, which would probably cause the same credential scanning issue we'd like to avoid.

Maybe we can allow passwordless login to local Flatcar builds (QEMU and raw). We don't want to do that on builds where the SSH endpoint is exposed to the internet though for obvious reasons.

johananl: Another point to consider is that we're already setting clear-text SSH passwords for other distros. Examples:

https://github.com/kubernetes-sigs/image-builder/blob/9330b3d349a952d0e799fc61aa931c7593f24b0b/images/capi/packer/ova/linux/centos/http/8/ks.cfg#L29

https://github.com/kubernetes-sigs/image-builder/blob/9330b3d349a952d0e799fc61aa931c7593f24b0b/images/capi/packer/ova/linux/ubuntu/http/base/preseed.cfg#L81

These files are checked into the repo and AFAICT the only difference between them and the Ignition file is the format and the fact that in the Ignition file we're specifying a *hash* rather than the password (which is arguably even safer...).

codenrhoden:
I tested `make gen-ignition` on macOS and it works.

> To test this PR, one needs to point a Flatcar build to the bootstrap.json Ignition file in the boot_command_prefix var and ensure the build succeeds.

Can you give an example of changing to point to the file that's generated at `files/flatcar/ignition/bootstrap.json` rather than pulling it from GitHub? If I understand correctly, the point here is to generate that bootstrap file dynamically at build time, but I don't see a change in the `packer.json` file that then *uses* that file (it still points to GitHub). Why is that?

johananl:
> > To test this PR, one needs to point a Flatcar build to the bootstrap.json Ignition file in the boot_command_prefix var and ensure the build succeeds.
> 
> Can you give an example of changing to point to the file that's generated at `files/flatcar/ignition/bootstrap.json` rather than pulling it from GitHub? If I understand correctly, the point here is to generate that bootstrap file dynamically at build time, but I don't see a change in the `packer.json` file that then _uses_ that file (it still points to GitHub). Why is that?

What I meant is to still point Packer to GitHub but **to the image-builder repo** rather than https://github.com/flatcar-linux/flatcar-packer-qemu. We need *some* way to get a JSON file onto the temporary VM at build time, and currently this is done by `curl`-ing the file from GitHub. We could get rid of the `curl` and look for another way to get the file onto the VM, however this would likely entail typing the contents of the file over the boot command, which is rather slow.

Let me look into that quickly and see if that's feasible.

Anyway, to your question, here is the change I was referring to:

```diff
diff --git a/images/capi/packer/qemu/qemu-flatcar.json b/images/capi/packer/qemu/qemu-flatcar.json
index bb10cccce..d07e3e591 100644
--- a/images/capi/packer/qemu/qemu-flatcar.json
+++ b/images/capi/packer/qemu/qemu-flatcar.json
@@ -1,6 +1,6 @@
 {
   "ansible_extra_vars": "ansible_python_interpreter=/opt/bin/python",
-  "boot_command_prefix": "sudo systemctl mask sshd.socket --now<enter>curl -sLo /tmp/ignition.json https://raw.githubusercontent.com/flatcar-linux/flatcar-packer-qemu/917f209e1afd262e71f41c65c1295f29c08cb8c6/ignition-builder.json<enter>sudo flatcar-install -d /dev/sda -C {{user `channel_name`}} -V {{user `release_version`}} -i /tmp/ignition.json<enter>sudo reboot<enter>",
+  "boot_command_prefix": "sudo systemctl mask sshd.socket --now<enter>curl -sLo /tmp/ignition.json https://raw.githubusercontent.com/kubernetes-sigs/image-builder/b9eaa82123d71104e0de1a87d19f0224800446e4/images/capi/packer/files/flatcar/ignition/bootstrap.json<enter>sudo flatcar-install -d /dev/sda -C {{user `channel_name`}} -V {{user `release_version`}} -i /tmp/ignition.json<enter>sudo reboot<enter>",
   "boot_media_path": "",
   "boot_wait": "120s",
   "build_name": "flatcar-{{env `FLATCAR_CHANNEL`}}-{{env `FLATCAR_VERSION`}}",
```

johananl:
>Let me look into that quickly and see if that's feasible.

Great, we can use the Packer HTTP server to serve the Ignition file locally the same way other distros are doing with Kickstart files. This seems to work well and has the added benefit of making the Flatcar build process more similar to other builds.

I'll ensure there are no side effects and will refactor.

/hold

johananl:
@codenrhoden I pushed a new commit which switches to serving the Ignition file using Packer's built-in HTTP server. This is much better and I'm not sure why we haven't considered this earlier 🙂

The following providers are affected by this change:

- QEMU
- raw
- OVA

This PR modifies QEMU only because the raw target for Flatcar is broken (fix in https://github.com/kubernetes-sigs/image-builder/pull/882) and OVA isn't merged yet (https://github.com/kubernetes-sigs/image-builder/pull/817). Once this PR is merged I'll rebase the raw and OVA PRs and align the boot command with QEMU.

johananl:
CI is broken for Flatcar because of https://github.com/kubernetes-sigs/image-builder/issues/906.

/hold

johananl:
> CI is broken for Flatcar because of #906.

With that resolved and the green CI, I think we should be good to merge.

Testing this PR should be a breeze now: `make build-qemu-flatcar`

johananl:
@CecileRobertMichon @codenrhoden I'd love to get this merged. This change set affects Flatcar only and is easy to test (see above). Please tell me if there is anything I can do to make reviewing this PR easier. Thanks! 🙂

codenrhoden:
/lgtm
If there are no objections from @CecileRobertMichon, I'll merge this in tomorrow.

PR #899, opened 2022-05-24
Update to Rocky Linux 8.6

Pearl1594:

The PR updates Rocky Linux ISO to the latest version - 8.6 

**Additional context**


kkeshavamurthy:
If an 8.6 AMI is available, can we bump that as well pls?

codenrhoden:
> If an 8.6 AMI is available, can we bump that as well pls?

@kkeshavamurthy I just tried out if there was an 8.6 available, but it doesn't look like it:

```
--> rockylinux-8: No AMI was found matching filters: {
  Filters: [
    {
      Name: "name",
      Values: ["Rocky-8-ec2-8.6-*"]
```

I think we should just pull this in as-is.

PR #900, opened 2022-05-24
Modify cloud.cfg to preserve hostname on reboot for OVAs

kkeshavamurthy:
We've noticed that FQDN hostname gets truncated on reboots on OVAs sometimes. 
Ref: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1044
Even though the bug says it's not reproducible anymore, we do see it. 

This fix addresses the problem by modifying `preserve_hostname: false` to `preserve_hostname: true` after firstboot of the OVA. 

**Additional context**
/assign @codenrhoden


codenrhoden:
I have a setup where I can recreate this issue, and after applying this change the hostname is stable across reboots. nice work @kkeshavamurthy 

/lgtm

PR #902, opened 2022-05-25
Add RockyLinux 8 to OVA CI

codenrhoden:
With more stuff coming in related to RHEL-8, and changes to RockyLinux 8 (#899) it would be nice if there was a CI target that would verify Rocky (which is at least an -EL derivative) was still building.

/assign @kkeshavamurthy 

**Additional context**

### DIFF
--- a/images/capi/scripts/ci-ova.sh
+++ b/images/capi/scripts/ci-ova.sh
@@ -22,7 +22,7 @@ CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
 export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
-TARGETS=("ubuntu-1804" "ubuntu-2004" "photon-3" "centos-7")
+TARGETS=("ubuntu-1804" "ubuntu-2004" "photon-3" "centos-7" "rockylinux-8")
 
 on_exit() {
   # kill the VPN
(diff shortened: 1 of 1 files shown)


codenrhoden:
ah, so yeah, exactly as @Pearl1594  said in #899, the 8.5 ISO is not available anymore. But that's because the path was given as just `8` instead of `8.5`.

Instead of `https://download.rockylinux.org/pub/rocky/8/isos/x86_64/Rocky-8.5-x86_64-minimal.iso` it would be better to use `https://download.rockylinux.org/pub/rocky/8.5/isos/x86_64/Rocky-8.5-x86_64-minimal.iso` so that things don't break when a new minor version comes out. I added that to this commit to get things moving along again.

That will be especially important once this is turned on for CI, which is what I'm trying to do here.

kkeshavamurthy:
If rockylinux-8 does not with floppy kickstart, we might have to build it from clone

codenrhoden:
> If rockylinux-8 does not with floppy kickstart, we might have to build it from clone

ah yes, thanks for the reminder. THat will be the faster route for now.

Long term, it would be nice to find a non-HTTP based way to kickstart EL-8 builds. For posterity, there's some interesting approaches in here: https://github.com/jetbrains-infra/packer-builder-vsphere/issues/276

images/capi/packer/ova/rockylinux-8.json
```diff
@@ -11,7 +11,7 @@
   "http_directory": "./packer/ova/linux/{{user `distro_name`}}/http/",
   "iso_checksum": "4eb2ae6b06876205f2209e4504110fe4115b37540c21ecfbbc0ebc11084cb779",
   "iso_checksum_type": "sha256",
-  "iso_url": "https://download.rockylinux.org/pub/rocky/8/isos/x86_64/Rocky-8.5-x86_64-minimal.iso",
+  "iso_url": "https://download.rockylinux.org/pub/rocky/8.5/isos/x86_64/Rocky-8.5-x86_64-minimal.iso",
```
Pearl1594, line 14: @codenrhoden should we use the latest Rocky version i.e., 8.6?

davidjumani, line 14:
```suggestion
  "iso_url": "https://download.rockylinux.org/pub/rocky/8/isos/x86_64/Rocky-8.6-x86_64-minimal.iso",
```
Agree with @Pearl1594 
the qemu provider also uses the latest version rn which is 8.6

kkeshavamurthy:
Thanks @codenrhoden 
/lgtm

PR #907, opened 2022-06-01
Fix SSH/scp/SFTP issues for OpenSSH 9.0+ and Flatcar stable

kopiczko:
This PR tries to fix various issues with Flatcar builds. This fixes multiple issues but at this point it isn't possible to test them in isolation.

Fixes https://github.com/kubernetes-sigs/image-builder/issues/905

Workaround for https://github.com/kubernetes-sigs/image-builder/issues/859

**Additional context**
Add any other context for the reviewers

PR #912, opened 2022-06-17
Improve bootstrap-flatcar.sh

kopiczko:
There were two issues I had:

- PYTHON_VERSION was not defined (only PYTHON3_VERSION) and `set -o nounset` helps catching that
- `set -o errexit` is the same as `set -e` so there is no change in that regard
- I think `set -o pipefail` just makes sense with `set -o errexit`

I also added:

- `set -x` - to make the debugging easier
- replaced `wget -O -` with `curl -sfL` to reduce the amount of output , because wget is printing all those dots and it's difficult to control that

Extracted from https://github.com/kubernetes-sigs/image-builder/pull/907

### DIFF
--- a/images/capi/packer/files/flatcar/scripts/bootstrap-flatcar.sh
+++ b/images/capi/packer/files/flatcar/scripts/bootstrap-flatcar.sh
@@ -1,13 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # This script installs PyPy as a Python interpreter on a Flatcar instance.
 
-set -e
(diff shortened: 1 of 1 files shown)


kopiczko:
CC @invidian @johananl

invidian (approved):
Looks good.

johananl (approved):
LGTM. Thanks @kopiczko!

invidian:
/test pull-azure-sigs

This should run to test Flatcar builds and I expect it to fail because of missing #907.

PR #914, opened 2022-06-20
capi: remove pki Ansible role

invidian:
Travis Rhoden said it is not used anymore and should be removed.

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>


invidian:
/retest

Hmm, it fails with the following error:
`[0;32m    vsphere-clone: fatal: [default]: UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo /tmp/.ansible `\"&& mkdir \"` echo /tmp/.ansible/ansible-tmp-1655714528.3113294-1027-246527224469686 `\" && echo ansible-tmp-1655714528.3113294-1027-246527224469686=\"` echo /tmp/.ansible/ansible-tmp-1655714528.3113294-1027-246527224469686 `\" ), exited with result 1", "unreachable": true}[0m
`
Seem odd.

codenrhoden:
that weird OVA failure will be fixed with #918

Issue #897, opened 2022-05-18
Ensure all Packer vars have a default value


johananl:
Following [this](https://github.com/kubernetes-sigs/image-builder/pull/882#discussion_r873743606) discussion, we might want to ensure all Packer vars have an explicit default even if the default value is an empty string:

From the Packer [docs](https://www.packer.io/docs/templates/legacy_json_templates/user-variables#usage):

>Even if you want a user variable to default to an empty string, it is best to explicitly define it. This explicitness helps reduce the time it takes for newcomers to understand what can be modified using variables in your template.

PR #917, opened 2022-06-23
Add option for custom pre/post roles

### LINKED ISSUE #894: Proposal: pre/post custom roles
I'd like to propose the idea of having both the `node.yml` and `firstboot.yml` playbook having a pre and post custom role. What I mean by this is each playbook having the opportunity to run two custom roles -- one as early as possible, and one as late as possible (for the node playbook, latest possible would mean right before the `sysprep` role, which is the one place we allow custom roles now).

Early on in the project, a bunch of flags came in to allow for early node setup around the idea of making builds possible in enterprise environments, air-gapped environments, and for folks using repo mirrors. This introduced a lot of the flags found in the [docs](https://image-builder.sigs.k8s.io/capi/capi.html#customization) around `extra_repos`, and disabling/re-enabling public repos. One thing I'm seeing in the project is more and more use cases for wanting to do some early OS-level customization **before** any installations begin.

A few examples:

- custom CA certs (https://github.com/kubernetes-sigs/image-builder/pull/720)
- configuring a PyPI mirror (https://github.com/kubernetes-sigs/image-builder/pull/850)
- A need to configure some custom DNS settings (in `/etc/systemd/resolved.conf` during **build** time, and undo these changes before finishing the images). This is currently not possible, and is what is prompting me to propose these changes rather than adding **another** flag for what is probably a highly specific use case.

Our one opportunity to run custom roles does not happen at the right time to cover the above use cases, but it does serve an important function and needs to stay. I'm proposing we generalize the approach and give each playbook an opportunity to essentially to customization and any needed undoing of that customization.

For example, I want to take the `node.yml` playbook from this:

```yaml
  tasks:
    - include_role:
        name: node
    - include_role:
        name: providers
    - include_role:
        name: containerd
    - include_role:
        name: kubernetes
    - include_role:
        name: load_additional_components
      when: load_additional_components | bool
    - include_role:
        name: "{{ role }}"
      loop: "{{ custom_role_names.split() }}"
      loop_control:
        loop_var: role
      when: custom_role | default(false)|bool
    - include_role:
        name: sysprep

```

to this:

```yaml
  tasks:
    - include_role:
        name: "{{ role }}"
      loop: "{{ node_custom_roles_pre.split() }}"
      loop_control:
        loop_var: role
      when: node_custom_roles_pre != ""
    - include_role:
        name: node
    - include_role:
        name: providers
    - include_role:
        name: containerd
    - include_role:
        name: kubernetes
    - include_role:
        name: load_additional_components
      when: load_additional_components | bool
    - include_role:
        name: "{{ role }}"
      loop: "{{ node_custom_roles_post.split() }}"
      loop_control:
        loop_var: role
      when: node_custom_roles_post != ""
    - include_role:
        name: sysprep
```

This would be a non-backwards compatible change, and so I want to open this up for discussion. The existing logic that deals with custom repositories (adding, removing, disabling public repos) is used heavily (at least by VMware, I'm not as sure outside of VMware), and I would like to keep it. However, I think the more useful way to do this would be to publish example/template roles within the repo that can be included in the proposed pre/post hooks to be included when needed. That way they are there for those who want/need them, but they aren't all there by default. That would get the required variables to use them out of files like `images/capi/packer/config/ansible-args.json`, and instead you would pass what you need via `ansible_user_vars`.

This would improve readability by having fewer variables that are passed (and are usually empty) every time, and instead only passing them when they are needed.

If we took this approach but didn't rename `custom_role_names`, we could technically make this backwards compatible. But I am proposing both renaming it, and eliminating the use of a separate boolean (`custom_role`) to control it, rather inferring the use based on whether the string is non-empty. This is to make the naming of the variables for the roles rational, like the following:

- node_custom_roles_pre
- node_custom_roles_post
- firstboot_custom_roles_pre
- firstboot_custom_roles_post

I think not renaming them to something consistent would be very confusing. But I am very curious to get others' thoughts on this. FIrst off, thoughts on the proposal, but also any feedback on how often people are using custom roles and the ability to manipulate repo definitions. Again, currently repo definitions are not modified via custom role, but I am proposing that we change that to happen in a custom role, **and** we publish example/template roles to do so that are usable out of the box.

/assign @CecileRobertMichon @jsturtevant @kkeshavamurthy 
/cc @voor @MaxRink 

Please ping anyone that you think may be interested. Just was naming people off the top of my head. :)

codenrhoden:
For each playbook (node, windows-node, and firstboot) and an option
pre/post hook for a custom role. This standardizes the way custom roles
are handled across all playbooks, and maintains backwards compatibility
by also making sure that "custom_role_names" is handled just like
"node_custom_roles_post".

The use of the boolean "custom_role" is no longer needed or supported,
as the custom roles are run as needed implicitly bases on whether roles
are defined or empty.

Fixes #894

**Additional context**
I made sure that the new vars had default values in Packer, in the spirit of #897.
The "custom_role" metadata that was present in the OVA and Azure (and only those two) is removed, as I didn't know how to do an "or" of four strings and pass through a boolean if non-empty in YAML. :) Also, I don't think any uses that metadata, but I want to point out that it's removed in case I am wrong.

/assign @kkeshavamurthy


codenrhoden:
OVA builds will be fixed with #918

codenrhoden:
/test pull-ova-all

OVA builds should be fixed now. azure-sigs will be blocked until the flatcar issues are resolved (currently fix is in #912)

CecileRobertMichon (commented):
/lgtm
/approve

PR #920, opened 2022-06-24
prevent package upgrades on node images

kkeshavamurthy:
Customers sometimes end up upgrading the node images unintentionally and this breaks a lot of the functionality since the VMs are supposed to be immutable. 
This PR just make the upgrades a little bit harder to do by locking all the packages in the `sysprep` role. A root user can still remove the hold/exclution if needed. 

**Additional context**

/assign @codenrhoden

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/redhat.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/redhat.yml
@@ -34,6 +34,19 @@
     masked: yes
   when: ansible_memory_mb.swap.total != 0 and ansible_distribution_major_version|int == 8
 
+- name: Get installed packages
+  package_facts:
+
+- name: create the package list
+  set_fact:
+    package_list: "{{ ansible_facts.packages.keys() | join(' ') }}"
+
+- name: exclude the packages from upgrades
+  lineinfile:
+    path: /etc/yum.conf
+    regexp: '^exclude='
+    line: exclude={{ package_list }}
+
 - name: Remove RHEL subscription
   block:
     - name: Remove subscriptions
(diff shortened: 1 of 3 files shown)


images/capi/ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -34,6 +34,19 @@
     masked: yes
   when: ansible_memory_mb.swap.total != 0 and ansible_distribution_major_version|int == 8
 
+- name: Get installed packages
+  package_facts:
```
codenrhoden, line 38: This looks odd, but I assume it's a built-in that just re-runs the facts module just to get package info (and re-populates `ansible_facts.packages`)?

images/capi/ansible/roles/sysprep/tasks/redhat.yml
```diff
@@ -34,6 +34,19 @@
...
+
+- name: create the package list
+  set_fact:
+    package_list: "{{ ansible_facts.packages.keys() | join(' ') }}"
+
+- name: exclude the packages from upgrades
+  lineinfile:
+    path: /etc/yum.conf
```
codenrhoden, line 46: at first I thought putting in the top level file was weird, but then I realized that putting it here means it apples to all repos that are defined in `yum.repos.d`, which is nice!

codenrhoden:
LGTM at first glance.
Let's make sure nothing breaks for Flatcar:
/test pull-azure-sigs

images/capi/ansible/roles/sysprep/tasks/redhat.yml
kkeshavamurthy: yup. It's probably run at the begininig, but needs to be re-run to get the updated list.

kkeshavamurthy:
/test pull-azure-sigs

Flatcar tests passed

PR #928, opened 2022-07-05
Adding containerd_timeout variable for Windows

knabben:
A few scenarios can have the win_get_url take more than the default 30 seconds for containerd download, this PR allows the user to increase the default timeout.

**Additional context**

### DIFF
--- a/images/capi/ansible/windows/roles/runtimes/tasks/containerd.yml
+++ b/images/capi/ansible/windows/roles/runtimes/tasks/containerd.yml
@@ -18,6 +18,7 @@
     dest: '{{ tempdir.stdout | trim }}\containerd.tar.gz'
     checksum: '{{ containerd_sha256 }}'
     checksum_algorithm: "sha256"
+    url_timeout: 300
   register: containerd
   retries: 5
   delay: 3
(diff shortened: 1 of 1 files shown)


images/capi/ansible/windows/roles/runtimes/tasks/containerd.yml
```diff
@@ -18,6 +18,7 @@
     dest: '{{ tempdir.stdout | trim }}\containerd.tar.gz'
     checksum: '{{ containerd_sha256 }}'
     checksum_algorithm: "sha256"
+    url_timeout: '{{ containerd_timeout }}'
```
jsturtevant, line 21: would it be easier to set this to a higher value here for this particular fetch?   Having a higher timeout here isn't goin to cause any issues I think. Either that or have a url_timeout value that could be provided to all url fetches?  

Otherwise could we use containerd_url_timeout?  As this field is exposed up to the user better to be a bit more explicit

knabben: I can hardcore here instead (since other values are already doing it).
Does 300s makes sense?

PR #938, opened 2022-07-19
use pause image from mcr.microsoft.com for azure images

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

This PR makes it so the pause/infra image used when building Azure VHDs comes from mcr.microsoft.com.

**Additional context**

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -385,7 +385,7 @@ $(GCE_VALIDATE_TARGETS): deps-gce
 
 .PHONY: $(AZURE_BUILD_VHD_TARGETS)
 $(AZURE_BUILD_VHD_TARGETS): deps-azure
-	. $(abspath packer/azure/scripts/init-vhd.sh) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-vhd.json)" -var-file="$(abspath packer/azure/$(subst build-azure-vhd-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
+	. $(abspath packer/azure/scripts/init-vhd.sh) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-vhd.json)" -var-file="$(abspath packer/azure/$(subst build-azure-vhd-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) -var-file="$(abspath packer/azure/common.json)" packer/azure/packer$(findstring -windows,$@).json 
 
 .PHONY: $(AZURE_VALIDATE_VHD_TARGETS)
 $(AZURE_VALIDATE_VHD_TARGETS): deps-azure
--- a/images/capi/packer/azure/common.json
+++ b/images/capi/packer/azure/common.json
@@ -1,0 +1,3 @@
+{
+  "pause_image": "mcr.microsoft.com/oss/kubernetes/pause:3.8"
+}


marosset:
I'm running some builds locally and have verified that this setting is being correctly picked up/passed to ansible

```bash
==> vhd-windows-2019-containerd: Executing Ansible: ansible-playbook -e packer_build_name="vhd-windows-2019-containerd" -e packer_builder_type=azure-arm -e ansible_winrm_server_cert_validation=ignore ansible_winrm_operation_timeout_sec=120 ansible_winrm_read_timeout_sec=150 --extra-vars runtime=containerd docker_ee_version=19.03.12 containerd_url=https://github.com/containerd/containerd/releases/download/v1.6.2/containerd-1.6.2-windows-amd64.tar.gz containerd_sha256=4fdf1850a132599cf679992d8405a54416a5ab2982959db20d969054c40a2b71 pause_image=mcr.microsoft.com/oss/kubernetes/pause:3.8
```

mboersma (approved):
/lgtm

images/capi/Makefile
```diff
@@ -385,7 +385,7 @@ $(GCE_VALIDATE_TARGETS): deps-gce
 
 .PHONY: $(AZURE_BUILD_VHD_TARGETS)
 $(AZURE_BUILD_VHD_TARGETS): deps-azure
-	. $(abspath packer/azure/scripts/init-vhd.sh) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-vhd.json)" -var-file="$(abspath packer/azure/$(subst build-azure-vhd-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
+	. $(abspath packer/azure/scripts/init-vhd.sh) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-vhd.json)" -var-file="$(abspath packer/azure/$(subst build-azure-vhd-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) -var-file="$(abspath packer/azure/common.json)" packer/azure/packer$(findstring -windows,$@).json
```
jsturtevant, line 388: will need to add this to the `AZURE_SIG` targets and `AZURE_VALIDATE` targets as well.

jsturtevant (commented):
Is this the right place for the value?  It means that anyone building azure images will not get the default pause image configured by the project: https://github.com/kubernetes-sigs/image-builder/blob/6f62980c20841e5648a03e016c6b520c0769462e/images/capi/packer/config/common.json#L14

This might be confusing to a user or they may wish to use the upstream images.

If we used a configuration step in our pipelines that produce the images we could overload it there like we do with the k8s version: 
https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/azure/.pipelines/build-vhd.yaml#L21

@CecileRobertMichon do you have an opinion?

images/capi/Makefile
marosset: Oh yea, let me add it to SIG and validate as well.

I'm OK if we want to have this get added in a configuration step but was thinking we should be intentional to use this for all of the images built for azure/CAPZ.

CecileRobertMichon:
> This might be confusing to a user or they may wish to use the upstream images.
> 
> If we used a configuration step in our pipelines that produce the images we could overload it there like we do with the k8s version

+1 to @jsturtevant's comment, this might not be the best place to configure it if we're just looking to change the image for the reference images we build. However, it might be worth making it easier to use all the MCR images long-term with some opt-in flag or variable for the Azure build as I could see that being a common use case not just for reference images but for first/third party users as well.

marosset:
> > This might be confusing to a user or they may wish to use the upstream images.
> > If we used a configuration step in our pipelines that produce the images we could overload it there like we do with the k8s version
> 
> +1 to @jsturtevant's comment, this might not be the best place to configure it if we're just looking to change the image for the reference images we build. However, it might be worth making it easier to use all the MCR images long-term with some opt-in flag or variable for the Azure build as I could see that being a common use case not just for reference images but for first/third party users as well.

ack, I'll look into different ways this can be specified.
Thanks!

kkeshavamurthy:
@marosset Are you still working on this?
/remove-lifecycle rotten

mboersma:
@jsturtevant @marosset should we keep this PR open?

marosset:
/close

we can re-activate if needed

Issue #814, opened 2022-02-22
Migrate Packer files from JSON to HCL


johananl:
**Is your feature request related to a problem? Please describe.**

JSON doesn't support comments whereas HCL does. Being unable to comment inline makes it harder to maintain Packer code because we can't explain tricky/unusual bits, add TODOs in context etc.

**Describe the solution you'd like**

As of `1.7.0`, the recommended config format for Packer is HCL. We should consider migrating. There is an [upgrade tool](https://learn.hashicorp.com/tutorials/packer/hcl2-upgrade).

**Describe alternatives you've considered**

We could stay with JSON and not use comments.

**Additional context**
None.

/kind feature

codenrhoden:
Thanks for starting this conversation, @johananl. It's definitely worth discussion, especially since the recommendation changed with Packer 1.7 as you pointed out. I know the project had talked about it once before (well over a year ago), but I can't find any notes that were captured on the topic. One thing that's nice about the JSON files is that they are easily manipulated with tools like `jq`, but we've talked about making some pretty sweeping changes in this area anyways, and HCL may as well be a part of that discussion.

CiraciNicolo:
Hi, 

Can we revamp the discussion? I'm interested in working on this since this can be a blocker for #1437 - JSON template are very limited as also indicated in the official packer [documentation](https://developer.hashicorp.com/packer/docs/templates/legacy_json_templates). Are you okay with that?

Regarding the issue raised by @codenrhoden about manipulation of file [this](https://github.com/mattolenik/hclq) project seems interesting.

jsturtevant:
doing a big change like this probably depends on the outcome of what we end up doing with https://github.com/kubernetes-sigs/image-builder/issues/1246.  Maybe its time to re-discuss the usage of packer

jsturtevant:
Why is this a blocker for #1437?

CiraciNicolo:
Hello! Sorry, I closed #1437 and re-open as #1442 - I mishandled the commits and did all the work on the main branch, causing issues. Regarding why is this a blocker: in #1442 I'm proposing to deprecate `containerd_version` user variable by introducing a standard `cri_version`. To implement this I would like to start gradually implementing the new variable, but maintaining the older one with more importance. JSON packer file does not allow complex evaluation as HCL2 does. 

To be more clear I would like to leverage HCL2 more complex function to keep the user un-aware of the change, without the needs to keep complex handling into the ansible files.

PR #942, opened 2022-07-30
expose advanced packer googlecompute auth options to users

### LINKED ISSUE #941: GCE: Expose advanced packer googlecompute auth options to users
We'd like to leverage Workload Identity Federation (in our case from Github to Google Cloud) to build CAPI images. Currently we're using a secret with a static JSON service account key that is mounted into the container in CI and point `GOOGLE_APPLICATION_CREDENTIALS` at it, as outlined in the GCP section of the docs: https://image-builder.sigs.k8s.io/capi/providers/gcp.html

Add uservars for the advanced auth options for the googlecompute packer builder: https://www.packer.io/plugins/builders/googlecompute#precedence-of-authentication-methods

While dumping static service account JSON keys into a file at build time totally does work, it's not exactly a great process to follow if better options are available and easily achieved.

**Additional context**
I'll mock this up on a fork and open a PR today or tomorrow.

/kind feature

itspngu:

Exposes some more of the knobs that come with the `googlecompute` packer builder.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #941 

**Additional context**
I'd have liked to make some more settings configurable, such as those related to [`iap`](https://www.packer.io/plugins/builders/googlecompute#use_iap) and [`oslogin`](https://www.packer.io/plugins/builders/googlecompute#use_os_login), but as far as I can tell that's not possible with the JSON format (if it is, please do point me in the right direction so I can add that) and thus blocked-ish by #814 

The choice of `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` as an environment variable name is in line with [what the Terraform does](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#impersonate_service_account) but not a standard for Google tooling (though I wish it was... x)) and can be removed or renamed if needed.

### DIFF
--- a/images/capi/hack/ensure-packer.sh
+++ b/images/capi/hack/ensure-packer.sh
@@ -20,7 +20,7 @@ set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-_version="1.7.2"
+_version="1.7.10"
 
 # Change directories to the parent directory of the one in which this
 # script is located.
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -1,9 +1,11 @@
 {
   "builders": [
     {
+      "account_file": "{{ user `service_account_key_file` }}",
       "disable_default_service_account": "{{ user `disable_default_service_account` }}",
       "image_family": "capi-{{user `build_name`}}-k8s-{{user `kubernetes_series` | clean_resource_name}}",
       "image_name": "cluster-api-{{user `build_name`}}-{{user `kubernetes_semver` | clean_resource_name}}-{{user `build_timestamp`}}",
+      "impersonate_service_account": "{{ user `impersonate_service_account` }}",
       "labels": {
         "build_timestamp": "{{user `build_timestamp`}}",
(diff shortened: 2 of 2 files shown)


itspngu:
I'm starting to think that this PR might not be the way to do it, there's too many loose ends. There were some interesting changes to the `googlecompute` builder that aren't part of any packer release yet (https://github.com/hashicorp/packer-plugin-googlecompute/releases/tag/v1.0.14),the used version of packer itself is one minor version behind upstream, and the (pending?) migration from JSON to HCL for the configuration files would bring its own benefits. Thoughts?

kkeshavamurthy:
@itspngu any updates on this?

Regarding migration from json to hcl, we have a lot of upstream and downstream pipelines tinkering the json files. Even though the process of conversion to HCL is easy, it'll break a lot things. So IMO, it might not be something that'll happen in the short term but we should definitely plan for it.

kkeshavamurthy:
Closing the PR since there has been no activity in 5 months. Please re-open when you work on it.

PR #947, opened 2022-08-11
updating dynamic port range for windows to address some collisions

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

This fixes some issues where ports inside the current dynamic range (33000-65535) are being used by Windows components.
This can random issues if container workloads are allocated ports that are already in use.
Lots of details in https://github.com/kubernetes/kubernetes/issues/107114

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes https://github.com/kubernetes/kubernetes/issues/107114

**Additional context**

/sig windows

### DIFF
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -159,8 +159,8 @@ command:
     exit-status: 0
     exec: netsh int ipv4 show dynamicportrange tcp
     stdout:
-    - "Start Port      : 33000"
-    - "Number of Ports : 32536"
+    - "Start Port      : 34000"
+    - "Number of Ports : 31536"
     timeout: 30000
 {{end}}
 
--- a/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
@@ -133,8 +133,8 @@
     data: 1
     type: dword
 
-- name: Expand dynamic port range to 33000-65535 to avoid port exhaustion
-  win_shell: netsh int ipv4 set dynamicportrange tcp 33000 32536
+- name: Expand dynamic port range to 34000-65535 to avoid port exhaustion
+  win_shell: netsh int ipv4 set dynamicportrange tcp 34000 31536
 
 - name: Add required Windows Features
   win_feature:


kkeshavamurthy:
@marosset This goss [check](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-command.yaml#L162) need to be modified as well

@knabben @jsturtevant any thoughts on the new port ranges?

kkeshavamurthy:
@marosset can you pls address the goss failures, we can get this merged.

jsturtevant:
this should resolve some flakes we are seeing in our upstream tests.  FYI @daschott

daschott:
@jsturtevant thanks for the ping here. I was looking at your comment here:
https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1122654452

This looks good if you are not experiencing port exhaustion. But if you are running out of ports due to heavy traffic load it might also make sense to define exclusion zone 30000 - 34000 and expand the range as needed (e.g. all the way down to 16385). Another alternative is to disable the SNAT, this would also reduce number of ports consumed.

claudiubelu:
> @jsturtevant thanks for the ping here. I was looking back at your concern here: [kubernetes/kubernetes#107114 (comment)](https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1122654452)
> 
> This looks great if you are not experiencing port exhaustion. But if you are running out of ports due to heavy traffic load it might also make sense to define exclusion zone 30000 - 34000 and expand the range as needed (e.g. all the way down to 16385). Another alternative is to disable the SNAT, this would also reduce number of ports consumed. But if you are not experiencing any port exhaustion in the e2e test clusters, then please ignore :)

From what I've seen (https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1208116303), defining exclusion zones does not prevent those ports from being used, but rather raise errors if there is an attempt at using them. This is the reason we're seeing ``An attempt was made to access a socket in a way forbidden by its access permissions`` errors in the Kubernetes CI signal, requests inside the containers are trying to use ports from the excluded ranges:

```
netsh int ipv4 show excludedportrange tcp

Protocol tcp Port Exclusion Ranges

Start Port    End Port
----------    --------
      5985        5985
      5986        5986
     33300       33363
     33364       33427
     33487       33550
     33551       33614
     33615       33678
     33679       33742
     47001       47001

* - Administered port exclusions.
```

IMO, that should have happened in the first place. Updating the port range to 34000 basically excludes ports from being used that are in the ``excludedportrange`` above. The only one that would remain would be ``47001``.

images/capi/packer/goss/goss-command.yaml
```diff
@@ -159,8 +159,8 @@ command:
     exit-status: 0
     exec: netsh int ipv4 show dynamicportrange tcp
     stdout:
-    - "Start Port      : 33000"
-    - "Number of Ports : 32536"
+    - "Start Port      : 34000"
+    - "Number of Ports : 31301"
```
jsturtevant, line 163: I think we can go up to 65535, since this is TCP: https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1339942696

jsturtevant:
Should we add the same port range for UDP (except shorten it for azure due to UDP restriction? https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq#what-protocols-can-i-use-within-vnets)  Maybe we can follow up in a seperate PR since we don't have UDP and would need to only do this port exclusion for Azure....

jsturtevant:
>@jsturtevant thanks for the ping here. I was looking back at your concern here: https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1122654452

>This looks great if you are not experiencing port exhaustion. But if you are running out of ports due to heavy traffic load it might also make sense to define exclusion zone 30000 - 34000 and expand the range as needed (e.g. all the way down to 16385). Another alternative is to disable the SNAT, this would also reduce number of ports consumed. But if you are not experiencing any port exhaustion in the e2e test clusters, then please ignore :)

>From what I've seen (https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1208116303), defining exclusion zones does not prevent those ports from being used, but rather raise errors if there is an attempt at using them. This is the reason we're seeing An attempt was made to access a socket in a way forbidden by its access permissions errors in the Kubernetes CI signal, requests inside the containers are trying to use ports from the excluded ranges:

@daschott Any ideas on why port exclusion doesn't seem to work in our case?  It looks like your suggestion of going all the way down to 16385 is what AKS is doing https://github.com/Azure/AgentBaker/blob/baaf54f3b2342d5f41ae76ecd09b082487898063/staging/cse/windows/configfunc.ps1#L116-L133 and I would expect they would see the same thing @claudiubelu is seeing in https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1208116303

daschott:
@jsturtevant Is the app picking a random port from the ephemeral port range, or is there some mechanism to define a port range in your application?

There is a script aka.ms/networkhealth which can be used to view how many port ranges are available, do your machines have enough ports available? How often does this issue occur?

jsturtevant:
> do your machines have enough ports available? How often does this issue occur?

I wouldn't think there is that many connections running in the test environment (@claudiubelu any thoughts on that?).  It occurs every 30-40 runs.

images/capi/packer/goss/goss-command.yaml
marosset: There were some discussions at https://github.com/kubernetes/kubernetes/issues/107114#issuecomment-1212284604 about only going up to 65,300

marosset: whoops mis-read that.
I can update.

marosset: updated

PR #948, opened 2022-08-12
upgrade packer to 1.8.3

mweibel:

Upgrade packer version to latest 1.8.x version. Among other changes, this allows to specify the `shared_image_gallery.storage_account_type` for azure builds.

### DIFF
--- a/images/capi/hack/ensure-packer.sh
+++ b/images/capi/hack/ensure-packer.sh
@@ -20,7 +20,7 @@ set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-_version="1.7.2"
+_version="1.8.3"
 
 # Change directories to the parent directory of the one in which this
 # script is located.


kkeshavamurthy:
Any reason to not update to the latest `1.8.3` version?

mweibel:
I somehow thought packer 1.8.x requires HCL definitions but after checking I believe this is not yet a requirement. I'll update the PR to use 1.8.3 to see if that works :+1:

kkeshavamurthy:
/lgtm

I wait a couple of days to approve to see if anyone has any thoughts on it.

jsturtevant:
/lgtm 
FYI @CecileRobertMichon @mboersma

CecileRobertMichon:
lgtm

mboersma:
/lgtm

Azure has actually been using Packer v1.8.1 through a [build image](https://github.com/deis/docker-go-dev/blob/main/rootfs/Dockerfile#L15) we use in internal pipelines, so this should be fine. (I'll update that image to v1.8.3 to match.)

Issue #960, opened 2022-09-02
Provide Ubuntu 22.04 support


andyzhangx:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Currently we only supports Provide Ubuntu 20.04

**Describe alternatives you've considered**

**Additional context**

/kind feature

joekr:
Sorry about that @andyzhangx I didn't mean to close the issue I should have used Ref instead of Fix in my PR.

dtzar:
@andyzhangx - this appears to be resolved by #1006

joekr:
@sriramandev should your PR have closed this issue? or simply referred to it. See the above comment were I accidentally closed this larger issue with my PR.

PR #961, opened 2022-09-02
Add Ubuntu 22.04 LTS for Azure

mboersma:

Adds Ubuntu 22.04 LTS build support to Azure.

Which issue(s) this PR fixes:

Refs #960

**Additional context**:

I've tested this and it builds successfully in our DevOps pipeline, but the resulting images need to be tested in a Kubernetes environment.


mboersma:
/hold cancel

I've finally done some local testing with this, by building a Ubuntu 22.04 managed image and provisioning CAPZ clusters. Seems to be working.

mboersma:
@andyzhangx @CecileRobertMichon @jackfrancis please take a look.

jackfrancis (approved):
/lgtm

jsturtevant:
/lgtm
looks like it is building successfully

```
--> vhd-ubuntu-2204: Azure.ResourceManagement.VMImage:
```

This add three more VMs we build for CI. Do we have a plan to remove some at any point?

andyzhangx (approved):
/lgtm
btw, are we also going to support Mariner 2.0 in this repo? thanks.

CecileRobertMichon:
> This add three more VMs we build for CI. Do we have a plan to remove some at any point?

+1, let's remove 18.04?

mboersma:
At one point discussed publishing 18.04 images [until Ubuntu stops](https://ubuntu.com/about/release-cycle) publishing updates, which I think is April 2023.  But I have no objection to stopping that earlier if we all agree.

I plan to publish Ubuntu 22.04 images for the most recent k8s patch releases as soon as this merges, and then continue to publish 22.04 going forward.

Also, the images we choose to publish are distinct from the tools in this repo. That is, we don't have to remove `make build-azure-vhd-ubuntu-1804` and friends in order to stop publishing those reference images.

mboersma:
> are we also going to support Mariner 2.0 in this repo?

That would be excellent. I don't think anyone is working on this currently, though.

jsturtevant:
> Also, the images we choose to publish are distinct from the tools in this repo. That is, we don't have to remove `make build-azure-vhd-ubuntu-1804` and friends in order to stop publishing those reference images.

Agreed, in addition we can run CI on a subset as well.

kkeshavamurthy:
I vote for keeping Ubuntu 18.04 as it's still being used. Since builds are being run in parallel it should not increase the ci time by too much. 
Also we since a few different behaviors between ubuntu 18 & 20. So the ci is useful.

CecileRobertMichon:
My concern is more $ and subscription quota, each CI job creates a VM for each image type. Right now we have:

```
vhd-centos-7: SUCCESS
vhd-ubuntu-1804: SUCCESS
vhd-windows-2019: SUCCESS
vhd-ubuntu-2004: SUCCESS
vhd-windows-2022-containerd: SUCCESS
vhd-ubuntu-2204: SUCCESS
vhd-windows-2019-containerd: SUCCESS
```

and

```
sig-windows-2019: SUCCESS
sig-centos-7-gen2: SUCCESS
sig-ubuntu-2204: SUCCESS
sig-ubuntu-2004-gen2: SUCCESS
sig-windows-2022-containerd: SUCCESS
sig-centos-7: SUCCESS
sig-flatcar: SUCCESS
sig-ubuntu-1804: SUCCESS
sig-flatcar-gen2: SUCCESS
sig-ubuntu-1804-gen2: SUCCESS
sig-ubuntu-2204-gen2: SUCCESS
sig-windows-2019-containerd: SUCCESS
sig-ubuntu-2004: SUCCESS
```

So 20 VMs for each PR run, which is a lot. If we could make CI a subset of these or at least make some of them optional and not run by default that would help. Not saying we should remove support for 18.04 right away, just concerned about CI.

mboersma:
I can remove `ubuntu-1804` from the `azure_targets.sh` file, which will remove it from CI and other "everything" targets like `build-azure-vhds`.

(Could we do this for `windows-2019` as well? It's mostly been replaced by the `-containerd` flavors is my understanding.)

jsturtevant:
> (Could we do this for windows-2019 as well? It's mostly been replaced by the -containerd flavors is my understanding.)

Yes I think so.  I believe we stopped running dockershim tests (@jackfrancis to confirm) in capz. Especially now that 1.22 is the latest release which supports containerd in capz

CecileRobertMichon:
flatcar failed

/retest
/lgtm

PR #968, opened 2022-09-08
Update pyOpenSSL to latest

### LINKED ISSUE #967: Ubuntu image builds fails during iptables-persistence package install
Run azure Ubuntu 1804/2004 SIG/VHD image build. Process fails during installation of iptables-persistent package (providers role - azure.yaml in Ansible)

Image build process to complete successfully 

Looks like the latest version of iptables-persistent package uses updated version of cryptography which needs later versions of openssl. This causes issue with flag 'X509_V_FLAG_CB_ISSUER_CHECK' and causes the installation to fail. Upgrading the pyOpenSSL package to the latest resolves the issue.
Important to note here is that if we try to install iptables-persistent package directly using apt, it sails through and only issue is when trying to install using Ansible.

image

/kind bug

sriramandev:
Signed-off-by: Sriraman S <ssriraman@vmware.com>

This PR updates the pyOpenSSL package to the latest. This will fix the issue of iptables-persistent package installation failing.

**Which issue(s) this PR fixes:** Fixes #967

**Additional context**
Looks like the latest version of iptables-persistent package uses updated version of cryptography which needs later versions of openssl. This causes issue with flag 'X509_V_FLAG_CB_ISSUER_CHECK' and causes the installation to fail. Upgrading the pyOpenSSL package to the latest resolves the issue.
Important to note here is that if we try to install iptables-persistent package directly using apt, it sails through and only issue is when trying to install using Ansible.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/debian.yml
+++ b/images/capi/ansible/roles/providers/tasks/debian.yml
@@ -18,6 +18,12 @@
     executable: pip3
     state: latest
 
+- name: upgrade pyOpenSSL to latest
+  pip:
+    name: pyOpenSSL
+    executable: pip3
+    state: latest
+
 - name: install Azure clients
   pip:
     executable: pip3


mboersma (approved):
/lgtm

images/capi/ansible/roles/providers/tasks/debian.yml
```diff
@@ -18,6 +18,12 @@
     executable: pip3
     state: latest
 
+- name: upgrade pyOpenSSL to latest
+  pip:
+    name: pyOpenSSL
```
kkeshavamurthy, line 23: Is it better to pin this to a stable version rather than installing the latest?

mboersma: That's probably safer, good point. [22.0.0](https://pypi.org/project/pyOpenSSL/) is the latest release.

sriramandev: The reason I didn't pin initially was because other packages are not pinned to specific versions. When they move to later versions, could cause issues with pyOpenSSL being older version and will need to explicitly make sure we keep updating the version here when new ones are available.

kkeshavamurthy: Seems like pyopenssl does not have have releases too often. So might be ok for now. We can revisit if we start having issues.

mboersma: I think that's reasonable. We should notice it quickly if a newer version of pyOpenSSL breaks something.

kkeshavamurthy:
/approve
Thanks for the fix @sriramandev

PR #972, opened 2022-09-09
Expose AWS_SESSION_TOKEN as a user variable

ThatsMrTalbot:
When uses STS credentials (in our case provided by vault) you also need to specify the session token. This PR exposes that as a user variable.

What this PR does / why we need it: Currently AWS AMIs cannot be built using STS based AWS credentials

**Additional context**

### DIFF
--- a/images/capi/packer/ami/packer.json
+++ b/images/capi/packer/ami/packer.json
@@ -59,6 +59,7 @@
         "source_ami": "{{user `source_ami`}}"
       },
       "temporary_security_group_source_cidrs": "{{ user `temporary_security_group_source_cidrs` }}",
+      "token": "{{ user `aws_session_token` }}",
       "type": "amazon-ebs",
       "user_data": "{{ user `user_data` }}",
       "vpc_id": "{{ user `vpc_id` }}"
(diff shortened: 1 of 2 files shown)


images/capi/packer/ami/packer.json
```diff
@@ -154,6 +155,7 @@
     "aws_region": "us-east-1",
     "aws_secret_key": "",
     "aws_security_group_ids": "",
+    "aws_session_token": "",
```
kkeshavamurthy, line 158: Can you pls add this to `packer-windows.json` as well.

ThatsMrTalbot: Sure, thats done

kkeshavamurthy: Thanks. Do you think it would be better if the default value reads the `AWS_SESSION_TOKEN` env var. User can override it of course.

ThatsMrTalbot: That would be good, but non of the other variables currently do this. AWS_REGION and AWS_ACCESS_KEY_ID for example. Would it be better to keep them all consistent?

ThatsMrTalbot: @kkeshavamurthy Do you have any thoughts on the above?

PR #980, opened 2022-09-20
Added qemu guest agent

isaacdorfman:
As discussed in:
https://github.com/kubernetes-sigs/image-builder/pull/847

I moved the adding of qemu-guest-agent into the image into a separate PR.

This branch is forked from the one in the previous PR so should only be merged after it.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -295,6 +295,7 @@ OCI_BUILD_NAMES			   ?= oci-ubuntu-1804 oci-ubuntu-2004 oci-oracle-linux-8
 DO_BUILD_NAMES 			?=	do-centos-7 do-ubuntu-1804 do-ubuntu-2004
 
 QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804 qemu-ubuntu-2004 qemu-centos-7 qemu-ubuntu-2004-efi qemu-rhel-8 qemu-rockylinux-8 qemu-flatcar
+QEMU_KUBEVIRT_BUILD_NAMES	:= $(addprefix kubevirt-,$(QEMU_BUILD_NAMES))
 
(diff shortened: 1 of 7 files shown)


isaacdorfman:
@davidvossel @seanschneeweiss FYI

seanschneeweiss:
I'll copy/paste [the same comment](https://github.com/kubernetes-sigs/image-builder/pull/847#discussion_r974145386) to this issue to have it within this disussion.


---
We asked Canonical on whether to use the qemu-guest-agent or not. There is no clear recommendation. It is a decision based on every individual use case.

> Qemu Guest Agent [1] is used by some of the Canonical customers and in general it is supported by the KVM (and OpenStack installations with virtualization based on KVM). Unfortunately there is no official Canonical documentation that provides pros and cons to make our client decision easier.
> Following are some items to take into consideration when using the guest agent:
> 1. it allows for file system access from the host to a VM
> 2. it allows for changing user's password on a VM, including root password
> 3. it allows for read/write file access within a VM
> 4. it provides another layer of connectivity between a VM and host and in some particular use cases this might be beneficial, ie. quiescence filesystem prior to snapshot in order to have consistent data
> 
> However it should be conscious, justified decision of a cloud user/operator to start using the agent as it exposed VM data to some extent.
> 
> [1] https://qemu-project.gitlab.io/qemu/interop/qemu-ga.html

For us personally, we won't be using the qemu-guest-agent. I'd recommend opt-in rather then opt-out.

kkeshavamurthy:
Seems like none of the providers (vsphere, aws, azure, gcp, openstack, etc) needs the `qemu-guest-agent`.

kkeshavamurthy:
I'm closing this PR since there has not been any updates in a long time. Please re-open when needed.

PR #984, opened 2022-09-27
debian kernel params console

qinqon:
Add console to kernel params

**Additional context**

Depends on:
- #890


images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -58,6 +58,7 @@ common_debs:
 - python3-netifaces
 - python3-pip
 - socat
+- qemu-guest-agent
```
usrbinkat, line 61: Addition of qemu-guest-agent to default packages I believe belongs to https://github.com/kubernetes-sigs/image-builder/pull/980

Should this line be removed from #984 ?

qinqon: This PR is rebases on top of the qemu guest agent, there is no dependency technically but we use both features, we have to review just the last commit.

qinqon:
Removed the dependency on guest-agent since this PR does not dpeend on that.

images/capi/ansible/roles/node/defaults/main.yml
qinqon: @usrbinkat I have remove the rebase since it does not really depend on that, can you rereview ?

seanschneeweiss:
Will https://github.com/kubernetes-sigs/image-builder/pull/998 fix the same?

hh:
/cc @CecileRobertMichon
Pinging @CecileRobertMichon as a cluster-api-maintainers human for `/approve` , feel free to kick it to another approver.

I chose from from https://github.com/kubernetes-sigs/image-builder/blob/master/OWNERS_ALIASES#L32-L36

hh:
@seanschneeweiss I'm fine with either approach.

kkeshavamurthy:
Do we still need this PR with https://github.com/kubernetes-sigs/image-builder/pull/998 merged?

PR #986, opened 2022-09-28
Add amazon_ssm_agent_rpm var to configure the RPM download URL

ThatsMrTalbot:

When building the AMI in an environment without internet access most options can be configured to pull from alternative repositories, however the amazon ssm agent is hard-coded in Ansible. This PR allows the URL to be configured. 

**Additional context**


images/capi/packer/config/common.json
```diff
@@ -1,4 +1,5 @@
 {
+  "amazon_ssm_agent_rpm": "https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm",
```
kkeshavamurthy, line 2: I suggest adding this to the ami/packer.json var section instead of a global common vars section.

ThatsMrTalbot: Sure, I was originally going to do that, but was not sure of the behaviour of the `{{ user "amazon_ssm_agent_rpm" }}` block in ansible-args.json when the variable is not defined (ie for other cloud providers)

PR #990, opened 2022-10-04
Update Azure pub tool to v0.3.3

mboersma:

Updates the version of the [`pub`](https://github.com/devigned/pub/releases/tag/v0.3.3) tool used to interact with Azure Marketplace. Also parameterizes the version with an environment variable so we won't need to open PRs like this in the future.

Which issue(s) this PR fixes: 

N/A

**Additional context**

I've tested this branch in an Azure pipeline job.

### DIFF
--- a/images/capi/packer/azure/scripts/new-disk-version.sh
+++ b/images/capi/packer/azure/scripts/new-disk-version.sh
@@ -6,6 +6,7 @@ echo "PWD: $PWD"
 
 OS=${OS:-"Ubuntu"}
 OS_VERSION=${OS_VERSION:-"18.04"}
+PUB_VERSION=${PUB_VERSION:-"v0.3.3"}
 
 required_env_vars=(
     "AZURE_CLIENT_ID"
@@ -13,6 +14,7 @@ required_env_vars=(
     "AZURE_TENANT_ID"
     "OS"
     "OS_VERSION"
+    "PUB_VERSION"
 )
 
 for v in "${required_env_vars[@]}"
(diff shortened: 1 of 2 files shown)


mboersma:
cc: @marosset @devigned @CecileRobertMichon @jackfrancis

images/capi/packer/azure/scripts/new-disk-version.sh
```diff
@@ -40,7 +42,7 @@ do
 done
 
 echo "Getting pub..."
```
marosset, line 44: should we echo the version here too?

marosset: (nit)

marosset:
I have a small nit but otherwise LGTM

images/capi/packer/azure/scripts/new-disk-version.sh
mboersma: We maybe should, but the `curl` command is preceded by `set -x` so it effectively logs the version:

```
Getting pub...
+ echo
+ echo 'Getting pub...'
+ set -x
+ curl -fsSL https://github.com/devigned/pub/releases/download/v0.3.3/pub_v0.3.3_linux_amd64.tar.gz -o pub
```

marosset: ahh, good enough

mboersma: Also--thanks for fixing `pub`! It was a bit dusty, nice to have it updated and hopefully working smoothly.

PR #993, opened 2022-10-11
Bump dependency versions

AverageMarcus:
Bumps `goss` and `Ansible` to latest versions to allow for building Ubuntu 22.04 images (with updated SSH server)


AverageMarcus:
Turns out this change isn't actually required for the Ubuntu 22.04 builds (just the latest build from the Dockerfile already in this repo publishing as a new version)

PR #1000, opened 2022-10-18
Ensure Azure storage accounts are private

CecileRobertMichon:
What this PR does / why we need it: By default, Azure storage accounts are public. 

Microsoft recommends to disallow public access to a storage account unless the scenario requires it. Disallowing public access helps to prevent data breaches caused by undesired anonymous access. More info: https://azure.microsoft.com/en-gb/updates/choose-to-allow-or-disallow-blob-public-access-on-azure-storage-accounts/

**Additional context**


CecileRobertMichon:
/assign @mboersma 

I don't believe any image-builder scenarios require blob public access for storage accounts and publishing should be fine since it uses a shared access signature to reach the containers but please let me know if you know otherwise.

CecileRobertMichon:
/hold 

@mboersma @marosset do we want to wait until we publish the latest 1.25.3 image first in case this interferes?

mboersma (commented):
/lgtm

We don't yet have an ETA on when that build will happen, so I don't think this needs to wait for it.

mboersma:
I'm not sure yet what's going on with the CI failure. It looks like some housekeeping error updating or invoking `pip` that probably has nothing to do with this PR. I'm investigating.

PR #1005, opened 2022-10-20
Add pipeline test job for Azure VHDs

### LINKED ISSUE #992: Run e2e/smoke tests against new VHDs before publishing them
Currently, the Azure image publishing pipelines doesn't run any sort of validation to ensure the image is not broken.

We want to add a stage to the image publishing pipeline that runs a "smoke test" to ensure the underlying image can be used to provision a CAPZ cluster successfully.

**Additional context**

/kind feature

willie-yao:
Currently, the Azure image publishing pipelines doesn't run any sort of validation to ensure the image is working. This PR adds a step in the image publishing pipeline after building the VHD that: 
- Applies the managed image to a cluster template.
- Builds a CAPZ cluster from the template using the [az capi cli extension](https://github.com/Azure/azure-capi-cli-extension).
- Ensures the nodes are running and ready. 

Which issue(s) this PR fixes: Fixes #992 

**Additional context**
cc @mboersma @CecileRobertMichon

### DIFF
--- a/images/capi/packer/azure/.pipelines/build-vhd.yaml
+++ b/images/capi/packer/azure/.pipelines/build-vhd.yaml
@@ -22,8 +22,15 @@ jobs:
   - script: |
       set -o pipefail
       make deps-azure
-      os=$(echo "$OS" | tr '[:upper:]' '[:lower:]') 
-      version=$(echo "$OS_VERSION" | tr '[:upper:]' '[:lower:]' | tr -d .)
+      os=$(echo "${OS}" | tr '[:upper:]' '[:lower:]')
+      version=$(echo "${OS_VERSION}" | tr '[:upper:]' '[:lower:]' | tr -d .)
+      export RESOURCE_GROUP_NAME="cluster-api-images"
+
+      # timestamp is in RFC-3339 format to match kubetest
+      export TIMESTAMP="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+      export JOB_NAME="${JOB_NAME:-"image-builder-vhd"}"
+      export TAGS="creationTimestamp=${TIMESTAMP} jobName=${JOB_NAME}"
+      printf "${TAGS}" | tee packer/azure/tags.out
       make build-azure-vhd-$os-$version | tee packer/azure/packer.out
     displayName: Building VHD
     workingDirectory: '$(system.defaultWorkingDirectory)/images/capi'
@@ -39,5 +46,5 @@ jobs:
   - template: delete-storage-account.yaml
   - script: |
       chown -R $USER:$USER .
-    displayName: cleanup - chown all files in work directory 
+    displayName: cleanup - chown all files in work directory
     condition: always()
--- a/images/capi/packer/azure/.pipelines/generate-sas.yaml
+++ b/images/capi/packer/azure/.pipelines/generate-sas.yaml
@@ -4,7 +4,10 @@ steps:
       RESOURCE_GROUP_NAME=$(jq -r '.builds[-1].custom_data.resource_group_name' manifest.json | cut -d ":" -f2)
       STORAGE_ACCOUNT_NAME=$(jq -r '.builds[-1].custom_data.storage_account_name' manifest.json | cut -d ":" -f2)
       OS_DISK_URI=$(cat packer/azure/packer.out | grep "OSDiskUri:" -m 1 | cut -d " " -f 2)
+      printf "${STORAGE_ACCOUNT_NAME}" | tee packer/azure/storage-account-name.out
+      printf "${OS_DISK_URI}" | tee packer/azure/vhd-base-url.out
       printf "${OS_DISK_URI}?" | tee packer/azure/vhd-url.out
+      printf "${RESOURCE_GROUP_NAME}" | tee packer/azure/resource-group-name.out
       az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
       az account set -s ${AZURE_SUBSCRIPTION_ID}
       ACCOUNT_KEY=$(az storage account keys list -g ${RESOURCE_GROUP_NAME} --subscription ${AZURE_SUBSCRIPTION_ID} --account-name ${STORAGE_ACCOUNT_NAME} --query '[0].value')
--- a/images/capi/packer/azure/.pipelines/stages.yaml
+++ b/images/capi/packer/azure/.pipelines/stages.yaml
@@ -28,6 +28,16 @@ stages:
       AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET_VHD)
       AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID_VHD)
 
+  - stage: test
+    condition: and(succeeded(), eq(variables.CLEANUP, 'False'))
+    jobs:
+    - template: test-vhd.yaml
+    variables:
+      AZURE_TENANT_ID: $(AZURE_TENANT_ID_VHD)
+      AZURE_CLIENT_ID: $(AZURE_CLIENT_ID_VHD)
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET_VHD)
+      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID_VHD)
+
   - stage: sku
     condition: and(succeeded(), eq(variables.CLEANUP, 'False'))
     jobs:
--- a/images/capi/packer/azure/.pipelines/test-vhd.yaml
+++ b/images/capi/packer/azure/.pipelines/test-vhd.yaml
@@ -1,0 +1,143 @@
+# Required pipeline variables:
+# - BUILD_POOL - Azure DevOps build pool to use
+# - CONTAINER_IMAGE - Dev container image URL to use. Should have Azure CLI, Packer, and Ansible.
+# - AZ_CAPI_EXTENSION_URL - URL to the Azure CAPI extension build.
+# - AZURE_TENANT_ID - tenant ID
+# - AZURE_CLIENT_ID - Service principal ID
+# - AZURE_CLIENT_SECRET - Service principal secret
+# - AZURE_SUBSCRIPTION_ID - Subscription ID used by the pipeline
+# - KUBERNETES_VERSION - version of Kubernetes to build the image with, e.g. `1.16.2`
+# - OS - target of build e.g. `Ubuntu/Windows`
+# - OS_VERSION - target of build e.g. `18.04/2004/2019`
+
+jobs:
+- job: test_vhd
+  container: $[ variables['CONTAINER_IMAGE'] ]
+  timeoutInMinutes: 120
+  strategy:
+    maxParallel: 0
+  pool:
+    name: $(BUILD_POOL)
+  steps:
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      source: current
+      artifact: publishing-info
+      path: $(system.defaultWorkingDirectory)/images/capi/packer/azure/vhd/
+  - script: |
+      set -x
+      set -e -o pipefail
+
+      VHD_RESOURCE_ID=$(jq -r .vhd_base_url $(system.defaultWorkingDirectory)/images/capi/packer/azure/vhd/vhd-publishing-info.json)
+      STORAGE_ACCOUNT_NAME=$(jq -r .storage_account_name $(system.defaultWorkingDirectory)/images/capi/packer/azure/vhd/vhd-publishing-info.json)
+      TAGS=$(jq -r .tags $(system.defaultWorkingDirectory)/images/capi/packer/azure/vhd/vhd-publishing-info.json)
+    
+      echo "##vso[task.setvariable variable=VHD_RESOURCE_ID]$VHD_RESOURCE_ID"
+      echo "##vso[task.setvariable variable=STORAGE_ACCOUNT_NAME]$STORAGE_ACCOUNT_NAME"
+      echo "##vso[task.setvariable variable=TAGS;]$TAGS"
+    displayName: Import variables from build vhd job
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - script: |
+      set -x
+      set -e -o pipefail
+
+      RANDOM=$(bash -c 'echo $RANDOM')
+      RESOURCE_GROUP="capi-testvmimage-${RANDOM}"
+      echo "${RESOURCE_GROUP}" is the group
+
+      # Azure CLI login
+      az login -u $AZURE_CLIENT_ID -p $AZURE_CLIENT_SECRET --service-principal --tenant $AZURE_TENANT_ID
+
+      # Find the VHD blob location from its storage account
+      AZURE_LOCATION=$(az storage account show --name "${STORAGE_ACCOUNT_NAME}" --query '[location]' -o tsv)
+
+      # Create the resource group
+      az group create --name "${RESOURCE_GROUP}" --location "${AZURE_LOCATION}" --tags "${TAGS}"
+
+      # Create a managed image from the VHD blob
+      OS_TYPE="Linux"
+      if [ "$OS" == "Windows" ]; then
+        OS_TYPE="Windows"
+      fi
+      az image create -n testvmimage -g "${RESOURCE_GROUP}" --os-type "${OS_TYPE}" --source "${VHD_RESOURCE_ID}"
+
+      # Pass the managed image resource ID on to the next step
+      IMAGE_ID=$(az image show -g "${RESOURCE_GROUP}" -n testvmimage --query '[id]' --output tsv)
+      echo "##vso[task.setvariable variable=RESOURCE_GROUP;]$RESOURCE_GROUP"
+      echo "##vso[task.setvariable variable=MANAGED_IMAGE_ID;]$IMAGE_ID"
+      echo "##vso[task.setvariable variable=AZURE_LOCATION;]$AZURE_LOCATION"
+    displayName: promote VHD blob to managed image
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - template: k8s-config.yaml
+  - script: |
+      set -x
+      set -e -o pipefail
+
+      export PATH=${PATH}:.local/bin
+      ./packer/azure/scripts/ensure-kustomize.sh
+
+      # Generate cluster template with kustomize
+      if [ "$OS" == "Windows" ]; then
+        kustomize build --load-restrictor LoadRestrictionsNone $(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/windows/ > $(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
+      else
+        kustomize build --load-restrictor LoadRestrictionsNone $(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/linux/ > $(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
+      fi
+      TEST_TEMPLATE=$(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
+      echo "##vso[task.setvariable variable=TEST_TEMPLATE;]$TEST_TEMPLATE"
+    displayName: generate cluster template
+    workingDirectory: '$(system.defaultWorkingDirectory)/images/capi'
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - script: |
+      set -x
+      set -e -o pipefail
+
+      os=$(echo "$OS" | tr '[:upper:]' '[:lower:]')
+
+      # Set up the Azure CLI Cluster API extension
+      # https://github.com/Azure/azure-capi-cli-extension/releases/download/az-capi-nightly/capi-0.0.vnext-py2.py3-none-any.whl
+      az extension add --yes --source "${AZ_CAPI_EXTENSION_URL}"
+
+      # Install required binaries
+      mkdir ~/test-binaries
+      export PATH=${PATH}:~/test-binaries
+      az capi install -a -ip ~/test-binaries
+
+      echo "##vso[task.setvariable variable=PATH;]$PATH"
+    displayName: Install and configure az capi extension
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - script: |
+      params=()
+      if [ "$OS" == "Windows" ]; then
+        params+=(--windows)
+      fi
+      
+      # Create a cluster
+      az capi create \
+        --yes \
+        --debug \
+        --name testvm \
+        --kubernetes-version="${KUBERNETES_VERSION}" \
+        --location="${AZURE_LOCATION}" \
+        --resource-group="${RESOURCE_GROUP}" \
+        --management-cluster-resource-group-name="${RESOURCE_GROUP}" \
+        --control-plane-machine-count=1 \
+        --node-machine-count=1 \
+        --template="${TEST_TEMPLATE}" \
+        --tags="${TAGS}" \
+        --wait-for-nodes=2 \
+        "${params[@]}"
+    displayName: Create a cluster
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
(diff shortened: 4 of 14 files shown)


images/capi/packer/azure/.pipelines/build-vhd.yaml
```diff
@@ -22,8 +22,16 @@ jobs:
   - script: |
       set -o pipefail
       make deps-azure
-      os=$(echo "$OS" | tr '[:upper:]' '[:lower:]') 
+      os=$(echo "$OS" | tr '[:upper:]' '[:lower:]')
       version=$(echo "$OS_VERSION" | tr '[:upper:]' '[:lower:]' | tr -d .)
+      get_random_region() {
+          local REGIONS=("eastus" "eastus2" "southcentralus" "westus2" "westeurope")
```
CecileRobertMichon, line 28: Did you verify that all regions in this list have sufficient quota for the VM size we are using to build test clusters in the Prow test sub? CAPZ uses a different list https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/hack/util.sh#L44 which is curated to have all the regions that have quota for those VM sizes + support for availability zones

CecileRobertMichon, line 28: Did you verify that all regions in this list have sufficient quota for the VM size we are using to build test clusters in the Prow test sub? CAPZ uses a different list https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/hack/util.sh#L44 which is curated to have all the regions that have quota for those VM sizes + support for availability zones

images/capi/packer/azure/.pipelines/build-vhd.yaml
```diff
@@ -22,8 +22,16 @@ jobs:
...
+          echo "${REGIONS[${RANDOM} % ${#REGIONS[@]}]}"
+      }
+      export AZURE_LOCATION=$(get_random_region)
+      RANDOM=$(bash -c 'echo $RANDOM')
+      export RESOURCE_GROUP_NAME="cluster-api-images-${RANDOM}"
```
CecileRobertMichon, line 33: Is the choice to use a separate RG per location for optimization purposes? Resources in a resource group don't need to be in the same location, it's only the metadata of that resource group that lives in that location https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/overview#resource-groups

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+      RESOURCE_GROUP="capi-testvmimage-${RANDOM}"
+
+      # Azure CLI login
+      az login -u $AZURE_CLIENT_ID -p $AZURE_CLIENT_SECRET --service-principal --tenant $AZURE_TENANT_ID
+
+      # Find the VHD blob location from its storage account
+      AZURE_LOCATION=$(az storage account show --name "${STORAGE_ACCOUNT_NAME}" --query '[location]' -o tsv)
+
+      # Create the resource group
+      az group create --name "${RESOURCE_GROUP}" --location "${AZURE_LOCATION}" --tags "DO-NOT-DELETE=CAPZ\ Reference\ Image"
```
CecileRobertMichon, line 42: Do we really want to add a "DO-NOT-DELETE" tag on the test RG? Isn't it meant to be ephemeral? 

If you don't want it to be deleted mid-job by the resource cleanup job, make sure you add the creationTimestamp and jobName tags instead https://github.com/willie-yao/image-builder/blob/azure-test-vhd/images/capi/scripts/ci-azure-e2e.sh#L68

images/capi/packer/azure/.pipelines/generate-sas.yaml
```diff
@@ -4,6 +4,8 @@ steps:
       RESOURCE_GROUP_NAME=$(jq -r '.builds[-1].custom_data.resource_group_name' manifest.json | cut -d ":" -f2)
       STORAGE_ACCOUNT_NAME=$(jq -r '.builds[-1].custom_data.storage_account_name' manifest.json | cut -d ":" -f2)
       OS_DISK_URI=$(cat packer/azure/packer.out | grep "OSDiskUri:" -m 1 | cut -d " " -f 2)
+      printf "${STORAGE_ACCOUNT_NAME}" | tee packer/azure/storage-account-name.out
+      printf "${OS_DISK_URI}" | tee packer/azure/vhd-base-url.out
```
CecileRobertMichon, line 8: was this a temporary debug statement meant to be cleaned up?

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+      az image create -n testvmimage -g "${RESOURCE_GROUP}" --os-type "${OS_TYPE}" --source "${VHD_RESOURCE_ID}"
+
+      # Pass the managed image resource ID on to the next step
+      IMAGE_ID=$(az image show -g "${RESOURCE_GROUP}" -n testvmimage --query '[id]' --output tsv)
+      echo "##vso[task.setvariable variable=MANAGED_IMAGE_ID;]$IMAGE_ID"
+      echo "##vso[task.setvariable variable=RESOURCE_GROUP;]$RESOURCE_GROUP"
+      echo "##vso[task.setvariable variable=AZURE_LOCATION;]$AZURE_LOCATION"
+    displayName: promote VHD blob to managed image
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
```
CecileRobertMichon, line 58: side note: we should look into using managed identity instead for this whole pipeline

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+    maxParallel: 0
+  pool:
+    name: $(BUILD_POOL)
+  steps:
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      source: current
+      artifact: publishing-info
+      path: $(system.defaultWorkingDirectory)/images/capi/packer/azure/vhd/
+  - script: |
```
CecileRobertMichon, line 26: instead of having one big script here, have you considered breaking it down into smaller steps?

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+        --debug \
+        --name testvm \
+        --kubernetes-version=$KUBERNETES_VERSION \
+        --location=$AZURE_LOCATION \
+        --resource-group=$RESOURCE_GROUP \
+        --management-cluster-resource-group-name=$RESOURCE_GROUP \
+        --control-plane-machine-count=1 \
+        --node-machine-count=1 \
+        --template=$TEST_TEMPLATE \
+        --tags="DO-NOT-DELETE=CAPZ\ Reference\ Image" \
```
CecileRobertMichon, line 97: same question here about the RG tags

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+      else
+        TEST_TEMPLATE=$(system.defaultWorkingDirectory)/images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
+      fi
+
+      # Create a cluster
+      az capi create \
+        --yes \
+        --debug \
+        --name testvm \
+        --kubernetes-version=$KUBERNETES_VERSION \
```
CecileRobertMichon, line 90:
```suggestion
        --kubernetes-version="${KUBERNETES_VERSION}" \
```

same thing for all other variables

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+        --location=$AZURE_LOCATION \
+        --resource-group=$RESOURCE_GROUP \
+        --management-cluster-resource-group-name=$RESOURCE_GROUP \
+        --control-plane-machine-count=1 \
+        --node-machine-count=1 \
+        --template=$TEST_TEMPLATE \
+        --tags="DO-NOT-DELETE=CAPZ\ Reference\ Image" \
+        "${params[@]}"
+      
+      # Wait for the workload nodes to be ready
```
CecileRobertMichon, line 100: Have we considered if there is a way to build this functionality into az capi? this is a very common scenario for testing...

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,124 @@
...
+      fi
+    displayName: test the managed image
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - script: |
+      set -x
+      set -e -o pipefail
+      
+      # Clean up the test resource group
+      az group delete -n "${RESOURCE_GROUP}" --yes --no-wait
```
CecileRobertMichon, line 122: is the RESOURCE_GROUP variable carried over from the previous step?

images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
```diff
@@ -0,0 +1,231 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
```
CecileRobertMichon, line 2: I'm a little concerned about having to maintain this static yaml in the image-builder repo over time.

I wonder if we can do anything to get rid of the boilerplate and only maintain the custom image part of it in here.

Some ideas:

1) Add custom image as a parameter to az capi
2) use kustomize with a source template from the capz repo to auto-generate the template with just the patch we need for the image (eg. https://github.com/microsoft/azure_arc/blob/main/azure_arc_k8s_jumpstart/cluster_api/capi_azure/artifacts/capz_kustomize/kustomization.yaml#L4) 
3) use ClusterClass and a json patch for the image - this one is a bit of a stretch since we don't publish a clusterclass template in CAPZ right now and there is no integration with az capi at the moment but maybe something to drive towards in the future (the template would become much simpler and look something like https://github.com/kubernetes-sigs/cluster-api/blob/main/test/infrastructure/docker/templates/cluster-template-development.yaml) 

Any other ideas for improving this?

images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
```diff
@@ -0,0 +1,231 @@
...
+metadata:
+  labels:
+    cni: calico
```
CecileRobertMichon, line 5: @mboersma is az capi using the cni label?

CecileRobertMichon (commented):
Sorry for the late review @willie-yao @mboersma, finally got around to looking at this post kubecon!

images/capi/packer/azure/.pipelines/test-vhd.yaml
CecileRobertMichon: https://github.com/kubernetes-sigs/image-builder/issues/1011

images/capi/packer/azure/.pipelines/build-vhd.yaml
willie-yao: We did not. I think I copied this from the [`ci-azure-e2e.sh`](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/scripts/ci-azure-e2e.sh#L57-L60) script. I'll go ahead and replace this with the CAPZ list

images/capi/packer/azure/.pipelines/build-vhd.yaml
willie-yao: I think we used a separate resource group the store the VHD vs the cluster itself. I'll have to double check with @mboersma for why we did this, but we also wanted to generate a random resource group name in case two pipelines are running at once.

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: Yes that was the goal. We noticed that the cleanup job was deleting our resources mid-job. I wasn't aware that these tags would prevent deletion, although we do cleanup the resource group after. It looks like this is the preferred way to go about it so we'll work on making that change!

images/capi/packer/azure/.pipelines/generate-sas.yaml
willie-yao: These statements are meant to write the storage account name and os disk uri to a file so we can retrieve them in [`vhd-publishing-info.yaml`](https://github.com/willie-yao/image-builder/blob/azure-test-vhd/images/capi/packer/azure/.pipelines/vhd-publishing-info.yaml)

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: I think we kept it in a bigger step since there are a lot of variables being used, and we have to manually export them to the next step each time like on [line 51](https://github.com/willie-yao/image-builder/blob/azure-test-vhd/images/capi/packer/azure/.pipelines/test-vhd.yaml#L51-L55). Would it be preferred to split into more steps and export the variables, or keep it as is?

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: This is being worked on and near completion in this [PR](https://github.com/Azure/azure-capi-cli-extension/pull/159). Once that is merged we can replace this block with that command.

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: Oh hmm, maybe we're actually misunderstanding how variables are passed between scripts. I'll double-check the logs and make changes accordingly.

images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
mboersma: Willie and I like option 2. which seems like it would be generally useful for `az capi` users. But currently with the extension, you can either use the built-in templating, or specify a custom `--template`. It applies `envsubst` to that template, but `kustomize` isn't part of that processing (yet).

Maybe this could be a separate work item? I agree that copy-and-paste could lead to drift here, but 2) might be more work than we want to tackle in this iteration.

images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
mboersma: If you're asking whether `az capi` uses this cni label with its built-in template, then yes it also does. Otherwise, I'm not understanding your question. :-|

images/capi/packer/azure/.pipelines/build-vhd.yaml
willie-yao: Update: we combined the whole pipeline to use the same RG

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: We broke it down into five steps: 

1. Import variables from build vhd job
2. promote VHD blob to managed image
3. Install and configure az capi extension
4. Create a cluster
5. Wait for workload nodes to be ready

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,139 @@
+# Required pipeline variables:
+# - BUILD_POOL - Azure DevOps build pool to use
+# - CONTAINER_IMAGE - Dev container image URL to use. Should have Azure CLI, Packer and Ansible.
```
marosset, line 3:
```suggestion
# - CONTAINER_IMAGE - Dev container image URL to use. Should have Azure CLI, Packer, and Ansible.
```

images/capi/packer/azure/.pipelines/test-vhd.yaml
```diff
@@ -0,0 +1,139 @@
...
+    env:
+      AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET)
+  - script: |
+      set -x
+      set -e -o pipefail
+
+      os=$(echo "$OS" | tr '[:upper:]' '[:lower:]')
+
+      # Set up the Azure CLI Cluster API extension
+      az extension add --yes --source https://github.com/Azure/azure-capi-cli-extension/releases/download/az-capi-nightly/capi-0.0.vnext-py2.py3-none-any.whl
```
marosset, line 75: Should we make the version of the az capi extension a pipeline variable?

images/capi/packer/azure/scripts/test-templates/cluster-template.yaml
CecileRobertMichon: we can do 2) right here in this repo, and then use the `--template` arg with the generated template just like we're doing now

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: @CecileRobertMichon We noticed that by combining the resource groups for the build and test step into one, we want the rg to be persisted indefinitely as the VHD is now stored in it. Would it be okay design-wise to have a separate resource group for each test that doesn't get deleted?

We also thought about using the default `cluster-api-images` resource group, but we would be restricted to `southcentralus` because the storage account location is tied to the resource group.

CecileRobertMichon: The main concern I have with using the same RG to host test resources and long-term persisted resources is that it makes it possible for us to leak resources: let's say the pipeline gets interrupted midway, and the cleanup step doesn't run. Those test cluster VMs will be running in the sub indefinitely. We have a periodic cleanup job for exactly this reason, but it's only able to delete entire resource groups, not individual resources.  Are we able to use a single RG for all images (maybe with a different storage account for each location) and one RG for each ephemeral test cluster that can be deleted (with appropriate tags).

willie-yao: That's definitely a better idea. I was concerned about storing all of the VHDs in one location but I didn't realize I can have a separate storage account for each location. Thanks for the suggestion!

images/capi/Makefile
```diff
@@ -67,6 +67,7 @@ deps-azure:
 	hack/ensure-jq.sh
 	hack/ensure-azure-cli.sh
 	hack/ensure-goss.sh
+	hack/ensure-kustomize.sh
```
kkeshavamurthy, line 70: is `kustomize` needed for azure image-buildig process?

willie-yao: Yes, we're using it to generate a cluster template to build our images on. It uses a base template from CAPZ. It isn't necessarily required since we can copy the exact template we want into this repo. However, if we ever decide to update the templates on CAPZ, we will have to change it here as well.

CecileRobertMichon: > is kustomize needed for azure image-buildig process or just the CI?

I believe @kkeshavamurthy is asking if customers will also need this to build images. The answer is no, this is only needed for our publishing pipeline image validation to build a cluster with the image before publishing it to make sure we don't publish a broken image (unless said customer uses the publishing pipeline template here that includes the test step).

kkeshavamurthy: Yes. Thanks @CecileRobertMichon & @willie-yao. 
Since it's not needed for image building, would it be possible to add this dependency in another location instead of the top level Makefile or make it opt-in? I'm not familiar with the Azure CI/Pipeline processes to suggest alternatives.

willie-yao: Sure! That sounds like a better idea. I'll move it somewhere within the Azure directory.

images/capi/packer/azure/.pipelines/test-vhd.yaml
willie-yao: I think this would be a good idea. Thoughts? @mboersma

mboersma:
One change we should make is to add a kustomize patch to the Windows template that scales the `MachineDeployment` for Linux to `replicas: 0`. That would avoid creating an unnecessary VM.

The harder problem is the chicken-and-egg issue with having only one `KUBERNETES_VERSION` variable. If we're only testing Windows, the k8s version really only applies to the Windows node(s). The control plane nodes need to use a released Linux version instead somehow...

marosset:
> One change we should make is to add a kustomize patch to the Windows template that scales the `MachineDeployment` for Linux to `replicas: 0`. That would avoid creating an unnecessary VM.
> 
> The harder problem is the chicken-and-egg issue with having only one `KUBERNETES_VERSION` variable. If we're only testing Windows, the k8s version really only applies to the Windows node(s). The control plane nodes need to use a released Linux version instead somehow...

Can you do this with env vars?

> export WORKER_MACHINE_COUNT=0
export WINDOWS_WORKER_MACHINE_COUNT=1

mboersma:
> Can you do this with env vars?
>
> export WORKER_MACHINE_COUNT=0
> export WINDOWS_WORKER_MACHINE_COUNT=1

We're re-using [this template](https://github.com/kubernetes-sigs/cluster-api-provider-azure/releases/download/v1.4.5/cluster-template-windows.yaml), which unfortunately reuses just `WORKER_MACHINE_COUNT` for both the Windows and Linux node pools. We could change that in our kustomize patches, but for this particular test pipeline we might as well just hard-code it to 1.

willie-yao:
@CecileRobertMichon Gentle ping. This should be ready for another round of reviews!

images/capi/packer/azure/scripts/test-templates/windows/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/releases/download/v1.4.5/cluster-template-windows.yaml
```
CecileRobertMichon, line 4: why is this using a template from release v1.4?

images/capi/packer/azure/scripts/test-templates/linux/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/kubernetes-sigs/cluster-api-provider-azure/releases/download/v1.4.5/cluster-template.yaml
```
CecileRobertMichon, line 4: same question here

images/capi/packer/azure/.pipelines/stages.yaml
```diff
@@ -28,6 +28,16 @@ stages:
       AZURE_CLIENT_SECRET: $(AZURE_CLIENT_SECRET_VHD)
       AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID_VHD)
 
+  - stage: test
+    condition: and(succeeded(), eq(variables.CLEANUP, 'False'))
```
CecileRobertMichon, line 32: I wonder if we want to run this regardless of whether this a "dry run" (and cleanup is true)? @mboersma wdyt?

images/capi/packer/azure/scripts/test-templates/windows/kustomization.yaml
willie-yao: Whoops, gonna bump this to v1.6.0

images/capi/packer/azure/.pipelines/stages.yaml
willie-yao: I think it would make sense to run this test regardless. I'll change this if @mboersma agrees!

mboersma: If `CLEANUP` is true, we run the `delete-storage-account.yaml` task right after building the VHD. So the storage account would be gone and there wouldn't be anything to test.

CecileRobertMichon: could we move the cleanup to after the test? And maybe add a separate variable to skip the test?

mboersma: We could move the cleanup, it's just a bit more refactoring because many other steps outside the scope of this PR check `CLEANUP`.

I don't think we need a `SKIP_TEST` variable because the test is already a separate stage. So when you run it in Azure DevOps, you can just un-click the test stage and things will run as they used to, without this test.

CecileRobertMichon: Got it, I think this is fine for now. If we change our mind later and realize it would be helpful to create a test cluster as part of the nightly dry run pipeline we can do that refactor then.

CecileRobertMichon (commented):
/lgtm

mboersma (approved):
/approve

@willie-yao used this current commit to build the last round of CAPZ reference images successfully, and it saved us from publishing some potentially faulty Windows VHDs. <3

mboersma:
/test pull-azure-vhds

Failure was transient a CentOS 7 403 error I've seen before. Works now:

```
wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```

PR #1012, opened 2022-11-02
capi,deb,kubevirt: override default netplan

rhrazdil:
The default netplan configuration created at cloud-init enables dhcp for a primary interface, that matches it's mac-address. However, when used with KubeVirt, the mac address changes on reboot, which mean that after the reboot, the DHCP is not enabled and thus IP is not acquired.

This commit overrides the default by enabling the dhcp for ethernet interfaces that match name for a pattern: enp*s*.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

**Additional context**


davidvossel (commented):
Before we merge this, can you run a vm build that uses this logic through the capk CI? You can use a PR to trigger this in capk repo itself.

For example
- build a new container disk with this logic and push it to your quay.io repo
- make the container publicly accessible in the quay.io ui
- make a pr to capk that overrides the normal ubuntu image with your custom image... by changing the line here in the functest.sh script, https://github.com/kubernetes-sigs/cluster-api-provider-kubevirt/blob/ccae28797861a0c232685823e4f58089b2d75433/hack/functest.sh#L18 
- then see how CI behaves.

davidvossel (commented):
/lgtm

yes! So glad this worked

images/capi/packer/qemu/packer.json
```diff
@@ -68,7 +68,7 @@
     },
     {
       "ansible_env_vars": [
-        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'"
+        "ANSIBLE_SSH_ARGS='{{user `existing_ansible_ssh_args`}} -o IdentitiesOnly=yes'", "KUBEVIRT={{user `kubevirt`}}"
```
kkeshavamurthy, line 71: I think `ANSIBLE_SSH_ARGS` is not the right place to pass non SSH related vars to the build.

rhrazdil: I'm not sure, IIUC, it would affect all images, wouldn't it?
This way, we're only adding this variable to qemu images

davidvossel (commented):
/lgtm

images/capi/packer/qemu/packer.json
kkeshavamurthy: True, it would  be passed for all builds  but I guess exercised only in qemu-kubevirt builds. I think this definitely a issue with `ansible-args.json`. It has become a common dumping location for all the ansible vars that used irrespective of if it's needed by the build.

kkeshavamurthy: We should find a cleaner solution

PR #1017, opened 2022-11-22
feat(gcp): user overridable image_name and image_family for gce

MadJlzz:

This PR adds the capability to override `image_name` and `image_family` for GCE images.
We would like to be able to have our own naming convention when storing images to standardize the way they can be found whatever the provider.

n.a

**Additional context**

I tried to keep the default naming behavior as much as possible but due to limitation of using functions within the `variables` block, I had to use `clean_resource_name` function on the entire variable instead of substrings:

https://github.com/kubernetes-sigs/image-builder/blob/03def50db05d589d409cbe9409912e1f6cdc51fe/images/capi/packer/gce/packer.json#L5-L6

What are you thoughts on this?

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -2,8 +2,8 @@
   "builders": [
     {
       "disable_default_service_account": "{{ user `disable_default_service_account` }}",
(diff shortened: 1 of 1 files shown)


mboersma (commented):
/lgtm

This looks like a benign way to allow those builder vars to be overriden.

PR #1021, opened 2022-11-29
Set FLATCAR_VERSION to latest available in Azure CI

### LINKED ISSUE #1002: Ensure FLATCAR_VERSION is available in Azure
The release of Flatcar images published to Azure is sometimes delayed, which creates a situation, where `hack/image-grok-latest-flatcar-version.sh` and `az vm image list --publisher kinvolk --sku stable --offer flatcar-container-linux-free --all -o json | jq -r '.[] | select( .sku == "stable") | .version' | sort | tail -n1` will return different versions, which makes CI fail for Flatcar until latest image is published.

Perhaps we should change `hack/image-grok-latest-flatcar-version.sh` to use the latter command in case we deploy to Azure.

CC @dongsupark @jepio

jepio:
The default `FLATCAR_VERSION=current` resolves to something that is not always available on Azure. Override its value to the latest version actually available in Azure's marketplace, to ensure that CI does not fail because of this.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1002

**Additional context**
Add any other context for the reviewers:
I have no better idea than doing it this way, I investigated conditionally returning a different value from image-grok-latest-flatcar-version but found no way to access the "target" from global scope in the Makefile.
Setting `FLATCAR_VERSION=latest` in the Azure CI script would also make it work, but is more hacky.

### DIFF
--- a/images/capi/scripts/ci-azure-e2e.sh
+++ b/images/capi/scripts/ci-azure-e2e.sh
@@ -78,6 +78,13 @@ trap cleanup EXIT
 
 make deps-azure
 
+# Latest Flatcar version is often available on Azure with a delay, so resolve ourselves
+az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
+get_flatcar_version() {
+    az vm image show --urn kinvolk:flatcar-container-linux-free:stable:latest --query 'name' -o tsv
+}
+export FLATCAR_VERSION="$(get_flatcar_version)"
+
 # Pre-pulling windows images takes 10-20 mins
 # Disable them for CI runs so don't run into timeouts
 export PACKER_VAR_FILES="packer/azure/scripts/disable-windows-prepull.json scripts/ci-disable-goss-inspect.json"


images/capi/scripts/ci-azure-e2e.sh
```diff
@@ -78,6 +78,12 @@ trap cleanup EXIT
 
 make deps-azure
 
+# latest Flatcar version is often available on Azure with a delay, so resolve ourselves
```
invidian, line 81: nit:
```suggestion
# Latest Flatcar version is often available on Azure with a delay, so resolve ourselves.
```

invidian (approved):
LGTM, thanks for taking care of it @jepio! It indeed seems like a good path for the time being.

jepio:
Fails due to missing `az login`, @mboersma how much do you hate adding that call to the ci script?

/test pull-azure-sigs

jepio:
@mboersma could you retrigger?

invidian:
> @mboersma could you retrigger?

I did already.

jepio:
centos-7 flaked, retry
/test pull-azure-sigs

mboersma:
This failed on a CentOS build which seems unrelated:

```
[0;32m    sig-centos-7: TASK [setup : add epel repo] ***************************************************[0m
[0;32m    sig-centos-7: fatal: [default]: FAILED! => {"changed": false, "msg": "Failure downloading https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm, HTTP Error 403: Forbidden"}
```

jsturtevant:
if we continue to see that, we could maybe add a retry to that task?

mboersma:
/retest-required

Ran up against the subscription's quota for `Standard_D2s_v3`.

PR #1028, opened 2022-12-02
images/capi/packer/azure: allow building images from various SIG images

invidian:

As far as I understand, right now it is only possible to build images
from Marketplace images. This patch actually enables building from the
community images, directly shared images and private SIGs.

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>

If one wants to build Flatcar image using SIG, one should create `packer.json` file with content similar to below:
```json
{
  "community_gallery_image_id": "/CommunityGalleries/<public galery name>/Images/flatcar-{{env `FLATCAR_CHANNEL`}}-amd64/Versions/{{env `FLATCAR_VERSION`}}",
  "image_offer": "",
  "image_publisher": "",
  "image_sku": "",
  "image_version": "",
  "plan_image_offer": "",
  "plan_image_publisher": "",
  "plan_image_sku": ""
}
```

For the time being though, let's keep building Flatcar from Marketplace image until we consider community galleries stable.

**Additional context**

Related to https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/2890.

CC @jepio @johananl


images/capi/packer/azure/scripts/init-sig.sh
```diff
@@ -34,44 +37,44 @@ SIG_TARGET=$1
...
-    create_image_definition ${SIG_TARGET} "win-2019-containerd" "V1" "Windows"
+    create_image_definition capi-${SIG_TARGET} "win-2019-containerd" "V1" "Windows"
   ;;
   windows-2022-containerd)
-    create_image_definition ${SIG_TARGET} "win-2022-containerd" "V1" "Windows"
+    create_image_definition capi-${SIG_TARGET} "win-2022-containerd" "V1" "Windows"
   ;;
   flatcar)
-    SKU="flatcar-${FLATCAR_CHANNEL}-${FLATCAR_VERSION}"
+    SKU="flatcar-${FLATCAR_CHANNEL}-amd64-capi-${KUBERNETES_SEMVER}-gen1"
```
invidian, line 61: I wonder if this is a good idea to do here. Perhaps it should be made externally configurable, so default image names remain consistent across all OSes.

jepio: I would think that we would republish the image version from the gallery produced by image-builder into the target gallery using the same source id.

invidian: Hmm, that may make a lot of sense, I'll give it a try. Thanks for the hint! Bummer I didn't get it before I wrote those patches :smile:

jepio: It might be possible to use an image version from one gallery as the source for another?

invidian: Yes, it should be possible according to docs:
```
    Add a new image version from another image version
        az sig image-version create --resource-group MyResourceGroup \
        --gallery-name MyGallery --gallery-image-definition MyImage \
        --gallery-image-version 1.0.0 \
        --image-version /subscriptions/00000000-0000-0000-0000-
        00000000xxxx/resourceGroups/imageGroups/providers/Microsoft.Compute/galleries/MyGallery/imag
        es/MyImageDefinition/versions/1.0.0
```

invidian: Works as expected, awesome. I'll mark as resolved.

invidian:
This is ready for merging I think.

jsturtevant:
Could add some docs to https://image-builder.sigs.k8s.io/capi/providers/azure.html on how to do this.

/lgtm

invidian:
>Could add some docs to https://image-builder.sigs.k8s.io/capi/providers/azure.html on how to do this.

Added some minimal docs. Let me know if you think there is a need for a whole section with some examples.

docs/book/src/capi/providers/azure.md
```diff
@@ -52,6 +52,7 @@ list, and greater explanation can be found in the
 
 | Variable | Description | Default |
 |----------|-------------|---------|
+| `community_gallery_image_id` | Use image from a Community gallery as a base image instead of default one from the marketplace. | `""` |
```
jsturtevant, line 55: if you supply this will it ignore the other fields, or do you need to blank those out too?

images/capi/packer/azure/packer-windows.json
```diff
@@ -61,6 +61,9 @@
       "os_disk_size_gb": "{{user `os_disk_size_gb`}}",
       "os_type": "Windows",
       "private_virtual_network_with_public_ip": "{{user `private_virtual_network_with_public_ip`}}",
+      "shared_image_gallery": {
+        "community_gallery_image_id": "{{ user `community_gallery_image_id` }}"
```
kkeshavamurthy, line 65: For completeness can we add the `direct_shared_gallery_image_id` option as well?
That way one could build from a private or community gallery?

invidian: I can add this as well, although actually for Flatcar images, we ended up with using just SIG images as a source, so I can just implement all possible variants.

invidian: Updated.

docs/book/src/capi/providers/azure.md
invidian: Added in docs, please have another look.

invidian:
While this PR adds support for all combinations, Flatcar team will be using building from the private SIGs for the time being and resulting images will be published into community SIGs.

kkeshavamurthy:
Thanks @invidian 
/lgtm

invidian:
I didn't expect such a quick merge, thanks!

PR #1030, opened 2022-12-12
Nutanix: Allow output image override if already exists

fad3t:
Allow support for the `force_deregister` flag in the Nutanix Packer plugin (cf. https://developer.hashicorp.com/packer/plugins/builders/nutanix/nutanix#environment-configuration).

> [force_deregister](https://developer.hashicorp.com/packer/plugins/builders/nutanix/nutanix#force_deregister) (boolean) - Allow output image override if already exists.


mboersma (approved):
/approve

PR #1031, opened 2022-12-12
Enable HNS and WCIFS fixes for WS2019

jsturtevant:
Signed-off-by: James Sturtevant <jstur@microsoft.com>

This resolves some OS level bugs in Windows Server.  These changes are coming from https://github.com/Azure/aks-engine/pull/5026 and aks vhds: https://github.com/Azure/AgentBaker/blob/ddbf61e7f6f42a0606e8c00b798b08315b32fd04/vhdbuilder/packer/configure-windows-vhd.ps1#L400-L416

**Additional context**

/sig windows
/assign @marosset

### DIFF
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -125,6 +125,18 @@ command:
     stdout:
     - "True"
     timeout: 30000
+  Check HNS Control Flag:
+    exit-status: 0
+    exec: powershell -command "(Get-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name HNSControlFlag).HNSControlFlag -eq 80"
+    stdout:
+    - True
+    timeout: 30000
+  Check WCIFS Flag:
+    exit-status: 0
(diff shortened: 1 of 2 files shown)


images/capi/packer/goss/goss-command.yaml
```diff
@@ -125,6 +125,18 @@ command:
     stdout:
     - "True"
     timeout: 30000
+  Check SMB CompartmentNamespace Flag:
+    exit-status: 0
+    exec: powershell -command "(Get-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Services\hns\State' -Name EnableCompartmentNamespace).EnableCompartmentNamespace -eq 0x50"
```
marosset, line 130: One slight difference between this and what was checked into the other repositories is that the other snippets are doing a binary or on 0x50 with the current value (if it is set).
From the packer logs for the other PRs it looks like value isn't set to anything in the base VHDs so I don't *think* it should be an issue. WDYT?

jsturtevant: oh, I missed that.  doing the OR is probably the safest if it was ever changed by the OS.  Not sure the likely hood of that.

jsturtevant: given we have a goss test that makes sure this value match our expectation and these fields should eventually be dropped I think it is ok to just set them

marosset (approved):
/lgtm

PR #1033, opened 2022-12-14
add support for ansible_user_vars in nutanix

fad3t:
* Add support for `ansible_user_vars` when building images using Nutanix.


mboersma (commented):
/ok-to-test

PR #1037, opened 2022-12-16
add containerd-wasm-shims to enable Wasm / WASI workloads

devigned:

This PR adds support for running Wasm workloads using both [Fermyon Spin](https://github.com/fermyon/spin) and [Deislabs SpiderLightning](https://github.com/deislabs/spiderlightning) via [Deislabs Containerd Wasm Shims](https://github.com/deislabs/containerd-wasm-shims). The same technology is what enables Azure Kubernetes Service to run Wasm / WASI workloads in [AKS WASI Node Pools](https://learn.microsoft.com/en-us/azure/aks/use-wasi-node-pools). 

Wasm / WASI workloads using a runtime class which informs containerd to delegate the pod workload to the specific shim as described in https://github.com/deislabs/containerd-wasm-shims#using-a-shim-in-kubernetes. 

**Additional context**
This PR stems from [this Slack discussion](https://kubernetes.slack.com/archives/CEX9HENG7/p1671044490720939) with @elmiko.


images/capi/packer/azure/packer.json
```diff
@@ -211,6 +211,7 @@
     "client_id": null,
     "client_secret": null,
     "community_gallery_image_id": "",
+    "containerd_additional_settings": "slight,spin",
```
jsturtevant, line 214: This field is being used as a way to pass arbitrary contained configuration bia base64 so I believe this will result in a bad toml file for containerd.  See line 25/33 in the `templates/etc/containerd/config.toml`

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -30,6 +30,13 @@
     dest: /tmp/containerd.tar.gz
     mode: 0600
 
+- name: download containerd-wasm-shims
```
jsturtevant, line 33: this should probably only be downloaded and installed if those config options are present?  Should this be added to the default install?  I see it is downloaded for everyone but only turned on in the containerd config for azure.

images/capi/packer/azure/packer.json
devigned: Thank you! I was worried about that. I'll create a new variable.

images/capi/ansible/roles/containerd/tasks/main.yml
devigned: I'll add the download conditionally based on setting: https://github.com/kubernetes-sigs/image-builder/pull/1037#discussion_r1052491837.

devigned:
/hold cancel

@jsturtevant I believe this is solid.

mboersma (commented):
/lgtm

images/capi/packer/goss/goss-command.yaml
```diff
@@ -10,6 +10,23 @@ command:
...
+    exit-status: 1
+    stdout: [ ]
+    stderr: ["io.containerd.slight.v1: InvalidArgument(\"Shim namespace cannot be empty\")"]
+    timeout: 0
+  containerd-shim-spin-v1:
+    exit-status: 1
+    stdout: [ ]
+    stderr: ["io.containerd.spin.v1: InvalidArgument(\"Shim namespace cannot be empty\")"]
+    timeout: 0
+  grep -E 'io\.containerd\.(slight|spin)\.v1' /etc/containerd/config.toml:
```
jsturtevant, line 24: nit: I think this wouldn't verify both are there are present only one of them (similar to comment above)

jsturtevant (commented):
/lgtm
/assign @kkeshavamurthy @mboersma

images/capi/packer/goss/goss-command.yaml
```diff
@@ -10,6 +10,23 @@ command:
     stdout: []
     stderr: []
     timeout: 0
+{{if ne .Vars.containerd_wasm_shims_runtimes ""}}
```
jsturtevant, line 13: It isn't garmented both runtime would be installed for this check so this could fail if only one is installed.  I am not sure this is a big deal at this point for the test.

images/capi/packer/goss/goss-command.yaml
devigned: You are 100% correct. I was having a hard time dealing with `.Vars.containerd_wasm_shims_runtime` as it's a comma delimited string (`"slight, spin"`). It's much easier to deal with it in Python in Ansible using a string contains.

Do you have a suggestion of how to handle it better in Goss?

images/capi/packer/goss/goss-command.yaml
devigned: It's unfortunate that the containerd wasm shims are packaged together into one release artifact. When you download one, you get both. 

I think in the future, I'd like to change the shims to be packaged individually and change the variables / tests in image-builder to allow for individual installation and verification.

I've also opened [an issue to add `--version` support](https://github.com/deislabs/containerd-wasm-shims/issues/60) for the shims. I didn't like the error test. The test does the job of verifying the shim is installed, on the path, and responding in a reasonable way, but I would much rather see something closer to the containerd version verification, so we can ensure we have the specific version installed that we expect.

mboersma: > I think in the future, I'd like to change the shims to be packaged individually and change the variables / tests in image-builder to allow for individual installation and verification.

I think it's acceptable to assume both shims for now, and to fix this when their packaging changes.

mboersma (approved):
/approve

There are a couple of small followup items identified here, but 👍🏻.

PR #1041, opened 2022-12-23
Fix Makefiles and update image-builder

mboersma:

The root `Makefile` has rusted a bit. The default targets apply to the `image-builder` code in the `v1` directory, but failed for me out of the box with Go 1.19 (and probably 1.17 and 1.18 I'm guessing).

This makes minimal changes to fix `make`, `make clean`, and get the image-builder code building again.

Which issue(s) this PR fixes: 

N/A

**Additional context**


kkeshavamurthy:
/lgtm

Are there any users of the image builder cli? I have not seen a commit go into it in a couple of years.

PR #1049, opened 2023-01-10
Unpin azure-cli to always install the latest

kkeshavamurthy:
Would be great to install the latest version of azure-cli instead of pinning to a particular version. 
This way we are able to get the latest features & security fixes when available. 

**Additional context**


mboersma (commented):
/lgtm

This may have been pinned to work around some conflicts with Python package dependencies, but installing the latest version appears to work fine. If we run into problems in the future, we may have to pin it again, but I think this is better than forgetting to update it.

PR #1051, opened 2023-01-11
Add support for Windows 2019/2022 to OCI capi provider image builder

joekr:
Adds support to build windows images to using with CAPOCI provider using cluster-api. 

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): No open issues

**Additional context**
the `oci_iad_windows_bm.json` file contents
```
{
  "availability_domain": "<myAD>",
  "build_name": "windows",
  "base_image_ocid": "ocid1.image.oc1.iad.aaaaaaaazt5nv5koc2sxjagzfvdv7gnza2te7gheiscw2ukdrwcdk4mwhyxq",
  "compartment_ocid": "<myCompartment",
  "kubernetes_deb_version": "1.24.9-00",
  "kubernetes_rpm_version": "1.24.9-0",
  "kubernetes_semver": "v1.24.9",
  "kubernetes_series": "v1.24",
  "ocpus": "128",
  "region": "us-ashburn-1",
  "shape": "BM.Standard.E4.128",
  "subnet_ocid": "<mySubnet>"
}
```

Build logs 
```
$ OPC_USER_PASSWORD="<mypassword>"  no_proxy=* PACKER_VAR_FILES=oci_iad_windows_bm.json make build-oci-windows-2019

...
...
...

==> oracle-oci: Provisioning with powershell script: /var/folders/yp/cx82s01j4m1d34krx_q97xs80000gn/T/powershell-provisioner2441768962
    oracle-oci: Using cloudbase-init unattend file for sysprep: C:\Program Files\Cloudbase Solutions\Cloudbase-Init\conf\Unattend.xml
    oracle-oci: IMAGE_STATE_COMPLETE
    oracle-oci: IMAGE_STATE_UNDEPLOYABLE
    oracle-oci: IMAGE_STATE_GENERALIZE_RESEAL_TO_OOBE
    oracle-oci: >>> Sysprep complete ...
==> oracle-oci: Creating image from instance...
==> oracle-oci: Updating image schema...
==> oracle-oci: Created image (ocid1.image.oc1.iad.XXXX).
==> oracle-oci: Terminating instance (ocid1.instance.oc1.iad.XXXX)...
==> oracle-oci: Terminated instance.
==> oracle-oci: Running post-processor: manifest
Build 'oracle-oci' finished after 33 minutes 47 seconds.

==> Wait completed after 33 minutes 47 seconds

--> oracle-oci: An image was created: 'cluster-api-windows-v1.24.9-1673378936' (OCID: ocid1.image.oc1.iad.XXXX) in region 'us-ashburn-1'
--> oracle-oci: An image was created: 'cluster-api-windows-v1.24.9-1673378936' (OCID: ocid1.image.oc1.iad.XXXX) in region 'us-ashburn-1'
```


images/capi/packer/oci/scripts/enable_second_nic.ps1
```diff
@@ -0,0 +1,42 @@
+
+$newNetAdapterName = "Ethernet 2"
+
+# check for two nics
+$netAdapters = Get-NetAdapter 
+$secondNic = $netAdapters[1]
```
joekr, line 6: - [ ] actually check that there are two nics

Right now this just assumes two. Need to add logic for two.

images/capi/ansible/windows/roles/cloudbase-init/tasks/main.yml
```diff
@@ -42,6 +42,14 @@
   win_template:
     src: templates/cloudbase-init.conf
     dest: '{{ programfiles.stdout | trim }}\Cloudbase Solutions\Cloudbase-Init\conf\cloudbase-init.conf'
+  when: not packer_builder_type.startswith('oracle-oci')
+
+# custom cloudbase-init file has modified plugins specific to OCI enabled plugins
+- name: Set up cloudbase-init-oci configuration 
+  win_template:
+    src: templates/cloudbase-init-oci.conf
+    dest: '{{ programfiles.stdout | trim }}\Cloudbase Solutions\Cloudbase-Init\conf\cloudbase-init.conf'
```
jsturtevant, line 51: what is the difference between these? most configuration should be configurable through `cloudbase_metadata_services` and `cloudbase_plugins` I think?

images/capi/packer/oci/packer-windows.json
```diff
@@ -0,0 +1,150 @@
...
+        "--extra-vars",
+        "{{user `ansible_common_vars`}}",
+        "--extra-vars",
+        "{{user `azure_extra_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_extra_vars`}}",
+        "--extra-vars",
+        "{{user `ansible_user_vars`}}",
+        "--extra-vars",
+        "gmsa_keyvault_url={{user `gmsa_keyvault_url`}}"
```
jsturtevant, line 71: I don't think you need this?

images/capi/packer/oci/packer-windows.json
```diff
@@ -0,0 +1,150 @@
...
+    "private_virtual_network_with_public_ip": "",
+    "runtime": "containerd",
+    "shape": "VM.Standard.E4.Flex",
+    "virtual_network_name": "",
+    "virtual_network_resource_group_name": "",
+    "virtual_network_subnet_name": "",
+    "vm_size": "",
+    "windows_service_manager": null,
+    "windows_updates_kbs": null,
+    "wins_url": "https://github.com/rancher/wins/releases/download/v{{user `wins_version`}}/wins.exe",
```
jsturtevant, line 147: could probably drop this since you are using contianerd.

images/capi/ansible/windows/roles/cloudbase-init/tasks/main.yml
joekr: I saw I could modify the `cloudbase_plugins` but it appeared to only be addition, I wouldn't figure out how to remove plugins OCI doesn't support https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/windows/roles/cloudbase-init/templates/cloudbase-init.conf#L35.

The OCI template removes 
```
cloudbaseinit.plugins.windows.createuser.CreateUserPlugin, 
cloudbaseinit.plugins.common.setuserpassword.SetUserPasswordPlugin, 
cloudbaseinit.plugins.windows.extendvolumes.ExtendVolumesPlugin,
```

images/capi/packer/oci/packer-windows.json
joekr: Removed

images/capi/packer/oci/packer-windows.json
joekr: Removed

images/capi/ansible/windows/roles/cloudbase-init/tasks/main.yml
jsturtevant: I think it might be better to move these out of the base template to avoid duplicating this file as much as we can.  what do you think?

joekr: Makes sense to me.

joekr: I've moved all the plugins to their respective `packer-windows.json` files and reduced the amount of default `cloudbase_plugins` in the `cloudbase-init.conf` file.

jsturtevant:
centos failed when updating with yum. I don't know much about yum and centos, would a retry to this command help flakes?

```
 [Errno -1] Metadata file does not match checksum\nTrying other mirror.\n\n\n One of the configured repositories failed (CentOS-7 - Updates (OpenLogic)),\n and yum doesn't have enough cached data to continue. At this point the only\n safe thing yum can do is fail. There are a few ways to work \"fix\" this:\n\n     1. Contact the upstream for the repository and get them to fix the problem.\n\n     2. Reconfigure the baseurl/etc. for the repository, to point to a working\n        upstream. This is most often useful if you are using a newer\n        distribution release than is supported by the repository (and the\n        packages for the previous distribution release still work).\n\n     3. Run the command with the repository temporarily disabled\n            yum --disablerepo=updates-openlogic ...\n\n     4. Disable the repository permanently, so yum won't use it by default. Yum\n        will then just ignore the repository until you permanently enable it\n        again or use --enablerepo for temporary usage:\n\n            yum-config-manager --disable updates-openlogic\n        or\n            subscription-manager repos --disable=updates-openlogic\n\n     5. Configure the failing repository to be skipped, if it is unavailable.\n        Note that yum will try to contact the repo. when it runs most commands,\n        so will have to try and fail each time (and thus. yum will be be much\n        slower). If it is a very temporary problem though, this is often a nice\n        compromise:\n\n            yum-config-manager --save --setopt=updates-openlogic.skip_if_unavailable=true\n\nfailure: repodata/filelists.sqlite.bz2 from updates-openlogic: [Errno 256] No more mirrors to try.\nhttp://olcentgbl.trafficmanager.net/centos/7/updates/x86_64/repodata/filelists.sqlite.bz2: [Errno -1] Metadata file does not match checksum\n", "obsoletes": 
```

/retest

joekr:
I had to push a change. After removing `"wins_url"` the image would build, but failed to join the cluster. Using `   "wins_url": "",` fixes the joining issue.

images/capi/packer/oci/scripts/set_bootstrap.sh
```diff
@@ -0,0 +1,25 @@
...
+# This script overrides the 'opc' password set in the winrm_bootstrap.txt file
+# This script is assumed to be run from the make file hence the pathing to the winrm_bootstrap.txt
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "Changing Password in winrm_bootstrap.txt"
+
+sed "s/(\[adsi\].*/([adsi](\"WinNT:\/\/\"+\$opcUser.caption).replace(\"\\\\\",\"\/\")).SetPassword(\"$OPC_USER_PASSWORD\")/g" packer/oci/scripts/winrm_bootstrap.txt | tee packer/oci/scripts/winrm_bootstrap.txt >/dev/null
```
jsturtevant, line 25: Is this going to leave the password in a file as plain text?

joekr: It will modify the password set in the `winrm_bootstrap.txt` to be set a launch. From my understanding I need to set a password at launch as our image process sets the password to automatically reset on launch and it won't be known. 

Here is an old example related issue: https://github.com/hashicorp/packer/issues/7033#issuecomment-486824142

I'm not sure how to use the `ENV` to set the value in `winrm_bootstrap.txt` without saving it to the file for launch.

joekr: I have updated how this works now. When the image is built we temporarily set the password to the user provided password, when the image create the file containing the provided password is deleted. I've also updated the book to call that out.

PR #966, opened 2022-09-07
Support setting oem_id for Flatcar QEMU images

AverageMarcus:
What this PR does / why we need it:

Allow the ability to set the `oem_id` value for Flatcar images built with QEMU, needed for OpenStack.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #937

**Additional context**
Currently untested but wanted to get feedback early that this is the appropriate approach to take.

PR #1054, opened 2023-01-13
capi/packer/qemu: Pass oem_id value to ansible

jepio:
Fixes an issue with PR #966 - the oem_id variable in packer is set to the value of env OEM_ID but this is not forwarded to ansible. Make sure this is done so that the ansible step is executed correctly.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #966 

**Additional context**


jepio:
@AverageMarcus
stumbled across this when we tried to set OEM_ID

AverageMarcus (commented):
/lgtm

PR #1062, opened 2023-01-24
Support building RHEL-8 Azure VHD and SIG Images

kkeshavamurthy:
Add support for building RHEL-8 Azure VHD & SIG Images. 

**Additional context**

### DIFF
--- a/images/capi/packer/azure/centos-7-gen2.json
+++ b/images/capi/packer/azure/centos-7-gen2.json
@@ -2,7 +2,7 @@
   "build_name": "centos-7-gen2",
   "distribution": "centos",
   "distribution_release": "centos-7",
-  "distribution_version": "7.7",
+  "distribution_version": "7",
   "image_offer": "CentOS",
   "image_publisher": "OpenLogic",
   "image_sku": "7_7-gen2"
--- a/images/capi/packer/azure/centos-7.json
+++ b/images/capi/packer/azure/centos-7.json
@@ -2,7 +2,7 @@
   "build_name": "centos-7",
   "distribution": "centos",
   "distribution_release": "centos-7",
-  "distribution_version": "7.7",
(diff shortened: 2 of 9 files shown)


kkeshavamurthy:
@jsturtevant @mboersma This should be good to go. PTAL.

images/capi/packer/azure/centos-7-gen2.json
```diff
@@ -2,7 +2,7 @@
   "build_name": "centos-7-gen2",
   "distribution": "centos",
   "distribution_release": "centos-7",
-  "distribution_version": "7.7",
+  "distribution_version": "7",
```
mboersma, line 5: This seems unrelated to the RHEL 8 changes. Is this so the build will use the latest 7.x version (currently CentOS 7.9)?

kkeshavamurthy: Right. `distribution_version` is used in image definition names, goss, ansible conditions etc. So I standardized it to `7` instead of `7.7` since it makes more sense. It'll also help in running the right tests in goss. `image_sku` is used to pick the right base images from the azure marketplace and it's left at `7.7`.

images/capi/packer/azure/centos-7.json
```diff
@@ -2,7 +2,7 @@
   "build_name": "centos-7",
   "distribution": "centos",
   "distribution_release": "centos-7",
-  "distribution_version": "7.7",
+  "distribution_version": "7",
   "image_offer": "CentOS",
   "image_publisher": "OpenLogic",
   "image_sku": "7.7"
```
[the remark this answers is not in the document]
kkeshavamurthy: `image_sku` is not linked to `distribution_version` AFAIK.

mboersma (commented):
/lgtm

Issue #924, opened 2022-06-27
Speed up Flatcar QEMU builds


invidian:
**Is your feature request related to a problem? Please describe.**

Right now Flatcar QEMU build downloads ISO image via Packer, then inside the VM image is downloaded again by Flatcar installation script to install Flatcar to the disk and reboot. This is disk and network intensive and could possibly be improved.

**Describe the solution you'd like**

Perhaps it would be possible to use `"disk_image"` and `"qemuargs"` Packer QEMU builder options to pass packer Ignition directly to the Packer QEMU builder to skip one level of indirection and get a VM directly booted with SSH available.

One blocker for this might be that IIRC right now Flatcar does not offer uncompressed QEMU disk images, so we would either have to make them available or download and unpack the image manually before we run Packer. (EDIT: reported https://github.com/flatcar-linux/Flatcar/issues/791)

Here is the PoC solution which seem to work:
```diff
diff --git images/capi/packer/qemu/packer.json images/capi/packer/qemu/packer.json
index 5c626e9c0..c30828f56 100644
--- images/capi/packer/qemu/packer.json
+++ images/capi/packer/qemu/packer.json
@@ -28,7 +28,13 @@
       "ssh_timeout": "2h",
       "ssh_username": "{{user `ssh_username`}}",
       "type": "qemu",
-      "vm_name": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}"
+      "vm_name": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}",
+      "disk_image": "{{user `disk_image`}}",
+      "qemuargs": [
+        [
+          "-fw_cfg", "name=opt/org.flatcar-linux/config,file=/<absolute path to image builder repo>/image-builder/images/capi/packer/qemu/flatcar/ignition-builder.json"
+        ]
+      ]
     }
   ],
   "post-processors": [
@@ -165,6 +171,7 @@
     "python_path": "",
     "qemu_binary": "qemu-system-x86_64",
     "ssh_password": "builder",
-    "ssh_username": "builder"
+    "ssh_username": "builder",
+    "disk_image": "false"
   }
 }
diff --git images/capi/packer/qemu/qemu-flatcar.json images/capi/packer/qemu/qemu-flatcar.json
index bb10cccce..120456f84 100644
--- images/capi/packer/qemu/qemu-flatcar.json
+++ images/capi/packer/qemu/qemu-flatcar.json
@@ -1,6 +1,5 @@
[...]

PR #1064, opened 2023-01-26
Add Cloudstack provider support for Flatcar

hrak:

This PR adds Cloudstack provider support for Flatcar Linux. Flatcar requires some support scripts to be present in the oem partition (/usr/share/oem) to support the Cloudstack metadata service. These files were taken from [Flatcar's own Cloudstack image](https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_cloudstack_image.bin.bz2)

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-coreos-cloudinit
+++ b/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-coreos-cloudinit
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+. /usr/share/oem/bin/cloudstack-dhcp
+
+DHCP_SERVER=$(get_dhcp_ip)
+USERDATA_URL="http://${DHCP_SERVER}/latest/user-data"
+
+block-until-url "${USERDATA_URL}"
+coreos-cloudinit --from-url="${USERDATA_URL}"
--- a/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-dhcp
+++ b/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-dhcp
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+get_dhcp_ip() {
+    local leases_dir="/run/systemd/netif/leases"
+    local found=0
+    while true; do
+        for leasefile in $(find "${leases_dir}" -type f -size +1c); do
+            dhcp_server_ip=$(cat $leasefile | awk -F= '/SERVER_ADDRESS/ { print $2 }')
+            if [[ -n "${dhcp_server_ip}" ]]; then
+                metadata_url="http://${dhcp_server_ip}/latest/meta-data/"
+                if curl --fail -s "${metadata_url}" >/dev/null; then
+                    echo $dhcp_server_ip
+                    found=1
+                    break
+                fi
+            fi
+        done
+        [[ $found -eq 0 ]] || break
+        sleep .5
+    done
+}
--- a/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-ssh-key
+++ b/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/cloudstack-ssh-key
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+. /usr/share/oem/bin/cloudstack-dhcp
+
+DHCP_SERVER=$(get_dhcp_ip)
+KEY_URL="http://${DHCP_SERVER}/latest/meta-data/public-keys"
+
+block-until-url "${KEY_URL}"
+curl --fail -s "${KEY_URL}" | update-ssh-keys -a cloudstack
--- a/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/flatcar-setup-environment
+++ b/images/capi/ansible/roles/providers/files/flatcar-oem-cloudstack/bin/flatcar-setup-environment
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+ENV=$1
+
+if [[ -z "$ENV" ]]; then
+    echo "usage: $0 /etc/environment" >&2
+    exit 1
+fi
+
+# Make sure that the file is writable
+touch $ENV
+if [[ $? -ne 0 ]]; then
+    echo "$0: unable to modify ${ENV}" >&2
+    exit 1
+fi
+
+sed -i -e '/^COREOS_PUBLIC_IPV4=/d' \
+    -e '/^COREOS_PRIVATE_IPV4=/d' \
+    "${ENV}"
+
+. /usr/share/oem/bin/cloudstack-dhcp
+
+DHCP_SERVER=$(get_dhcp_ip)
+METADATA_URL="http://${DHCP_SERVER}/latest/meta-data/"
+
+block-until-url "${METADATA_URL}"
+
+PUBLIC_IP=$(curl --fail -s "${METADATA_URL}public-ipv4")
+echo COREOS_PUBLIC_IPV4=${PUBLIC_IP} >> $ENV
+
+PRIVATE_IP=$(curl --fail -s "${METADATA_URL}local-ipv4")
+echo COREOS_PRIVATE_IPV4=${PRIVATE_IP} >> $ENV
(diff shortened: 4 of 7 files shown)


invidian:
@mboersma thanks for assigning me, but perhaps we should get something more knowledgeable about Flatcar internals to look into this one. Maybe @jepio or @pothos would find a minute to have a look?

@hrak thanks for the PR. As you mentioned [Flatcar's own Cloudstack image](https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_cloudstack_image.bin.bz2), I wonder, could we customize `iso_url` perhaps when building an image for different provider? This way, we wouldn't have to copy-paste and play keep-up-to-date game with Flatcar images, but just inherit what they provide. I don't know how feasible that is though.

I'm talking about these:
```console
$ git grep flatcar-linux.net
packer/ova/flatcar.json:  "iso_checksum": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso.DIGESTS.asc",
packer/ova/flatcar.json:  "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
packer/qemu/qemu-flatcar.json:  "iso_checksum": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso.DIGESTS.asc",
packer/qemu/qemu-flatcar.json:  "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
packer/raw/raw-flatcar.json:  "iso_checksum": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso.DIGESTS.asc",
packer/raw/raw-flatcar.json:  "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
```

Could you give it a try?

hrak:
> @hrak thanks for the PR. As you mentioned [Flatcar's own Cloudstack image](https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_cloudstack_image.bin.bz2), I wonder, could we customize `iso_url` perhaps when building an image for different provider? This way, we wouldn't have to copy-paste and play keep-up-to-date game with Flatcar images, but just inherit what they provide. I don't know how feasible that is though.
> 
> Could you give it a try?

Flatcar does not supply provider-specific iso images, only VM images. What they do is build a VM image based on the iso and then build a OEM-specific ebuild into the oem partition in the 'image_to_vm' build stage (see [here](https://github.com/flatcar/scripts/blob/ffa5527012e995c9ff972c4e92f888736cf78569/build_library/vm_image_util.sh#L471))

The ebuild containing the oem-specific files can be found [here](https://github.com/flatcar/coreos-overlay/tree/main/coreos-base/oem-cloudstack). The files have not really changed in a long time.

invidian:
>Flatcar does not supply provider-specific iso images, only VM images. What they do is build a VM image based on the iso and then build a OEM-specific ebuild into the oem partition in the 'image_to_vm' build stage (see [here](https://github.com/flatcar/scripts/blob/ffa5527012e995c9ff972c4e92f888736cf78569/build_library/vm_image_util.sh#L471))

> The ebuild containing the oem-specific files can be found [here](https://github.com/flatcar/coreos-overlay/tree/main/coreos-base/oem-cloudstack). The files have not really changed in a long time.

TIL, thanks!

>So actually, this does not just apply to Cloudstack, but to any provider supported by image-builder. Which makes me wonder how people actually currently use Flatcar based capi images on any of the supported cloud providers, since i don't see [oem-azure](https://github.com/flatcar/coreos-overlay/tree/main/coreos-base/oem-azure), [oem-gce](https://github.com/flatcar/coreos-overlay/tree/main/coreos-base/oem-gce) etc. referenced anywhere.

At least on Azure, we use Azure-specific Flatcar images as a base, this is why there is no need for special handling of oem I think.

>btw, building based on an image instead of an ISO could be an option if the "disk_image": "true" flag would be supported as suggested in https://github.com/kubernetes-sigs/image-builder/issues/924.

Interesting. So then you would just point it to cloudstack disk image and that's it, correct? That seem like a nice win. Could you give it a try or at least leave a comment in #924 about this please? So we can collect more use cases for this, so it's maybe easier to prioritize to work on at some point.

Otherwise, I think the PR looks good then.

jepio:
@hrak I'm not excited about copying files out of the Flatcar repos and keeping them here. Not that the files change often, but if they no one would remember to keep them in sync. Here are two approaches that I think would be better:
* fetching them from github at runtime from the proper tag that matches the release
* passing the right OEM_ID to `flatcar-install -o`. You'd have to try this, but this should install the cloudstack image during the packer run.

pothos:
Using Packer, and specially in the way it's done here with the ISO "bootstrap" phase makes things more complicated than needed^^ As was already said, I also think the best options would be to start the right image or at least pass the `-o cloudstack` flag to `flatcar-install` (which is what we also do with VMware).
By the way, since cloudstack isn't covered by the Flatcar release tests, things may be not in the best state and if you know what to improve please file an issue and if you can suggest an easy way how to cover it in the Flatcar release tests that would also help.

hrak:
I'm back, and a little wiser after opening a related issue on the Flatcar project. It appears [cloud-init on Flatcar does not support Jinja templates](https://github.com/flatcar/Flatcar/issues/953), which cluster-api requires for cloud-init, so its a dead end road.

Someone pointed out to me in that issue that ignition is the way to go for cluster-api + Flatcar. This is working for me now, including metadata and installing ssh keys! Description of the process can be found in https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/216#issue-1565006039

Based on my findings, what could be implemented as a cloudstack provider for flatcar is those two drop-ins that i now send using ignition. They override the metadata provider name to be `cloudstack-metadata`. This is needed because coreos-metadata only knows providers `cloudstack-configdrive` and `cloudstack-metadata` not `cloudstack` like the oem id says.

hrak:
> Thanks, it looks like cloudstack could be switched to afterburn (coreos-metadata).

Indeed, the only thing is that using `oem_id=cloudstack` currently doesn't work because afterburn only knows about platforms `cloudstack-configdrive` and `cloudstack-metadata`. Hence the override systemd dropins that set `Environment=COREOS_METADATA_OPT_PROVIDER=--provider=cloudstack-metadata` in my ignition cfg.

This was [recently addressed for openstack](https://github.com/coreos/afterburn/pull/462), which had a similar problem. For cloudstack this is still an [open issue](https://github.com/coreos/afterburn/issues/856).

> A fix for the same issue was once necessary for openstack https://github.com/flatcar/bootengine/pull/21/files. Could you PR that for openstack?

It doesn't look like this is needed for Cloudstack, i don't have any issues with the hostname being set with Flatcar using afterburn on Cloudstack.

hrak:
> Thanks, it looks like cloudstack could be switched to afterburn (coreos-metadata). A fix for the same issue was once necessary for openstack https://github.com/flatcar/bootengine/pull/21/files. Could you PR that for openstack?

@jepio Looking into this some more now. It looks like the hostname is currently being set by whatever is supplied by DHCP, which works, but i was wondering if there is any advantage in doing this same Openstack approach for Cloudstack?

jepio:
> > Thanks, it looks like cloudstack could be switched to afterburn (coreos-metadata). A fix for the same issue was once necessary for openstack https://github.com/flatcar/bootengine/pull/21/files. Could you PR that for openstack?
> 
> @jepio Looking into this some more now. It looks like the hostname is currently being set by whatever is supplied by DHCP, which works, but i was wondering if there is any advantage in doing this same Openstack approach for Cloudstack?

What I intended to suggest was opening a PR to Flatcar to improve support for cloudstack, so that the OEM support does not need to be fixed downstream in image-builder. That way it works for anyone using cloudstack.

The openstack approach just seems similar to me, so you could reuse some of the same units/scripts after extending them a bit.

jsturtevant:
Checking in - Is this still a relevant PR?

hrak:
> Checking in - Is this still a relevant PR?

Sorry for the delay, no i think this can be closed, we have a solution now with some systemd drop-ins, and i will explore fixing this upstream.

PR #1074, opened 2023-02-10
Add note about Windows in OCI capi docs

joekr:
Want to make sure to clearly call out a caveat about instance shapes so users don't build using the wrong hardware. 

**Additional context**


mboersma (commented):
/lgtm

PR #1081, opened 2023-02-23
Allow Customizing values for Azure SIG Image Definition

primeroz:

At the moment the `init-sig.sh` script has many hardcoded values and does not allow customizing values. 

This PR allows customizing values using Environment Variables while defaulting to the current behaviour

**Additional context**

When creating FLATCAR image definition i want to be able to create

* image definition `capi-flatcar-stable-$(KUBERNETES_VERSION)-gen2`
* have one image for each `flatcar` version in the image definition which is organized by k8s version

with this PR i just need to export the right Environment variables to achieve this


mboersma (approved):
/lgtm
/approve

PR #1087, opened 2023-03-01
Do not use same variable for managed_image_storage_account_type and SIG storage_account_type

primeroz:

SIG `storage_account_type` support `Standard_ZFS` but `managed_image_storage_account_type` does _not_, 
using the same variable means we can't use Zone redundancy for SIG Images

This PR Split the variable for the 2 fields to allow us to specify ZFS storage for SIG Images

default for new variable is to copy the old variable value so it should be backward compatible for anyone setting this already

**Additional context**

At the moment you get

```
build-capz-image-1.24.10-r4e901b36c57fb10aaa0c9e5854a979c2a-pod step-build-image Error: Failed to prepare build: "sig-flatcar-gen2"
build-capz-image-1.24.10-r4e901b36c57fb10aaa0c9e5854a979c2a-pod step-build-image 
build-capz-image-1.24.10-r4e901b36c57fb10aaa0c9e5854a979c2a-pod step-build-image 1 error(s) occurred:
build-capz-image-1.24.10-r4e901b36c57fb10aaa0c9e5854a979c2a-pod step-build-image 
build-capz-image-1.24.10-r4e901b36c57fb10aaa0c9e5854a979c2a-pod step-build-image * The managed_image_storage_account_type "Standard_ZRS" is invalid
```


images/capi/packer/azure/packer.json
```diff
@@ -254,6 +254,7 @@
     "kubernetes_semver": null,
     "kubernetes_series": null,
     "kubernetes_source_type": null,
+    "managed_image_storage_account_type": "",
```
AverageMarcus, line 257: To ensure this doesn't break for anyone it might be worth defaulting this to the `storage_account_type` value.
```suggestion
    "managed_image_storage_account_type": "{{user `storage_account_type`}}",
```

primeroz: make sense , thanks :+1:

mboersma (approved):
/lgtm
/approve

PR #1090, opened 2023-03-02
add /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg to all Azure VMs

nawazkh:
- this applies the below struct to enable Azure CNI on CAPZ workloads created on Azure
```
datasource:
  Azure:
    apply_network_config: false
```

- This PR is a part of the PR https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3204 solving https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/467

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+++ b/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
@@ -1,0 +1,16 @@
+## We need the below configuration on Azure flavor VMs to enable DHCP on eth0
+## instead of Azure's IMDS endpoint. This configuration supports CNIs that
+## use extra available IPs on the host to assign them to the scheduled pods.
+## CNIs operating in overlay mode do not get affected by this configuration update.
+## We apply this configuration to the VM before its first boot, and not via
+## cloud-init's VM user data(customData), because network will get setup before reading
+## customData. Please refer for more context:
+## - https://github.com/kubernetes-sigs/image-builder/pull/1090#issuecomment-1468552870
+## - https://github.com/kubernetes-sigs/image-builder/pull/1090#issuecomment-1462944511
+##
+## Also refer to the PR adding this configuration for complete history and conversation.
+## - https://github.com/kubernetes-sigs/image-builder/pull/1090
+##
+datasource:
+  Azure:
+    apply_network_config: false
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -65,3 +65,14 @@
     - netbase
     - nfs-common
   when: ansible_os_family == "Debian"
+
+## refer to ../files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+## for more context on below file addition
+- name: Create azure-vnet config file
+  copy:
+    src: files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+    dest: /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+    owner: root
+    group: root
+    mode: 0644
+  when: ansible_os_family == "Debian"
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -324,6 +324,12 @@ ubuntu:
       iptables -C FORWARD -d 168.63.129.16/32 -p tcp -m tcp --dport 80 -m comment --comment "block traffic to 168.63.129.16 for cve-2021-27075" -j DROP:
         exit-status: 0
         timeout: 0
+      grep -z -q -E '^datasource:\n\s\sAzure:\n\s\s\s\sgiapply_network_config:\sfalse$' /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg || echo $?:
+        exit-status: 0
+        stdout: []
+        stderr: []
+        timeout: 0
+        title: "Check exact for the contents in /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg"
     package:
       open-vm-tools:
       linux-cloud-tools-virtual:


jsturtevant:
Is this a configuration file we want on every node regardless of CNI?

nawazkh:
> Is this a configuration file we want on every node regardless of CNI?

I will be testing this out using Calico to confirm the cluster is up and running with the above configurations.

nawazkh:
I was able to create the workload cluster with the above configuration and with Calico as CNI. Everything look good so far.

Poking around on testing the workload cluster created with the above configuration.

images/capi/ansible/roles/providers/tasks/azure.yml
```diff
@@ -65,3 +65,12 @@
...
   when: ansible_os_family == "Debian"
+
+- name: Create azure-vnet config file
+  copy:
+    src: files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+    dest: /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
+    owner: root
+    group: root
+    mode: 0644
+  when: ansible_os_family == "Debian" # TODO: Should this be on other OS_Families too ?
```
nawazkh, line 76: I am not clear about `when: ansible_os_family == "Debian"`. 
Should we be including more OS families in here? 
Or maybe add `ansible_distribution == "Microsoft"` ?

jsturtevant:
> I was able to create the workload cluster with the above configuration and with Calico as CNI. Everything look good so far.
> 
> Poking around on testing the workload cluster created with the above configuration.

I am more wondering if this is the right place to set up configuratoin for a specific CNI.  Should the CNI be setting this up? I am not aware of other CNI specific configuration that we put on the node, typically they are responsible for their own configuration and set up.  

We do drop the [basic set of CNI's on the node](https://github.com/kubernetes-sigs/image-builder/blob/c5d66c378e0b03047865d27cf2ee33d5966dd1a3/images/capi/ansible/roles/kubernetes/tasks/url.yml#L23-L31) though we don't set up the configuration for them. 

this maybe this is ok, just wondering if it is the right place.

images/capi/ansible/roles/providers/tasks/azure.yml
jsturtevant: This depends on if we want to include it with Flatcar and if we can do it this way for flatcar or need a different approach

jsturtevant:
> This looks good overall. We should see if we can also support Flatcar, but IDK if that file location is writable on that distro.

maybe @invidian knows?

images/capi/ansible/roles/providers/tasks/azure.yml
nawazkh: What about `ansible_os_family = RedHat` ? Is it similar to `Debian` ?

invidian:
>>This looks good overall. We should see if we can also support Flatcar, but IDK if that file location is writable on that distro.

>maybe @invidian knows?

`/etc` is writable, yes. Mainly `/usr` is read-only on Flatcar.

>I am more wondering if this is the right place to set up configuration for a specific CNI. Should the CNI be setting this up? I am not aware of other CNI specific configuration that we put on the node, typically they are responsible for their own configuration and set up.

If this file is only for Azure CNI, I don't think it should be added to _all_ image-builder based images. For Azure-specific images, _maybe_ it make sense, but still it feels to me that this should be done by extending Node initialization configuration (e.g. via extra files via cloud-init or Ignition), instead of being done in image-builder. image-builder currently installs `kubeadm` and related binaries, suggesting that the default implied bootstrap provider is CABPK, which seems fine for reference, but I do see that other bootstrap providers would like to provide their own binaries into the images. The same applies to CNIs, although it's already one level higher in hierarchy and actually CNIs are not required to form a cluster.

To summarize, I have a feeling that image-builder is not the right place to put this configuration, but I might be wrong of course.

>We do drop the [basic set of CNI's on the node](https://github.com/kubernetes-sigs/image-builder/blob/c5d66c378e0b03047865d27cf2ee33d5966dd1a3/images/capi/ansible/roles/kubernetes/tasks/url.yml#L23-L31) though we don't set up the configuration for them.

I think those are different, as they are required by kubelet and Kubernetes, as described in https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#network-plugin-requirements.

jsturtevant:
> To summarize, I have a feeling that image-builder is not the right place to put this configuration, but I might be wrong of course.

I tend to agree, this should be handled by the CNI itself IMO

CecileRobertMichon:
> I tend to agree, this should be handled by the CNI itself IMO

This is a tough one because AFAIK it needs to be applied before the first boot, and we have no way of telling which CNI will be used with which image, unless we start building images for each CNI, but then that explodes the matrix.

FWIW, AKS Engine applies this config on all VMs, regardless of CNI: https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_install.sh#L257, called here https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_main.sh#L85

What would be the alternative?

nawazkh:
Meanwhile, I am testing if I can patch a node's `cloud.cfg.d` using a daemonset. 
I had tested this earlier and did not achieve desired results, but this time, I am trying to see if I can force reboot the node on which the pod was scheduled.

jsturtevant:
>This is a tough one because AFAIK it needs to be applied before the first boot

this could be reason enough. Do we know what this configuration does? Are there any side affects?  I found this https://bugs.launchpad.net/cloud-init/+bug/1818302 which suggests it isn't really CNI related https://github.com/Azure/aks-engine/issues/761

> unless we start building images for each CNI, but then that explodes the matrix.

definitely not suggesting this.  On the flipside, I also think it's not great to start bringing in CNI configuration for every CNI to image-builder

> What would be the alternative?

I was thinking the DS that installs the CNI could apply the configuration but sounds like @nawazkh already tried that and it didn't work.  If I understand correctly this looks like it is [cloud-int configuration ](https://cloudinit.readthedocs.io/en/latest/reference/network-config.html) so this makes sense.

jsturtevant:
found the docs on this: https://cloudinit.readthedocs.io/en/latest/reference/datasources/azure.html#configuration

CecileRobertMichon:
> I was able to create the workload cluster with the above configuration and with Calico as CNI

@nawazkh did you build an image from this branch and test that you were able to build a cluster while using that image for all VMs? If not that would be a good exercise

nawazkh:
> @nawazkh did you build an image from this branch and test that you were able to build a cluster while using that image for all VMs? If not that would be a good exercise

Hmm, I haven't built a cluster off of this branch. Will try!
Earlier, I was hinting at applying patch and rebooting the VM and then applying Calico. The workload cluster came up just fine.

nawazkh:
> Meanwhile, I am testing if I can patch a node's `cloud.cfg.d` using a daemonset.
> I had tested this earlier and did not achieve desired results, but this time, I am trying to see if I can force reboot the node on which the pod was scheduled.

TL;DR, could not find an easier way to reboot a VM/node. 

I poked around a bit using reboot from a script but that did not reboot the node per se. 
The other way is to cordon, drain and reboot a node using a script running on the master/control plane node. I think going that route would be quite a hack unless you all think otherwise.

nawazkh:
I also tried using Calico CNI and with `apply_network_config:false` to each of the VMs of the workload cluster and I was able  get the cluster is up and running. No errors showed per se. But as Cecile suggested, I will try testing out the cluster using this branch's image.

images/capi/ansible/roles/providers/tasks/azure.yml
CecileRobertMichon: this TODO still needs to be resolved

nawazkh:
@invidian @jsturtevant @CecileRobertMichon , Apologies for making this PR sound very specific to Azure CNI. While this PR does help enabling Azure CNI, it is also a configuration replication on Azure VMs like AKS engine does to unblock customers wanting to assign the extra IPConfigs on their VMs to the scheduled pods.

#### Further explanation and Summary:
##### How does this PR help Azure CNI ?
- Without this PR, `apply_network_config` will be set to `true` by default, meaning the VM will be using the network configuration described by Azure's IMDS endpoint instead of DHCP on `eth0`. Due to this, when the user creates multiple IPConfigs on a VM, the extra IPConfigs created get assigned to the host OS. We need these extra IPConfigs to be assigned to the pods scheduled on the node instead. (Thanks to @dthorsen on pointing this out!)
- Since Azure CNI is one of the CNIs which uses the available extra IPConfig to assign it to the pods, this PR helps Azure CNI.
- A user could also run into this issue if they were using any other CNI; CNIs which assign extra IPs created in the VM to the scheduled pods. 

#### Why `apply_network_config:false` ? 
- Tracing the history, `apply_network_config:false` was first set from Ubuntu `18.x` distor due to https://github.com/Azure/aks-engine/issues/761 via the PR https://github.com/Azure/aks-engine/pull/223. 
- Also please take a look at [the workaround suggested by the _az linux team_](https://github.com/Azure/aks-engine/pull/223/files#r266089530)
  - Moreover, as @CecileRobertMichon pointed out, AKS engine is doing the same regardless of the CNI. 
  > AKS Engine applies this config on all VMs, regardless of CNI: https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_install.sh#L257, called here https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_main.sh#L85

#### What about testing ? 
- I tested the images generated off of this branch using Azure CNI and Calico, and I did not notice any disruption in bringing up the workload cluster from CAPZs perspective. The core-dns pods were up and running, and the cluster was provisioned as expected.

#### Concerns and Consensus?
- The concerns around updating network config at the image level instead of achieving it via a daemonset on using Azure CNI (or a similar CNI) is valid. I will follow up with Azure CNI via an issue and also link it here for reference.
- Concensus : [Team's discussion](https://teams.microsoft.com/l/message/19:883ad42d34d148f1b41b2f730db17c4f@thread.skype/1677878005706?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e121dbfd-0ec1-40ea-8af5-26075f6a731b&parentMessageId=1677878005706&teamName=Azure%20Container%20Compute&channelName=Upstream%20Infra&createdTime=1677878005706&allowXTenantAccess=false)

kkeshavamurthy:
getting some networking folks in vmware to take a look at this change to see if it breaks any of our stuff. 
/hold

jsturtevant:
Thanks @nawazkh for the detailed notes!

rbtr:
> I tend to agree, this should be handled by the CNI itself IMO

I disagree - this is not something that can be handled by the CNI, as it is boot-time config. The CNI rebooting the Node would be extremely aberrant behavior. IMO this should be handled by the infra orchestrator...I consider it similar to the prereq that `net.ipv4.ip_forward` is set in the kernel.

I suspect that `apply_network_config:false` should be the default and that other CNIs, especially any CNI operating in overlay mode, just don't care about this config.

dthorsen:
I came here to say the same thing as @rbtr. The impact of this config to other CNI's should be nil. Also, regardless of the CNI in use, this setting being enabled in Azure results in undesired behavior for any machine that joins with multiple IPConfigs. Notably it was not possible to join a CAPZ AzureMachine to a cluster with multiple IPConfigs until recently, so this is only coming up now.

The way the Azure CNI is designed, the additional IPs must _not_ be in use by the host OS in order for them to be available for assignment to pods. Unfortunately with this setting on, early during boot the host OS grabs all the available VM IPs and assigns them to the host OS. Then notably, if this is done on a control plane machine, the 2nd or 3rd control plane machine will fail to bootstrap, as node to node network communication may occur over any of the IPs. etcd does not like receiving connections from member IPs it doesn't expect.

Given that it wasn't even possible to build CAPZ nodes that had multiple IPConfigs until very recently, there should be no cases of people depending on this behavior anyway.

kkeshavamurthy:
Thanks for the explanation everyone. 
/hold cancel

invidian:
Before we merge, has this been checked if it's possible to deliver `apply_network_config:false` via cloud-init?

invidian:
>@invidian That would be true for many (most) cloud-init configs. However, for networking related things, my understanding is that networking must be up in order for cloud-init to reach the IMDS and fetch the customData in order to parse it and execute. So no unfortunately, this would mean the network configuration is already applied to the host by the time any cloud-init config we supply via customData is run.

Thanks. I guess that make sense. Not a great design though... What do you think about documenting this reasoning as a comment in the code, so it's easier to find why such specific thing is there? It seem like a good idea to me.

Regardless, I think the patch should still be changed to only apply to Azure flavors, not all VMs.

I would also check how this should be handled by Flatcar, which does not use cloud-init at all.

Finally, should we also have some validation to make sure this patch does what it suppose to do?

CecileRobertMichon:
> Thanks. I guess that make sense. Not a great design though... What do you think about documenting this reasoning as a comment in the code, so it's easier to find why such specific thing is there? It seem like a good idea to me.

+1

> Finally, should we also have some validation to make sure this patch does what it suppose to do?

we should be able to add to https://github.com/devigned/image-builder/blob/8be13ef6192f88926270d39cb5247bcb308c9cd6/images/capi/packer/goss/goss-vars.yaml#L157 to test it in image-builder, then the actual e2e behavior with multiple IPs will be tested in e2e in CAPZ (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3204) 

> I would also check how this should be handled by Flatcar, which does not use cloud-init at all.

Let's keep this scoped to Azure Ubuntu VMs for now, and we can opt-in Flatcar later if needed?

jsturtevant:
Thanks for everyone's input. It is now clear to me this isn't an Azure CNI configuration only, but more documentation in the comments makes sense (maybe with a link to the discussion here too)

This patch does apply only to Azure VM's since it is in the provider Azure file.  

Adding a goss test is a great idea. @nawazkh can you add one? let me know if you need help figuring that out. It can be a bit overwhelming

nawazkh:
Thank you everyone for pitching in their thoughts and perspectives I have added a comment in the patch file, please take a look to improve its appropriateness. I will squash the commits before merging.

> I would also check how this should be handled by Flatcar, which does not use cloud-init at all.

Is there a way I can test this manually? Do you have any pointers on testing flatcar images for this purpose?

> Adding a goss test is a great idea. @nawazkh can you add one?

+1, thats a great idea! I will reach out to you in a bit to get more context on goss test.

invidian:
>Is there a way I can test this manually? Do you have any pointers on testing flatcar images for this purpose?

I guess you can use `flatcar` flavor of CAPZ with Azure CNI for testing? And if you want to use custom-built images, you need to change the template to use the SIG image you built.

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -324,6 +324,11 @@ ubuntu:
       iptables -C FORWARD -d 168.63.129.16/32 -p tcp -m tcp --dport 80 -m comment --comment "block traffic to 168.63.129.16 for cve-2021-27075" -j DROP:
         exit-status: 0
         timeout: 0
+      grep -z -q -E '^datasource:\n\s\sAzure:\n\s\s\s\sgiapply_network_config:\sfalse$' /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg && test $? -eq 0 && echo "DHCP on eth0 is used" || echo "DHCP is not used":
```
jsturtevant, line 327: Is there a reason to use a [command ](https://github.com/goss-org/goss/blob/master/docs/manual.md#command)vs using the [goss file helper](https://github.com/goss-org/goss/blob/master/docs/manual.md#file)?

jsturtevant: There is a section that should work here https://github.com/kubernetes-sigs/image-builder/blob/2011a94f61f3ddc5d638481a328c49e0dcf5238c/images/capi/packer/goss/goss-files.yaml. This is for Windows but I believe it should work the same way with ubuntu: https://github.com/kubernetes-sigs/image-builder/blob/2011a94f61f3ddc5d638481a328c49e0dcf5238c/images/capi/packer/goss/goss-vars.yaml#L472-L473

nawazkh: No specific reason to use `command` over `goss file helper`.

1. I did not know about `goss file helper`, but I will probe into it. I am trying to exactly match the contents of the file with the given regex, paying attention to the number of spaces between the strings. I wonder if I can regex match the contents of the file using `goss file helper` too.
1. I was hoping to check for contents of the 15_azure-vnet.cfg file and match it exactly with 
```yaml
datasource:
   Azure:
     apply_network_config: false
```

jsturtevant: This looks like it is working now. I don't have any particular reason to use the file helper vs grep and it looks like we've used grep in other places. The file helper was added for windows and doesn't look adopted for Linux yet.  

The works for me. @mboersma any objections?

jsturtevant:
> I would also check how this should be handled by Flatcar, which does not use cloud-init at all.

What does it use? is there a similar setting that could be used by flatcar?

Is this something we want to block this PR on?

nawazkh:
@invidian Do you think we could let this PR merge for "Debian" flavor under Azure offering and follow up with another issue for flatcar image ?

invidian:
:+1: for what @mboersma said. Is there plans for CAPZ to use Azure CNI as a default? In such case, we would have to make sure Flatcar is supported as well.

Perhaps an issue in https://github.com/flatcar/flatcar can be created about this as well, to let team know about this work and possibly have some support.

CecileRobertMichon:
> Is there plans for CAPZ to use Azure CNI as a default

no. In fact there is no "default" when it comes to CNIs as Cluster API infra providers don't install CNI, it's up to the user to choose and install a CNI.

nawazkh:
> Is there plans for CAPZ to use Azure CNI as a default?

I don't think we plan on broadcasting Azure CNI as default, we rather ask the community to bring in their own CNI. For the most part as per our documentation, we use Calico CNI(which uses Overlay network) with CAPZ. That being said we don't really favor it.
This PR mostly facilitates folks willing to use Azure CNI. 
(@CecileRobertMichon @mboersma Please correct me if I am wrong)

--- 

> Perhaps an issue in https://github.com/flatcar/flatcar can be created about this as well, to let team know about this work and possibly have some support.

I dont really know if flatcar needs such a change. I will test and confirm if Azure CNI doesn't work with flatcar.

images/capi/ansible/roles/providers/tasks/azure.yml
nawazkh: I removed the TODO in favor of testing flatcar images and following it up via separate issue if needed.

CecileRobertMichon (commented):
/lgtm

nawazkh:
Three of the tests seem to be failing
```
photon-3: FAILED. See logs in the artifacts folder.
photon-4: FAILED. See logs in the artifacts folder.
ubuntu-2004: FAILED. See logs in the artifacts folder.
```

I haven't touched `photon` related configs in this PR. 
From `ubuntu-2004` artifact, I see the below error
```
[1;32m==> vsphere-iso.vsphere: Clear boot order...[0m
[1;32m==> vsphere-iso.vsphere: Power off VM...[0m
[1;32m==> vsphere-iso.vsphere: Deleting Floppy image ...[0m
[1;32m==> vsphere-iso.vsphere: Destroying VM...[0m
[1;31mBuild 'vsphere-iso.vsphere' errored after 32 minutes 28 seconds: Timeout waiting for IP.[0m

==> Wait completed after 32 minutes 28 seconds

==> Some builds didn't complete successfully and had errors:
--> vsphere-iso.vsphere: Timeout waiting for IP.

==> Builds finished but no artifacts were created.
make: *** [Makefile:416: build-node-ova-vsphere-ubuntu-2004] Error 1
```

This looks more of a timeout error to me, so will be retesting it. Can someone check if this is really a timeout error? @jsturtevant

kkeshavamurthy:
@nawazkh 
#1121 should fix the photon ci failures. Please rebase.

nawazkh:
> @nawazkh
> #1121 should fix the photon ci failures. Please rebase.

Waiting on the above PR to be merged..

nawazkh:
> I dont really know if flatcar needs such a change. I will test and confirm if Azure CNI doesn't work with flatcar.

We ran into some some errors while testing Azure CNI with flatcar. That is being tracked at  https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/3387.

That being said. Flatcar related changes can be achieved (if needed) via another PR and flatcar (related changes) should not necessarily be blocking this PR. This PR is for Azure's Debian offering only :) 

Will wait for PR to go green !

nawazkh:
PRs green! Can @jsturtevant @CecileRobertMichon @invidian please give it a look?

images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/15_azure-vnet.cfg
```diff
@@ -0,0 +1,16 @@
+## We need the below configuration on Azure flavor VMs to enable DHCP on eth0
```
jsturtevant, line 1: Thanks for leaving the context here 💯

Issue #991, opened 2022-10-05
cluster-node-image-builder container fails in Azure pipeline


mboersma:
**What steps did you take and what happened:**

When running Azure image-builder tasks from this repository in pipelines, we want to use the container produced by `make -C images/capi docker-build` as recommended in the [Image Builder book](https://image-builder.sigs.k8s.io/capi/container-image.html#examples). But in practice it errors out, as shown in the below log.

This same startup script with `useradd` runs successfully when backed by the old [deis/go-dev](https://github.com/deis/docker-go-dev) container.

**What did you expect to happen:**

**Anything else you would like to add:**

```
Starting: Initialize containers
/usr/bin/docker version --format '{{.Server.APIVersion}}'
'1.41'
Docker daemon API version: '1.41'
/usr/bin/docker version --format '{{.Client.APIVersion}}'
'1.41'
Docker client API version: '1.41'
/usr/bin/docker ps --all --quiet --no-trunc --filter "label=d03b08"
/usr/bin/docker network prune --force --filter "label=d03b08"
/usr/bin/docker pull k8s.gcr.io/scl-image-builder/cluster-node-image-builder-amd64:v0.1.13
v0.1.13: Pulling from scl-image-builder/cluster-node-image-builder-amd64
675920708c8b: Pulling fs layer
...
b955c67dff1f: Pull complete
Digest: sha256:cd095a2718006fffeced8acdf296e65b108999e1c742672d6dea83ee45c6e060
Status: Downloaded newer image for k8s.gcr.io/scl-image-builder/cluster-node-image-builder-amd64:v0.1.13
k8s.gcr.io/scl-image-builder/cluster-node-image-builder-amd64:v0.1.13
/usr/bin/docker info -f "{{range .Plugins.Network}}{{println .}}{{end}}"
bridge
host
ipvlan
macvlan
null
overlay
/usr/bin/docker network create --label d03b08 vsts_network_14e080230a5f4d178a8f7af4a7ce14d9
063dbf403c87b9cf2d4fef33f83d6c772495242ce91f2baf87fe13b254efacb7
/usr/bin/docker inspect --format="{{index .Config.Labels \"com.azure.dev.pipelines.agent.handler.node.path\"}}" k8s.gcr.io/scl-image-builder/cluster-node-image-builder-amd64:v0.1.13
[...]

PR #1092, opened 2023-03-03
docker: support qemu target builds

tormath1:
this commit adds qemu amd64 binaries to build qemu images from the Docker image.

example:
```
docker run -it --net host --rm -e OEM_ID=openstack cluster-node-image-builder-amd64 build-qemu-flatcar
```

_Note_: For now it can only run builds without acceleration because it still uses the `imagebuilder` user but as `imagebuilder` needs to be in the KVM group and we don't know in advance the group ID of `kvm` group on the host: we would need to build and run the image with `root` user to allow KVM acceleration, this can be done in a follow-up PR.

**Additional context**

We could preserve the `imagebuilder` user by creating it with the uid/gid 0 but in the end it's creating another "root" user...

### DIFF
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -17,7 +17,20 @@
 ARG BASE_IMAGE=docker.io/library/ubuntu:latest
 FROM $BASE_IMAGE
 
-RUN apt-get update && apt-get install -y apt-transport-https build-essential ca-certificates curl git jq python3-pip rsync unzip vim wget \
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+	apt-transport-https \
+	build-essential \
+	ca-certificates \
+	curl \
+	git \
+	jq \
+	python3-pip \
+	rsync \
+	unzip \
+	vim \
+	wget \
+	qemu-system-x86 \
+	qemu-kvm \
     && useradd -ms /bin/bash imagebuilder \
     && apt-get purge --auto-remove -y \
     && rm -rf /var/lib/apt/lists/*


images/capi/Dockerfile
```diff
@@ -17,26 +17,22 @@
...
-RUN apt-get update && apt-get install -y apt-transport-https build-essential ca-certificates curl git jq python3-pip rsync unzip vim wget \
-    && useradd -ms /bin/bash imagebuilder \
+RUN apt-get update && apt-get install -y apt-transport-https build-essential ca-certificates curl git jq python3-pip rsync unzip vim wget qemu-system-x86 \
     && apt-get purge --auto-remove -y \
     && rm -rf /var/lib/apt/lists/*
 
 ARG ARCH
 ARG PASSED_IB_VERSION
 
-USER imagebuilder
```
mboersma, line 28: I would also like to use the default `root` user in this image, because we would like to use this container image in our Azure DevOps pipelines and that's a requirement. (See #991.)

But using a specific non-root user in a Docker container is still generally considered a "best practice" I believe, so I'd like to hear from others as to whether this change is acceptable.

jsturtevant: Can we default it but make it overridable for the use cases it is needed for?

images/capi/Dockerfile
```diff
@@ -17,26 +17,22 @@
 ARG BASE_IMAGE=docker.io/library/ubuntu:latest
 FROM $BASE_IMAGE
 
-RUN apt-get update && apt-get install -y apt-transport-https build-essential ca-certificates curl git jq python3-pip rsync unzip vim wget \
-    && useradd -ms /bin/bash imagebuilder \
+RUN apt-get update && apt-get install -y apt-transport-https build-essential ca-certificates curl git jq python3-pip rsync unzip vim wget qemu-system-x86 \
```
mdbooth, line 20: could we also add qemu-kvm?

mdbooth (commented):
I think I'd prefer to see the user changes in a separate PR. Could we reduce this one to just adding qemu?

mboersma (commented):
/lgtm
/assign @jsturtevant @kkeshavamurthy

kkeshavamurthy:
maybe someone in the qemu approvers list need to approve this?

mdbooth:
Pretty sure I can't, but

/approve

mdbooth:
Despite what the bot says, I think it needs an approver from https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/OWNERS, which is just `cluster-api-maintainers`, i.e.:

@CecileRobertMichon
@enxebre
@fabriziopandini
@sbueringer
@vincepri

mboersma:
>  I think it needs an approver from https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/OWNERS

This really shouldn't be the case, see
https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners
We don't set `no_parent_owners: false`, so I think it should be incorporating the root OWNERS/OWNERS_ALIASES files. And in fact, it always has done and it's only recently that this behavior seems to have changed.

sbueringer:
Yup top-level approval should 100% be enough. Can we open an issue in e.g. test-infra to get some awareness / maybe someone has time to investigate?

Let's see if that helps for now
/approve

Issue #1080, opened 2023-02-23
Ubuntu-2204, qemu - Cloud-init configuration is disabled by the subiquity autoinstaller 


matofeder:
Subiquity autoinstaller does not allow further Ubuntu-2204, qemu image configuration by the cloud-init NoCloud or Cloud datasources. 

**What steps did you take and what happened:**

Steps to reproduce the issue:

1. Build qemu Ubuntu 22.04
```bash
make build-qemu-ubuntu-2204
```
2. Create custom cloud-init config e.g.:
```bash
cat >user-data <<EOF
#cloud-config
users:
  - name: test
    plain_text_passwd: test
    groups: sudo
    lock-passwd: false
    sudo: ALL=(ALL) NOPASSWD:ALL
    shell: /bin/bash
EOF
```
3. Create a local disc from the `user-data` file
```bash
cloud-localds seed.img user-data
```
4. Create a new virtual machine using the `seed.img` and ubuntu22.04, e.g.: 
```
virt-install \
  --name ubuntu2204_cloud_init_test \
  --memory 2048 \
  --vcpus 2 \
  --disk ubuntu-2204-kube-v1.23.15,device=disk,bus=virtio \
  --disk seed.img,device=cdrom \
  --os-variant ubuntu22.04 \
  --noautoconsole \
  --network network=default,model=virtio \
  --import
```

5. User `test` which should be created by `cloud-init` via `seed.img` is not there

What happened:

Note: 
   Inspection of the ubuntu22.04 image has been done using an unlocked `builder` user:
```diff
-      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'usermod -L {{user `ssh_username`}} && {{user `shutdown_command`}}'",
+      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c '{{user `shutdown_command`}}'",
``` 

1. Subiquity autoinstaller creates cloud-init `99-installer.cfg ` file that restricts data sources only to `None`

```bash
root@localhost:~# cat /etc/cloud/cloud.cfg.d/99-installer.cfg 
datasource:
  None:
    metadata:
      instance-id: 35edaee4-d23a-4a90-8afd-27d827d25cb4
    userdata_raw: "#cloud-config\ngrowpart:\n  mode: 'off'\nlocale: en_US.UTF-8\n\
[...]

PR #1105, opened 2023-03-11
Adding vmware datasource configuration when building ubuntu 22.04 OVA

### LINKED ISSUE #1104: Add the datasource configuration when building Ubuntu 22.04 OVA's
When creating clusters on the vSphere environment using Ubuntu 22.04 OVA the datasource for `cloud-init` is [nocloud](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html) instead of [vmware](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html) this is because [ds-identify](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L837) script checks if there is any data source present in the kernel command, this causes Cloud-Init to pick NoCloud datasource as we are using [Auto Install](https://ubuntu.com/server/docs/install/autoinstall) from Ubuntu 22.04, so need to add the `ds-identify.cfg` in `/etc/cloud/` with `datasource: VMware` to set the appropriate cloud data source.
```
  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/22.04.efi/'<enter><wait>initrd /casper/initrd<enter><wait>boot<enter>",
```

```
dscheck_NoCloud() {
    local fslabel="cidata CIDATA" d=""
    case " ${DI_KERNEL_CMDLINE} " in
        *\ ds=nocloud*) return ${DS_FOUND};;
    esac
    case " ${DI_DMI_PRODUCT_SERIAL} " in
        *\ ds=nocloud*) return ${DS_FOUND};;
    esac
```

Along with this need to remove the below could init file as per this [KB](https://kb.vmware.com/s/article/80934) [Similar issue at #1080 ]
- `/etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg`
- `/etc/cloud/cloud.cfg.d/99-installer.cfg`

The cloud-Init data source should be `VMware` instead of `NoCloud`

**Environment:**

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi)):

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`): Ubuntu 22.04

/kind bug

DimpleRajaVamsi:
- This PR adds the `ds-identify.cfg` with the datasource as VMware when building Ubuntu 22.04 OVA this is required to address the issue mentioned in #1104 
- Removes the cloud-init config files as per [KB](https://kb.vmware.com/s/article/80934)
- Adds the ability to disable ipv6 during the installation similar to #648 where it disabled ipv6 during boot for Ubuntu 20.04

### Testing
- Created Ubuntu 20.04 and Ubuntu 22.04 OVA
- Verified the OVA by creating Kubernetes clusters.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1104 

**Additional context**
While I removed some files related to the cloud-init configurations as per [KB](https://kb.vmware.com/s/article/80934), this is also affecting other clouds as per #1080.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
@@ -43,3 +43,11 @@
     - { src: files/etc/networkd-dispatcher/routable.d/20-chrony.j2, dest: /etc/networkd-dispatcher/routable.d/20-chrony }
     - { src: files/etc/networkd-dispatcher/off.d/20-chrony.j2, dest: /etc/networkd-dispatcher/off.d/20-chrony }
     - { src: files/etc/networkd-dispatcher/no-carrier.d/20-chrony.j2, dest: /etc/networkd-dispatcher/no-carrier.d/20-chrony }
+
+- name: Create cloud-init datasource config file
+  copy:
+    dest: /etc/cloud/ds-identify.cfg
+    force: true
+    content: |
+      datasource: VMware
+  when: ansible_distribution_version is version('22.04', '>=')
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -93,3 +93,15 @@
   file:
     state: absent
     path: /etc/udev/rules.d/70-persistent-net.rules
+
+- name: Remvoing subiquity disable disable cloud-init networking config
+  file:
+    path: /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg
+    state: absent
+  when: ansible_distribution_version is version('22.04', '>=')
+
+- name: Removing 99-installer.cfg which sets the cloud-init datasource to None
+  file:
+    path: /etc/cloud/cloud.cfg.d/99-installer.cfg
+    state: absent
+  when: ansible_distribution_version is version('22.04', '>=')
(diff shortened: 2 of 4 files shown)


images/capi/ansible/roles/providers/tasks/vmware-ubuntu.yml
```diff
@@ -43,3 +43,23 @@
...
+  file:
+    path: /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg
+    state: absent
+  when: ansible_distribution_version is version('22.04', '>=')
+
+- name: Removing 99-installer.cfg which sets the cloud-init datasource to None
+  file:
+    path: /etc/cloud/cloud.cfg.d/99-installer.cfg
+    state: absent
+  when: ansible_distribution_version is version('22.04', '>=')
```
kkeshavamurthy, line 57: Are these 2 files need for the firstbootup after the installation? If not, we should prob remove them in the installer step [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ova/linux/ubuntu/http/22.04.efi/user-data#L108)
@matofeder Does this solve #1080?

kkeshavamurthy: I should've worded it better. 
Glad that the proposed soln when implement for qemu solves #1080 . 
@DimpleRajaVamsi let's move this file removal to the installer step if you are ok with it.

matofeder: Hmm, it seems that removing files in the installer step, did not help. Subiquity probably creates these files after [late-commands](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ova/linux/ubuntu/http/22.04.efi/user-data#L103). Anyway, the latest version in this PR ad5bdb2329f61f020962f1738580deda5b4152a0 works also for **qemu** build and fixes #1080 (tested).

DimpleRajaVamsi: Thanks, @matofeder for testing. 
@kkeshavamurthy as @matofeder mentioned I hit an issue when I am removing the files in the installer step so moved the removal to [sysprep](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi/ansible/roles/sysprep) role.

PR #1107, opened 2023-03-13
Check for kube-proxy before building Azure Windows image

### LINKED ISSUE #1106: Check for Windows kube-proxy images during Azure VHD build pipeline

Today (and unfortunately) the Windows kube-proxy image used by CAPI/CAPZ clusters is built and published out of https://github.com/kubernetes-sigs/sig-windows-tools.
This means that it is possible to build and publish new reference VHDs without having corresponding kube-proxy images for a given K8s patch version which can result in cluster provisioning failures or unpredictable behavior.

Until SIG-Windows has automated building / publishing of new kube-proxy images (which depends on work being doing in https://github.com/kubernetes-sigs/windows-service-proxy) we should add a check to the pipelines that build Windows VHDs to ensure that a corresponding kube-proxy image is available for the K8s patch version the VHD is being built for.

**Additional context**

We have a github action at https://github.com/kubernetes-sigs/sig-windows-tools/actions/workflows/test-kube-proxy-images.yaml that checks to make sure we have kube-proxy images published as new patch versions come out.

/kind feature
/sig windows
/cc @mboersma @jsturtevant

mboersma:

Adds a pre-flight check for the existence of the Windows kube-proxy image, before building and publishing the correlated Windows Kubernetes reference image.

Which issue(s) this PR fixes: 

Fixes #1106

**Additional context**

### DIFF
--- a/images/capi/packer/azure/.pipelines/build-vhd.yaml
+++ b/images/capi/packer/azure/.pipelines/build-vhd.yaml
@@ -20,6 +20,19 @@ jobs:
   steps:
   - template: k8s-config.yaml
   - script: |
+      set -euo pipefail
+      kube_proxy_url="sigwindowstools/kube-proxy:v${KUBERNETES_VERSION/+/_}-calico-hostprocess"
+      echo "Checking for Windows kube-proxy image $kube_proxy_url"
+      if ! stderr="$(docker pull $kube_proxy_url 2>&1 > /dev/null)"; then
+        # It's a Windows image, so expect an error after pulling it on Linux
+        if [[ $stderr != *"cannot be used on this platform"* ]]; then
+          echo "Failed to pull kube-proxy image: $stderr"
+          exit 1
+        fi
+      fi
+    displayName: Checking for Windows kube-proxy image
+    condition: and(eq(variables['PREFLIGHT_CHECKS'], 'true'), eq(variables['OS'], 'Windows'))
+  - script: |
       set -o pipefail
       make deps-azure
       os=$(echo "${OS}" | tr '[:upper:]' '[:lower:]')


mboersma:
/approved cancel

(I don't like this setting where maintainers' PRs are already approved, maybe we should change it?)

images/capi/packer/azure/.pipelines/build-vhd.yaml
```diff
@@ -19,6 +19,20 @@ jobs:
     name: $(BUILD_POOL)
   steps:
   - template: k8s-config.yaml
+  - script: |
+      set -euo pipefail
+      if [ "$OS" == "Windows" ]; then
+        kube_proxy_url="sigwindowstools/kube-proxy:v${KUBERNETES_VERSION/+/_}-calico-hostprocess"
+        echo "Checking for Windows kube-proxy image $kube_proxy_url"
+        if ! stderr="$(docker pull $kube_proxy_url 2>&1 > /dev/null)"; then
```
willie-yao, line 27: In this line, are we checking if there is no error with the docker pull? If so, would the next if statement always be skipped since there isn't an error?

Also, will this return the correct error in the case where kube-proxy doesn't pull correctly? I'm definitely misunderstanding something here but I just wanted to make sure!

mboersma: Yes, if `docker pull` succeeds, we should skip the rest. Also if `docker pull` fails but with that specific "wrong platform" error, it's a success. If `docker pull` fails with any other error, the script fails.

I tested these variations in a DevOps pipeline and it seems to be working as expected.

marosset (approved):
/lgtm

images/capi/packer/azure/.pipelines/build-vhd.yaml
jsturtevant: > Yes, if docker pull succeeds, we should skip the rest.

why? If we are expecting an error and it pulls correctly shouldn't we fail?

willie-yao (approved):
/lgtm

images/capi/packer/azure/.pipelines/build-vhd.yaml
mboersma: I guess I'm assuming that this script could be run on Windows platform, or that the image might evolve to be multi-architecture. Then the `docker pull` would pass, but otherwise we expect the "wrong platform" error.

We could change this to always assume the script is on Linux and the image is Windows-only, thus `docker pull` succeeding would be unexpected and should fail. I think the behavior will be the same either way, but LMK what is preferable.

mboersma:
/hold

I think I should make this a toggleable step via an env var like `PREFLIGHT_CHECK`. I can imagine someone would want to build an unreleased image version or just do the build step for purposes other than publishing.

marosset:
> /hold
> 
> I think I should make this a toggleable step via an env var like `PREFLIGHT_CHECK`. I can imagine someone would want to build an unreleased image version or just do the build step for purposes other than publishing.

Oh - good ideas.
Maybe name the flag `PREFLIGHT_CHECKS` in case we want to add more in the future?

mboersma:
/hold cancel

I updated this so it's opt-in behavior if `PREFLIGHT_CHECKS == true` and tested it on the CAPZ ref image pipeline.

mboersma:
@marosset @willie-yao I made changes so this step is skipped unless `OS == Windows` and `PREFLIGHT_CHECKS == true`. PTAL when time permits.

PR #1113, opened 2023-03-15
fix Nutanix CAPI doc

tuxtof:

fix/update some part of the Nutanix CAPI procedure

### DIFF
--- a/docs/book/src/capi/providers/nutanix.md
+++ b/docs/book/src/capi/providers/nutanix.md
@@ -1,6 +1,6 @@
-# Building CAPI Images for Nutanix Cloud Platform (NCP)
+# Building CAPI Images for Nutanix Cloud Platform
 
-## Install Ansible and Packer
+## Prerequisites for Nutanix builder
 
 ```bash
 # If you don't have the image-builder repository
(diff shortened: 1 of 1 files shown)


AverageMarcus (commented):
I've no specific knowledge of Nutanix but generally looks ok. I'll 👍  but would be best to have another set of eyes on this too.

/lgtm

docs/book/src/capi/providers/nutanix.md
```diff
@@ -10,24 +10,27 @@ $ cd image-builder/images/capi/
...
     "nutanix_endpoint": "Prism Central IP / FQDN",
-    "nutanix_port": "9440",
     "nutanix_insecure": "false",
-    "nutanix_username": "PrismCentral_Username",
     "nutanix_password": "PrismCentral_Password",
-    "nutanix_cluster_name": "Name of PE Cluster",
+    "nutanix_port": "9440",
     "nutanix_subnet_name": "Name of Subnet",
-    "force_deregister": "true",
-    "image_name": "Name of Destination Image"
```
kkeshavamurthy, line 27: is `image_name` not needed anymore?

docs/book/src/capi/providers/nutanix.md
```diff
@@ -10,24 +10,27 @@ $ cd image-builder/images/capi/
...
+    "nutanix_username": "PrismCentral_Username",
+    "scp_extra_vars": ""
 }
 ```
 
+if you are using a recent `OpenSSH_9` version adding the `-O` value in `scp_extra_vars` may be necessary for servers that do not implement recent SFTP protocol.
```
mboersma, line 32:
```suggestion
If you are using a recent `OpenSSH_9` version, adding the `-O` value in `scp_extra_vars` may be necessary for servers that do not implement a recent SFTP protocol.
```

docs/book/src/capi/providers/nutanix.md
tuxtof: it was removed in the last PR but we forgot to fix the doc
image_name are by default generated based on the `build_name-kube-version` schema

tuxtof:
@jsturtevant OK for me

docs/book/src/capi/providers/nutanix.md
tuxtof: thanks @mboersma

mboersma (approved):
/lgtm
/approve

PR #1117, opened 2023-03-20
Allow all maintainers to approve docs changes

mdbooth:
This had somehow become only image-builder-windows-maintainers. Reset to project maintainers until a some wider set of owners can be agreed.

### DIFF
--- a/docs/OWNERS
+++ b/docs/OWNERS
@@ -1,4 +1,17 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+# Maintainers are expected to refrain from approving documentation changes
+# which don't relate to their area of maintenance.
 approvers:
+  - cluster-api-aws-maintainers
+  - cluster-api-azure-maintainers
+  - cluster-api-cloudstack-maintainers
+  - cluster-api-digitalocean-maintainers
+  - cluster-api-gcp-maintainers
+  - cluster-api-maintainers
+  - cluster-api-nutanix-maintainers
+  - cluster-api-openstack-maintainers
+  - cluster-api-vsphere-maintainers
+  - image-builder-maintainers
+  - image-builder-raw-maintainers
   - image-builder-windows-maintainers


mdbooth:
/cc @jsturtevant @perithompson 

This looks weird. Don't know if you're intentionally the only docs approvers? If you are, how about we create an image-builder-docs-maintainers alias with you in it, in addition to the project maintainers?

AverageMarcus:
Is there a need for this OWNERS file? Is the top level not also suitable for docs here?

jsturtevant:
The OWNERS Files are hierarchical so this isn't giving _only_ Windows maintainers rights to approve but it _adds_ the ability for Windows maintainers to approve the docs plus everyone in the top level owners file: https://github.com/kubernetes-sigs/image-builder/blob/2011a94f61f3ddc5d638481a328c49e0dcf5238c/OWNERS#L3-L8

Maybe we need to add `cluster-api-maintainers` to the docs as well? But I don't think dropping windows maintainers is right here.

mdbooth:
On second thoughts we're already bad enough at keeping OWNERS_ALIASES up to date, adding another one doesn't seem like a good plan (unless it could be defined in terms of other aliases, which it seems it can't).

Ideally we don't want to require 'root' approval on all docs changes. How about listing all `*-maintainers` aliases here and trusting folks not to approve changes not relevant their own stuff?

jsturtevant:
@mboersma @kkeshavamurthy @AverageMarcus thoughts? This seems like a reasonable approach since the approvers have the ability to approve the code changes for their own areas. The other approach would make folks reviewers and get a few more active folks into docs approval area.

mboersma:
I think this is reasonable. It's simpler and we should assume that reviewers will stay within their areas of knowledge.

kkeshavamurthy:
sounds good to me. 
/lgtm

AverageMarcus (approved):
/lgtm

docs/OWNERS
```diff
@@ -1,4 +1,18 @@
...
+  - cluster-api-digitalocean-maintainers
+  - cluster-api-gcp-maintainers
+  - cluster-api-maintainers
+  - cluster-api-nutanix-maintainers
+  - cluster-api-openstack-maintainers
+  - cluster-api-vsphere-maintainers
+  - image-builder-maintainers
+  - image-builder-raw-maintainers
   - image-builder-windows-maintainers
+  - kops-maintainers
```
jsturtevant, line 18: do we have a kops section of the book?

mboersma: No. Maybe this is in here because of `images/kube-deploy`?

```shell
% git grep -i kops
OWNERS_ALIASES:  kops-maintainers:
OWNERS_ALIASES:    # kOps maintainers, but members of kubernetes, not of kubernetes-sigs:
images/kube-deploy/OWNERS:  - kops-maintainers
images/kube-deploy/imagebuilder/templates/1.11-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.12-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.13-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.14-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.15-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.16-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.17-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
images/kube-deploy/imagebuilder/templates/1.18-stretch.yml:       # Issue: https://github.com/kubernetes/kops/issues/10122
```

jsturtevant: should we keep there here? I'm thinking we drop this...

mboersma: Agreed, it looks old / doesn't seem relevant now.

mdbooth: Updated to remove kops-maintainers

mdbooth:
@jsturtevant What needs to happen to bump that to an approve to we can clear this off the decks?

mboersma (approved):
/approve

PR #1118, opened 2023-03-23
Adding the possibility to have flatcar updated after deployment.

lanord:

Adding the possibility to have flatcar updated after deployment.

**Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #**

**Additional context**

New ansible variables for `extra-vars.json` to have update-engine reenable and pointing to a Nebraska instance
- *update_flatcar* this is to enable the update capability of Flatcar.
- *update_group* is to assign a node to a update group in nebraska (ex: stable, beta, custom one)
- *update_url* is to point to a Nebraska update server (ex: https://public.update.flatcar-linux.net/v1/update/)

### DIFF
--- a/images/capi/ansible/roles/sysprep/defaults/main.yml
+++ b/images/capi/ansible/roles/sysprep/defaults/main.yml
@@ -15,3 +15,5 @@
 extra_repos: ""
 pip_conf_file: ""
 remove_extra_repos: false
+update_flatcar: false
+update_url: "https://public.update.flatcar-linux.net/v1/update/"
--- a/images/capi/ansible/roles/sysprep/tasks/flatcar.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/flatcar.yml
@@ -22,12 +22,30 @@
     path: /etc/machine-id
     state: absent
 
+- name: Reactivate update-engine to enable updates from Nebraska
+  systemd:
+    name: update-engine
+    state: stopped
+    enabled: yes
+    masked: no
+  when: update_flatcar
+
+- name: Point to nebraska update server for OS and application updates
+  copy:
+    dest: /etc/flatcar/update.conf
+    content: |
+      GROUP={{update_group}}
+      SERVER={{update_url}}
+    mode: 0644
+  when: update_flatcar and (update_group is defined)
+
 - name: Stop and mask update-engine to freeze the image version
   systemd:
     name: update-engine
     state: stopped
     enabled: no
     masked: yes
+  when: not update_flatcar
 
 - name: Stop and mask the locksmith reboot manager since it depends on update-engine
   systemd:


images/capi/ansible/roles/sysprep/tasks/flatcar.yml
```diff
@@ -22,12 +22,46 @@
...
+  systemd:
+    name: machine-alias
+    daemon_reload: yes
+    enabled: yes
+  when: update_flatcar
+
+- name: Reactivate update-engine to enable updates from Nebraska
+  systemd:
+    name: update-engine
+    state: started
```
jepio, line 44: It probably shouldn't be `started` at this time, because that could result in the build vm updating during the build.

images/capi/ansible/roles/sysprep/defaults/main.yml
```diff
@@ -15,3 +15,6 @@
 extra_repos: ""
 pip_conf_file: ""
 remove_extra_repos: false
+update_flatcar: false
+update_group: "stable"
```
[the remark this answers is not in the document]
lanord: Will have to look back to be sure but as this is disable and remove in the packer step of the build I wanted to make sure the the configuration file would at least point to a default _stable_ chanel when you just enable this feature and not give the chanel you desire.

images/capi/ansible/roles/sysprep/files/etc/systemd/system/machine-alias.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Requires=coreos-metadata.service
+After=coreos-metadata.service
+Description=Script to populate machine alias in /etc/flatcar/update.conf
+
+[Service]
+EnvironmentFile=/run/metadata/flatcar
+Type=oneshot
+ExecStart=/bin/bash -c 'sed -i "/MACHINE_ALIAS=.*/d" /etc/flatcar/update.conf && echo "MACHINE_ALIAS=\"${COREOS_CLOUDSTACK_LOCAL_HOSTNAME}\"" | tee -a /etc/flatcar/update.conf'
```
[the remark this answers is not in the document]
lanord: This is to have hostname based name instead of uuid when the machine report to a Nebraska instance. I could add a condition for cloudstack provider to make it more generic

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
lanord: This is the syspep step which happen after the image is already build with packer and this will make the image either ready to report and be updated if you point to a private / internal Nebraska or be updated with the latest _stable_ chanel on the official flatcar nebraska instance if you did not specify anything more then the _update_flatcar_ flag to true.

lanord: After re reading I see what you mean. Will change to `stopped` instead of `started`

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
```diff
@@ -22,12 +22,48 @@
...
+  when: update_flatcar and
+    provider is defined and
+    provider is search('cloudstack')
+
+- name: Enable machine-alias.service unit
+  systemd:
+    name: machine-alias
+    daemon_reload: yes
+    enabled: yes
+  when: update_flatcar
```
jepio, line 41: This won't work if `update_flatcar` is enabled on any other platform (it will fail to enable a unit that is only created on cloudstack).

If you really want to have this functionality in image-builder, it would need to be an extra parameter (update_flatcar_machine_alias).

lanord:
I remove the part about the machine-alias. This will make it more generic and as suggested I'll move this to the cluster template instead.

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
```diff
@@ -22,12 +22,30 @@
...
+    state: stopped
+    enabled: yes
+    masked: no
+  when: update_flatcar
+
+- name: Point to nebraska update server for OS and application updates
+  copy:
+    dest: /etc/flatcar/update.conf
+    content: |
+      GROUP={{update_group}}
```
jepio, line 37: this line also needs removing

jepio (commented):
LGTM after the last comment is addressed.

jepio:
Yeah, I think that makes sense.

images/capi/ansible/roles/sysprep/tasks/flatcar.yml
invidian: I think we can use templates to conditionally add `GROUP`.

invidian (commented):
Hmm, can't `/etc/flatcar/update.conf` and `update-engine.service` be configured with the Ignition as well?

lanord:
> Hmm, can't `/etc/flatcar/update.conf` and `update-engine.service` be configured with the Ignition as well?

My reasoning for this to be in the _sysprep step_ is to not impact the building of the image. This would make the image ready to be upgraded if desire before deploying it.

lanord:
Any news on this ?

invidian:
@lanord could you elaborate on your usecase? As I said, I'm not sure why this must be added into image-builder and cannot be done via Ignition.

lanord:
> @lanord could you elaborate on your usecase? As I said, I'm not sure why this must be added into image-builder and cannot be done via Ignition.

Not sure I understand the question exactly. Do you mean to use ignition after the image is build ?

jepio:
> > @lanord could you elaborate on your usecase? As I said, I'm not sure why this must be added into image-builder and cannot be done via Ignition.
> 
> Not sure I understand the question exactly. Do you mean to use ignition after the image is build ?

Yes - you can include your own ignition (CLC) in the capi yaml.

lanord:
> > > @lanord could you elaborate on your usecase? As I said, I'm not sure why this must be added into image-builder and cannot be done via Ignition.
> > 
> > 
> > Not sure I understand the question exactly. Do you mean to use ignition after the image is build ?
> 
> Yes - you can include your own ignition (CLC) in the capi yaml.

Ok will look into that and probably close this if I can make it work.

jsturtevant:
@lanord did the alternative approach work for you?

lanord:
Yes it did I'll close this.

PR #1126, opened 2023-04-04
Fix owners aliases for tide

TerryHowe:
Looks like there is a format error in the owners aliases file that is preventing automated merges.


jsturtevant:
/lgtm
/approve

@TerryHowe thanks for the help!

TerryHowe:
Someone might need to manually merge this

PR #1128, opened 2023-04-05
Parameterize OS_VERSION in Azure nightly smoke test

mboersma:

Parameterizes two variables to fix pipeline failures. 

This nightly job had a hard-coded `OS_VERSION`, but Windows 2019 (with Docker EE) is no longer supported and its Nuget feed apparently disappeared on March 28. This allows us to replace it with `2022-containerd`, for example, which builds fine.

Which issue(s) this PR fixes:

N/A

**Additional context**

I've tested this change in the wild, works as expected.


jsturtevant:
/approve

Might be time to remove the Windows Docker related files in this repo

PR #1129, opened 2023-04-12
Add support for RHEL 8 to GCP capi provider image builder

kahun:
Adds support to build rhel 8 images to using with CAPG provider using cluster-api.

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -6,7 +6,7 @@
       "image_name": "{{user `image_name` | clean_resource_name}}",
       "labels": {
         "build_timestamp": "{{user `build_timestamp`}}",
-        "distribution": "ubuntu",
+        "distribution": "{{user `distribution`}}",
         "distribution_release": "{{user `distribution_release`}}",
         "distribution_version": "{{user `distribution_version`}}",
         "kubernetes_version": "{{user `kubernetes_semver` | clean_resource_name}}"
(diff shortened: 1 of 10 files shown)


jsturtevant:
the failure was centos which should be addressed in https://github.com/kubernetes-sigs/image-builder/pull/1135
/retest

images/capi/packer/gce/packer.json
```diff
@@ -6,17 +6,17 @@
...
         "build_timestamp": "{{user `build_timestamp`}}",
-        "distribution": "ubuntu",
+        "distribution": "{{user `distribution`}}",
         "distribution_release": "{{user `distribution_release`}}",
         "distribution_version": "{{user `distribution_version`}}",
         "kubernetes_version": "{{user `kubernetes_semver` | clean_resource_name}}"
       },
       "machine_type": "{{ user `machine_type` }}",
-      "name": "{{user `build_name`}}",
+      "name": "{{ user `build_name` }}",
```
mboersma, line 15: This is a nitpick, but if we're trying to be consistent in templates, I think we should omit these spaces. That's how we have it already for the majority of template variables, and that's how HashiCorp has it [in their docs](https://developer.hashicorp.com/packer/docs/templates/legacy_json_templates/user-variables#usage).

mboersma (commented):
/lgtm

images/capi/packer/gce/packer.json
kahun: I've removed the spaces & also an unnecesary symlink

PR #1135, opened 2023-04-14
Remove centos7 from Azure CI

mboersma:

Azure CI's `centos7`-based jobs have begun failing with this error:

```
azure-arm.vhd-{{user `build_name`}}: fatal: [default]: FAILED! =>
{"changed": false, "msg": 
"Failure downloading https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm,
HTTP Error 403: Forbidden"}
```

The link in Fedora's [Quick Start doc](https://docs.fedoraproject.org/en-US/epel/#_quickstart) is the same and is broken. 

I can't tell if this is a transient problem or not, but I would argue we don't need to be testing CentOS 7 with each PR change regardless. Please feel free to disagree!

**Additional context**


mboersma:
Looks like the permissions issue has been fixed, so this PR isn't strictly necessary. I'd still be interested in whether others think we should continue to test CentOS 7. (We don't support versions 8 or 9 either.)

PR #1139, opened 2023-04-20
Add OS_VERSION to goss for qemu

primeroz:

currently a local `qemu-flatcar` build fails with

```
==> qemu: Goss validate failed
==> qemu: Inspect mode on : proceeding without failing Packer
==> qemu: Running GOSS render command: cd /tmp/goss &&  /tmp/goss-0.3.16-linux-amd64 --gossfile goss/goss.yaml --vars /tmp/goss/goss-vars.yaml --vars-inline '{"ARCH":"amd64","OS":"flatcar","PROVIDER":"qemu","containerd_version":"1.6.18","kubernetes_cni_deb_version":"1.2.0-00","kubernetes_cni_rpm_version":"1.2.0","kubernetes_cni_source_type":"http","kubernetes_cni_version":"1.2.0","kubernetes_deb_version":"1.23.15-00","kubernetes_rpm_version":"1.23.15","kubernetes_source_type":"http","kubernetes_version":"1.23.15"}' render > /tmp/goss-spec.yaml
==> qemu: could not read json data in goss/goss-package.yaml: template: test:60:26: executing "test" at <.Vars.OS_VERSION>: map has no entry for key "OS_VERSION"
==> qemu: Goss render failed
==> qemu: Inspect mode on : proceeding without failing Packer
==> qemu: Running GOSS render debug command: cd /tmp/goss &&  /tmp/goss-0.3.16-linux-amd64 --gossfile goss/goss.yaml --vars /tmp/goss/goss-vars.yaml --vars-inline '{"ARCH":"amd64","OS":"flatcar","PROVIDER":"qemu","containerd_version":"1.6.18","kubernetes_cni_deb_version":"1.2.0-00","kubernetes_cni_rpm_version":"1.2.0","kubernetes_cni_source_type":"http","kubernetes_cni_version":"1.2.0","kubernetes_deb_version":"1.23.15-00","kubernetes_rpm_version":"1.23.15","kubernetes_source_type":"http","kubernetes_version":"1.23.15"}' render -d > /tmp/debug-goss-spec.yaml
==> qemu: could not read json data in goss/goss-package.yaml: template: test:60:26: executing "test" at <.Vars.OS_VERSION>: map has no entry for key "OS_VERSION"
```

Setting the same value that is set on the `azure/packer.json` for `OS_VERSION`

**Additional context**

### DIFF
--- a/images/capi/packer/qemu/packer.json
+++ b/images/capi/packer/qemu/packer.json
@@ -128,6 +128,7 @@
       "vars_inline": {
         "ARCH": "amd64",
         "OS": "{{user `distro_name` | lower}}",
+        "OS_VERSION": "{{user `distribution_version` | lower}}",
         "PROVIDER": "qemu",
         "containerd_version": "{{user `containerd_version`}}",
         "kubernetes_cni_deb_version": "{{ user `kubernetes_cni_deb_version` }}",
--- a/images/capi/packer/qemu/qemu-centos-7.json
+++ b/images/capi/packer/qemu/qemu-centos-7.json
@@ -3,6 +3,7 @@
   "boot_command_prefix": "<tab> text ks=",
   "boot_command_suffix": "/7/ks.cfg<enter><wait>",
   "build_name": "centos-7",
+  "distribution_version": "7",
   "distro_arch": "amd64",
   "distro_name": "centos",
   "distro_version": "7",
--- a/images/capi/packer/qemu/qemu-flatcar.json
+++ b/images/capi/packer/qemu/qemu-flatcar.json
@@ -6,6 +6,7 @@
   "build_name": "flatcar-{{env `FLATCAR_CHANNEL`}}-{{env `FLATCAR_VERSION`}}",
   "channel_name": "{{env `FLATCAR_CHANNEL`}}",
   "crictl_source_type": "http",
+  "distribution_version": "{{env `FLATCAR_CHANNEL`}}-{{env `FLATCAR_VERSION`}}",
   "distro_name": "flatcar",
   "guest_os_type": "linux-64",
   "http_directory": "./packer/files/flatcar/ignition/",
--- a/images/capi/packer/qemu/qemu-rhel-8.json
+++ b/images/capi/packer/qemu/qemu-rhel-8.json
@@ -2,6 +2,7 @@
   "boot_command_prefix": "<tab> text inst.ks=",
   "boot_command_suffix": "/8/ks.cfg<enter><wait>",
   "build_name": "rhel-8",
+  "distribution_version": "8",
   "distro_name": "rhel",
   "distro_version": "8",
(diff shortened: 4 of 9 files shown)


images/capi/packer/qemu/packer.json
```diff
@@ -128,6 +128,7 @@
       "vars_inline": {
         "ARCH": "amd64",
         "OS": "{{user `distro_name` | lower}}",
+        "OS_VERSION": "{{user `distribution_version` | lower}}",
```
AverageMarcus, line 131: Where does `distribution_version` come from? I don't see it listed in the variables below but I do see some references to `distro_version` in other files - https://github.com/kubernetes-sigs/image-builder/blob/9b101ee17fa0db4a2f3e3116d6b039e94b79abd5/images/capi/packer/qemu/qemu-centos-7.json#L8

Is that the same?

primeroz: GOOS is working so i guess is expanding to empty 

i will add it explicitely :+1:

primeroz: added

AverageMarcus:
Is `distro_version` and `distribution_version` a different thing? Why would the user provide both and when would they be different?

primeroz:
> distro_version

honestly is a bit of a mess and i am not 100% sure

* some `json` vars have `distro_version` some `don't` 
* most ( all ? ) `json` vars have `distribution_version` 
* in some `distribution_versions` is set to the same value of `distro_version` but in most is not 
* GOSS seems to expect both to be set , with `distro_version` being optional and `distribution_version` required ( reason why it was failing for QEMU builds )

AverageMarcus:
/lgtm

Generally looks good as far as I can tell but I'm not really that clear on this so going to need someone else to review also. (And I don't have approval permission anyway)

primeroz:
actually `distribution_version` is only really used to check on some windows specific tests https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-command.yaml#L138 

So probably this GOSS setup can be cleaned up quite a bit , but in this PR i was only looking to get GOSS to work in qemu since i use it to test `custom roles` 

but if we think this is pointless and would be better to cleanup the use of variables I can close this and just ignore the the goss error since the image gets built anyway

AverageMarcus:
I'm always in favour of cleaning things up if we can improve consistency. But I feel I'm missing enough context to say for sure, maybe we take this discussion to Slack to get other's thoughts?

primeroz:
> I'm always in favour of cleaning things up if we can improve consistency. But I feel I'm missing enough context to say for sure, maybe we take this discussion to Slack to get other's thoughts?

Sure, i ll be honest i did not dig deep into it , i just followed up on the goss error

kkeshavamurthy:
@primeroz Can you pls squash the commits.

primeroz:
@kkeshavamurthy done :+1:

PR #1142, opened 2023-04-24
AARCH64 / ARM64 CAPI image build for Ubuntu 22.04

ader1990:

Added example on how to build an ARM64 CAPI image using QEMU.
How to run:

```bash

# under root

apt-get install qemu-system-arm qemu-efi qemu-kvm libvirt-daemon-system -y

# create disk images
pushd /var/lib/libvirt/images/
dd if=/dev/zero of=capi.fd bs=1M count=64
dd if=/dev/zero of=capi-nvmram.fd bs=1M count=64
dd if=/usr/share/qemu-efi-aarch64/QEMU_EFI.fd of=capi.fd conv=notrunc
popd

PACKER_VAR_FILES=arm64.json ARCH=arm64 PACKER_LOG=1 make build-qemu-ubuntu-2204-arm64

# vnc listener is set conveniently to 0.0.0.0:5903 for quicker debug, can be removed later on

# Image should be present at
# ./output/ubuntu-2204-arm64-kube-v1.24.11/ubuntu-2204-arm64-kube-v1.24.11
```

Several parts need to be refactored, any help would be great.

To do's:

1. linux-cloud-tools-virtual debian package does not exist on ARM64 (an exception needs to be added)
2. Replace qemuargs with Packer keywords. I did check on Packer support for ARM64 + QEMU-KVM, but could not see any way to replace the specific qemuargs with Packer equivalent.
3. Cleanup the code after the above have been implemented

Ubuntu 22.04 ARM64 CAPI image with K8s version 1.24 was created and was tested using Tinkerbell Cluster API Provider.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -338,7 +338,7 @@ OPENSTACK_BUILD_NAMES	?=	openstack-ubuntu-2004 openstack-ubuntu-2204 openstack-f
 
 OSC_BUILD_NAMES 			?=	osc-ubuntu-2004
 
-QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804 qemu-ubuntu-2004 qemu-ubuntu-2204 qemu-centos-7 qemu-ubuntu-2004-efi qemu-rhel-8 qemu-rockylinux-8 qemu-flatcar
+QEMU_BUILD_NAMES			?=	qemu-ubuntu-1804 qemu-ubuntu-2004 qemu-ubuntu-2204 qemu-centos-7 qemu-ubuntu-2004-efi qemu-rhel-8 qemu-rockylinux-8 qemu-flatcar qemu-ubuntu-2204-arm64
 QEMU_KUBEVIRT_BUILD_NAMES	:= $(addprefix kubevirt-,$(QEMU_BUILD_NAMES))
 
 RAW_BUILD_NAMES                        ?=      raw-ubuntu-1804 raw-ubuntu-2004 raw-ubuntu-2004-efi raw-flatcar raw-rhel-8
@@ -700,6 +700,7 @@ build-openstack-ubuntu-2204: ## Builds Ubuntu 22.04 OpenStack image
 build-openstack-flatcar: ## Builds Flatcar OpenStack image
 build-openstack-all: $(OPENSTACK_BUILD_TARGETS)
 
+build-qemu-ubuntu-2204-arm64: ## Builds Ubuntu 22.04 QEMU ARM64 image
 build-qemu-flatcar: ## Builds Flatcar QEMU image
 build-qemu-ubuntu-1804: ## Builds Ubuntu 18.04 QEMU image
 build-qemu-ubuntu-2004: ## Builds Ubuntu 20.04 QEMU image
@@ -830,11 +831,11 @@ validate-node-ova-local-base-ubuntu-1804: ## Validates Ubuntu 18.04 Base Node OV
 validate-node-ova-local-base-ubuntu-2004: ## Validates Ubuntu 20.04 Base Node OVA w local hypervisor
 validate-node-ova-local-base-ubuntu-2204: ## Validates Ubuntu 22.04 Base Node OVA w local hypervisor
 
-validate-qemu-flatcar: ## Validates Flatcar QEMU image packer config
 validate-qemu-ubuntu-1804: ## Validates Ubuntu 18.04 QEMU image packer config
 validate-qemu-ubuntu-2004: ## Validates Ubuntu 20.04 QEMU image packer config
 validate-qemu-ubuntu-2004-efi: ## Validates Ubuntu 20.04 QEMU EFI image packer config
 validate-qemu-ubuntu-2204: ## Validates Ubuntu 22.04 QEMU image packer config
+validate-qemu-ubuntu-2204-arm64: ## Validates Ubuntu 22.04 QEMU ARM64 image packer config
 validate-qemu-centos-7: ## Validates CentOS 7 QEMU image packer config
 validate-qemu-rhel-8: ## Validates RHEL 8 QEMU image
 validate-qemu-rockylinux-8: ## Validates Rocky Linux 8 QEMU image packer config
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -91,7 +91,6 @@ common_virt_rpms:
 - open-vm-tools
 
 common_virt_debs:
-- linux-cloud-tools-virtual
 - linux-tools-virtual
 - open-vm-tools
 
--- a/images/capi/ansible/roles/setup/tasks/debian.yml
+++ b/images/capi/ansible/roles/setup/tasks/debian.yml
@@ -19,7 +19,16 @@
     mode: 0644
   # OCI Base images have the required apt sources list embedded inside the image, adding the sources list
   # from this repo leads to build failures(especially in Arm), hence ignoring the step.
-  when: packer_builder_type != "oracle-oci"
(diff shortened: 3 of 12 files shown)


mihaelabalutoiu:
Also, we need support building ARM64 CAPI images for Flatcar. 
Thank you.

jsturtevant:
> Perhaps we can add it as a separate PR.

makes sense to me, keeps the change set smaller. Would that rely on some of the ansible/goss changes here?

images/capi/Makefile
```diff
@@ -830,11 +831,11 @@ validate-node-ova-local-base-ubuntu-1804: ## Validates Ubuntu 18.04 Base Node OV
 validate-node-ova-local-base-ubuntu-2004: ## Validates Ubuntu 20.04 Base Node OVA w local hypervisor
 validate-node-ova-local-base-ubuntu-2204: ## Validates Ubuntu 22.04 Base Node OVA w local hypervisor
 
-validate-qemu-flatcar: ## Validates Flatcar QEMU image packer config
```
invidian, line 833: This seem like an unrelated change?

images/capi/ansible/roles/setup/tasks/debian.yml
```diff
@@ -19,7 +19,16 @@
...
-  when: packer_builder_type != "oracle-oci"
+  when: packer_builder_type != "oracle-oci" and ansible_architecture != "aarch64"
+
+- name: Put templated sources.list in place for ARM64
+  template:
+    src: etc/apt/arm64.sources.list.j2
+    dest: /etc/apt/sources.list
+    mode: 0644
+  # OCI Base images have the required apt sources list embedded inside the image, adding the sources list
+  # from this repo leads to build failures(especially in Arm), hence ignoring the step.
```
invidian, line 30: `Arm` -> `ARM`?

images/capi/ansible/roles/setup/tasks/debian.yml
```diff
@@ -19,7 +19,16 @@
...
+  when: packer_builder_type != "oracle-oci" and ansible_architecture == "aarch64"
```
invidian, line 31: Given that this is mainly for arm64, would it make sense to swap the conditionals, so we kind of do select + filter? It seem to me it makes it easier to reason about, as it reads as "For ARM64 architecture except builders of type oracle-oci".

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -551,4 +550,5 @@ windows:
         exists: true
         filetype: file
         contains:
-        - "metadata_services=cloudbaseinit.metadata.services.base.EmptyMetadataService"
\ No newline at end of file
+        - "metadata_services=cloudbaseinit.metadata.services.base.EmptyMetadataService"
+
```
invidian, line 554: I think trailing newline is not needed here?

images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -91,7 +91,6 @@ common_virt_rpms:
 - open-vm-tools
 
 common_virt_debs:
-- linux-cloud-tools-virtual
```
invidian, line 94: I'd expect not having this package should cause some validation to fail. If not, perhaps we should have some validation for functionality provided by this package?

Also, if this package is missing for ARM architecture, perhaps we should create an upstream issue for making it available?

images/capi/packer/qemu/packer.json
```diff
@@ -23,12 +23,74 @@
       "net_device": "virtio-net",
       "output_directory": "{{user `output_directory`}}",
       "qemu_binary": "{{user `qemu_binary`}}",
+      "qemuargs": [
```
invidian, line 26: Why are all those additional options now needed? If those changes are generic for supporting ARM, could we put them into separate commit with some explanation perhaps, so it is easier to reason about them in the future?

images/capi/packer/qemu/qemu-ubuntu-2204-arm64.json
```diff
@@ -0,0 +1,15 @@
+{
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz console=ttyAMA0 autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/' --- <enter><wait>initrd /casper/initrd<enter><wait>boot<enter>",
+  "boot_type": "uefi",
+  "build_name": "ubuntu-2204-arm64",
+  "distro_name": "ubuntu",
+  "guest_os_type": "ubuntu-64",
+  "http_directory": "./packer/qemu/files/22.04-arm64",
```
invidian, line 7: Perhaps this path should point to `packer/qemu/linux/ubuntu/http/22.04.arm64/`, similar to where other http directories are placed?

images/capi/packer/qemu/packer.json
```diff
@@ -23,12 +23,74 @@
...
+        ]
+      ],
       "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'usermod -L {{user `ssh_username`}} && {{user `shutdown_command`}}'",
       "ssh_password": "{{user `ssh_password`}}",
       "ssh_timeout": "2h",
       "ssh_username": "{{user `ssh_username`}}",
       "type": "qemu",
-      "vm_name": "{{user `vm_name`}}"
+      "vm_name": "{{user `vm_name`}}",
+      "vnc_bind_address": "0.0.0.0",
```
invidian, line 91: This also seem unrelated to ARM changes, could this be placed in separate commit/PR?

invidian (commented):
Awesome work @ader1990, great to see some initial ARM support for CAPI images. Overall PR looks good to me, I've left few comments/questions, which I think clarifying would help.

images/capi/Makefile
ader1990: yes, will be reverted.

images/capi/ansible/roles/node/defaults/main.yml
ader1990: These packages are a bundle up of old days paravirtualized software, like tools + drivers (I think now all drivers are upstreamed already) + services. I think we can create an upstream bug to see what are the alternatives or if these are even required for arm.

images/capi/ansible/roles/setup/tasks/debian.yml
ader1990: I had to keep backwards compatibility, maybe the conditionals can be aggregated somewhere else?

images/capi/packer/goss/goss-vars.yaml
ader1990: This file did not have consistent file endings with the other files.

images/capi/packer/qemu/packer.json
ader1990: These are just for the working example to be usable in a more dev friendly way.

images/capi/packer/qemu/packer.json
ader1990: These are the main changes required to be able to start the qemu vm, which are not supported yet as far as I know by packer. Once these changes can be upstreamed in Packer, we can reuse. Without this part, the VM does not even start.

images/capi/packer/qemu/qemu-ubuntu-2204-arm64.json
ader1990: images/capi/packer/qemu/files/22.04-arm64/user-data should be reusable in some way for any Ubuntu UEFI, I think this can be another PR too.

ader1990 (commented):
Mainly, I am thinking to split this PR into at least 3 parts:

1. tooling support for arm64
2. qemu + packer support for arm64
3. Ubuntu related UEFI parts that can be reused by x64 too

PR #1144, opened 2023-04-25
Avoid errors when RHEL subscription user and pass are not provided

iamjanr:

Avoid errors when RHEL subscription username and password are not provided

Tasks: RHEL subscription & Remove subscriptions

### DIFF
--- a/images/capi/ansible/roles/setup/tasks/redhat.yml
+++ b/images/capi/ansible/roles/setup/tasks/redhat.yml
@@ -18,7 +18,10 @@
     username: "{{ lookup('env', 'RHSM_USER') }}"
     password: "{{ lookup('env', 'RHSM_PASS') }}"
     auto_attach: true
-  when: ansible_distribution == "RedHat"
+  when: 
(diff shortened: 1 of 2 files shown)


iamjanr:
> Silly question: are these RHEL subscription steps optional then? That is, if I don't set `RHSM_USER`, building will skip these attach-to-subscription and detach tasks, but the image will still build?

Hi @mboersma, the answer is 'yes' with this fix those steps will be skipped and will build the image

kahun:
> Silly question: are these RHEL subscription steps optional then? That is, if I don't set `RHSM_USER`, building will skip these attach-to-subscription and detach tasks, but the image will still build?

Most providers works by default with a pay-as-you-go model (at least AWS, GCP, Azure) and optionally you can bring your own subscription.

kkeshavamurthy:
> Most providers works by default with a pay-as-you-go model (at least AWS, GCP, Azure) and optionally you can bring your own subscription.
I did not know this about RHEL-8. TIL. Thanks

/test pull-ova-all

PR #1147, opened 2023-04-26
feat: Enable ability to build GPU drives during image build

drew-viles:

This has been raised based [on the chat](https://docs.google.com/document/d/1YIOD0Nnid_0h6rKlDxcbfJaoIRNO6mQd9Or5vKRNxaU/edit#heading=h.msjsxsi4pgqp) in the recent office hours.

NVIDIA GPU drivers can be compiled by the GPU operator when it is deployed without much fuss however this adds time to the deployment of any new nodes - a pretty significant amount of time.

By (optionally) baking it into the image, the GPU operator can be deployed much fast by skipping the need for driver compilation.

Because of the changes to the NVIIDA license structure [noted here](https://docs.nvidia.com/grid/news/vgpu-software-license-server-eol-notice/index.html), there is now the requirements for a `.tok` file to be used for licensing. This means we can't just have a basic endpoint to pull the `.run` or `.tok` from as these contain sensitive licensing information. As a result, this new addition presumes that the user will have an S3 endpoint to pull this information from. Since there are multiple paid and free solutions to this, it didn't seem unreasonable to take this approach.

A new S3 additional component has been added and the NVIDIA module makes use of this to acquire the required resources.

**There is no book on the NVIDIA role as I wasn't sure where it would be wanted as it's not a provider, but I have updated the additional components section to include the S3 addition - maybe a `community_roles` section could be introduced where things like this could be stored?**

**Additional context**

If I've missed anything with regards to the inclusion of the s3 feature in `load_additional_modules`, please let me know and I'll get that corrected.

This has only been tested on Ubuntu as of this PR.


drew-viles:
FYI @mboersma  @AverageMarcus @jsturtevant

AverageMarcus (commented):
LGTM so far but it would be nice to have some sort of documentation around the nvidia vars and how to include it in your build. 

But I'm not sure _where_ that should go exactly. 🤔

drew-viles:
Couldn't agree more. I don't want to dump it in with nothing to explain how it would be used. Let's see what others may have to say but I think a new place to put things like this and the security bits I mentioned on the call would be a good direction to go in. A sort of "optional addons" section.

AverageMarcus:
I wonder if a README.md in the `images/capi/ansible/roles/nvidia/` directory would be enough for now.

My knowledge of Ansible isn't all that great and I've just realised I don't actually know if all other roles we have listed are included or they require some changes by the user to include them (like this one will I think?)

drew-viles:
Valid point. I can always start with that and then if it needs some adjustment we can move it into the book. I'm just wondering how many people will go hunting into the roles directory or would head straight for the book.

I'll get a readme in for now anyway, at least that's something!

drew-viles:
I'm just going to draft this for now as we've just noticed some of our latest images are not actually working with GPUs even though the image build including the NVIDIA role is completing without any errors. Once I've validated it's not this new role, I'll remove the draft status.

drew-viles:
ok got to the root of the problem. So originally the module was being included as part of `node_custom_roles_post`.

The problem we were having was the kernel had been updated via the dist-upgrade that runs during the `setup` role.
Because the driver was installed after the install of the new kernel, the driver was only being compiled for the previous kernel meaning that upon spinning up the image, the new kernel was active but the driver wasn't configured to work with it.
It purely came down to the DKMS hook not running when the new kernel was installed because it didn't know about the driver. Makes sense.

Anyway, as a result there has been a light adjustment to move it into the `node_custom_roles_pre` section now.

Glad it came up now before a merge :-D 

It's the first time we'd hit this as the latest kernel was released around 17th/18th April and we'd not built any new images yet until today.

AverageMarcus (commented):
/lgtm

drew-viles:
Just done a json-sort. I really need to setup a precommit to do that for me :facepalm:

AverageMarcus (commented):
/lgtm

drew-viles:
Sorry @AverageMarcus I've just spotted a missing "d" off the name as shown in the image.

I've just adjusted this - ran a test and it's worked locally. It must have been a mis-key last night when I made the last push.

image

I've just pushed this change now and I can 100% confirm it's functionality as I've just built an image which is working as expected.

drew-viles:
I've been doing some testing today around using CephS3 and noticed this wasn't working so I've just added a patch that allows ceph to work. Simple 1-line fix really that tells the Ansible module it's dealing with ceph.

AverageMarcus (commented):
/lgtm

images/capi/ansible/roles/nvidia/README.md
```diff
@@ -0,0 +1,26 @@
+# NVIDIA GPU driver installation
+
+To install the NVIDIA GPU driver as part of the image build process, you must have a `.run` file and `.tok` file from NVIDIA ready and available from an S3 endpoint
```
mboersma, line 3:
```suggestion
To install the NVIDIA GPU driver as part of the image build process, you must have a `.run` file and `.tok` file from NVIDIA ready and available from an S3 endpoint.
```

images/capi/ansible/roles/nvidia/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+Then all you need to do is reference as part of your packer file.
```
mboersma, line 5:
```suggestion
Then all you need to do is reference those files in your packer file.
```

images/capi/ansible/roles/nvidia/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+
+The role has to be installed via the `node_custom_roles_pre` option to avoid a known issue where should a dist-upgrade install a new kernel, 
+the driver won't work with it when the image is booted. This is because the DKMS hook doesn't get run due to the driver 
+being installed after the kernel has been installed. To get around this, we install the driver first.
+
+The `nvidia` custom role makes use of the `s3->load_additional_components` role so that it can fetch the items required from an S3 endpoint.
+
+The reasoning behind requiring an S3 endpoint was due to the fact NVIDIA will soon (July) no longer support an internal licensing server being hosted by a customer .
```
mboersma, line 23:
```suggestion
The reasoning behind requiring an S3 endpoint was due to the fact NVIDIA will soon (July 2023) no longer support an internal licensing server being hosted by a customer.
```

images/capi/ansible/roles/nvidia/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+- name: unload nouveau
+  modprobe:
+    name: nouveau
+    state: absent
+  ignore_errors: true
+
+- name: Add NVIDIA package signing key
+  ansible.builtin.apt_key:
+    url: https://nvidia.github.io/libnvidia-container/gpgkey
+  when: ansible_distribution == "Ubuntu"
```
mboersma, line 26:
```suggestion
  when: ansible_distribution == "Debian"
```

Should this be `Debian` instead? We generally prefer this if it's not specific just to Ubuntu, because it covers both distros. (Although technically I don't think we build actual Debian anywhere in the project, it seems a better default.)

Same comment for the other references to `Ubuntu` added in this file.

mboersma (commented):
This looks great, I just had some very minor doc comments and a question about maybe using `Debian` for distro.

drew-viles:
Thanks @mboersma - I will get onto these changes at some point this week. Just been a busy one so I've had little time unfortunately.

images/capi/ansible/roles/nvidia/tasks/main.yml
drew-viles: I've updated theses and spotted one more place it should be set too.

AverageMarcus:
/lgtm
/approve

Thanks for the persistence in getting this done @drew-viles! 😁  Lets get it merged in!

AverageMarcus:
What is going on with these tests?!

/retest-required
/hold for tests passing

drew-viles:
> What is going on with these tests?!
> 
> /retest-required /hold for tests passing

No idea. They seem a bit flaky tbh but I'm happy to refire as needed.

drew-viles:
So on the packer failure, it looks like it's centos-7 via Digital Ocean where it's failing.
`Error: Failed to initialize build "centos-7"
error initializing builder 'digitalocean': Unknown builder digitalocean
make: *** [Makefile:494: validate-do-centos-7] Error 1`

AverageMarcus:
Ah! We [recently bumped the version of packer](https://github.com/kubernetes-sigs/image-builder/pull/1179) (a patch version though 🤨 ) and theres some notes in the [release notes](https://github.com/hashicorp/packer/releases/tag/v1.8.7) about DigitalOcean:

> packer-plugin-digitalocean: The Digital Ocean Packer plugin has been handed
over to the Digital Ocean team. New releases for this plugin are available
at https://github.com/digitalocean/packer-plugin-digitalocean.

Related PR: https://github.com/hashicorp/packer/pull/12376

I'll open an issue for this now and look at getting a PR up shortly.

AverageMarcus:
Looks like that PR has fixed it. 

@drew-viles Once that PR is merged in you'll need to rebase this PR then it should be good to merge 😄 🤞

drew-viles:
> Looks like that PR has fixed it.
> 
> @drew-viles Once that PR is merged in you'll need to rebase this PR then it should be good to merge 😄 🤞

Cool, all pushed in, let's see what happens :-p

Thanks @AverageMarcus !

AverageMarcus:
/lgtm

🤞

PR #1148, opened 2023-04-27
Add support for Confidential VM images

### LINKED ISSUE #1154: Support building Azure Confidential VM image based on Ubuntu 22.04 LTS for CVMs
I would like to build and publish an [Azure Confidential VM image](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support) based on Ubuntu 22.04 LTS for CVMs, so that I can use it with [CAPZ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265). Currently supported CVM base images are `Windows 2019`, `Windows 2022` and `Ubuntu 20.04 LTS` for CVMs).

I would like to build and publish a CAPI Azure CVM image based on Ubuntu 22.04 LTS for CVMs.

I have tried the following 2 scenarios:
* When trying to build such an image with a VM with `vTPM` disabled, running `apt-get dist-upgrade` fails because of [this](https://bugs.launchpad.net/ubuntu/+source/nullboot/+bug/1993256) `nullboot` bug(?). 
* When trying to build such an image with a VM with `vTPM` enabled, running packer complains about [this](https://github.com/hashicorp/packer-plugin-azure/pull/257/files#diff-542c8fccfc4385f43d11c723777a061dd22251a34db06cb5499f5d626df71326R1072). 

**Additional context**
* Related [PR](https://github.com/kubernetes-sigs/image-builder/pull/1148) for support for CVM images.
* Related [question](https://github.com/hashicorp/packer-plugin-azure/issues/290#issuecomment-1535194074) on the [packer-plugin-azure](https://github.com/hashicorp/packer-plugin-azure) repository.

/kind feature

mresvanis:

This PR adds support for [Azure Confidential VM images](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#size-support). These images are necessary in order to enable the [Cluster API Provider Azure](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265) support nodes hosted on Confidential VMs. 

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1154

**Additional context**

Related [CAPZ PR](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265).

NOTE: Building a CAPI Ubuntu 22.04 CVM image from a VM instance with vTPM disabled fails due to [this](https://bugs.launchpad.net/ubuntu/+source/nullboot/+bug/1993256) bug(?) in nullboot. At the same time, Azure (and thus also packer) does not [support](https://github.com/hashicorp/packer-plugin-azure/pull/257/files#diff-542c8fccfc4385f43d11c723777a061dd22251a34db06cb5499f5d626df71326R1072) publishing trusted launch images to [managed images](https://github.com/hashicorp/packer-plugin-azure/pull/257#pullrequestreview-1287854834) (i.e. when `vTpmEnabled: true`, we can't publish a managed image, which is currently what we do now for all linux images). Because of this we are currently unable to build an CAPI Ubuntu 22.04 for CVM image.

### DIFF
--- a/docs/book/src/capi/providers/azure.md
+++ b/docs/book/src/capi/providers/azure.md
@@ -42,6 +42,18 @@ make build-azure-sig-ubuntu-1804-gen2
 
 Generation 2 images may only be used with Shared Image Gallery, not VHD.
 
+### Confidential VM Images
+
+Confidential VMs require specific generation 2 OS images. The naming pattern of those images includes the suffix `-cvm`. For example:
+
+```bash
+# Ubuntu 20.04 LTS for Confidential VMs
+make build-azure-sig-ubuntu-2004-cvm
+
+# Windows 2019 with containerd for Confindential VMs
+make build-azure-sig-windows-2019-containerd-cvm
+```
+
 ### Configuration
 #### Common Azure options
 
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -333,9 +333,11 @@ GCE_BUILD_NAMES			   ?= gce-ubuntu-1804 gce-ubuntu-2004 gce-ubuntu-2204
 VHD_TARGETS := $(shell grep VHD_TARGETS azure_targets.sh | sed 's/VHD_TARGETS=//' | tr -d \")
 SIG_TARGETS := $(shell grep SIG_TARGETS azure_targets.sh | sed 's/SIG_TARGETS=//' | tr -d \")
 SIG_GEN2_TARGETS := $(shell grep SIG_GEN2_TARGETS azure_targets.sh | sed 's/SIG_GEN2_TARGETS=//' | tr -d \")
+SIG_CVM_TARGETS := $(shell grep SIG_CVM_TARGETS azure_targets.sh | sed 's/SIG_CVM_TARGETS=//' | tr -d \")
 AZURE_BUILD_VHD_NAMES	   ?= $(addprefix azure-vhd-,$(VHD_TARGETS))
 AZURE_BUILD_SIG_NAMES	   ?= $(addprefix azure-sig-,$(SIG_TARGETS))
 AZURE_BUILD_SIG_GEN2_NAMES ?= $(addsuffix -gen2,$(addprefix azure-sig-,$(SIG_GEN2_TARGETS)))
+AZURE_BUILD_SIG_CVM_NAMES ?= $(addsuffix -cvm,$(addprefix azure-sig-,$(SIG_CVM_TARGETS)))
 
 OCI_BUILD_NAMES			   ?= oci-ubuntu-1804 oci-ubuntu-2004 oci-ubuntu-2204 oci-oracle-linux-8 oci-oracle-linux-9 oci-windows-2019 oci-windows-2022
 
@@ -373,8 +375,10 @@ AZURE_BUILD_VHD_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_VHD_NAMES))
 AZURE_VALIDATE_VHD_TARGETS	:= $(addprefix validate-,$(AZURE_BUILD_VHD_NAMES))
 AZURE_BUILD_SIG_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_SIG_NAMES))
 AZURE_BUILD_SIG_GEN2_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_SIG_GEN2_NAMES))
+AZURE_BUILD_SIG_CVM_TARGETS	:= $(addprefix build-,$(AZURE_BUILD_SIG_CVM_NAMES))
 AZURE_VALIDATE_SIG_TARGETS	:= $(addprefix validate-,$(AZURE_BUILD_SIG_NAMES))
 AZURE_VALIDATE_SIG_GEN2_TARGETS	:= $(addprefix validate-,$(AZURE_BUILD_SIG_GEN2_NAMES))
+AZURE_VALIDATE_SIG_CVM_TARGETS	:= $(addprefix validate-,$(AZURE_BUILD_SIG_CVM_NAMES))
 DO_BUILD_TARGETS 	:= $(addprefix build-,$(DO_BUILD_NAMES))
 DO_VALIDATE_TARGETS 	:= $(addprefix validate-,$(DO_BUILD_NAMES))
 OPENSTACK_BUILD_TARGETS	:= $(addprefix build-,$(OPENSTACK_BUILD_NAMES))
@@ -462,6 +466,10 @@ $(AZURE_BUILD_SIG_TARGETS): deps-azure
 $(AZURE_BUILD_SIG_GEN2_TARGETS): deps-azure
 	. $(abspath packer/azure/scripts/init-sig.sh) $(subst build-azure-sig-,,$@) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-sig-gen2.json)" -var-file="$(abspath packer/azure/$(subst build-azure-sig-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
 
+.PHONY: $(AZURE_BUILD_SIG_CVM_TARGETS)
+$(AZURE_BUILD_SIG_CVM_TARGETS): deps-azure
+	. $(abspath packer/azure/scripts/init-sig.sh) $(subst build-azure-sig-,,$@) && packer build $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-sig-cvm.json)" -var-file="$(abspath packer/azure/$(subst build-azure-sig-,,$@).json)" -only="$(subst build-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
+
 .PHONY: $(AZURE_VALIDATE_SIG_TARGETS)
 $(AZURE_VALIDATE_SIG_TARGETS): deps-azure
 	packer validate $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-sig.json)" -var-file="$(abspath packer/azure/$(subst validate-azure-sig-,,$@).json)" -only="$(subst validate-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
@@ -469,6 +477,10 @@ $(AZURE_VALIDATE_SIG_TARGETS): deps-azure
 .PHONY: $(AZURE_VALIDATE_SIG_GEN2_TARGETS)
 $(AZURE_VALIDATE_SIG_GEN2_TARGETS): deps-azure
 	packer validate $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-sig-gen2.json)" -var-file="$(abspath packer/azure/$(subst validate-azure-sig-,,$@).json)" -only="$(subst validate-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring windows,$@).json
+
+.PHONY: $(AZURE_VALIDATE_SIG_CVM_TARGETS)
+$(AZURE_VALIDATE_SIG_CVM_TARGETS): deps-azure
+	packer validate $(if $(findstring windows,$@),$(PACKER_WINDOWS_NODE_FLAGS),$(PACKER_NODE_FLAGS)) -var-file="$(abspath packer/azure/azure-config.json)" -var-file="$(abspath packer/azure/azure-sig-cvm.json)" -var-file="$(abspath packer/azure/$(subst validate-azure-sig-,,$@).json)" -only="$(subst validate-azure-,,$@)" $(ABSOLUTE_PACKER_VAR_FILES) packer/azure/packer$(findstring -windows,$@).json
 
 .PHONY: $(DO_BUILD_TARGETS)
 $(DO_BUILD_TARGETS): deps-do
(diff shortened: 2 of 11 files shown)


jsturtevant (commented):
This is really cool!  Couple thoughts:

- Are there VHD targets for Confidential VMs?
- We might need to add something like https://github.com/kubernetes-sigs/image-builder/blob/9b101ee17fa0db4a2f3e3116d6b039e94b79abd5/images/capi/scripts/ci-azure-e2e.sh#L40-L46 to get CI running

mresvanis:
@jsturtevant thank you for your time and review!

> * Are there VHD targets for Confidential VMs?

I tried to create a VHD from the CVM images with the current `packer` configuration, but I got the following error:

```sh
Build 'azure-arm.vhd-{{user `build_name`}}' errored after 13 minutes 2 seconds: Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details." D
etails=[{"code":"Conflict","message":"{\r\n  \"error\": {\r\n    \"code\": \"OperationNotAllowed\",\r\n    \"message\": \"The specified platform image is not supported for creating a Virtual Machine with unmanaged disks. Please refer to the disallowed VM disk types at https://docs.mi
crosoft.com/en-us/rest/api/compute/virtualmachineimages/get#vmdisktypes. \"\r\n  }\r\n}"}]
```

Then I tried looking into it and I'm not quire sure we can get VHD from a managed disk via `packer` (although this is supported by Azure). Any ideas on what I'm missing or how we should move forward?

My two cents would be that since VHDs are slowly getting deprecated, we could skip adding such targets for CVMs. But I probably missing CAPI's current policy on VHDs and their usecases. 

> * We might need to add something like https://github.com/kubernetes-sigs/image-builder/blob/9b101ee17fa0db4a2f3e3116d6b039e94b79abd5/images/capi/scripts/ci-azure-e2e.sh#L40-L46
>    to get CI running

Great suggestion, I missed the e2e CI script altogether. I have tried to suggest a solution for the CI CVM targets, keeping in mind that CVM images are only available in 4 Azure [locations](https://azure.microsoft.com/en-us/explore/global-infrastructure/products-by-region/?products=virtual-machines) currently (i.e. `eastus`, `westus`, `westeurope`, `northeurope`). Do you think [this](https://github.com/kubernetes-sigs/image-builder/pull/1148/files#diff-60a9e2fc101dd24789022a3c3c241f8d6b25029f870c28b9709c8dccf7838888) change makes sense?

jsturtevant:
> My two cents would be that since VHDs are slowly getting [deprecated](https://developer.hashicorp.com/packer/plugins/builders/azure/arm#deprecation-warning), we could skip adding such targets for CVMs. But I am probably missing CAPI's current policy on VHDs and their usecases.

Yes I generally think that is fine, though we don't have a good way to add an [e2e test](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265#issuecomment-1513212110) in capz with the sig images or maybe that has changed @CecileRobertMichon @mboersma?

looks like your ubuntu CMM image is failing (windows passed): https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/1148/pull-azure-sigs/1651961055783424000/artifacts/azure-sigs/ubuntu-2204-cvm.log:

```
[0;32m    azure-arm.sig-{{user `build_name`}}: TASK [setup : perform a dist-upgrade] ******************************************[0m
[0;32m    azure-arm.sig-{{user `build_name`}}: fatal: [default]: FAILED! => {"attempts": 5, "changed": false, "msg": "'/usr/bin/apt-get dist-upgrade ' failed: E: Sub-process /usr/bin/dpkg returned an error code (1)\n", "rc": 100, "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nCalculating upgrade...\nThe following package was automatically installed and is no longer required:\n  libfreetype6\nUse 'sudo apt autoremove' to remove it.\nThe following packages have been kept back:\n  libglib2.0-0 libglib2.0-bin libglib2.0-data\n0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.\n1 not fully installed or removed.\nAfter this operation, 0 B of additional disk space will be used.\nSetting up nullboot (0.4.0-0ubuntu0.22.04.1) ...\r\n2023/04/28 14:53:28 cannot trust boot assets used for current boot: open /sys/kernel/security/tpm0/binary_bios_measurements: no such file or directory\r\ndpkg: error processing package nullboot (--configure):\r\n installed nullboot package post-installation script subprocess returned error exit status 1\r\nErrors were encountered while processing:\r\n nullboot\r\nneedrestart is being skipped since dpkg has failed\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Calculating upgrade...", "The following package was automatically installed and is no longer required:", "  libfreetype6", "Use 'sudo apt autoremove' to remove it.", "The following packages have been kept back:", "  libglib2.0-0 libglib2.0-bin libglib2.0-data", "0 upgraded, 0 newly installed, 0 to remove and 3 not upgraded.", "1 not fully installed or removed.", "After this operation, 0 B of additional disk space will be used.", "Setting up nullboot (0.4.0-0ubuntu0.22.04.1) ...", "2023/04/28 14:53:28 cannot trust boot assets used for current boot: open /sys/kernel/security/tpm0/binary_bios_measurements: no such file or directory", "dpkg: error processing package nullboot (--configure):", " installed nullboot package post-installation script subprocess returned error exit status 1", "Errors were encountered while processing:", " nullboot", "needrestart is being skipped since dpkg has failed"]}[0m
```

jsturtevant:
> Great suggestion, I missed the e2e CI script altogether. I have tried to suggest a solution for the CI CVM targets, keeping in mind that CVM images are only available in 4 Azure [locations](https://azure.microsoft.com/en-us/explore/global-infrastructure/products-by-region/?products=virtual-machines) currently (i.e. `eastus`, `westus`, `westeurope`, `northeurope`). Do you think [this](https://github.com/kubernetes-sigs/image-builder/pull/1148/files#diff-60a9e2fc101dd24789022a3c3c241f8d6b25029f870c28b9709c8dccf7838888) change makes sense?

Yes generally looks like a good way to handle it 👍

mresvanis:
> looks like your ubuntu CMM image is failing (windows passed): https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/1148/pull-azure-sigs/1651961055783424000/artifacts/azure-sigs/ubuntu-2204-cvm.log:****

@jsturtevant it seems that `dist-upgrade` on Ubuntu 22.04 for CVM fails due to [this](https://bugs.launchpad.net/ubuntu/+source/nullboot/+bug/1993256) bug in `nullboot`. In the change I proposed, I was trying to create the image via a non-confidential VM and the absence of TPM leads to the `dist-upgrade` failing. At the same time, I tried to use a VM size that supports setting its `SecurityType` to `ConfidentialVM` or `TrustedLaunch`, but packer currently does not seem to support setting a `SecurityType` for a VM.

Since this doesn't seem to be an issue with Ubuntu 20.04 for CVMs, I would drop support for Ubuntu 22.04 for CVMs until there is either a packer update that supports setting `SecurityType` on a VM or `nullboot` 's bug is resolved. What do you think?

jsturtevant:
Sounds reasonable to me.  Could you create an issue in Packer to add that field and maybe an issue to track adding 2204 here?

jsturtevant:
I think we can move forward with this and revisit when 22.04 becomes more adopted. Hopefully the bug is resolved by then.

lgtm code wise.  @mresvanis is it worth updating the book to point out CVM images and any caveats?

> Yes I generally think that is fine, though we don't have a good way to add an [e2e test](https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/3265#issuecomment-1513212110) in capz with the sig images or maybe that has changed @CecileRobertMichon @mboersma?

If we need e2e tests in capz we might need to figure this out @mboersma @nawazkh 

/assign @mboersma

mresvanis:
@jsturtevant the `nullboot` issue (i.e. trying to run `nullbootctl` without the vTPM enabled fails) hit us in Ubuntu 20.04 LTS for CVM as well. That's why I implemented a workaround [here](https://github.com/kubernetes-sigs/image-builder/pull/1148/files#diff-c51aa260a3e0c1b4bae66f7454c144c7c2b297d1acecccf340525cf4231547dc) and added the Ubuntu 22.04 LTS for CVMs back to the Azure SIG CVM build targets. 

This workaround is not great, but it provides users with CAPZ Ubuntu images for confidential VMs. Please

jsturtevant:
> This workaround is not great, but it provides users with CAPZ Ubuntu images for confidential VMs. Do you think this is an acceptable workaround?

I don't really know enough about that that is doing. Could you give a summary?

mresvanis:
> > This workaround is not great, but it provides users with CAPZ Ubuntu images for confidential VMs. Do you think this is an acceptable workaround?
> 
> I don't really know enough about that that is doing. Could you give a summary?

Absolutely, the issue we face when building CAPZ images based on `Ubuntu {20,22}.04 LTS for CVMs` is due to `nullboot`'s post-install script expecting vTPM to be enabled. This post-install script is run after `apt-get distupgrade` and fails because we cannot have vTPM enabled in the VMs we're building the images from. This leads to `apt-get dist-upgrade` to fail, which in turn results into the image build failure.

To resolve the issue, i.e. to have a successful `apt-get dist-upgrade` with `nullboot` and vTPM disabled, I did the following:

- hold the `nullboot` package, i.e. do not update it when `apt-get dist-upgrade`
- temporarily adjust the `nullboot` post-install script to not fail when vTPM is disabled (this is accomplished by adding 2 additional fields in `nullbootctl`)
- run `apt-get dist-upgrade`
- change the `nullboot` post-install script back to its original
- unhold the `nullboot` package, i.e. when `apt-get update | dist-upgrade` runs, it will be updated to its latest version

 This is configured to only run for CVM images. What do you think?

nawazkh:
@mresvanis Thank you for the explanation, makes sense to me. Though I am not so well versed with image-builder flow, your approach sounds good to me. Few questions:
- What are the maintainable components in this PR we need to be mindful of as the OS offerings increase or change going ahead?
- Is there a scope to add a test to validate the changes in being generated in the new image? I once added a [goss test](https://github.com/nawazkh/image-builder/blob/367a5602fb61804499eb7fa54907b910cef28de9/images/capi/packer/goss/goss-vars.yaml#L327-L332C47) in one of the PRs I worked on https://github.com/kubernetes-sigs/image-builder/pull/1090 ; I am not sure, but can we add something similar ?

mresvanis:
@nawazkh thank you for reviewing this change! 

> * What are the maintainable components in this PR we need to be mindful of as the OS offerings increase or change going ahead?

If I understand your question correctly, ideally we would just run the regular image-builder flow on top of specific CVM images without any additional changes to them. For example, exactly what this change adds for the Windows CVM images.
 
> * Is there a scope to add a test to validate the changes in being generated in the new image? I once added a [goss test](https://github.com/nawazkh/image-builder/blob/367a5602fb61804499eb7fa54907b910cef28de9/images/capi/packer/goss/goss-vars.yaml#L327-L332C47) in one of the PRs I worked on [add /etc/cloud/cloud.cfg.d/15_azure-vnet.cfg to all Azure VMs #1090](https://github.com/kubernetes-sigs/image-builder/pull/1090) ; I am not sure, but can we add something similar ?

Very good point! I tried to think about how we could test the changes of CVM images. But since we are not doing anything more other than base the CAPI images to the Ubuntu and Windows CVM images, without any CVM specific configuration needed, we should be good to go. If we were to test/check for specific CVM image configuration, already included in thos CVM base images, I believe there is no way (at least [not at the time of this writing](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-faq-amd#do-all-os-images-work-with-confidential-vms-))  to consistently test that the CVM images we are basing upon the CAPI CVM images, include everything that is needed, for them to be successfully used in CVM instances. We can only trust that the Ubuntu and Windows base CVM images have everything we need.

IMHO the holding/unholding `nullboot` package workaround should be covered by the inclusion of the CVM image building in the e2e CI tests. I am not sure I covered your questions, but please feel to let me know for more details or questions, I'll be happy to try and help more.

nawazkh:
^ Thank you for the detailed explanation and answer @mresvanis. I am happy with the changes, explanation makes sense to me, and I would love to "lgtm" this PR. 

"looks good to me!"

Delegating to @jsturtevant and @mboersma for their thoughts :)

mresvanis:
> > * change the `nullboot` post-install script back to its original
> > * unhold the `nullboot` package, i.e. when `apt-get update | dist-upgrade` runs, it will be updated to its latest version
> 
> This should only be needed for the 22.04 LTS images, right?

Unfortunately, we have the same issue with Ubuntu 20.04 LTS as well. We saw the issue appearing in both distributions gradually, as the latest version of the `nullboot` package was being gradually released.
 
> A few more questions about the workaround:
> 
> * What happens next time a user runs an upgrade? will they run into this issue as well?

Azure CVMs won't boot unless their OS image adheres to specific requirements (which are not yet [published](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-faq-amd#do-all-os-images-work-with-confidential-vms-), but I'm confident that [vTPM PCR measurements](https://www.usenix.org/legacy/event/sec06/tech/full_papers/berger/berger.pdf) are a hard requirement). The Ubuntu CVM images are specific to the Azure CVMs, so users would use them only on those supported VM sizes. These VM sizes require vTPM to be enabled, otherwise Azure just refuses to create them (there is no point in using a CVM if it cannot be [attested](https://learn.microsoft.com/en-us/azure/confidential-computing/guest-attestation-confidential-vms) to). If a user tries to create a CVM with a non-compliant OS images, then the VM just won't boot.
 
When a user runs an `apt-get dist-upgrade` on a CVM using one of the Ubuntu CVM images, everything should work as expected, because vTPM will be enabled (as it's a hard requirement to be so). 

> * Do the CVM images need a TPM? If we can't publish the image due to the azure bug are we getting anything from building them?

As described above, CVMs require vTPM to be enabled and these OS images are a hard requirement, in order to be able to boot those VMs (and after boot for users to be able to attest them).

> I'm wondering if we can get away with just skipping 22.04 until the issues have been resolved upstream.

That was my initial thought as well, but since the `nullboot` package was updated also in Ubuntu 20.04 LTS, we would have to skip both Ubuntu images (which would leave us with only the Windows ones and I think having both Windows and Linux images is quite useful). 

> Thanks for your patience and help understanding all the nuances.

Absolutely no worries, we have to be sure we're adding reliable and useful changes. In addition, CVMs are a quite recent field :)

jsturtevant:
> These VM sizes require vTPM to be enabled, otherwise Azure just refuses to create them

So the VM sizes we are using to build this images don't have the vTPM enabled? But we expect users to select VM sizes in Azure that do make sense when creating VMs from the images?  

Would we have this issue if the VM size we are building from has a vTPM?

mresvanis:
@jsturtevant thank you for digging in, great questions and those are exactly what I was also trying to figure out. Please let me know if the following makes sense:

> So the VM sizes we are using to build this images don't have the vTPM enabled?

Exactly. We can't have [SecureBoot or vTPM enabled](https://github.com/hashicorp/packer-plugin-azure/pull/257/files#diff-542c8fccfc4385f43d11c723777a061dd22251a34db06cb5499f5d626df71326R1072) when creating a managed image. 

> But we expect users to select VM sizes in Azure that do make sense when creating VMs from the images?

That's correct. The CVM images target Azure Confidential VMs only, but currently we can't use such VMs to build the respective CAPZ managed images.

> Would we have this issue if the VM size we are building from has a vTPM?

The VM size we are building from can have a vTPM, but not an enabled one ([here](https://github.com/hashicorp/packer-plugin-azure/pull/257/files#diff-542c8fccfc4385f43d11c723777a061dd22251a34db06cb5499f5d626df71326R1072) is the error from the packer plugin). 

It is the combination of the following two issues that lead me to this implementation:

- `nullboot` by default requires vTPM to be enabled when being configured (which happens with `apt-get dist-upgrade`)
- `packer-plugin-azure` (and also Azure) does not allow us to create a managed image with vTPM enabled

jsturtevant:
Thanks for unraveling that for me. I think I finally understand the various bugs :-)

/lgtm
/assign @mboersma

mboersma (approved):
/lgtm
/approve

I'm sorry I didn't review this earlier! I've finally caught up–thank you for the detailed explanation of the complexities here.

PR #1159, opened 2023-05-12
Removing ssh user (builder) from OVA vsphere-iso

knabben:
The user builder exists after the image-builder job is finished, the script removes the sudoers file and the user before shutdown the system.

This PR is only related to OVA and vsphere-iso hypervisor. 

**Additional context**
The current setup does not passes STIG controls for a not used user.

builder user is removed in the final OVA:
```
root@nobu-cv4hl-p2t8k:/home# ls -lah
total 12K
drwxr-xr-x  3 root root 4.0K May 30 23:18 .
drwxr-xr-x 19 root root 4.0K May 30 23:18 ..
drwx------  4 capv capv 4.0K May 30 23:19 capv               
root@nobu-cv4hl-p2t8k:/home# ls /etc/sudoers.d/
90-cloud-init-users  README
```


images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -232,3 +232,17 @@
     name: ssh
     enabled: yes
   when: ansible_os_family == "Debian"
+
+- name: Remove temporary builder user
+  ansible.builtin.user:
+    name: builder
+    state: "absent"
+    remove: true
+  when: ansible_os_family == "Ubuntu"
```
mboersma, line 241:
```suggestion
  when: ansible_os_family == "Debian"
```

This isn't specific to Ubuntu, is it? It's generally preferable to use "Debian" to cover both distros.

Is this specific to the distro, though? Is there also a `builder` user left around on RHEL, for example?

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -232,3 +232,17 @@
...
+
+- name: Remove builder sudoers file
+  file:
+    state: absent
+    path: /etc/sudoers.d/builder
+  when: ansible_os_family == "Ubuntu"
```
mboersma, line 247:
```suggestion
  when: ansible_os_family == "Debian"
```

images/capi/ansible/roles/sysprep/tasks/main.yml
knabben: This is not specific for Ubuntu. Fixing it.

kkeshavamurthy: Would it be better to remove the user in the shutdown command [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/ova/packer-node.json#L18)

images/capi/packer/ova/packer-node.json
```diff
@@ -108,7 +108,7 @@
       "remote_password": "{{user `remote_password`}}",
       "remote_type": "{{user `remote_type`}}",
       "remote_username": "{{user `remote_username`}}",
-      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'usermod -L {{user `ssh_username`}} && {{user `shutdown_command`}}'",
+      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'userdel -f -r {{user `ssh_username`}} && rm -f /etc/sudoers.d/{{user `ssh_username` }} && {{user `shutdown_command`}}'",
```
kkeshavamurthy, line 111: Can this change be added to the other builder types in the same file as well

knabben: `vsphere-clone` documentation say the user is kept.
is this required for `vsphere-vmx`?

kkeshavamurthy: The following builder `name`s will require the change as the output will be the final capi image. 
```
vmware-vmx
vmware-iso
vsphere-iso
vsphere-clone
```

The `builder` user will be retained in 
```
vmware-iso-base
vsphere-iso-base
```

basically where ever you see the `usermod -L` command needs your changes.

PR #989, opened 2022-10-03
Remove old build targets from Azure test matrix

mboersma:
What this PR does / why we need it:

In preparing to [add Ubuntu 22.04 support](https://github.com/kubernetes-sigs/image-builder/pull/961#issuecomment-1263846684) to Azure Cluster API images, we realized that our PR [test matrix is huge](https://github.com/kubernetes-sigs/image-builder/pull/961#issuecomment-1263846684) (up to 20 items, requiring 20 Azure VMs to be provisioned for each CI run). This ~~mostly~~ removes `ubuntu-1804` and `windows-2019` from that matrix, as they are both on the verge of being sunsetted.

Which issue(s) this PR fixes:

N/A

**Additional context**

See also #961

cc: @CecileRobertMichon @jackfrancis @jsturtevant @kkeshavamurthy

PR #1163, opened 2023-05-22
Skip Ubuntu 18.04 in ova CI tests

mboersma:

Removes `ubuntu-18.04` from the script that is run as part of the default `pull-ova-all` CI job. Ubuntu 18.04 is no longer supported, and this builder has been flaky with unhelpful errors such as:

```
[1;31m==> vsphere-iso.vsphere: error typing a boot command (code, down) `23, false`: ServerFaultCode: A general system error occurred: Invalid fault[0m
```

Which issue(s) this PR fixes:

**Additional context**

See also #989


kkeshavamurthy:
I've been seeing a lot of failures last couple of weeks or so and was thinking of removing 18.04 from the ci list. 
Thanks for the PR @mboersma

PR #1167, opened 2023-05-23
[azure] Update pip and az before installing extension

mboersma:

The CAPZ reference image pipeline began failing to install the `az capi` extension recently. Debugging shows this to be a deep dependency error that is out of our hands (either in the `az` CLI itself or in the implementation of `az extension add`), but ensuring that `pip` and `az` are updated first is a workaround that succeeds.

Which issue(s) this PR fixes:

N/A

**Additional context**


mboersma:
/hold

For reviewing (I don't like that my PRs are pre-approved; we should discuss at office hours if we like this policy.)

nojnhuh (commented):
/lgtm

Even if this might not be the most targeted fix for this particular issue, it's probably good hygiene in any case.

PR #1150, opened 2023-05-03
Disable Flatcar updates during image creation

johananl:
What this PR does / why we need it: This PR ensures Flatcar doesn't auto-update during image creation.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #854 

**Additional context**
None

PR #1172, opened 2023-06-01
Refactor Ignition config files

johananl:

Following https://github.com/kubernetes-sigs/image-builder/pull/1150#issuecomment-1571721560, we're replacing the generic bootstrap Ignition config with two separate configs - one for cloud platforms which have their own mechanism for authorizing SSH keys during machine bootstrap and another for platforms on which we have to rely on SSH password auth.

**Additional context**
We need to add this change set separately from modifying consumer code because we need fully-qualified and merged GitHub blobs to exist before we can modify Packer builders to use the new files.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -968,6 +968,6 @@ json-sort: ## Sort all JSON files alphabetically
 ## --------------------------------------
 ##@ Ignition
 .PHONY: gen-ignition
-ignition_files = bootstrap
+ignition_files = bootstrap bootstrap-pass-auth bootstrap-cloud
 gen-ignition: deps-ignition ## Generates Ignition files from CLC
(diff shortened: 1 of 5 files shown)


invidian (approved):
/lgtm

johananl:
@jsturtevant looks like we've got an LGTM from Mateusz. Do we need more feedback from people with Ignition knowledge? If so, I can ask some more colleagues to weigh in.

I don't anticipate any strong objections to this PR since we're _adding_ new Ignition files without touching consumer code at this point.

Thanks!

mboersma (approved):
/lgtm
/approve

I'm also not an expert, but I tested this branch locally some and poked at `make gen-ignition`.

jsturtevant:
nope good to go, just missed the notification :-)
/approve

PR #1175, opened 2023-06-02
Cleanup unmaintained projects

### LINKED ISSUE #1143: Removing unmaintained projects from image-builder

I would like to propose removing the currently unmaintained projects and simplify this repository.  The readme states there are [three projects](https://github.com/kubernetes-sigs/image-builder/tree/master#what-is-the-image-builder) in this repository but this also a command line tool.  There has been little to no activity on the projects beyond the Cluster API image builder:

- [Image builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi) 
  - actively maintained
- [kube-deploy/imagebuilder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/kube-deploy/imagebuilder)
   - not actively maintained since 2021: https://github.com/kubernetes-sigs/image-builder/commits/master/images/kube-deploy
   - only owners are [kops](https://github.com/kubernetes-sigs/image-builder/blob/master/images/kube-deploy/OWNERS)?
 - [konfigadm](https://github.com/kubernetes-sigs/image-builder/tree/master/images/konfigadm)
   - no commits since initial project structure: https://github.com/kubernetes-sigs/image-builder/commits/master/images/konfigadm
- [cli tool](https://github.com/kubernetes-sigs/image-builder/tree/master/v1)
  - no major commits in over a year (and no feature development past initial commit: https://github.com/kubernetes-sigs/image-builder/commits/master/v1) 
  - We provide a docker image and guidance on how to use that to do deploys. This seems to be working for the majority of users
    
There has been feedback that the project is difficult to approach as folks are not sure where to begin.  Part of this is due to three projects being in the repository and it is difficult to figure out what changes are needed.

Another part of this is that the current maintainers of the project have no experience with the tools outside image-builder folder or have a clear vision for where those projects should go.

Remove the code from this project and maybe move the CAPI folder up in the code tree structure (this might be breaking change or difficult so lower prior than remove code).

We could continue to leave them with no commits
 - doesn't address the issue of making it hard to know where and how to begin.
 - creates confusion
 
We could move them to new github repositories.
  - there are not clear maintainers for the projects so this doesn't achieve much
  - If in the future someone wants to the code could still be moved

**Additional context**

Please comment and let us know if there are any reservations or concerns.  I propose we have a lazy consensus period (maybe a month?) and pin this issue to the repository for enough visibility to get any feedback/concerns.

/kind cleanup

AverageMarcus:
## What this PR does / why we need it:

Removes the following from the repo:
* konfigadm
* kube-deploy
* Abandoned image-builder CLI

## Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): 

Fixes #1143

**Additional context**
I removed each project in its own commit so it should be easy to revert a single project if we ever need to bring them back.


README.md
```diff
@@ -6,19 +6,13 @@
 
 The Image Builder is a collection of cross-provider Kubernetes virtual machine image building utilities.
 
-There are currently 3 distinct tools in this repo:
+There is currently 1 distinct tools in this repo:
```
drew-viles, line 9: It's so minor - and I'm being picky😄, but can "tools" be changed to "tool" since we're only hosting one in this project? 😉
With that in mind, is this line needed at all now?

drew-viles (changes requested):
Good stuff!
Separating them out into their own commits seems like a logical idea for a back out process if it's required.
I only have the one doc comment from my side.

README.md
AverageMarcus: Updated :) And changed the Slack link to point to the `#image-builder` one rather than `#sig-cluster-lifecycle` too.

Makefile
```diff
@@ -17,17 +17,6 @@ SHELL := /usr/bin/env bash
 CWD := $(shell pwd)
 BIN := image-builder
```
jsturtevant, line 18: should this be dropped too?

Makefile
```diff
@@ -17,17 +17,6 @@ SHELL := /usr/bin/env bash
...
 
-.PHONY: clean $(BIN)
-
-default: $(BIN)
-
-$(BIN):
-	$(MAKE) -C v1
-
-clean:
-	rm -f dist/
```
jsturtevant, line 28: should `v1` target get dropped too?

jsturtevant (commented):
thanks! couple minor suggestions

Makefile
AverageMarcus: Yeah, this is the `v1` cli that was abandoned.

Makefile
jsturtevant: looks like it is only referenced in https://github.com/kubernetes-sigs/image-builder/blob/master/v1/Makefile so I think it is safe to drop

AverageMarcus: 👍  Removed!

PR #1189, opened 2023-06-14
Lock packages versions

DvdChe:
Additional task in `kubernetes` role to  lock packages versions of Kubernetes components to avoid undesired upgrade.

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/debian.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/debian.yml
@@ -34,3 +34,13 @@
       - "kubeadm={{ kubernetes_deb_version }}"
       - "kubectl={{ kubernetes_deb_version }}"
       - "kubernetes-cni={{ kubernetes_cni_deb_version }}"
+
+- name: Lock packages versions
+  dpkg_selections:
+    name: "{{ item }}"
+    selection: hold
+  loop:
+    - kubeadm
+    - kubelet
+    - kubectl
+    - kubernetes-cni
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -28,7 +28,23 @@ command:
     timeout: 0
 {{end}}
 {{if eq .Vars.kubernetes_source_type "pkg"}}
+{{if or (eq .Vars.OS "debian") (eq .Vars.OS "ubuntu") }}
+{{- range mkSlice "kubeadm" "kubelet" "kubectl" }}
+  apt-mark showhold | grep {{.}}:
+    exit-status: 0
+    stderr: []
+    timeout: 0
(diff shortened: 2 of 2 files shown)


jsturtevant:
lgtm, I guess it might be nice to have a goss check that it is locked in https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-package.yaml

Looking at [goss package](https://github.com/goss-org/goss/blob/master/docs/manual.md#package) parameters I am not sure there is a clean way to do it, might need to run a command?

DvdChe:
> lgtm, I guess it might be nice to have a goss check that it is locked in https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/goss/goss-package.yaml
> 
> Looking at [goss package](https://github.com/goss-org/goss/blob/master/docs/manual.md#package) parameters I am not sure there is a clean way to do it, might need to run a command?

We can run `apt-mark showhold` and the output will show which package has its version pinned. output would be something like 
```
#: apt-mark showhold
kubectl
kubeadm
kubelet
kubernetes-cni
```

If you want me to implement this step in goss test, that's ok, i'll see how to do it so

kkeshavamurthy:
We already run apt-mark in sysprep role. That should already be locking in all the installed packages.

mboersma:
@DvdChe as @kkeshavamurthy pointed out, this [existing task](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/sysprep/tasks/debian.yml#L20-L21) seems like it has this covered:

```
- name: apt-mark all installed packages
  shell: dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
```

Have you seen the kubernetes packages (or others) being upgraded anyway? Just curious what motivated this change.

DvdChe:
> @DvdChe as @kkeshavamurthy pointed out, this [existing task](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/ansible/roles/sysprep/tasks/debian.yml#L20-L21) seems like it has this covered:
> 
> ```yaml
> - name: apt-mark all installed packages
>   shell: dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
> ```
> 
> Have you seen the kubernetes packages (or others) being upgraded anyway? Just curious what motivated this change.

Sorry for late response. I had undesired upgrades while I was troubleshooting issues ( weren't related with this project ). I was running packer with `-on-error=ask` parameter. Error occured before sysprep role and I was replaying the whole playbook from the beginning during my fix/retries sessions and because of [this step](https://github.com/kubernetes-sigs/image-builder/blob/d37da2a5307c1a5cc4b04190a55518935e6edc48/images/capi/ansible/roles/setup/tasks/debian.yml#L75),  kubernetes was bumped to its latest version

Issue #1198, opened 2023-06-17
Cleanup references of ubuntu 18.04.


wasup-yash:
**What steps did you take and what happened:**
The 1804 image has been deleted from `Ubuntu-os-cloud` project. 

**What did you expect to happen:**

referencing this https://github.com/kubernetes/kubernetes/issues/118718#issuecomment-1595711655

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi):

AverageMarcus:
@wasup-yash Can you confirm for me if this is:

* just related to the automated tests?
* only an issue for the GCE provider?
* applies to all providers?

If I understand correctly based on the linked issue, the problem is with the [cluster-api-provider-gcp](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/scripts/ci-conformance.sh) periodic test failing. We can remove ubuntu-18.04 from image-builder but that wont stop the CAPG test from failing until they update their script.

I think that the ubuntu-18.04 base images are still available on other providers (e.g. the AMIs are still available on AWS) so removing all references to ubuntu-18.04 might not be appropriate without first consulting users on Slack etc.

wasup-yash:
> @wasup-yash Can you confirm for me if this is:
> 
> * just related to the automated tests?
> * only an issue for the GCE provider?
> * applies to all providers?
> 
> If I understand correctly based on the linked issue, the problem is with the [cluster-api-provider-gcp](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/scripts/ci-conformance.sh) periodic test failing. We can remove ubuntu-18.04 from image-builder but that wont stop the CAPG test from failing until they update their script.
> 
> I think that the ubuntu-18.04 base images are still available on other providers (e.g. the AMIs are still available on AWS) so removing all references to ubuntu-18.04 might not be appropriate without first consulting users on Slack etc.

Yes you are right it will not be good to remove all references and as the Issue Mentioned it is related to the `Cluster-api-provider-gcp` so it will be good to remove it from the image-builder and as soon they update their script it will probably fix the issue.

AverageMarcus:
This has been done for GCP. We have https://github.com/kubernetes-sigs/image-builder/issues/1232 to cover removing Ubuntu 18.04 (and other EOL OSs) from other providers so I'm going to close this issues as done. 😄

PR #1200, opened 2023-06-19
Remove Ubuntu-18.04 from GCE provider

AverageMarcus:

Removed Ubuntu 18.04 from the GCE provider as the base image is no longer available.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Towards #1198

**Additional context**
I have left Ubuntu 18.04 in the other providers for now as its not clear if they are being used or not. As this it is no longer possible to build it for GCE there is no (further) harm in removing it completely.


mboersma (commented):
/lgtm

PR #1201, opened 2023-06-19
Remove VMWare open-vm-tools from qemu images

### LINKED ISSUE #1131: QEMU build of Ubuntu 22.04 installs a package related to VMWare
I noticed that during a `qemu` build of Ubuntu 22.04 running the following command:
```
make build-qemu-ubuntu-2204
```
within `images/capi` directory we're installing `open-vm-tools` as an additional package here:

https://github.com/kubernetes-sigs/image-builder/blob/2e96edbc8edcfde228860c2b9dff55342b851ee1/images/capi/packer/ova/linux/ubuntu/http/22.04/user-data#L62-L63

https://github.com/kubernetes-sigs/image-builder/blob/2e96edbc8edcfde228860c2b9dff55342b851ee1/images/capi/packer/ova/linux/ubuntu/http/22.04.efi/user-data#L83-L84

The PR that added those files (#1003) mentions adding support for VMWare and `open-vm-tools` is specific to VMWare. I guess the problem stems from the fact that the `linux` file [here](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/qemu/linux) is a symlink to `../ova/linux`? It feels like we need to split things up a bit or copy-paste basically, if possible.

Not to have a VMWare specific tool installed as part of the Ubuntu autoinstaller by default.

N/A

**Environment:**
N/A

AverageMarcus:

Removes the VMWare tools `open-vm-tools` from the qemu image builds

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1131

**Additional context**
I'm not sure if these tools would ever be required when building qemu images so I'd like some input from others before merging.

/hold


AverageMarcus:
> Another option would be to have some sort of hack script to remove the open-vm-tools from the ks files before the builds begin.

I feel like that could get confusing as the state of the files in the directories wouldn't match what is actually applied. I'd rather be explicit I think.

Ideally, as a long term solution we'd have something that installs a bunch of "default" linux stuff and then be able to add provider-specific extras on top of those. But I don't have enough experience with the current setup to know how / if we'd refactor it to that.

AverageMarcus:
/retest

(I want to check CI still works as expected following branch rename)

AverageMarcus:
/test pull-packer-validate
/test json-sort-check

(I want to check CI still works as expected following branch rename)

PR #1204, opened 2023-06-20
Flatcar: ARM64: fix make deps

mihaelabalutoiu:
What this PR does / why we need it:

* Fixed the make dependencies to ensure compatibility with both architectures.

PR #1205, opened 2023-06-21
Fix host arch detection for arm64

mboersma:

#1204 didn't take into account that `uname -m` returns `arm64` on Apple Silicon. So after that change, I was unable to run image-builder targets locally:

```shell
% noproxy=* make -C images/capi build-azure-sig-flatcar-gen2
hack/ensure-ansible.sh
unsupported HOSTARCH=arm64
make: *** [deps-azure] Error 1
```

Which issue(s) this PR fixes:

N/A

**Additional context**

### DIFF
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -32,6 +32,9 @@ case "${_hostarch}" in
 *aarch64*)
   HOSTARCH=arm64
   ;;
+*arm64*)
+  HOSTARCH=arm64
+  ;;
 *x86_64*)
   HOSTARCH=amd64
   ;;


AverageMarcus (commented):
/lgtm
/approve

AverageMarcus:
/test pull-ova-all

There looks to be an issue with our test:

> The pod could not start because it could not mount the volume "private-key": secret "clusterapi-provider-vsphere-ci-prow" not found

AverageMarcus:
Yup, same thing. I think it is likely related to https://github.com/kubernetes/test-infra/pull/29752

I guess we need to figure out where that secret comes from and make sure its included in the new EKS cluster (assuming we actually do need it)

mboersma:
/override pull-ova-all

I'm confident this is a safe fix to a general problem, and I think it's ok to skip this job while we sort out the "clusterapi-provider-vsphere-ci-prow" secret. I opened #1206 to track it and @AverageMarcus is pursuing a fix in #sig-testing.

PR #1207, opened 2023-06-22
Nutanix: update RedHat family

tuxtof:

- add RHEL8 & 9 support 
- update documentation for RHEL
- fix iscsi-init bug for redhat family
- fix Rocky Linux 8 & 9 URL
- split nutanix ansible provider role in two files
- take care of ipvs dependency
- fix OS_VERSION missing causing  goss issue


tuxtof:
Hello @AverageMarcus @kkeshavamurthy if you have 5 min to a quick review
thanks

docs/book/src/capi/providers/nutanix.md
```diff
@@ -63,5 +68,21 @@ The `nutanix` sub-directory inside `images/capi/packer` stores JSON configuratio
...
+| `rhel-8.json`       | Settings for RedHat Enterprise Linux 8 image  |
+| `rhel-9.json`       | Settings for RedHat Enterprise Linux 9 image  |
 | `flatcar.json`      | Settings for Flatcar Linux image (beta)       |
 | `windows-2022.json` | Settings for Windows Server 2022 image (beta) |
+
+### OS specific options
+
+#### RHEL
+
+You need to set your `image_url` value correctly in your rhel-(8|9).json file with a working Red Hat Enterprise Linux KVM Guest Image url.
```
AverageMarcus, line 80: I think it'd be better to remove the `image_url` from the `rhel-8.json` and `rhel-9.json` in the repo and update these docs to explain how to use the `PACKER_VAR_FILES` environment variable to supply the user-required variables, ideally with an example of the `image_url` being provided.

images/capi/packer/nutanix/rockylinux-8.json
```diff
@@ -8,7 +8,7 @@
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8",
   "extra_rpms": "python3",
   "guest_os_type": "Linux",
-  "image_url": "https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-GenericCloud-Base-8.7-20221130.0.x86_64.qcow2",
```
AverageMarcus, line 11: This is unrelated to "update RedHat family` so can we leave this out of this PR please? If you think we should switch to a rolling latest tag can you please open an issue / different PR to discuss it.

images/capi/packer/nutanix/rockylinux-9.json
```diff
@@ -7,7 +7,7 @@
   "distro_name": "rockylinux",
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-9",
   "guest_os_type": "Linux",
-  "image_url": "https://dl.rockylinux.org/pub/rocky/9/images/x86_64/Rocky-9-GenericCloud-Base-9.1-20221130.0.x86_64.qcow2",
```
AverageMarcus, line 10: This is unrelated to "update RedHat family` so can we leave this out of this PR please? If you think we should switch to a rolling latest tag can you please open an issue / different PR to discuss it.

docs/book/src/capi/providers/nutanix.md
tuxtof: @AverageMarcus this is how all the other image builder platform who support RedHat are doing , why we will do differently ??

images/capi/packer/nutanix/rockylinux-8.json
tuxtof: actual link is dead, latest is just to avoir to udpate repo all the time
let me put it in different PR

images/capi/packer/nutanix/rockylinux-9.json
tuxtof: actual link is dead, latest is just to avoir to udpate repo all the time
let me put it in different PR

docs/book/src/capi/providers/nutanix.md
AverageMarcus: Maybe I'm misunderstanding but I don't see any of the other providers using placeholder values for the URL in their packer json files?

tuxtof: you have it in qemu images/capi/packer/qemu/qemu-rhel-8.json => iso_url
ova images/capi/packer/ova/rhel-8.json  => iso_url

AverageMarcus: Wow we really don't have consistency here 😞 

Ok, I guess we're not making things any worse here and the docs are good enough for people to get it working.  I'll not block on this but I'm gonna open an issue to try and get consistency of RHEL over the different providers. 🙂

AverageMarcus:
/lgtm

Just waiting on tests to pass for approval :)

docs/book/src/capi/providers/nutanix.md
tuxtof: why you say it is not consistent ?
everybody are using external image , coming from different place/format  depending of the provider

AverageMarcus: The user experience is slightly different. 

In this PR, the job wont run at all without the user changing the value. In OVA it'll work as long as there is an iso found in the root. I'm not sure if the qemu one would run or not with just the file name in that field.

AverageMarcus:
> ServerFaultCode: A general system error occurred: Invalid fault

Likely a flake.

/test pull-ova-all

docs/book/src/capi/providers/nutanix.md
tuxtof: for me in all three case , you need to put image somewhere and mention it,
ova builder support file: URL not the case with the nutanix one for now , so no way to propose a default choice who almost work (as long as there is an iso found in the root)
But this PR works as long as there is a correct URL set

tuxtof:
@AverageMarcus all tests are ok now

PR #1209, opened 2023-06-22
Add Azure packer config file

jsturtevant:
What this PR does / why we need it:

This allows us to bump the Azure packer plugin to get new bug fixes and features.  The 1.4.3 version of the Azure packer plugin resolves a long standing bug where VMs didn't delete properly: https://github.com/hashicorp/packer-plugin-azure/issues/125

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #

**Additional context**
Add any other context for the reviewers

PR #1210, opened 2023-06-22
Move packer installs to the Dependency targets

jsturtevant:

During #1209 I noticed a we had a few different ways of installing plugins, this brings them all to the same place.

**Additional context**


AverageMarcus (commented):
/lgtm
/approve

PR #1213, opened 2023-06-26
Nutanix: fix broken rocky link

tuxtof:

Fix Rocky Linux broken link, switch to latest release to avoir futur issue

### DIFF
--- a/images/capi/packer/nutanix/rockylinux-8.json
+++ b/images/capi/packer/nutanix/rockylinux-8.json
@@ -8,7 +8,7 @@
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8",
   "extra_rpms": "python3",
   "guest_os_type": "Linux",
-  "image_url": "https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-GenericCloud-Base-8.7-20221130.0.x86_64.qcow2",
+  "image_url": "https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-GenericCloud-Base.latest.x86_64.qcow2",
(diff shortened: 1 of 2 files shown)


tuxtof:
@AverageMarcus this is the PR for rocky link update

images/capi/packer/nutanix/rockylinux-8.json
```diff
@@ -8,7 +8,7 @@
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8",
   "extra_rpms": "python3",
   "guest_os_type": "Linux",
-  "image_url": "https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-GenericCloud-Base-8.7-20221130.0.x86_64.qcow2",
+  "image_url": "https://dl.rockylinux.org/pub/rocky/8/images/x86_64/Rocky-8-GenericCloud-Base.latest.x86_64.qcow2",
```
AverageMarcus, line 11: It looks like the image does still exist but now available at `http://dl.rockylinux.org/vault/rocky/8.7/images/x86_64/Rocky-8-GenericCloud-Base-8.7-20221130.0.x86_64.qcow2` instead.

I'd like to get other peoples opinion before switching to using `latest` as it causes re-runs of the same configuration to potentially result in different images which I'm not too keen on. I'm going to open a thread on Slack to gather thoughts.

tuxtof: this is already how image builder work as a

yum update / apt update is performed during each build

so each result is different with always the last security update

starting from latest allow to have the exact same result just improve the build time

but open to have other peoples opinion

AverageMarcus: Slack discussion: https://kubernetes.slack.com/archives/C01E0Q35A8J/p1687764381019339

PR #1215, opened 2023-06-29
Add fallback for tools that don't have darwin/arm64 binaries

AverageMarcus:

Adds a helper function to fallback to `amd64` on an M1/M2 Mac when a tool doesn't provide an arm64 binary.

**Additional context**

### DIFF
--- a/images/capi/hack/ensure-goss.sh
+++ b/images/capi/hack/ensure-goss.sh
@@ -22,6 +22,9 @@ set -o pipefail
 
 source hack/utils.sh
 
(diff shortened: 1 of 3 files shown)


architchandra:
Struggling with this exact issue for the past 2-3 hours. Please release this soon. Thanks!

mboersma (approved):
/lgtm
/approve

This seems to work on my M1 MacBook. (I hadn't seen any problems with this previously because I had `kustomize` installed through homebrew and that was being found in my $PATH.)

PR #1223, opened 2023-07-11
fix: making nvidia standalone due to conflicts with load_additional_components  

drew-viles:
I've been trying to get some additional container images baked into the image I'm building to speed up deployment time and I've hit a snag.

During the process I've had multiple failed builds and it stems from the following bits within the `nvidia` role.
```
  ansible.builtin.include_role:
    name: load_additional_components
```

When the `nvidia` role is loaded in via `node_custom_roles_pre`, which is the recommended approach to avoid issues with dkms/kernel updates during the image build, it also calls the `load_additional_components` role. 

You may or may not already have spotted the problem, but I'll continue my tale!

If you're trying to do anything else that also requires that role, then it'll try to do it in the pre-steps. This is obviously bad if you're doing what I am which is pulling a container image as `containerd` and the associated tools are non-existent at this time. 
I tried to get a few workarounds in place to avoid having the S3 code essentially twice but there is nothing that can be done without a bunch of conditionals which in my opinion would mess up the code base.

This PR basically makes `nvidia` a self-contained role and it allows the `load_additional_components` role to run during the natural course of the build rather than whenever `nvidia` is called.

Hope that makes sense!

### DIFF
--- a/images/capi/ansible/roles/nvidia/README.md
+++ b/images/capi/ansible/roles/nvidia/README.md
@@ -8,7 +8,7 @@ An example of the fields you need are defined below. Make sure to review and cha
 
 ```json
 {
-  "ansible_user_vars": "additional_s3=true nvidia_s3_url=https://s3-endpoint nvidia_bucket=nvidia nvidia_bucket_access=ACCESS_KEY nvidia_bucket_secret=SECRET_KEY nvidia_installer_location=NVIDIA-Linux-x86_64-525.85.05-grid.run nvidia_tok_location=client_configuration_token.tok gridd_feature_type=4"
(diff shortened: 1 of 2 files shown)


mboersma:
I think I see what you're talking about, and it doesn't seem unreasonable to me to bring the S3 dependency directly into the nvidia role as a workaround. I'd love to hear what others think. (Sorry, reviewing has been a bit slow with a couple maintainers on summer vacation.)

drew-viles:
No problem @mboersma - no rush, just something I noticed the other day and thought "ah, didn't hit that test case when this first went in" 🤦. Test suite should be added to our list of "nice to haves" 😄  

I'm happy for there to be an s3 component in the `load_additional_components`, it's more that if `nvidia` is used and you want to use any element of `load_additional_components` too, it causes a bit of unhappiness in the run due to order of operations getting flipped around.

All I'm thinking is the `nvidia` role has it's on little s3 task that is separate from the `load_additional_components` so it remains self-contained.

mboersma (commented):
/lgtm

AverageMarcus:
Fine with me!
/lgtm

Issue #905, opened 2022-05-30
Workaround SSH error "no matching host key type found. Their offer: ssh-rsa"


invidian:
**What steps did you take and what happened:**

Right now, if one builds the CAPI images with OpenSSH version 8.8+ on the host system, the environment variable below is required to make provisioning work. This is because OpenSSH starting from version [8.8](https://www.openssh.com/txt/release-8.8) has disabled `ssh-rsa` algorithm from default configuration and this is what Ansible plugin for Packer [still generates](https://github.com/hashicorp/packer-plugin-ansible/issues/69).

```console
ANSIBLE_SSH_ARGS="-oHostKeyAlgorithms=+ssh-rsa -oPubkeyAcceptedAlgorithms=+ssh-rsa"
```

**What did you expect to happen:**

Builds to work on modern systems out of the box.

**Anything else you would like to add:**

I suggest we commit those arguments as default ones until upstream issue is not resolved.

As an alternative, we could try disabling Ansible provisioner proxy, bypassing local keys generation completely. I don't know what side effects that creates though.

**Environment:**

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi), [kube-deploy/imagebuilder](https://github.com/kubernetes-sigs/image-builder/tree/master/images/kube-deploy/imagebuilder), [konfigadm](https://github.com/kubernetes-sigs/image-builder/tree/master/images/konfigadm)): Image Builder for Cluster API

Additional info for Image Builder for Cluster API related issues:

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`): Arch Linux
- Packer Version: 1.8.1
- Packer Provider:
- Ansible Version: 2.11.5

/kind bug
[One or more /area label. See https://github.com/kubernetes-sigs/cluster-api/labels?q=area for the list of labels]
 
CC @kopiczko

AverageMarcus:
Looks like this problem also prevents being able to use Ubuntu 22.04 as a base VM. Doesn't look like the env var can be used as a workaround in that situation (at least it didn't work for me).

Meecr0b:
we successfully implemented a workaround in our environment by directly setting
```
"existing_ansible_ssh_args": "-o HostKeyAlgorithms=+ssh-rsa -o PubkeyAcceptedKeyTypes=+ssh-rsa"
```

PR #1224, opened 2023-07-12
images/capi/packer/config/ansible-args.json: remove ssh-rsa workaround

invidian:

In commit 76671ca21b2d31e062c489bc93d4da03c8c19c3f we added a workaround for packer-plugin-ansible not supporting keys other than ssh-rsa, which broke building images for clients using newer versions of OpenSSH.

This should no longer be needed with the latest version of packer ansible plugin:
https://github.com/hashicorp/packer-plugin-ansible/issues/69#issuecomment-1631493391

**Additional context**

This is a follow up to issue #905.


mboersma (commented):
/lgtm

Thanks for following up on this @invidian!

mboersma (approved):
/approve

PR #1225, opened 2023-07-12
Add chrony hook for Openstack

mpiscaer:
Add the networkd-dispatcher for chrony, this will make sure that Openstack images, check for the DNS srv record and DHCP option 42.

The current Openstack images do not install the NTP servers in chrony when DHCP option 42 set or the _ntp._udp DNS SRV records are there. To fix this the networkd-dispatcher is needed.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/openstack.yml
+++ b/images/capi/ansible/roles/providers/tasks/openstack.yml
@@ -31,3 +31,16 @@
     state: stopped
     enabled: false
   when: ansible_os_family == "Debian"
+
+- name: Copy networkd-dispatcher scripts to add DHCP provided NTP servers
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: a+x
+  vars:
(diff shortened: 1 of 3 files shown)


mnaser:
@mpiscaer : this is good on the surface but if you want to resolve this for the images we use in the Cluster API, it would be best to make the change in the `qemu` section, since that is what we use there.

Also, it would be good to introduce GOSS tests for the both OpenStack and QEMU:

https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/goss/goss-vars.yaml#L204-L207C22

I think you'd be able to slot it in here for Ubuntu both for QEMU and OpenStack:

mnaser (changes requested):
needs some goss tests + would be nice to add to the `qemu` image.

AverageMarcus:
/lgtm

@mpiscaer Could you please [squash your commits](https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md#squash-commits)?

PR #1229, opened 2023-07-18
adds cdrom_type and http_ip to packer node variables.

jvrahav:

ide device is deprecated in the latest linux kernel as of linux 5.2. currently image builder defaults  cdrom to ide for ova builder. This change adds cdrom_type variable enabling users to set SATA as the backing device for cd rom when building images using latest kernel.

http_ip is required to enable user to pass the ip that packer should connect to download kickstarter file.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes # 1228


AverageMarcus:
/ok-to-test

Does this also make sense for QEMU builds or is it something specific to OVA?

jvrahav:
> QEMU

Based on https://developer.hashicorp.com/packer/plugins/builders/qemu#cdrom_interface QUEMU builder by default uses virtio or virtio-iscsi. OS i believe these parameters are not required to be exposed for QEMU at the moment.

I'm no QEMU expert, if we believe it makes sense for QEMU, i will be happy to add it

images/capi/packer/ova/packer-node.json
```diff
@@ -457,6 +466,7 @@
     "folder": "",
     "guest_os_type": null,
     "http_directory": "./packer/ova/linux/{{user `distro_name`}}/http/",
+    "http_ip": "",
```
kkeshavamurthy, line 469: When this var is not passed, will the default be picked up?

jvrahav: if its unset packer will automatically discover and set ip

https://developer.hashicorp.com/packer/plugins/builders/vsphere/vsphere-iso#http_ip

images/capi/packer/ova/packer-node.json
```diff
@@ -1,11 +1,13 @@
 {
   "builders": [
     {
+      "cdrom_type": "{{user `cdrom_type`}}",
```
kkeshavamurthy, line 4: `vmware-iso` and `vmware-vmx` builders do not have `cdrom_type` var. I think it's `cdrom_adaptor_type`

jvrahav: I have added cdrom_adapter_type for iso.
These variables are not relevant for `vmware-vmx` hence i have removed them. I dont see an equivalent for vmx

images/capi/packer/ova/packer-node.json
```diff
@@ -38,6 +38,7 @@
         "{{user `boot_command_suffix`}}"
       ],
       "boot_wait": "{{user `boot_wait`}}",
+      "cdrom_type": "{{user `cdrom_type`}}",
```
kkeshavamurthy, line 41: adaptor here as well.

images/capi/packer/ova/packer-node.json
```diff
@@ -444,6 +451,8 @@
     "base_output_dir": "./output/{{user `base_build_version`}}",
     "build_timestamp": "{{timestamp}}",
     "build_version": "{{user `build_name`}}-kube-{{user `kubernetes_semver`}}",
+    "cdrom_adapter_type": "",
+    "cdrom_type": "",
```
kkeshavamurthy, line 455: maybe the the default to `ide` explicitly so that it's clear. It can be over ridden the configs when needed.

kkeshavamurthy:
Looks good. Pls squash the commit and we'll get this merged.

jvrahav:
> Looks good. Pls squash the commits and we'll get this merged.

@kkeshavamurthy  done

kkeshavamurthy:
/approve
Congrats on the first PR to the repo @jvrahav !! Keep them coming.

PR #1241, opened 2023-08-02
Update containerd to 1.6.22

hrak:

Update containerd to 1.6.22

Changes [can be found here](https://github.com/containerd/containerd/releases/tag/v1.6.22)

**Additional context**

### DIFF
--- a/images/capi/packer/config/containerd.json
+++ b/images/capi/packer/config/containerd.json
@@ -1,7 +1,7 @@
 {
   "containerd_additional_settings": null,
   "containerd_cri_socket": "/var/run/containerd/containerd.sock",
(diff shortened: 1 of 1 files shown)


hrak:
Question: should we consider switching to containerd 1.7 in the near future?

mboersma:
/ok-to-test

> should we consider switching to containerd 1.7 in the near future?

Generally we should try to keep critical packages like that fresh, so I'd say yes. But IDK if there are specific migration issues or if 1.7 is problematic for any particular provider (chances are good it breaks *something*, right?).

Unless someone has a specific objection, I think we should try 1.7 soon.

mboersma (commented):
/lgtm

mboersma (approved):
/approve

I opened #1244 to track updating to containerd v1.7.

PR #1279, opened 2023-08-25
Remove inline provisioner for GCE Ubuntu

AverageMarcus:

The CAPG tests are currently failing - see https://kubernetes.slack.com/archives/C8TSNPY4T/p1692992049432829

Turns out this block of code never did anything due to previously having:
```
"if [ $BUILD_NAME != \"ubuntu-1804\" ] || [ $BUILD_NAME != \"ubuntu-2004\" ]; then exit 0; fi",
```
Which always matched regardless of what the build name was. 

As it was never actually used I've removed it completely.

**Additional context**

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -24,17 +24,6 @@
   ],
   "provisioners": [
     {
-      "environment_vars": [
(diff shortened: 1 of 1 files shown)


AverageMarcus:
Rate limit :( 

/retest

PR #1291, opened 2023-09-08
Add dracut cmd to generate initramfs with drivers for RHEL raw

### LINKED ISSUE #1287: Redhat raw images do not have all the required drivers in initrd
Built a rhel-8 raw image, and used the resulting image to run a baremetal cluster using Tinkerbell provider. The image failed to bootup. Upon inspection it looks like the `initrd` image does not include all the required drivers needed. After adding the drivers already available on the OS on dracut.conf, the image boots up properly.

The built raw image's initrd to include all the required drivers

Raw image builds use KVM to create a vm and run the image building on the vm, and export the disks. This is causing `initrd` to only consume drivers required to run on a KVM virtualized environment. An ansible task to force dracut to rebuild initrd with host mode set to false should be added to raw providers.
```
dracut --force --host-only=false
```
This will make the initrd image built to consume all drivers, and not only the ones required for running on a KVM virtual environment.

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi))

/kind bug

vignesh-goutham:
Raw builds are run on KVM vms. RHEL 8.x ships with dracut that builds initramfs with default `host-only` mode set to true. This causes the initramfs to only include drivers that are required to run kvm infrastructure. When the disks are exported and booted on a metal server, the OS fails to boot up due to this missing drivers. 

This change runs dracut with `--no-hostonly` to rebuild initramfs to include all drivers

This PR also fixes another issue that bubbled up during testing the dracut changes. The `build_type` was not set to `raw` in the rhel json file, causing providers role to run qemu tasks as opposed to raw tasks.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1287 

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/raw.yml
+++ b/images/capi/ansible/roles/providers/tasks/raw.yml
@@ -35,6 +35,10 @@
     - cloud-utils-growpart
   when: ansible_os_family == "RedHat"
 
+- name: Run dracut cmd to regenerate initramfs with all drivers - needed when converting to different hypervisor templates
+  shell: dracut --force --no-hostonly
+  when: ansible_os_family == "RedHat"
+
 #- name: Unlock password
 #  replace:
 #    path: /etc/cloud/cloud.cfg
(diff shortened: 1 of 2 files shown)


AverageMarcus:
> ServerFaultCode: A general system error occurred: Invalid fault

Looks like a flake

/retest

mboersma (commented):
/lgtm++

PR #1305, opened 2023-09-22
Disable colour output with ansible-lint

AverageMarcus:

The output with colour doesn't look great in the prow logs (see https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/934/pull-lint/1705217812173164544) so we would be better disabling them and just having the plain output.

**Additional context**

### DIFF
--- a/images/capi/scripts/ci-lint.sh
+++ b/images/capi/scripts/ci-lint.sh
@@ -28,4 +28,4 @@ cd "${CAPI_ROOT}" || exit 1
 export PATH=${PWD}/.local/bin:$PATH
 export PATH=${PYTHON_BIN_DIR:-"${HOME}/.local/bin"}:$PATH
 
-make lint
+NO_COLOR=1 make lint


AverageMarcus:
@mboersma It looks like there is an unskippable failure 😕 

```
                 Rule Violation Summary                   
 count tag                    profile rule associated tags 
     1 syntax-check[specific] min     core, unskippable    
Failed: 1 failure(s), 0 warning(s) on 161 files.
```

mboersma:
We could change this just in the CI job by changing scripts/ci-lint.sh to do:

`NO_COLOR=1 make lint`

AverageMarcus:
> We could change this just in the CI job by changing scripts/ci-lint.sh to do:
> 
> `NO_COLOR=1 make lint`

I like that better. I'll update

mboersma (commented):
/lgtm

Thanks for noticing this! After #1306 merges, I just want to see the actual output from CI before approving.

mboersma (approved):
/approve

PR #1306, opened 2023-09-22
Ensure Windows and common deps are installed before ansible-lint

mboersma:

Fixes a missed dependency for the `make lint` target that prevented it from running cleanly in CI.

Which issue(s) this PR fixes: 

N/A

**Additional context**

I also made a trivial change in the /ansible directory to trigger the linter prow job.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -181,6 +181,8 @@ deps-proxmox: deps-common
 
 .PHONY: deps-lint
 deps-lint: ## Installs/checks dependencies for image-builder code linting
+deps-lint: deps-common
+	hack/ensure-ansible-windows.sh
 	hack/ensure-ansible-lint.sh
 
 ## --------------------------------------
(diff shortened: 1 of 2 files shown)


images/capi/Makefile
```diff
@@ -181,6 +181,8 @@ deps-proxmox: deps-common
 
 .PHONY: deps-lint
 deps-lint: ## Installs/checks dependencies for image-builder code linting
+deps-lint: deps-common
+	hack/ensure-ansible-windows.sh
```
AverageMarcus, line 185: Does this actually check if we're on windows or not?

mboersma: No, it doesn't. But I think we want to install the Windows ansible modules in order to lint all the sources, regardless of platform. The job is currently failing because it doesn't know what the `win_shell` task is, and I can see from #1306 that this fixed things.

AverageMarcus: Oh, looks like it's not for installing on windows but for building windows specific things.

PR #1314, opened 2023-09-28
Expose ssh_proxy variables for vSphere builder

erkanerol:

In my environment, vSphere API is accessible only via proxy. It is not enough to set classic proxy variables (e.g. https_proxy, no_proxy etc.). Packer's ssh communicator only supports SOCKS proxy and it is configuration is separate. See https://developer.hashicorp.com/packer/docs/communicators/ssh

I couldn't find another way to set `ssh_proxy_host` and `ssh_proxy_port` without exposing them in builder's configuration as I do in this PR. 

**Open Points:**
1. Is there a way to set these variables without any change in the code?
2. Is there a standard across builders in the whole repo? Should I add this change to every builder?

### DIFF
--- a/images/capi/packer/ova/packer-common.json
+++ b/images/capi/packer/ova/packer-common.json
@@ -23,6 +23,8 @@
   "remote_username": "",
   "skip_compaction": "false",
   "ssh_password": "builder",
+  "ssh_proxy_host": "",
+  "ssh_proxy_port": "",
   "ssh_timeout": "60m",
   "ssh_username": "builder",
   "vmx_version": "15",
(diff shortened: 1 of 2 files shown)


mboersma (commented):
> Is there a way to set these variables without any change in the code?

I'm not a Packer expert, but I read the docs again and did some searching. As far as I can tell, this is the way to do it and there aren't any environment variables or another mechanism to use.

> Is there a standard across builders in the whole repo? Should I add this change to every builder?

I don't see any existing use of `ssh_proxy_*` and I don't think we have any precedent for this.

I don't think this needs to be added globally, although we often do that with newly exposed vars.

erkanerol:
> I don't think this needs to be added globally, although we often do that with newly exposed vars

Then my PR is ready to review and merge :)

mboersma (commented):
/lgtm
/assign @AverageMarcus

PR #1315, opened 2023-09-28
Fix path issues in post-processor of vsphere

erkanerol:
 
The current code mainly works but there is a failing scenario. When `./output` directory is a symbolic link, `../..` refers to the original path instead of `images/capi` path and post-processing fails.
 
The other uses of this script (`image-build-ova.py`) and other scripts used in post-processors are already called as `./hack/<script>`. See https://github.com/kubernetes-sigs/image-builder/pull/1315/files#diff-e5a52b8d4ed5f9c2b3af687e2e7ca510249ead5a7071b3827c698e46da551461R335

### DIFF
--- a/images/capi/packer/ova/packer-node.json
+++ b/images/capi/packer/ova/packer-node.json
@@ -322,8 +322,7 @@
         "vsphere-iso-base"
       ],
       "inline": [
-        "cd {{user `output_dir`}}",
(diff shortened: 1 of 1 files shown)


mboersma (commented):
/lgtm

Although it's a bit confusing why we were `cd`ing to a subdirectory, then referencing scripts in the parent directory in the first place, this change seems like it would avoid the symlink problem mentioned.

/assign @chrischdi @sbueringer 
For a vSphere perspective.

chrischdi (commented):
Change makes sense to me 👍 

/lgtm

mboersma (approved):
/approve

PR #1317, opened 2023-09-29
Flatcar: Use OS default volume/exec symlink

pothos:
The usage of a bind mount does not work with systemd-sysext managing the
    /usr mount tree because the mounts may get lost on an extension reload.
    Flatcar starts to use extensions by default for cloud integration
    software. Fortunately, we don't to define an own systemd-sysext
    extension to bring in the symlink under /var because the symlink is
    present in all recent Flatcar versions.

(Old:
The usage of a bind mount does not work with systemd-sysext managing the /usr mount tree because the mounts may get lost on an extension reload. Flatcar starts to use extensions by default for cloud integration software.
Instead of using a bind mount, create a systemd-sysext extension that brings in a symlink to a folder under /var.)

`systemd-sysext refresh` will cause the bind mount to be lost, but it's also not needed anymore and only causes problems when there are two possible locations.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes a future issue.

**Additional context**
Add any other context for the reviewers:
https://www.freedesktop.org/software/systemd/man/systemd-sysext.html


mboersma:
/ok-to-test

Could you include "flatcar" in the commit message so it's clear where this change is targeted? (The whole image-builder repo is effectively just for CAPI now.)

pothos:
Now I understood that it was an actual failure because of swapped arguments. Fixed it.

pothos:
The Ansible docs say that `src` is the link destination and `dest` is the symlink itself, is this correct? Other link usages in this repo here seem to say otherwise.
 https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-src

pothos:
Since this was open long enough, all Flatcar versions now have a good default behavior and the bind mount is not needed anymore nor any other workaround. We should get this in quick because the bind mount would cause two possible locations to be used for the next major Flatcar Stable release.

AverageMarcus (commented):
/lgtm

@mboersma @jsturtevant are you both happy with this?

images/capi/ansible/roles/setup/tasks/flatcar.yml
```diff
@@ -14,28 +14,6 @@
...
-  file:
-    path: "{{ item }}"
-    state: directory
-  with_items:
-    - /opt/libexec
-    - /opt/libexec.work
-
-- name: Create usr-libexec.mount unit
-  copy:
-    src: etc/systemd/system/usr-libexec.mount
```
mboersma, line 27: We should also remove the `images/capi/ansible/roles/setup/files/etc/systemd/system/usr-libexec.mount` since it's no longer referenced.

jsturtevant: I see that file was removed.  Am I missing something else?

mboersma: Nope, I just read the diff wrong. 👍🏻

mboersma (approved):
/lgtm
/approve

dalees:
This patch requires a certain version of Flatcar - kubelet fails to start `kube-controller-manager` when combined with an older version of Flatcar (we built with `DISTRIB_RELEASE=3510.2.5`).

I acknowledge that the version we built with is old (July 2023), and we should be pulling the latest during builds but it seems worth mentioning the flatcar version requirement of this PR, lest others encounter this problem.

mboersma:
> it seems worth mentioning the flatcar version requirement of this PR

@dalees that's a great point that I wasn't aware of when approving this. Do you know more specifically from which version of Flatcar this will work? We could at least document it as a requirement.

pothos:
Right, the required version now is [3602.2.1](https://www.flatcar.org/releases#release-3602.2.1) from 2023-10-25

PR #1320, opened 2023-10-10
Update pip install to handle "error: externally-managed-environment"

mboersma:

Adds a new `pip3_install` bash function that will retry installation if a certain error is raised.

Recently the e2e build environment was updated such that [PEP 668](https://peps.python.org/pep-0668/) became implemented. This means that `pip3 install <package>` raises an error telling the user to do `apt-get install python-<package>` instead. [Here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/1319/pull-azure-vhds/1711845171575918592) is one example of that failure.

This makes sense for distro vendors, but image-builder doesn't control the distro and has specific requirements for some tools (Packer, ansible, ansible-lint).

Which issue(s) this PR fixes:

Fixes #

**Additional context**

I looked at the recommended approach–using a Python virtualenv–but it's not a great fit with our set of scripts and would require more significant refactoring.

### DIFF
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -100,6 +100,19 @@ ensure_py3() {
   fi
 }
 
+pip3_install() {
+  ensure_py3
+  if output=$(pip3 install --disable-pip-version-check --user "${@}" 2>&1); then
+    echo "$output"
+  elif [[ $output == *"error: externally-managed-environment"* ]]; then
+    >&2 echo "warning: externally-managed-environment, retrying pip3 install with --break-system-packages"
+    pip3 install --disable-pip-version-check --user --break-system-packages "${@}"
+  else
+    >&2 echo "$output"
+    exit 1
+  fi
+}
+
 hostarch_without_darwin_arm64() {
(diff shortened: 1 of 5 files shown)


images/capi/hack/utils.sh
```diff
@@ -100,6 +100,19 @@ ensure_py3() {
   fi
 }
 
+pip3_install() {
+  ensure_py3
+  if output=$(pip3 install --disable-pip-version-check --user "${@}" 2>&1); then
+    echo "$output"
+  elif [[ $output == *"error: externally-managed-environment"* ]]; then
+    >&2 echo "warning: externally-managed-environment, retrying pip3 install with --break-system-packages"
+    pip3 install --disable-pip-version-check --user --break-system-packages "${@}"
```
AverageMarcus, line 109: 😆 
```
--break-system-packages
```
What a great flag name

AverageMarcus (commented):
/lgtm
/approve

Do you happen to know if the E2E environment changes were announced anywhere? Is there something we should be keeping an eye on for potential problems?

PR #1326, opened 2023-10-16
add az-cli for capz windows images

DannyBrito:
This PR adds azure cli install to CAPZ windows images.

**Additional context**

### DIFF
--- a/images/capi/ansible/windows/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/windows/roles/providers/tasks/azure.yml
@@ -46,3 +46,18 @@
     members: "{{ users }}"
   vars:
     users: "{{ wire_server_users.split(',') if (wire_server_users is defined) and (wire_server_users | length > 0) else [] }}"
+
+- name: Download Azure CLI MSI installer
+  ansible.windows.win_get_url:
+    url: https://aka.ms/installazurecliwindowsx64
+    dest: C:\azure-cli.msi
+
+- name: Install Azure CLI
+  ansible.windows.win_package:
+    path: C:\azure-cli.msi
+    state: present
+
+- name: Clean up Azure CLI Installer file
+  ansible.windows.win_file:
+    path: C:\azure-cli.msi
+    state: absent
--- a/images/capi/packer/goss/goss-command.yaml
+++ b/images/capi/packer/goss/goss-command.yaml
@@ -220,7 +220,7 @@ command:
     - True
     stderr: []
     timeout: 30000
-  
+
   # this could be moved to place for other providers if they want to install it
   Key Vault gMSA binary is installed:
     exec: powershell -command "Test-Path -Path  C:\Windows\System32\CCGAKVPlugin.dll"
@@ -246,6 +246,9 @@ command:
     stdout:
     - "ICcgDomainAuthCredentials"
     timeout: 30000
+  Check Azure CLI is installed via alias:
+    exec: powershell -command "az"
+    exit-status: 0
 {{end}}
 
 {{ if ne .Vars.ssh_source_url "" }}


mboersma:
Looks good judging by the output in the Azure windows jobs. (The `ova` failure is an unrelated flake, we can run it again.)

images/capi/ansible/windows/roles/providers/tasks/azure.yml
```diff
@@ -46,3 +46,9 @@
     members: "{{ users }}"
   vars:
     users: "{{ wire_server_users.split(',') if (wire_server_users is defined) and (wire_server_users | length > 0) else [] }}"
+
+- name: Install Azure CLI via MSI Installer
+  ansible.windows.win_package:
+    path: https://azcliprod.blob.core.windows.net/msi/azure-cli-2.53.0-x64.msi
```
mboersma, line 52: Is there a "latest" link to the installer? When we install `az` for Linux, we just do `pip3 install azure-cli` so we get the latest version available when we build. It would be nice if we could do that here too, otherwise we'll need to remember to update this.

images/capi/packer/goss/goss-command.yaml
```diff
@@ -246,6 +246,9 @@ command:
     stdout:
     - "ICcgDomainAuthCredentials"
     timeout: 30000
+  Check Azure CLI is installed:
+    exec: powershell -command "& 'C:\\Program Files\\Microsoft SDKs\\Azure\\CLI2\\wbin\\az.cmd'"
```
mboersma, line 250: Is it worth changing this command to verify that `az` is available in the %PATH%?

This verifies that it's at an expected location, but in kubernetes-sigs/cluster-api-provider-azure#4141 we just run `az login --identity` which doesn't specify the full path to the command.

images/capi/ansible/windows/roles/providers/tasks/azure.yml
mboersma: From the Azure CLI [Windows installation page](https://learn.microsoft.com/cli/azure/install-azure-cli-windows), it appears that https://aka.ms/installazurecliwindowsx64 is a link to "latest x64 installer." Could we try that for `path:`?

DannyBrito: it might be a bit of a problem, to my understanding, we need [`product_id` field](https://docs.ansible.com/ansible/2.7/modules/win_package_module.html#win-package-module:~:text=This%20SHOULD%20be%20set%20when%20the%20package%20is%20not%20an%20MSI%2C%20or%20the%20path%20is%20a%20url%20or%20a%20network%20share%20and%20credential%20delegation%20is%20not%20being%20used) when path is an URL and I think that ID will be different to each version

DannyBrito: There is another possible flow for this by replacing the `win_package` task for 3 tasks to unpin version:
1. Download MSI Installer
2. Install it.
3. Clean it

Let me know if prefer that I could look into that!

mboersma: If that flow avoids the need for `product_id`, then let's do try that. Then we wouldn't have to remember to update the version (because spoiler alert: we won't remember to update it :smile:).

jsturtevant:
Is this something we want on every image produced or should it be configurable?  It sounds like we install it on linux?

mboersma:
> Is this something we want on every image?

I think we only want it when building images for the Azure provider, and that's how it is for LInux as well AFAICT. (Despite the generic name `debian.yml`, it looks like it's only included by `azure.yml`.)

images/capi/ansible/windows/roles/providers/tasks/azure.yml
DannyBrito: ok new flow is updated in branch.

mboersma (commented):
/lgtm
/assign @jsturtevant

PR #1332, opened 2023-10-25
QEMU RockyLinux 9 and Ubuntu 23.04

feitnomore:
This PR adds RockyLinux 9 as well as Ubuntu 23.04 LTS support for QEMU.

**Additional context**
This PR adds support for Rocky Linux 9.2 and Ubuntu 23.04 LTS for QEMU CAPI.
@AverageMarcus F.Y.I.

### DIFF
--- a/images/capi/packer/qemu/packer.json
+++ b/images/capi/packer/qemu/packer.json
@@ -8,6 +8,7 @@
         "{{user `boot_command_suffix`}}"
       ],
       "boot_wait": "{{user `boot_wait`}}",
+      "cpu_model": "host",
       "cpus": "{{user `cpus`}}",
       "disk_compression": "{{ user `disk_compression`}}",
       "disk_discard": "{{user `disk_discard`}}",
--- a/images/capi/packer/qemu/qemu-ubuntu-2204.json
+++ b/images/capi/packer/qemu/qemu-ubuntu-2204.json
@@ -1,5 +1,5 @@
 {
-  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/22.04/'<enter><wait>initrd /casper/initrd<enter><wait>boot<enter>",
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/22.04/'<enter><wait>initrd /casper/initrd<enter><wait><wait><wait>boot<enter>",
   "build_name": "ubuntu-2204",
   "distribution_version": "2204",
   "distro_name": "ubuntu",
--- a/images/capi/packer/qemu/qemu-ubuntu-2304.json
+++ b/images/capi/packer/qemu/qemu-ubuntu-2304.json
@@ -1,0 +1,13 @@
+{
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/23.04/'<enter><wait><wait><wait><wait>initrd /casper/initrd<wait><wait><enter><wait>boot<enter>",
+  "build_name": "ubuntu-2304",
+  "distribution_version": "2304",
+  "distro_name": "ubuntu",
+  "guest_os_type": "ubuntu-64",
+  "iso_checksum": "c7cda48494a6d7d9665964388a3fc9c824b3bef0c9ea3818a1be982bc80d346b",
+  "iso_checksum_type": "sha256",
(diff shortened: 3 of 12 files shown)


images/capi/packer/qemu/packer.json
```diff
@@ -8,6 +8,7 @@
         "{{user `boot_command_suffix`}}"
       ],
       "boot_wait": "{{user `boot_wait`}}",
+      "cpu_model": "host",
```
AverageMarcus, line 11: Will this effect other OSs? This file is used by all the QEMU builds.

images/capi/packer/qemu/qemu-ubuntu-2204.json
```diff
@@ -1,5 +1,5 @@
 {
-  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/22.04/'<enter><wait>initrd /casper/initrd<enter><wait>boot<enter>",
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/22.04/'<enter><wait>initrd /casper/initrd<enter><wait><wait><wait>boot<enter>",
```
AverageMarcus, line 2: Is this related to this PR?

images/capi/packer/qemu/qemu-ubuntu-2304.json
```diff
@@ -0,0 +1,13 @@
+{
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz --- autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/23.04/'<enter><wait><wait><wait><wait>initrd /casper/initrd<wait><wait><enter><wait>boot<enter>",
+  "build_name": "ubuntu-2304",
+  "distribution_version": "2304",
+  "distro_name": "ubuntu",
+  "guest_os_type": "ubuntu-64",
+  "iso_checksum": "c7cda48494a6d7d9665964388a3fc9c824b3bef0c9ea3818a1be982bc80d346b",
+  "iso_checksum_type": "sha256",
+  "iso_url": "https://releases.ubuntu.com/23.04/ubuntu-23.04-live-server-amd64.iso",
```
AverageMarcus, line 9: Does this ISO change with new updates to 23.04? If so the checksum will eventually not match, right?

AverageMarcus (commented):
Looks good. Just a few small comments / questions.

Aside from that, could you please squash your commits before we merge? 😄

images/capi/packer/qemu/packer.json
feitnomore: So, I don't think this will really cause any side effect. The newer kernels were suffering from a `kernel panic` while booting the installation, and doing some Google research this was the solution I've found.
I've built `qemu-ubuntu-2004`, `qemu-ubuntu-2204`, `qemu-ubuntu-2304`, `qemu-rockylinux-8` and `qemu-rockylinux-9` on a KVM nested virtualization, and everything worked fine.

images/capi/packer/qemu/qemu-ubuntu-2204.json
feitnomore: So, while testing, depending on the performance, I've got issues. If you are using a slower system, you may need to wait a bit more. In my case, I needed to add this, as well as increase some timeouts on `ansible.cfg` so that I could have success. The `ansible.cfg` changes I have not commited.

images/capi/packer/qemu/qemu-ubuntu-2304.json
feitnomore: Unfortunately, I don't see a specific file for 23.04.1 or .2 as we have for 22. The only file I see is this one.

AverageMarcus: 😞  This is starting to become a recurring issue for us. 

@mboersma did we ever come up with any way of handling this when there isn't an `old.ubuntu.com` link available? 

I find it so weird that Ubuntu doesn't produce static images for their current releases.

feitnomore: https://releases.ubuntu.com/23.04/SHA256SUMS might be a way. A possible solution could be to load the values from that file dynamically. 
```
curl -sS https://releases.ubuntu.com/23.04/SHA256SUMS | grep "ubuntu-23.04-live-server-amd64.iso" | awk '{print $1}'
```

AverageMarcus: I'm more concerned about reproducible builds. If the ubuntu iso is updated with a breaking change it could break for image-builder users without there being any changes / releases to image-builder.

We currently doing have a way around this I don't think, it's just really annoying. 😞

AverageMarcus:
I'm not going to block based on the iso issue as we have that elsewhere already. I'll open an issue to address the problem more generally.

/lgtm

@feitnomore before approval could you please squash the commits? :)

PR #1337, opened 2023-11-06
Set Flatcar OS_TYPE as  other4xLinux64Guest

erkanerol:
vSphere gives an error while importing OVA file because of the template. I started a thread in the Kubernetes' Slack about this issue https://kubernetes.slack.com/archives/C01E0Q35A8J/p1698754064351559

~My understanding is the value should be `other3xLinux64Guest`.  I changed the OVF file of the image manually and tested importing. It worked. The value in upstream Flatcar repo is also `other3xLinux64Guest`. See https://github.com/flatcar/scripts/blob/main/build_library/template_vmware.ovf#L138~

See the discussions below.

### DIFF
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -114,7 +114,7 @@ def main():
                  "rhel8-64": {"id": "80", "version": "8", "type": "rhel8_64Guest"},
                  "rockylinux-64": {"id": "80", "version": "", "type": "rockylinux_64Guest"},
                  "ubuntu-64": {"id": "94", "version": "", "type": "ubuntu64Guest"},
-                 "flatcar-64": {"id": "100", "version": "", "type": "other3xLinux64Guest"},
+                 "flatcar-64": {"id": "100", "version": "", "type": "other4xLinux64Guest"},
                  "Windows2019Server-64": {"id": "112", "version": "", "type": "windows2019srv_64Guest"}}
 
     # Create the OVF file.
--- a/images/capi/hack/image-build-ova.py
+++ b/images/capi/hack/image-build-ova.py
@@ -114,7 +114,7 @@ def main():
                  "rhel8-64": {"id": "80", "version": "8", "type": "rhel8_64Guest"},
                  "rockylinux-64": {"id": "80", "version": "", "type": "rockylinux_64Guest"},
                  "ubuntu-64": {"id": "94", "version": "", "type": "ubuntu64Guest"},
-                 "flatcar-64": {"id": "100", "version": "", "type": "otherLinux64Guest"},
+                 "flatcar-64": {"id": "100", "version": "", "type": "other3xLinux64Guest"},
                  "Windows2019Server-64": {"id": "112", "version": "", "type": "windows2019srv_64Guest"}}
 
     # Create the OVF file.
--- a/images/capi/packer/ova/flatcar.json
+++ b/images/capi/packer/ova/flatcar.json
@@ -21,5 +21,5 @@
   "systemd_prefix": "/etc/systemd",
   "sysusr_prefix": "/opt",
   "sysusrlocal_prefix": "/opt",
-  "vsphere_guest_os_type": "other3xLinux64Guest"
+  "vsphere_guest_os_type": "other4xLinux64Guest"
 }


mboersma:
Makes sense to me. @erkanerol this is still a draft PR; if you want to mark it "Ready for Review" I think we can merge it.

cc: @invidian @jepio @johananl

jepio:
looks good to me as well, thanks!

invidian (approved):

ohauer:
woudn't it be better to use other4xLinux64Guest instead other3xLinux64Guest in this case?

erkanerol:
Now, we have a better understanding of this value. Thanks a lot @ohauer. As you said, the kernel version in the latest Flatcar is `5.x`  so the value should be `other5xLinux64Guest`.

There is a minimum version map in the [referred](https://github.com/vmware/govmomi/blob/main/vim25/types/enum.go#L8814) code
```
minAPIVersionForEnumValue["VirtualMachineGuestOsIdentifier"] = map[string]string{
...
"other3xLinux64Guest":        "5.5",
"other4xLinux64Guest":        "6.7",
"other5xLinux64Guest":        "7.0.1.0",
"other6xLinux64Guest":        "8.0.0.1",
...
```

I tested `other5xLinux64Guest` in my vSphere environment (7.0.3) and it worked too. 

I don't know if `other5xLinux64Guest ` works for vSphere `6.7` and if we really need to care `6.7` even though it is not supported officially [anymore](https://blogs.vmware.com/vsphere/2020/06/announcing-extension-of-vsphere-6-7-general-support-period.html). 

For the sake of supporting `6.7`, we can go with `other4xLinux64Guest` even though the kernel version is `5.x`  in the newest Flatcar images but I don't know what will be the negative side effect of this mismatch.

ohauer:
The usage of the otherNxLinuxGuest can affect the data that is provided from the vmware-tools inside the VM to the host system.
I found an older interesting article on the VMware site [KB71306](https://kb.vmware.com/s/article/71306) that shows a side effect of setting the otherNxLinuxGuest to a higher version without appropriate VMware tools installed.
So setting this to a lower version has less impact then setting it to a higher version and not havving the latest vmware-tools installed

erkanerol:
@ohauer Then I think the best value is `other4xLinux64Guest`. What do you think?

ohauer:
> @ohauer Then I think the best value is `other4xLinux64Guest`. What do you think?
yes, for the current time it looks like the best option / compromise

akutz:
Hi @ravindravmw,

Can you possibly comment on the Tools compatibility regarding the guest OS ID used to deploy a VM?

akutz:
cc @jsavanyo4vmw as well for any help you may be able to provide. John, FWIW, I'm pinging a few folks internally on the tools and VMX teams who may have more info as well.

akutz:
@erkanerol / @ohauer,

The following is from one of the engineers on the Tools team:

> Some background:  there are two GuestOS identifiers:  what the VM creator specifies, and what the guest reports. In a perfect world these match.  The complexity is that new values traditionally need a newer vSphere to have the new guestOS constants in the API (that's the root issue with the KB linked in the thread).  Sine these are API constants, that means if some new Linux distro comes out, unless its a Big Name its got to lie if it wants everything to look right in the API/UI.  And even if its a Big Name, if the version came out after the vSphere, there won't be a constant for it.
>
> There's all sorts of minor tweaks in the host side code based on these values.  The VMX and UI will tune stuff (I'm vague on details, but things like default configs when creating the VM, devices, etc).  We've also got some minor, subtle stuff in guestOps -- if it says its Windows but its not, weird problems happen in filenames (backslash conversion) and certain ops may not work.
>
> We added some extra config options and details to handle the random distro scenario.  William sums it up here: https://williamlam.com/2023/11/new-detailed-guestos-data-in-vsphere-8-0-update-2.html

I'm still digging into the specific question, but hopefully the above helps somewhat.

akutz:
One more bit of info from the same engineer:

> Here the tools.conf commentary on how to change what Tools reports.
https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/tools.conf#L357
Its a bit terse on what those values mean though.

This would enable the image to configure Tools to override what was injected when the VM was deployed.

ohauer:
@akutz
Thanks for the research!
The ability to set the guest operating system in tools.conf opens up a new possibility to adapt this depending on the environment.

erkanerol:
In light of the latest information shared in the PR, I updated the value as `other4xLinux64Guest` to stay on the safe side.

I created another issue to improve the whole mechanism https://github.com/kubernetes-sigs/image-builder/issues/1339

Thanks everyone!

AverageMarcus (commented):
/lgtm
/approve

Thank you so much everyone for the detailed investigation and summary. I've certainly learnt something from it all! 🙂

Issue #1312, opened 2023-09-27
All ansible actions should use fully-qualified collection name


mboersma:
**What steps did you take and what happened:**

Run `make -C images/capi lint` and see that there are numerous warnings from `ansible-lint` that the project is ignoring.

In particular, there are many "FQCN" warnings. It's safer and generally preferable to use the full name of ansible actions–that is, `community.windows.win_firewall_rule` instead of just `win_firewall_rule`, for example.

**What did you expect to happen:**

Running `make -C images/capi lint` shouldn't result in any `fqcn[action]` warnings.

**Anything else you would like to add:**

One way to accomplish this:

```shell
ansible-lint --fix=fqcn images/capi/ansible
make -C images/capi lint-ignore
git add images/capi
git commit -m "Ansible actions use fully-qualified collection names"
```

These commands:
- Tell the linter to fix this `fqcn` type of warning automatically
- Regenerate the `.ansible-lint-ignore` file so these warnings won't be ignored in the future
- Add all the changed files to version control and commit them.

Then run `make -C images/capi lint` to ensure it's passing, and create a PR.

**Environment:**

N/A

/kind bug
/help
/good-first-issue

mboersma:
Let's try to fix #1313 first, since fixing this will also include those changes and it would be saner to review separately.

khannakshat7:
Hey @mboersma, I would like to work on this once #1313 is fixed.
/assign

pranavp2005:
I am new to contributing to opensource. I would like to take this up as my first issue.
/assign

mboersma:
@pranavp2005 I think @khannakshat7 claimed this one first, just fyi.

I think we should wait for #1338 to merge before making this change. Otherwise `ansible-lint --fix=fqcn` will include several unrelated formatting changes.

pranavp2005:
> @pranavp2005 I think @khannakshat7 claimed this one first, just fyi.
> 
> I think we should wait for #1338 to merge before making this change. Otherwise `ansible-lint --fix=fqcn` will include several unrelated formatting changes.

Okay, thank you for informing me.

mboersma:
@pranavp2005 we haven't heard anything from @khannakshat7 so if you're still available to take this on, please do. Sorry to slow things down.

mboersma:
Here are updated commands that should help to create this PR:

```console
cd images/capi
rm .ansible-lint-ignore
ansible-lint --fix=fqcn ansible
make lint-ignore
[...]

PR #1338, opened 2023-11-06
Bump ansible-lint to v6.21.1

mboersma:

Updates ansible-lint to v6.21.1 and fixes some basic formatting issues so that `make lint-fix` is a no-op.

Which issue(s) this PR fixes:

Refs #1312

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/photon.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/photon.yml
@@ -21,4 +21,8 @@
 - name: Install Kubernetes
   command: tdnf install {{ packages }} --nogpgcheck -y
   vars:
-    packages: kubelet-{{ kubernetes_rpm_version }} kubeadm-{{ kubernetes_rpm_version }} kubectl-{{ kubernetes_rpm_version }} kubernetes-cni-{{ kubernetes_cni_rpm_version }}
+    packages: >-
+      kubelet-{{ kubernetes_rpm_version }}
+      kubeadm-{{ kubernetes_rpm_version }}
+      kubectl-{{ kubernetes_rpm_version }}
+      kubernetes-cni-{{ kubernetes_cni_rpm_version }}
--- a/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
@@ -130,7 +130,7 @@
     path: HKLM:\SYSTEM\CurrentControlSet\Services\hns\State
     state: present
     name: HNSControlFlag
-    data: 0x50
+    data: 80
     type: dword
   when: distribution_version == "2019"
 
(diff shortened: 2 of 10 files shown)


mboersma:
Looks like my changes to the RPM vars failed. I'll take another look.

mboersma:
What I learned in getting this to work is that the best way to format a long Jinja2 (Python) expression is to use the YAML folding operator (`>`). Other approaches such as quotes with backslashes are legal, but will be rewritten by `ansible-lint`.

If the expression defines an ansible variable, remove the quotes (`"`) and use the YAML folding operator with the trailing dash to omit a final newline.

```yaml
rpms: >-
  {{ (common_photon_rpms + lookup('vars', 'photon_' + ansible_distribution_major_version + '_rpms' )
    + lookup('vars', 'common_' + build_target + '_photon_rpms')) }}
```

images/capi/ansible/roles/kubernetes/tasks/photon.yml
```diff
@@ -21,4 +21,8 @@
 - name: Install Kubernetes
   command: tdnf install {{ packages }} --nogpgcheck -y
   vars:
-    packages: kubelet-{{ kubernetes_rpm_version }} kubeadm-{{ kubernetes_rpm_version }} kubectl-{{ kubernetes_rpm_version }} kubernetes-cni-{{ kubernetes_cni_rpm_version }}
+    packages: >-
```
AverageMarcus, line 24: This format is so much nicer!

images/capi/ansible/windows/roles/systemprep/tasks/main.yml
```diff
@@ -130,7 +130,7 @@
     path: HKLM:\SYSTEM\CurrentControlSet\Services\hns\State
     state: present
     name: HNSControlFlag
-    data: 0x50
+    data: 80
```
AverageMarcus, line 133: Is this still valid? Seems odd that it was ever defined in hex if it is.

AverageMarcus (commented):
LGTM just one question

images/capi/ansible/windows/roles/systemprep/tasks/main.yml
mboersma: I *think* it's still valid. `80` decimal equals `0x50` in hex, and the docs for the Ansible [`win_regedit`](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_regedit_module.html#id2) task says:

> DWORD and QWORD values should either be represented as a decimal number or a hex value.

I would prefer to leave it alone, but `ansible-lint --fix=none` dislikes the hex version and wanted to rewrite it as `0120`.

AverageMarcus: Ha! That's even worse! 🙈 ok. Can we make sure it's called out somehow in the release notes in case it does cause a problem?

mboersma: Yes, good idea. Maybe @jsturtevant could also take a look, since he knows Windows better than I do.

jsturtevant: using decimal should be fine.  I also think we might be able to drop these flags now.  I will need to confirm with some of the networking team

mboersma:
@AverageMarcus could you give this another review? I think it will unstick #1312 and other future "help wanted" linting updates.

PR #1340, opened 2023-11-08
Add Goss output files to .gitignore

mboersma:

Answers the question "How many times will you mistakenly check these log files in to version control before you add them to .gitignore?" 7. About seven is the answer.

Which issue(s) this PR fixes:

Fixes #

**Additional context**

### DIFF
--- a/images/capi/.gitignore
+++ b/images/capi/.gitignore
@@ -7,3 +7,7 @@
 /.local/bin/
 manifest.json
 **.DS_Store
+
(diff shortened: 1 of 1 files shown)


AverageMarcus (commented):
> Answers the question "How many times will you mistakenly check these log files in to version control before you add them to .gitignore?" 7. About seven is the answer.

This made me laugh! 😆 

/lgtm
/approve

PR #1352, opened 2023-12-05
made gridd and license optional

### LINKED ISSUE #1351: Nvidia role requires a license file and gridd feature version
A quick Bug Report to go with a PR I'm raising.

So, we've started using GPU passthrough at EscherCloud and we don't need the gridd conf or license file to be supplied for the GPU to work as the license we have is to make MIG function correctly. In the process of building images for these nodes we've noticed we can't get a build to work if the gridd feature version and license file are not supplied as they are currently required.

To get around it we have to supplied these files even though they're not required.

/kind bug

drew-viles:
It enables the license file and gridd template to be optional.

This is required because when using passthrough instead of MIG, the gridd feature configuration and license file isn't required so rather than adding items to the image that don't need to be there (and to prevent it as a requirement if someone doesn't have the license file too), the steps in the role are simply ignored if the two values mentioned in the readme update are not supplied.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #1351

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/nvidia/README.md
+++ b/images/capi/ansible/roles/nvidia/README.md
@@ -5,6 +5,7 @@ To install the NVIDIA GPU driver as part of the image build process, you must ha
 Then all you need to do is reference those files in your packer file.
 
 An example of the fields you need are defined below. Make sure to review and change any fields where required.
+If the gridd configuration or licensing .tok file are not required then you can omit the `gridd_feature_type` and `nvidia_tok_location` respectively. 
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


mboersma (commented):
/lgtm

I can't test this directly but it makes sense.

cotzo:
hello. this task still requires the drivers to exist on an S3 bucket. can't we modify so it supports bringing public drivers? thanks

PR #1309, opened 2023-09-26
Update containerd_wasm_shims to v0.9.1

mboersma:
What this PR does / why we need it:

Updates the containerd-wasm-shims to the latest pre-release [v0.9.1](https://github.com/deislabs/containerd-wasm-shims/releases/tag/v0.9.1), which adds a new Wasm runtime. Also handles the new packaging where each shim is in its own archive.

Which issue(s) this PR fixes: 

Fixes #

**Additional context**

I've tested locally with Ubuntu 22.04 images built from this PR and everything seems to work. (I'm hitting the local ClusterIP service endpoints because of security restrictions on Azure dev accounts.)

```shell
$ sudo KUBECONFIG=/etc/kubernetes/admin.conf kubectl get svc
NAME           TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)   AGE
kubernetes     ClusterIP   10.96.0.1        <none>        443/TCP   9m14s
wasm-lunatic   ClusterIP   10.106.103.176   <none>        80/TCP    9s
wasm-slight    ClusterIP   10.97.72.208     <none>        80/TCP    9s
wasm-spin      ClusterIP   10.98.26.207     <none>        80/TCP    9s
wasm-wws       ClusterIP   10.109.105.219   <none>        80/TCP    9s
$ curl http://10.106.103.176/hello  # lunatic
Hello :)
$ curl http://10.97.72.208/hello  # slight
hello world!
$ curl http://10.98.26.207/hello  # spin
Hello world from Spin!
$ curl http://10.109.105.219/hello  # wws
<!DOCTYPE html>
<head>
  <title>Wasm Workers Server</title>
  <meta name="viewport" content="width=device-width,initial-scale=1">
  <meta charset="UTF-8">
  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/water.css@2/out/water.css">
  <style>
    body { max-width: 1000px; }
    main { margin: 5rem 0; }
    h1, p { text-align: center; }
    h1 { margin-bottom: 2rem; }
    pre { font-size: .9rem; }
    pre > code { padding: 2rem; }
    p { margin-top: 2rem; }
  </style>
</head>
<body>
  <main>
    <h1>Hello from Wasm Workers Server </h1>
    <pre><code>Replying to /hello
Method: GET
User Agent: curl/7.81.0
Payload: -</code></pre>
    <p>
      This page was generated by a JavaScript file running in WebAssembly.
    </p>
  </main>
</body>
$ 
```

[...]

PR #1364, opened 2023-12-14
Remove containerd-wasm-shim tarballs after installing

mboersma:

Cleans up the invididual tarballs used to install containerd Wasm shims. This was an oversight in #1309.

Which issue(s) this PR fixes:

Fixes #

**Additional context**

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -176,10 +176,12 @@
     path: /tmp/containerd.tar.gz
     state: absent
 
-- name: Delete containerd-wasm-shims tarball
+- name: Delete containerd-wasm-shims tarballs
   file:
-    path: /tmp/containerd_wasm_shims.tar.gz
+    path: /tmp/{{ item }}_wasm_shims.tar.gz
     state: absent
+  when: containerd_wasm_shims_runtimes | length > 0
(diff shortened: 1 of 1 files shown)


images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -176,10 +176,12 @@
     path: /tmp/containerd.tar.gz
     state: absent
 
-- name: Delete containerd-wasm-shims tarball
+- name: Delete containerd-wasm-shims tarballs
   file:
-    path: /tmp/containerd_wasm_shims.tar.gz
+    path: /tmp/{{ item }}_wasm_shims.tar.gz
     state: absent
+  when: containerd_wasm_shims_runtimes | length > 0
```
jsturtevant, line 183: should we include `ansible_os_family != "Flatcar"`  like we do on unpacking?

mboersma: Good question–I don't think we need to, because Flatcar shares the `Download containerd-wasm-shims` task with other distros. 

https://github.com/kubernetes-sigs/image-builder/blob/25a576b52f576ac513deb889c744b132279b7313/images/capi/ansible/roles/containerd/tasks/main.yml#L33-L43

It's only the `Unpack` task that differs, so this cleanup should work in both cases.

mboersma (commented):
/approve

PR #1368, opened 2024-01-02
Delete all rotated logs

### LINKED ISSUE #381: Delete log files that end in .1 from the machine
Recently https://github.com/kubernetes-sigs/image-builder/pull/331 went in which is a vast improvement in the log behavior, but some logs are rotated by appending a number rather than targz the files.  Just add another entry -- or get fancy with regex for numbers -- to remove anything with a `.1` after it:

```
- name: Delete all logrotated log by rename
  shell:
    cmd: |
      find /var/log -type f -name '*.1' -exec rm {} +
```

hugoprudente:

This regex matches the folllowing patterns found in ubuntu, centos, rhel

```bash
messages-20231224
cloud-init.log.gz.2
dnf.log.1
script_exporter_setup.log-20231215.xz
client.log-20231119.gz
```

Fixes #381 

**Additional context**

Find uses by default `Emacs Regular Expressions` [docs](https://man7.org/linux/man-pages/man1/find.1.html) to to avoid use complex cases like `([0-9]$|gz$)` I went with a more minimalist approach. 

* https://www.gnu.org/software/emacs/manual/html_node/emacs/Regexps.html

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -198,10 +198,10 @@
       find /var/log -type f -iname '*.log' | xargs truncate -s 0
   when: ansible_os_family != "Flatcar"
 
-- name: Delete all logrotated log zips
+- name: Delete all logrotated logs
   shell:
     cmd: |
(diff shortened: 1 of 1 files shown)


mboersma (commented):
/lgtm

PR #1383, opened 2024-01-27
pin packer version to 1.9.5

### LINKED ISSUE #1379: Enforce the use of a supported Packer version

We're unable to support Packer v1.10.0+ due to https://github.com/kubernetes-sigs/image-builder/issues/1246. At time of writing, the latest version of Packer that we support is `v1.9.5` and is [what is installed](https://github.com/kubernetes-sigs/image-builder/blob/8f3ec42484d78a843bad8054fd6a7e410843bc40/images/capi/hack/ensure-packer.sh#L24) if Packer isn't found on the machine.

But if Packer `v1.10+` is found on the machine image-builder will just happily attempt to use that version which can then lead to issues such as https://github.com/kubernetes-sigs/image-builder/issues/1378

We should restrict the version used and if v1.10+ is found we should still download `v1.9.5`. We'll need to update the Makefile also to make sure if uses the locally downloaded binary instead of the system-wide version of Packer.

Alternatively we could not download the version and instead show an error to the user but this means the user then needs to figure out how to downgrade their version.

**Additional context**

/kind feature

kranurag7:
if packer version is greater than 1.9.5 in semver terms then downgrade packer.

Fixes #1379 

### with centos container 
```bash
docker run --rm -it centos:latest sh
# put hack/utils.sh & hack/ensure-packer.sh and +x permissions. 
sh-4.4# ./packer.sh 
Packer is already installed, checking version...
unexpected packer found
```
### with 1.10.0 version of packer 
```bash
$ ./hack/ensure-packer.sh
Packer is already installed, checking version...
unsupported packer version (1.10.0) found
current packer version: 1.10.0 is not supported
Downgrading packer to 1.9.5
Installing packer v1.9.5 in .local/bin
packer_1.9.5_linux_amd64.zip: OK
Archive:  packer_1.9.5_linux_amd64.zip
  inflating: packer
```

**Additional context**

### DIFF
--- a/images/capi/hack/ensure-packer.sh
+++ b/images/capi/hack/ensure-packer.sh
@@ -38,8 +38,28 @@ source hack/utils.sh
 # invoked, so we are timeboxing it to 10 seconds. This shouldn't be the
 # case with Packer installed from Hashicorp releases, which should give
 # us a version number. This helps us distinguish the two Packer executables.
-if (command -v packer && timeout 10 packer version) >/dev/null 2>&1; then exit 0; fi
 
+if (command -v packer) >/dev/null 2>&1; then
+  echo "Packer is already installed, checking version..."
+  # if it's not the hashicorp packer, return "unexpected packer found"
+  if !(timeout 10 packer version) >/dev/null 2>&1; then
+    echo "unexpected packer found";
+    echo "downloading hashicorp packer version v1.9.5"
+  fi
+  existing_packer_version=$(packer version | head -1 | cut -d 'v' -f 2; exit 0)
+  echo "existing packer version: $existing_packer_version"
+  if [ "$existing_packer_version" != "$_version" ]; then
+    echo "unsupported packer version ($existing_packer_version) found"
+    echo "current packer version: $existing_packer_version is not supported"
+    echo "Downgrading packer to ${_version}"
+  else
+    echo "Packer version is as expected"
+    echo "Packer version $existing_packer_version is already installed"
+    exit 0
+  fi
+fi
+
+echo "Installing packer v${_version} in .local/bin"
 mkdir -p .local/bin && cd .local/bin
 
 SED="sed"


images/capi/hack/ensure-packer.sh
```diff
@@ -38,8 +38,23 @@ source hack/utils.sh
...
 # us a version number. This helps us distinguish the two Packer executables.
-if (command -v packer && timeout 10 packer version) >/dev/null 2>&1; then exit 0; fi
 
+if (command -v packer) >/dev/null 2>&1; then
+  echo "Packer is already installed, checking version..."
+  # if it's not the hashicorp packer, return "unexpected packer found"
+  if !(timeout 10 packer version) >/dev/null 2>&1; then
+    echo "unexpected packer found";
+    exit 1;
+  fi
```
AverageMarcus, line 48: We don't want to exit here do we? We want to continue in this script and download Packer.

images/capi/hack/ensure-packer.sh
```diff
@@ -38,8 +38,23 @@ source hack/utils.sh
...
+  existing_packer_version=$(packer version | head -1 | cut -d 'v' -f 2)
+  if [ "$existing_packer_version" != "$_version" ]; then
+    echo "unsupported packer version ($existing_packer_version) found"
+    echo "current packer version: $existing_packer_version is not supported"
+    echo "Downgrading packer to ${_version}"
+  fi
```
AverageMarcus, line 54: We need to exit this script (successfully) if the version is as expected.

mboersma:
I'm seeing this in the logs:

```shell
hack/ensure-packer.sh
Packer is already installed, checking version...
make: *** [Makefile:61: deps-common] Error 141
```

kranurag7:
> I'm seeing this in the logs:
> 
> ```shell
> hack/ensure-packer.sh
> Packer is already installed, checking version...
> make: *** [Makefile:61: deps-common] Error 141
> ```

Hey. sorry for being late here and because of that the patch was open for a while now. :( 
w.r.t. this, I changed the following and I think this should be good now. (thanks to `set -x` 🙃) 
```diff
-   existing_packer_version=$(packer version | head -1 | cut -d 'v' -f 2)
+   existing_packer_version=$(packer version | head -1 | cut -d 'v' -f 2; exit 0)
```

mboersma (approved):
lgtm

@kranurag7 could you squash the commits into one and force-push?

images/capi/hack/ensure-packer.sh
```diff
@@ -38,8 +38,29 @@ source hack/utils.sh
...
 # us a version number. This helps us distinguish the two Packer executables.
-if (command -v packer && timeout 10 packer version) >/dev/null 2>&1; then exit 0; fi
 
+if (command -v packer) >/dev/null 2>&1; then
+  echo "Packer is already installed, checking version..."
+  # if it's not the hashicorp packer, return "unexpected packer found"
+  if !(timeout 10 packer version) >/dev/null 2>&1; then
+    echo "unexpected packer found";
+    echo "downloading hashicorp packer version v1.9.5"
+    exit 0
```
drew-viles, line 48: Wouldn't we need to break out of the `if` and continue to download packer rather than exiting here? Maybe wrap line 50-60 in an else to achieve this?

kranurag7: I think we can just remove `exit 0` on line 48 here and everything works. Reason being, my assumption was that we wanted to exit if the some other packer on centos was found but I think even in that case, we want to continue to download so just removing this will do the work. Let me know if that's okay and I'll remove that. 

Please suggest if you think there's a better approach here.

drew-viles: I think that makes perfect sense! Let's just remove that one line 😃 It'll flow through to the next conditional anyway on line 51

kranurag7: Thanks for the review Drew, I have addressed it via https://github.com/kubernetes-sigs/image-builder/pull/1383/commits/b37d35219f53fec2f511c7549cba56fe4ac2f86e

drew-viles:
/hold

Actually @kranurag7 can you resquash the commits again please?

AverageMarcus (commented):
/lgtm

I'll leave it to @drew-viles to unhold :)

drew-viles:
Sorted. Thanks for this and your patience with us! @kranurag7

PR #1385, opened 2024-01-29
Use specific CNI package only if defined

### LINKED ISSUE #1196: Update CNI plugins to v.1.3.0

CNI plugins 1.3.0 [is released](https://github.com/containernetworking/plugins/releases/tag/v1.3.0) but does not appear yet on the location that image-builder uses.

Update CNI plugins to 1.3.0 once they appear at https://storage.googleapis.com/k8s-artifacts-cni/

This issue is mostly a reminder to check https://storage.googleapis.com/k8s-artifacts-cni/ every now and then

**Additional context**

/kind feature

### LINKED ISSUE #1363: Can't build Kubernetes v1.29.0 image

Trying to build an Ubuntu image for the new k8s v1.29.0 leads to an error finding the v1.2.0 version of the `kubernetes-cni` .deb package. It appears that verion v1.3.0 is now required alongside the v1.29.0 packaging (at least for Debian-ish distros).

But updating to use the v1.3.0 `kubernetes-cni` leads to problems building older versions of Kubernetes. This needs to be made conditional so image-builder can handle all supported versions of k8s.

**Environment:**

Project ([Image Builder for Cluster API](https://github.com/kubernetes-sigs/image-builder/tree/master/images/capi):

- OS (e.g. from `/etc/os-release`, or `cmd /c ver`):
- Packer Version:
- Packer Provider:
- Ansible Version:

/kind bug

mboersma:

For RPM- and deb-based distros, fixes the "which version of CNI goes with this version of Kubernetes?" problem by removing the specific cni version package by default, allowing the package manager to choose a compatible version.

This *should* be backward-compatible in that it will continue to request a specific CNI package version if that variable is specified by the user.

Which issue(s) this PR fixes:

Fixes #1363
Fixes #1196

**Additional context**

I'm not sure `null`-ing out these two variables and adding a bit of conditional python is the cleanest way to handle this: ideas welcome! I guess one alternative would be to remove these two variables entirely.

Make targets that build kubernetes from source should be able to override several variables to achieve v1.29+ builds, but will need to have knowledge already about which CNI version matches up.

I have tried this successfully with k8s v1.26.7 and v1.29.1.

See also https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/#what-are-significant-differences-between-the-google-hosted-and-kubernetes-package-repositories

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/debian.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/debian.yml
@@ -33,4 +33,4 @@
       - kubelet={{ kubernetes_deb_version }}
       - kubeadm={{ kubernetes_deb_version }}
       - kubectl={{ kubernetes_deb_version }}
-      - kubernetes-cni={{ kubernetes_cni_deb_version }}
+      - kubernetes-cni{{ '='+kubernetes_cni_deb_version if kubernetes_cni_deb_version else '' }}
--- a/images/capi/ansible/roles/kubernetes/tasks/mariner.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/mariner.yml
@@ -31,7 +31,7 @@
       - kubelet-{{ kubernetes_rpm_version }}
       - kubeadm-{{ kubernetes_rpm_version }}
       - kubectl-{{ kubernetes_rpm_version }}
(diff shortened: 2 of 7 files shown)


mboersma:
/retitle Use specific CNI package only if defined

AverageMarcus (commented):
/lgtm
/approve

PR #1389, opened 2024-01-31
patch: qemu efi fixes

drew-viles:
# What this PR does / why we need it:

* Set partition size for root to be -1 to prevent issues with disk sizes when someone wants to use the "disk_size" parameter in packer.
* Set the efibootmgr to look for the correct boot entry instead of hard coding it. This prevents random failures when the installer is run again on installation completion and reboot. The qemu method (or Ubuntu) sometimes fails to eject the "CDROM" and as a result, boot from disk fails if the boot order is not correct. This would cause the build to hang on "Waiting for SSH to become available..." indefinitely.

### Testing
After changing these two values I've ran multiple, consistent builds where the disk size matches a values passed in via `disk_size` and I never get a hanging "Waiting for SSH to become available..." anymore.

### DIFF
--- a/images/capi/packer/qemu/linux/ubuntu/http/22.04.efi.qemu/user-data
+++ b/images/capi/packer/qemu/linux/ubuntu/http/22.04.efi.qemu/user-data
@@ -52,7 +52,7 @@ autoinstall:
         type: format
         id: format-0
       - device: disk-sda
-        size: 8023703552
+        size: -1
         wipe: superblock
         number: 2
         preserve: false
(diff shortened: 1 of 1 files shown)


mboersma (commented):
/lgtm

I can't easily test this, but your explanation is convincing and the code changes make sense. It would be nice to get a review from one of the qemu reviewers assigned.

mboersma:
@chrischdi @seanschneeweiss do either of you have time to take a look at this PR? In the absence of a more expert opinion, I'm inclined to approve it.

chrischdi:
Change seems reasonable to me 👍 

/approve

PR #1400, opened 2024-02-08
Set HOME_MODE to 0700 in Flatcar images

weseven:
## What this PR does / why we need it:
This sets `HOME_MODE` to 0700 in /etc/login.defs.

Currently it is unset in Flatcar (coming from Gentoo that keeps the upstream setting).
When `HOME_MODE` is unset, useradd uses the default umask of 022 when creating home directories for users.
This causes home directories for all users created through ignition in Flatcar to be world readable.

Setting `HOME_MODE` to 0700 restrict home permissions to the owner, as is default in RHEL derived distros ([Ubuntu uses 0750](https://git.launchpad.net/ubuntu/+source/shadow/tree/debian/login.defs#n156)).

### DIFF
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -133,3 +133,7 @@
     src: usr/local/bin/etcd-network-tuning.sh
     dest: "{{ external_binary_path }}/etcd-network-tuning.sh"
     mode: "0755"
+
+- name: Set default HOME_MODE in login.defs (Flatcar)
+  shell: sed -ri "/^#?HOME_MODE\>/s:.*:HOME_MODE 0700:" /etc/login.defs
+  when: ansible_os_family == "Flatcar"
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -135,5 +135,5 @@
     mode: "0755"
 
 - name: Set default HOME_MODE in login.defs (Flatcar)
-  shell: sed -ri "/^#?HOME_MODE\>/s:.*:HOME_MODE 0700:" /etc/login.defs
+  shell: sed -ri "s/^#?HOME_MODE\>.*/HOME_MODE 0700/" /etc/login.defs
   when: ansible_os_family == "Flatcar"


mboersma (commented):
Makes sense to me. Let's see if some Flatcar experts can also check this.

/cc @jepio @invidian

invidian:
> Makes sense to me. Let's see if some Flatcar experts can also check this.

I wonder if this is something Flatcar upstream would be interested or what can be done using Ignition, since that would be probably more desirable than per-project hardening like in this case.

weseven:
I can give more context on that, since I made two PRs upstream:
- [this is the RFE on Flatcar](https://github.com/flatcar/Flatcar/issues/1353), the login.defs configuration though comes from the gentoo sys-apps/shadow package 
- [this is a PR to set the default on the gentoo package](https://github.com/gentoo/gentoo/pull/35231), but gentoo developers lean towards keeping the upstream default (`HOME_MODE` unset).

In flatcar the file can be changed through ignition, true, but from my tests users created through the same ignition file will still have world readable home directories (that is, the useradd calls during ignition do not use the overridden value).

I do believe setting `HOME_MODE` to 0700 is a safer default, and in fact world readable home directories are a basic check in security assessments. Keeping it unset will have a vanilla flatcar image fail the basic checks (e.g., CIS Benchmarks). 

This might also be useful to apply to other distros: RedHat based distributions already set it to 0700, Ubuntu sets it to 0750, but [ Debian for example](https://salsa.debian.org/debian/shadow/-/blob/master/debian/login.defs?ref_type=heads#L156) leaves it commented as the [upstream package](https://github.com/shadow-maint/shadow/blob/3e59e9613ec40c51c19c7bb5c28468e33a4529d5/etc/login.defs#L207).

invidian:
If this cannot be done via Ignition, I guess the change is fine, although it might be breaking for users which rely on that read access (I hit similar issue with Mariner recently). I always have a feeling that upstream image-builder should provide minimal viable images while hardening should be done for specific use-cases, similar to all other customizations, but I might be wrong, so I will let others decide :)

drew-viles:
My two cents (from someone, who in reality, knows zero about Flatcar).

I'd be happy for this to go in because as a general rule because it hardens security and I would like to see a more security focused image coming out of the builder. Whilst we shouldn't be dictating the security posture of everyone, we should be providing as secure an image as is feasibly possible without hindering peoples development and deployments.

That being said, I'd also like to see how it goes forward in the Flatcar repo because if it's fixed upstream, it seems a bit redundant to then have it set within image-builder too.

images/capi/ansible/roles/node/tasks/main.yml
```diff
@@ -133,3 +133,7 @@
     src: usr/local/bin/etcd-network-tuning.sh
     dest: "{{ external_binary_path }}/etcd-network-tuning.sh"
     mode: "0755"
+
+- name: Set default HOME_MODE in login.defs (Flatcar)
+  shell: sed -ri "/^#?HOME_MODE\>/s:.*:HOME_MODE 0700:" /etc/login.defs
```
jepio, line 138: to make it slightly more (could drop the `\>` thing as well
```suggestion
  shell: sed -ri "s/^#?HOME_MODE\>.*/HOME_MODE 0700/" /etc/login.defs
```

jepio:
> In flatcar the file can be changed through ignition, true, but from my tests users created through the same ignition file will still have world readable home directories (that is, the useradd calls during ignition do not use the overridden value).
> 

FWIW: the way to do it through ignition would be like this:
```
variant: flatcar
version: 1.0.0
storage:
  directories:
  - path: /home/jeremi
    mode: 0700
passwd:
  users:
  - name: jeremi
```
(I couldn't get it to work for the default core user though).

> That being said, I'd also like to see how it goes forward in the Flatcar repo because if it's fixed upstream, it seems a bit redundant to then have it set within image-builder too.

I'd say it's probably a good idea - it's a better default and will be available faster if the change is done in image-builder.

jepio (approved):

mboersma (approved):
/lgtm

Seems like we have a rough consensus that making this change here is worthwhile, although ideally it could be done upstream or in ignition.

mboersma (approved):
/approve

PR #1402, opened 2024-02-12
patch: new GPU role to support both AMD and NVIDIA drivers

drew-viles:

This PR contains an addition to support adding the AMD GPU drivers into an image.
More specifically it installs the `amdgpu-installer` which can then be passed a `--usecase` that allows the user to define what they want to install. IE: `dkms`, `graphics` and much more based on the [ROCm documentation](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/amdgpu-install.html#use-cases)

I've moved the Nvidia and AMD tasks into a new GPU role so that should anyone want to down the line, more options can be added - as before it's an optional role.

I've tested both an NVIDIA and AMD build using the OpenStack provisioner and both have completed successfully.

N/A

**Additional context**


mboersma (approved):
/lgtm
/approve

I'm not a GPU expert, but I've worked with the NVIDIA support in CAPZ and this looks like a safe addition. Nice work @drew-viles!

PR #1373, opened 2024-01-11
CAPI: Add option to automatically set system-reserved using formulas shared with EKS and GKE.

randomvariable:
What this PR does / why we need it:

Automatically set system reserved if it's not been specified in /etc/sysconfig/kubelet . This prevents OOM in many cases. This is opt-in and requires `kubernetes_enable_automatic_resource_sizing` to be set.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #

**Additional context**
Add any other context for the reviewers

PR #1403, opened 2024-02-13
patch: fix boolean condition for kubernetes_enable_automatic_resource_sizing

kolorful:
## Change description

- Fix a bug where kubernetes_enable_automatic_resource_sizing is set to "false" by default but it is interpreted as non-falsey String

## Related issues
- https://github.com/kubernetes-sigs/image-builder/pull/1373

## Additional context
- Duirng our cluster upgrade (only two patch versions), we discovered `/run/kubelet/extra-args.env` (content: `KUBELET_EXTRA_ARGS= --system-reserved=cpu=230m,memory=5052Mi`) is added which cause node to have way less allocatable memory. We've patched it internally and verified the fix works.
- Ref: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#conditionals-based-on-variables

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/main.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/main.yml
@@ -81,7 +81,7 @@
     owner: root
     group: root
     mode: "0755"
-  when: kubernetes_enable_automatic_resource_sizing
+  when: kubernetes_enable_automatic_resource_sizing | bool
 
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm

AverageMarcus:
/lgtm
/approve

Thanks!

PR #1407, opened 2024-02-13
Remove unneeded pip and package upgrade

mboersma:
## Change description

Removes a stanza we used to set things up for installing the Azure CLI without a `pip` error, but this workaround is ancient and doesn't seem to be needed now.

## Related issues

- Fixes #


drew-viles (approved):
/lgtm

As this only touches Azure specific builds, and you know much more about this than I, I will trust you've done the due diligence as I can't test this myself. 😃 Seems sensible though from my POV.

AverageMarcus (commented):
/lgtm
/approve

PR #1412, opened 2024-02-26
Update qemu ubuntu 22.04

mujtaba-ahmed12:
Ubuntu has released newer version of ISO. Version in config is no longer available.

https://releases.ubuntu.com/22.04/ubuntu-22.04.3-live-server-amd64.iso?checksum=sha256%3Aa4acfda10b18da50e2ec50ccaf860d7f20b389df8765611142305c0e911d16fd

## Related issues

- Fixes #

### DIFF
--- a/images/capi/packer/qemu/qemu-ubuntu-2204-efi.json
+++ b/images/capi/packer/qemu/qemu-ubuntu-2204-efi.json
@@ -5,9 +5,9 @@
   "firmware": "OVMF.fd",
   "guest_os_type": "ubuntu-64",
   "http_directory": "./packer/qemu/linux/ubuntu/http/22.04.efi.qemu",
-  "iso_checksum": "a4acfda10b18da50e2ec50ccaf860d7f20b389df8765611142305c0e911d16fd",
+  "iso_checksum": "45f873de9f8cb637345d6e66a583762730bbea30277ef7b32c9c3bd6700a32b2",
   "iso_checksum_type": "sha256",
-  "iso_url": "https://releases.ubuntu.com/22.04/ubuntu-22.04.3-live-server-amd64.iso",
+  "iso_url": "https://releases.ubuntu.com/22.04/ubuntu-22.04.4-live-server-amd64.iso",
   "os_display_name": "Ubuntu 22.04",
   "shutdown_command": "shutdown -P now",
   "unmount_iso": "true"


AverageMarcus:
I didn't realise that Ubuntu unpublished ISOs when a new one was out 😳 Has that always been the case?

/lgtm
/assign @mboersma

chrischdi:
> I didn't realise that Ubuntu unpublished ISOs when a new one was out 😳 Has that always been the case?
> 
> /lgtm /assign @mboersma

I think in the past we also considered to refer from `old-releases` which then does not get removed https://github.com/kubernetes-sigs/image-builder/pull/1251#discussion_r1297373965

AverageMarcus:
ah yes! I remember now.

mboersma (approved):
/lgtm
/approve

pawcykca:
Changes from this PR have been overridden by #1420 - czy to było zamierzone?

chrischdi:
> Changes from this PR have been overridden by #1420 - was that intentional?

I think you refer to this change:

https://github.com/kubernetes-sigs/image-builder/pull/1420/files#diff-224d60561647ff209dc20bb7fcc336faeb55ebea98e49b005c2cb8af13eb22a9R10

Which did change `https://releases.ubuntu.com/22.04/ubuntu-22.04.4-live-server-amd64.iso` to `https://old-releases.ubuntu.com/releases/22.04/ubuntu-22.04.3-live-server-amd64.iso`.

I think that's intended to have a stable url which does not go away, xref:

https://github.com/kubernetes-sigs/image-builder/pull/1412#issuecomment-1964587464

AverageMarcus:
Yeah, we don't really have a good solution to this. 😞  The `releases.ubuntu.com` URLs end up no longer working when a new release is available so image-builder would suddenly stop working for people without any changes being made.

The `old-releases.ubuntu.com` don't go away but they also don't include the very latest release. 

I'm not sure how to solve this.

pawcykca:
In this situation information about this PR in last release is a little bit confusing.

AverageMarcus:
I've started a discussion on Slack about this: https://kubernetes.slack.com/archives/C01E0Q35A8J/p1710405050849849

Going to see what solutions we can come up with to avoid weirdness going forward. I agree that right now it's confusing and not really how we want it to work.

PR #1415, opened 2024-02-27
fix: Increase waiting for GRUB

CiraciNicolo:
## Change description

- Increase the waiting for the next GRUB subcommand, this align the waiting time to Ubuntu 2304. The wait time is currently not enough, and the next subcommand is typed incorrectly.
- Update Ubuntu ISO image to latest available

## Related issues

- Fixes #

## Additional context

This fixes an issue that caused a load panic in QEMU and causing a timeout by packer.

This PR is a duplicate of #1414 since I used a wrong email address for the commit, this fixes that issue.

PR #1420, opened 2024-03-01
Update ISO URLs to latest, add script to update ISO checksums

abhay-krishna:
* Updating a few ISO URLs pointing to images that got removed from the downloads mirror
* Bump some Ubuntu ISO URLs to the latest stable point releases
* Add script to update Ubuntu ISO URL checksums

PR #1421, opened 2024-03-02
Fix qemu ubuntu 22.04 URL

weizhouapache:
## Related issues

- Fixes wrong URL for ubuntu22.04
see https://github.com/kubernetes-sigs/image-builder/pull/1415#issuecomment-1974763022


AverageMarcus:
Thanks @weizhouapache but we currently have an open PR to fix this (and update a bunch of other URLs) here: https://github.com/kubernetes-sigs/image-builder/pull/1420

weizhouapache:
> Thanks @weizhouapache but we currently have an open PR to fix this (and update a bunch of other URLs) here: #1420

Ok, thanks @AverageMarcus 
closing this then

PR #1427, opened 2024-03-13
Add raw image building support for rhel-9 and rhel-9 efi

### LINKED ISSUE #1119: Support of RHEL 9
Can you please support RHEL 9 with image builder?

/kind feature

vignesh-goutham:
## Change description

Adding Image building support on Raw provider for RHEL-9 and RHEL-9 EFI. As part of the change, the kickstart files for RHEL raw are now symlinked to qemu's RHEL kickstart files as opposed ova's, since raw provider is much more closer to qemu than ova. 

CPU Model for raw is set to `host` as opposed to the default `qemu64`. While this is mainly required for RHEL 9 image building, this works well for other raw image builds as well.

## Related issues

- Partially Fixes #1119

### DIFF
--- a/images/capi/packer/raw/linux/rhel
+++ b/images/capi/packer/raw/linux/rhel
@@ -1,1 +1,0 @@
-../../ova/linux/centos
--- a/images/capi/packer/raw/packer.json
+++ b/images/capi/packer/raw/packer.json
@@ -8,6 +8,7 @@
         "{{user `boot_command_suffix`}}"
       ],
       "boot_wait": "{{user `boot_wait`}}",
+      "cpu_model": "{{user `cpu_model`}}",
       "cpus": "{{user `cpus`}}",
       "disk_compression": "{{ user `disk_compression`}}",
       "disk_discard": "{{user `disk_discard`}}",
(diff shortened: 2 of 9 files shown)


images/capi/packer/raw/linux/rhel
```diff
@@ -1 +1 @@
-../../ova/linux/centos
```
AverageMarcus, line 1: I'm wondering if we should replace this symlink with just copying over the actual files. We've been caught out before where a change to the target files ended up breaking something else unexpectedly. 

@mboersma What do you think? Duplicate the files rather than potential surprises in the future?

The alternative would be to move the files out of a specific provider and make it clear that they are shared with multiple providers.

images/capi/packer/raw/packer.json
```diff
@@ -8,6 +8,7 @@
         "{{user `boot_command_suffix`}}"
       ],
       "boot_wait": "{{user `boot_wait`}}",
+      "cpu_model": "host",
```
AverageMarcus, line 11: Can we update this to use a `user` directive so that people can override it if needed?

vignesh-goutham: Sure thing

images/capi/packer/raw/linux/rhel
AverageMarcus: Lets copy them over to actual files. I'd like to get rid of these symlinks unless they're _really_ clear that they are being shared.

AverageMarcus: Once that's done, would you mind squashing your commits down please?

PR #1429, opened 2024-03-15
Add NIC exclusion to VMware tools conf in VMware Windows OVA

### LINKED ISSUE #1428: VMware Windows node OVA does not exclude addresses from CNI interfaces
## Environment

- **Make target:** `make build-node-ova-vsphere-windows-2022-efi`
- **Run using container image? (Y/N): N**
- **Environment vars:**
```

```
- **Vars file:**
```

```

* Create a Windows node OVA
* The VM instance created from the OVA may contains addresses on interfaces created by CNI
* This may result improper node IP reported by vSphere CPI

* Exclude the addresses reported by CNI

## Relevant log output

---

/kind bug

laozc:
## Change description
Currently the node created from VMware Windows OVA could report multiple NIC addresses as CNI may create multiple NICs on nodes. This breaks the assumption from vSphere CPI which may result in wrong IP address to be set on Node.
This commit fixes the issue by setting exclude-nics for VMware tools.

- Fixes #1428

### DIFF
--- a/images/capi/ansible/windows/roles/providers/files/vmware/vmtools/tools.conf
+++ b/images/capi/ansible/windows/roles/providers/files/vmware/vmtools/tools.conf
@@ -1,0 +1,2 @@
+[guestinfo]
+exclude-nics=vEthernet*,antrea-*,cali*,cilium*,lxc*,ovs-system,flannel*,veth*,vxlan_sys_*,genev_sys_*,gre_sys_*,stt_sys_*,????????-??????
--- a/images/capi/ansible/windows/roles/providers/tasks/vmware.yml
+++ b/images/capi/ansible/windows/roles/providers/tasks/vmware.yml
@@ -1,3 +1,6 @@
+# Copyright 2024 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 
@@ -9,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-- ansible.builtin.include_tasks: azure.yml
-  when: packer_builder_type.startswith('azure')
-
-- ansible.builtin.include_tasks: vmware.yml
-  when: packer_builder_type is search('vmware') or packer_builder_type is search('vsphere')
+- name: Create provider vmtools config file
+  ansible.windows.win_copy:
+    src: vmware/vmtools/tools.conf
+    dest: 'C:\ProgramData\VMware\VMware Tools\tools.conf'


images/capi/ansible/windows/roles/providers/files/vmware/vmtools/tools.conf
```diff
@@ -0,0 +1,2 @@
+[guestinfo]
+exclude-nics=vEthernet*,antrea-*,cali*,cilium*,lxc*,ovs-system,flannel*,veth*,vxlan_sys_*,genev_sys_*,gre_sys_*,stt_sys_*,????????-??????
```
AverageMarcus, line 2: Are those `????` at the end really correct?

AverageMarcus:
Looks ok from what I know but I'm not really comfortable with Windows so assigning James.

/assign @jsturtevant 

I think that failed test is a flake so rerunning...

/test pull-azure-vhds

images/capi/ansible/windows/roles/providers/files/vmware/vmtools/tools.conf
laozc: The config file for Linux also has the same setting.
https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/providers/files/etc/vmware-tools/tools.conf
I guess there could be some CNI plugins out there which may set the interface name to this strange format.
So I left them here to keep consistency until we met issues.

AverageMarcus: Weird. I wonder what its for 🤔  If its there for Linux then I'm fine with it. Just curious.

jsturtevant:
I am not familiar with VMWare config files but looks ok
/lgtm

/assign @knabben 
for a VMware related work

knabben:
yes, looks fine.. confirming with @laozc in what scenario he is running it, if this is validated already, +1 for the change.

laozc:
We noticed that the issue would appear when Antrea CNI is used.
There are multiple NICs created by Antrea agent when the workload is deployed to Windows node.
The IP addresses collected by VMware tools show up in a random order and the node IP becomes unreliable.

images/capi/ansible/windows/roles/providers/files/vmware/vmtools/tools.conf
randomvariable: Though it's not harmful, I'm guessing the first part of ????????-?????? is about predictive interface names, e.g. `enp5s0f1`, and the second part is some subinterface format. I should just ask internally. Anyway, it probably wasn't relevant for Windows.

PR #1431, opened 2024-03-18
Revert updates to Ubuntu 2004 ISO URLs

abhay-krishna:
This PR reverts updates to Ubuntu 20.04 URLs changed in #1420 since we need to figure out a way to migrate these to the subiquity installer from the debian-installer preseed approach before the URLs can be changed.


mboersma (approved):
/lgtm

mboersma (approved):
/approve

PR #1432, opened 2024-03-18
Place RHEL kickstart files for raw builds in `http` directory

abhay-krishna:
By default, image-builder expects the kickstart configs to be placed inside the `<OS>/http/<OS version>` folder to match the `http_directory` field. This PR just moves the existing version folders (containing the KS config files) to the `http` directory.


thebsdbox (approved):
LGTM

drew-viles (approved):
/lgtm - aligns with what we're seeing in the qemu setup.

Also, note for the future, it looks like there is no packer config for RHEL 7 so it's currently not used by anything that I can see.

abhay-krishna:
> /lgtm - aligns with what we're seeing in the qemu setup.
> 
> 
> 
> Also, note for the future, it looks like there is no packer config for RHEL 7 so it's currently not used by anything that I can see.

Yup I saw but I will include the removal in another cleanup PR where I will be removing the Ubuntu 18.04 preseed files. Noticed those ones lying around too. If you would like me to include both the cleanups in this PR, let me know.

mboersma (approved):
/lgtm
/approve

PR #1433, opened 2024-03-19
Replace symlink OS folders with actual folders, remove unused kickstart/preseed configs

abhay-krishna:
This PR extends some of the work done in #1427 to replace the OS folder symlinks with the actual folders and content to isolate changes to a specific OS or OS version and avoid breaking something unintentionally. It also removes unused autoinstall config files for builds whose Packer config doesn't exist.


AverageMarcus:
Wow the GitHub PR view really makes this hard to follow doesn't it? 😆 

I took a look at the branch directly and looks good to me :)

/lgtm
/assign @mboersma

mboersma (approved):
/lgtm
/approve

PR #1437, opened 2024-03-28
Support other flavour of CRI

CiraciNicolo:
## Change description

This PR implements support for different CRI implementations, instead of the default containerd. At this moment the only other CRI implementation supported is CRI-O. By default containerd is still the default CRI, to start using CRI-O references the packer configuration called "qemu-ubuntu-2204-crio.json".

Currently the only operating systems supported are debian based and centos based. It should also work on rhel but not tested.

The implementation can be done following this steps:
- create an ansible role for your CRI flavour
- implement default packer variables
- create builder target and makefile target

Both ansible role and default values should be called as your flavour, in this way you can set the ```cri_flavour``` build variable and all tasks/var should be included.

Be also careful to correctly default the variables in ```images/capi/packer/config/ansible-args.json``` and implement the goss specs.

## Related issues
No issues are addressed by this PR

This PR also includes the following fixes:
- Fix incorrect boolean evalutation for GVisor installation if enabled. Be careful that this fixes could change the expected behaviour if you are leveraging Gvisor

PR #1442, opened 2024-04-11
Support other flavour of CRI

CiraciNicolo:
## Change description

This PR implements support for different CRI implementations, instead of the default containerd. At this moment the only other CRI implementation supported is CRI-O. By default containerd is still the default CRI, to start using CRI-O references the packer configuration called "qemu-ubuntu-2204-crio.json".

Currently the only operating systems supported are debian based and centos based. It should also work on rhel but not tested.

The implementation can be done following this steps:
- create an ansible role for your CRI flavour
- implement default packer variables
- create builder target and makefile target

Both ansible role and default values should be called as your flavour, in this way you can set the ```cri_flavour``` build variable and all tasks/var should be included.

Be also careful to correctly default the variables in ```images/capi/packer/config/ansible-args.json``` and implement the goss specs.

## Related issues
No issues are addressed by this PR

This PR also includes the following fixes:
- Fix incorrect boolean evalutation for GVisor installation if enabled. Be careful that this fixes could change the expected behaviour if you are leveraging Gvisor

## Additional context

This PR is a de facto continuation of #1437, but all my commits have been cherry picked to a dedicated branch.


CiraciNicolo:
Hi! Am I misunderstanding, or are there problems in the pipeline?

```
Powering off VirtualMachine:vm-276295... Error: The attempted operation cannot be performed in the current state (Powered off).
```

mboersma:
/retest

I *think* that's a flake. We'll see.

CiraciNicolo:
Regarding this PR, we were discussing the feasibility of deprecating some variables. Before merging, can we continue with that? It would be pretty nice to unify the variables.

PR #1444, opened 2024-04-19
photon: Ensure grub2 and gawk are installed

randomvariable:
## Change description

Can help if user wants to change kernel options post-boot

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -71,6 +71,8 @@ common_photon_rpms:
   - dbus-python3
   - distrib-compat
   - ebtables
+  - gawk
+  - grub2
   - net-tools
   - openssl-c_rehash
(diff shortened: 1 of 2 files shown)


mboersma:
Should we install these by default though? I think the overall idea is to produce minimal images, so you could argue this is a user-specific addition.

mboersma:
Could we make this an optional install that defaults to false, like in #1452?

drew-viles:
I'd agree these should be optional additions as not everyone is going to want/need these packages installed.

mboersma:
Another way to make this an optional install is to use the `debug_tools` variable that we have defined elsewhere (although it only seems to be used currently in Windows builds). I did this in #1458, for example.

randomvariable:
I'll make these changes, apparently the Photon team says grub2-mkconfig isn't used, which is interesting.

PR #1447, opened 2024-04-21
set python version to 3.9 minimum

cpanato:
## Change description

- set python version to 3.9 minimum

getting this error when building `rhel-8` for capg 

```
Python 3.6.x is no longer officially supported by the Google Cloud CLI\nand may not function correctly. 
```

job: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-image-builder-gcp-all-nightly/1781957253432086528

/assign @mboersma @dims 

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/roles/python/defaults/main.yml
+++ b/images/capi/ansible/roles/python/defaults/main.yml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ---
-pypy_python_version: "3.6"
+pypy_python_version: "3.9"
 pypy_version: 7.2.0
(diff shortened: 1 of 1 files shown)


cpanato:
/hold

looks like there is not the place : https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/1447/pull-image-builder-gcp-all/1782072065260523520

:/

dims:
/close 

( based on last comment from @cpanato )

Issue #1422, opened 2024-03-04
Add custom files into the image


drew-viles:
## Is your feature request related to a problem? Please describe.
A user has requested that we support adding custom files into the image builder code.
the EKS Anywhere supports this already so we should be able to replicate something similar to what they have done over there in this project.

## Additional context

---

/kind feature

hugoprudente:
I have the same requirement and I'm the following capabilities to add my extra-role that adds my files and are able to template them too.

```yaml
node_custom_roles_pre
node_custom_roles_post
```

I'm not sure if doing like EKSA is doing would be generic enough as some times you need to template things and not only `file` copy into it. 

So maybe include on the documentation examples on how to achieve copy of files and templating of files would cover this with more flexibility.

PR #1452, opened 2024-04-26
Adding optional security tools

drew-viles:
## Change description

This adds some optional security tooling to the image if users/orgs require it.
Some orgs may not allow privileged pods to run in their environments to run things like the trivy operator or falco. This PR allows the user to install the binaries for those tools directly into the image. 

This PR doesn't support the configuring of Falco via any configuration files as we shouldn't be telling people how to configure this. This should be done using another method, for example copying in files using a feature as requested in [this Issue](https://github.com/kubernetes-sigs/image-builder/issues/1422) once it has been worked on and completed (which I'll try and find the time for asap), or pulling them from an S3 endpoint etc.

I've been using this in my repo for near on a year now and have tested builds both with and without the tools installed so I don't expect any issues including this role as an optional addon.

I'm happy to wait until the next office hours to discuss this if needs be 😃

### DIFF
--- a/images/capi/ansible/roles/security/tasks/falco.yml
+++ b/images/capi/ansible/roles/security/tasks/falco.yml
@@ -1,0 +1,57 @@
+# Copyright 2024 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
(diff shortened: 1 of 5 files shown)


images/capi/ansible/roles/security/tasks/falco.yml
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+- name: Add Falco package signing key
+  ansible.builtin.apt_key:
+    url: https://falco.org/repo/falcosecurity-packages.asc
+    state: present
+  when: ansible_distribution == "Ubuntu"
```
mboersma, line 21: It might be more generalizable if these were `when: ansible_os_family == "Debian"`, unless there's something actually Ubuntu-specific about these packages. (I think image-builder prefers this test rather than Ubuntu as a distro, but we're not very consistent.)

mboersma (commented):
This looks good to me. I just had one question about checking for Ubuntu vs. Debian.

This has been helpful to you in the real world for a while, so I think others will find value here too. 👍🏻

mboersma (approved):
/lgtm

mboersma (approved):
/approve

Issue #1406, opened 2024-02-13
Support Ubuntu 24.04 LTS (Noble Numbat)


mboersma:
## Describe the solution you'd like

The next long-term-support release of Ubuntu Linux will arrive at the end of April. image-builder should be ready to support it, although it may require multiple PRs and contributions or testing from several stakeholders.

## Additional context

---

/kind feature

mboersma:
/remove-lifecycle stale

We're pretty close to calling this done!

AverageMarcus:
@mboersma Whats outstanding on this?

mboersma:
/close

> What's outstanding on this?

Just #1503 for OVA and RAW, but we don't need to block the epic on those. The idea was to get *most* providers to suport 24.04, and I think we're there.

PR #1453, opened 2024-04-26
add Ubuntu 24.04 LTS support for Nutanix

tuxtof:
## Change description

This PR add the support of Ubuntu 24.04 LTS for the Nutanix plaform

## Additional context

one global deb dependency `python3-distutils` was removed because useless for old distri and no more exist in Ubuntu 24.04 LTS

## Related issues
 * Refs [Support Ubuntu 24.04 LTS (Noble Numbat) #1406](https://github.com/kubernetes-sigs/image-builder/issues/1406)

### DIFF
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -58,7 +58,6 @@ common_debs:
   - libnetfilter-acct1
   - libnetfilter-cttimeout1
   - libnetfilter-log1
-  - python3-distutils
   - python3-netifaces
   - python3-pip
   - socat
--- a/docs/book/src/capi/providers/nutanix.md
+++ b/docs/book/src/capi/providers/nutanix.md
@@ -46,7 +46,7 @@ Corresponding env variables
 | Variable              | Description                                                    | Default                             |
 |-----------------------|----------------------------------------------------------------|-------------------------------------|
 | `force_deregister`    | Allow output image override if already exists.                 | `false`                             |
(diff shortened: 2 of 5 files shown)


mboersma:
Refs #1406

mboersma:
> one global deb dependency python3-distutils was removed because useless for old distri and no more exist in Ubuntu 24.04 LTS

Thanks! I noticed this too; I have a PR I'll open soon adding 24.04 support for Azure.

images/capi/ansible/roles/node/defaults/main.yml
```diff
@@ -58,7 +58,6 @@ common_debs:
   - libnetfilter-acct1
   - libnetfilter-cttimeout1
   - libnetfilter-log1
-  - python3-distutils
```
mboersma, line 61: AFAICT removing distutils doesn't cause any problems for earlier versions of the distro. I'm not sure exactly why it's here; I tried to dig back into the `git` history but couldn't find a specific justification. So hopefully it's fine to remove, since it doesn't exist in 24.04.

mboersma (approved):
/lgtm

images/capi/ansible/roles/node/defaults/main.yml
tuxtof: yes i searched too, was added during an OVA vSphere PR but no justification
i try to rebuild 20.04 and  22.04 and no problem without the package

mboersma (approved):
/approve

PR #1454, opened 2024-04-26
Add Ubuntu 24.04 LTS for Azure

mboersma:
## Change description

Adds support for Ubuntu 24.04-based builds on Azure.

## Related issues

- Refs #1406
- See #961 for prior art

PR #1458, opened 2024-05-01
Don't install Azure CLI by default

### LINKED ISSUE #1330: Don't install azure client by default

When building images for Azure (that is, running `make build-azure-<foo>`), the `az` command-line client is installed on each Linux (and soon, Windows) image.

The `az` client is convenient for development and troubleshooting, but isn't appropriate for production images. This should be an option that must be turned on explicitly.

See also #1326

/kind bug

mboersma:
## Change description

The `az` CLI isn't necessary for Azure images to run, it's more of a troubleshooting or debugging tool, so let's not install it by default. You can still install it by enabling the `debug_tools` variable, for example:

```shell
PACKER_FLAGS="--var 'debug_tools=true'" noproxy=* \
  make -C images/capi build-azure-vhd-mariner-2
```

## Related issues

- Fixes #1330
- Unblocks #1454 (since there's no .deb feed for `az` on Ubuntu 24.04 yet)
- See also #1326

## Additional context

I wasn't sure how to test conditionally for the `azure-cli` package in Goss, so I removed it to match the default. If you have an idea how to keep this as a conditional test, please let me know!

Also this doesn't apply to Windows images currently because I wasn't positive that `az` wasn't actually required there. But maybe I should extend this behavior to Windows images as well? Feedback welcome.


jsturtevant:
> FYI this doesn't apply to Windows currently. Should it?

What doesn't apply?  We do install azure cli on Windows: https://github.com/kubernetes-sigs/image-builder/pull/1326

mboersma:
Sorry, I meant that I didn't wire this up to *not* install on Windows; it still does so by default. Say the word and I'll try to make Windows images have the same default and option.

mboersma:
Updated to act the same way in Windows.

jsturtevant:
/hold 
for the tests

PR #1465, opened 2024-05-10
Add Azure Linux 3 for Azure

mboersma:
/kind feature
/area provider/azure

## Change description

Adds support for the upcoming [Azure Linux](https://github.com/microsoft/azurelinux) 3 release and changes references from "Mariner" to "Azure Linux" for consistency.

Azure Linux 3 doesn't release until July AFAICT, but the source image feed looks stable and everything works, so I thought I'd jump the gun and add this now. But if it's better to wait for the official release (or to prefix the make targets with "Preview" or somesuch), please comment here.

## Related issues

N/A

## Additional context

I've tested this locally, but I left Azure Linux 3 out of CI tests, since it's not released yet. Once it's officially released, we can consider it stable enough to add to image-builder tests.


images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -16,7 +16,7 @@
   when: ansible_os_family == "Debian"
 
 - ansible.builtin.import_tasks: redhat.yml
-  when: ansible_os_family in ["Common Base Linux Mariner", "RedHat"]
+  when: ansible_os_family in ["Common Base Linux Mariner", "Microsoft Azure Linux", "RedHat"]
```
mboersma, line 19: Unfortunately, v3 has a completely different `ansible_os_family` string, as well as `ansible_distribution`. So we need to check for both.

images/capi/azure_targets.sh
```diff
@@ -1,8 +1,8 @@
-VHD_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 mariner-2 rhel-8 windows-2019-containerd windows-2022-containerd"
-VHD_CI_TARGETS="ubuntu-2204 ubuntu-2404 mariner-2 windows-2019-containerd windows-2022-containerd"
-SIG_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 mariner-2 rhel-8 windows-2019-containerd windows-2022-containerd flatcar"
-SIG_CI_TARGETS="ubuntu-2204 ubuntu-2404 mariner-2 windows-2019-containerd windows-2022-containerd flatcar"
-SIG_GEN2_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 mariner-2 flatcar"
-SIG_GEN2_CI_TARGETS="ubuntu-2204 ubuntu-2404 mariner-2 flatcar"
+VHD_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 azurelinux-2 azurelinux-3 rhel-8 windows-2019-containerd windows-2022-containerd"
```
mboersma, line 1: I added `azurelinux-3` to the build targets but not the CI ones. This way it's possible to build Azure Linux 3, but image-builder CI won't fail if there are hiccups in the prerelease versions.

images/capi/Makefile
```diff
@@ -656,6 +659,13 @@ build-azure-sig-ubuntu-2204-cvm: ## Builds Ubuntu 22.04 CVM managed image in Sha
 build-azure-sig-ubuntu-2404-cvm: ## Builds Ubuntu 24.04 CVM managed image in Shared Image Gallery
 build-azure-vhds: $(AZURE_BUILD_VHD_TARGETS) ## Builds all Azure VHDs
 build-azure-sigs: $(AZURE_BUILD_SIG_TARGETS) $(AZURE_BUILD_SIG_GEN2_TARGETS) $(AZURE_BUILD_SIG_CVM_TARGETS) ## Builds all Azure Shared Image Gallery images
+# Deprecated targets, for backward compatibility
+build-azure-vhd-mariner-2: ## Deprecated: use build-azure-vhd-azurelinux-2
+	$(MAKE) build-azure-vhd-azurelinux-2
+build-azure-sig-mariner-2: ## Deprecated: use build-azure-sig-azurelinux-2
+	$(MAKE) build-azure-sig-azurelinux-2
+build-azure-sig-mariner-2-gen2: ## Deprecated: use build-azure-sig-azurelinux-2-gen2
+	$(MAKE) validate-azure-sig-azurelinux-2-gen2
```
AverageMarcus, line 668: 👍

images/capi/packer/azure/azurelinux-2.json
```diff
@@ -1,6 +1,6 @@
 {
-  "build_name": "mariner-2",
-  "distribution": "mariner",
+  "build_name": "azurelinux-2",
```
AverageMarcus, line 2: Would we consider this a breaking change as I think this changes the name of final image, right?

mboersma: I think you're right, if someone was depending on that naming pattern. I was just trying to keep the `Makefile` targets intact for backward-compatibility. The `build_name` also needs to match the build target name AFAICT, so reverting this to `mariner-2` might mean reworking the Makefile as well.

I was hoping to have this be consistently `azurelinux`, to make it easier to navigate (and also to simplify some code I wrote here: kubernetes-sigs/cluster-api-provider-azure#4832). image-builder hasn't had a 1.0 release yet, so I think technically we can break things...but let's not if it's avoidable. Let me see how I can improve this, good catch @AverageMarcus!

images/capi/ansible/roles/containerd/tasks/main.yml
mboersma: FYI, I'm working on a PR for ansible/ansible that adds Azure Linux detection in general, and sets the distro fact to `Azure`. This would simplify these checks. (Thanks for the Flatcar team for having blazed this trail already!)

images/capi/packer/azure/azurelinux-2.json
mboersma: I reworked this PR to leave the existing `mariner-2` targets intact, just adding forward-compatible `azurelinux-2` targets and new `azurelinux-3` support. This seems unlikely to break anyone's usage.

AverageMarcus (commented):
/lgtm

@mboersma is this PR good to go or are you wanting to wait on the changes to Ansible first?

AverageMarcus:
👍 
/approve

PR #1467, opened 2024-05-13
Fix PowerVS builds by including wasm-shims file

### LINKED ISSUE #1466: Building CAPI powervs and qemu images fail
## Environment

- **Make target:** `make build-powervs-centos-8`
- **Run using container image? (Y/N):** N
- **Environment vars:**
```

```
- **Vars file:**
```
{
    "account_id": "XXX",
    "apikey": "XXX",
    "capture_cos_access_key": "XXX",
    "capture_cos_bucket": "XXX",
    "capture_cos_region": "us-south",
    "capture_cos_secret_key": "XXX",
    "key_pair_name": "amulya-ssh-pubkey",
    "kubernetes_deb_version": "1.29.4-1.1",
    "kubernetes_rpm_version": "1.29.4",
    "kubernetes_semver": "v1.29.4",
    "kubernetes_series": "v1.29",
    "region": "mon",
    "service_instance_id": "XXX,
    "ssh_private_key_file": "XXX",
    "zone": "mon01"
}
```

`ANSIBLE_SSH_ARGS="-o HostKeyAlgorithms=+ssh-rsa" PACKER_VAR_FILES=variables.json make build-powervs-centos-8`

```
 powervs: TASK [containerd : Copy in containerd config file etc/containerd/config.toml] ***
    powervs: An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleError: An unhandled exception occurred while templating '{
{ containerd_wasm_shims_runtime_versions | from_json }}'. Error was a <class 'json.decoder.JSONDecodeError'>, original message: Expecting value: line 1 column 1 (char 0)
    powervs: fatal: [default]: FAILED! => changed=false
    powervs:   msg: 'AnsibleError: An unhandled exception occurred while templating ''{{ containerd_wasm_shims_runtime_versions | from_json }}''. Error was a <class ''json.decoder.JSONDecode
Error''>, original message: Expecting value: line 1 column 1 (char 0)'
    powervs:
    powervs: PLAY RECAP *********************************************************************
    powervs: default                    : ok=33   changed=27   unreachable=0    failed=1    skipped=264  rescued=0    ignored=0
    powervs:
```

Image to build successfully

## Relevant log output

<details>
<summary>Log Output</summary>

```
<!--
If you have any relevant build logs that could help debug this issue please include them here
but MAKE SURE ANY SENSITIVE INFO IS REMOVED!
-->
```
</details>

---

/kind bug

Amulyam24:
## Change description

Removes unused variable `containerd_wasm_shims_runtime_versions` while copying containerd config file.

## Related issues

- Fixes #1466

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -285,6 +285,7 @@ COMMON_POWERVS_VAR_FILES := packer/config/kubernetes.json \
 					packer/config/cni.json \
 					packer/config/ppc64le/cni.json \
 					packer/config/containerd.json \
+					packer/config/wasm-shims.json \
                     packer/config/ppc64le/containerd.json \
                     packer/config/ansible-args.json \
                     packer/config/goss-args.json \


mboersma:
This should be a no-op overall: since `containerd_wasm_shims_runtimes` defaults to empty, it won't actually install anything. But it seems that `/images/capi/packer/config/wasm-shims.json` isn't included in the powervs/centos build.

What if instead of removing these lines in `main.yml`, we add packer/config/wasm-shims.json` to `COMMON_POWERVS_VAR_FILES`? I think that will fix the build.

https://github.com/kubernetes-sigs/image-builder/blob/50a070ab39f94bd77c04edb91b06946aeb08cef5/images/capi/Makefile#L283

Amulyam24:
> This should be a no-op overall: since `containerd_wasm_shims_runtimes` defaults to empty, it won't actually install anything. But it seems that `/images/capi/packer/config/wasm-shims.json` isn't included in the powervs/centos build.
> 
> What if instead of removing these lines in `main.yml`, we add `packer/config/wasm-shims.json` to `COMMON_POWERVS_VAR_FILES`? I think that will fix the build.
> 
> https://github.com/kubernetes-sigs/image-builder/blob/50a070ab39f94bd77c04edb91b06946aeb08cef5/images/capi/Makefile#L283

Since `containerd_wasm_shims_runtime` is empty, I feel it is better to remove the var as a fix. Any particular reason that it was added?

mboersma:
> Any particular reason that it was added?

In #1037, the idea was that any distro could install the containerd-wasm-shims by populating the `containerd_wasm_shims` variable with one or more Wasm backends, but that by default that variable is empty. I think we just forgot to take into account PowerVS builds.

The simple and IMHO consistent fix would be to include that file. You could certainly argue it's cleaner not to include it all, since I don't think the containerd-wasm-shims are built for that architecture yet, but removing it from this file obviously breaks most other builds. Maybe you could just define `containerd_wasm_shims_runtime_versions` as an empty variable in the PowerVS JSON config file?

Amulyam24:
> > Any particular reason that it was added?
> 
> In #1037, the idea was that any distro could install the containerd-wasm-shims by populating the `containerd_wasm_shims` variable with one or more Wasm backends, but that by default that variable is empty. I think we just forgot to take into account PowerVS builds.
> 
> The simple and IMHO consistent fix would be to include that file. You could certainly argue it's cleaner not to include it all, since I don't think the containerd-wasm-shims are built for that architecture yet, but removing it from `images/capi/ansible/roles/containerd/tasks/main.yml` obviously breaks most other builds. Maybe you could just define `containerd_wasm_shims_runtime_versions` as an empty variable in the PowerVS JSON config file?

Thanks @mboersma for clarifying, updated the changes. PTAL

mboersma (approved):
/lgtm

mboersma:
/retitle Fix PowerVS builds by including wasm-shims file

mboersma:
Thanks again @Amulyam24! We don't have coverage of all targets in CI unfortunately, so we have to rely on intrepid community members like you to let us know when something is broken.

PR #1477, opened 2024-05-27
Support setting image_name user var for OVA image name output

AverageMarcus:
## Change description

Adds support to OVA for providing an `image_name` user variable to set the name of the output image file. If not set, the default of `vm_name` should still be used.
 
This is in-line with other providers e.g. [openstack](https://github.com/kubernetes-sigs/image-builder/blob/0cce6da44a01c8663bd29e5cefb23e958c3c4077/images/capi/packer/openstack/packer.json#L8) and [nutanix](https://github.com/kubernetes-sigs/image-builder/blob/0cce6da44a01c8663bd29e5cefb23e958c3c4077/images/capi/packer/nutanix/packer.json#L11).

## Related issues

N/a


mboersma:
@sachinphogat sorry about the regression!

Could you open a GitHub issue for this? It just makes it easier to track, assign, and fix. 😄

PR #1484, opened 2024-06-05
proxmox-ubuntu-22.04: Make boot command wait times more robust

tbabej:
This increases the wait times after pressing Enter key from 5 second to 10 seconds during the boot command insertion, which more results in more robust VM startup process. Previous value of 5 seconds can result in failure due to insufficient wait time:

    ==> proxmox-iso.ubuntu-2204: Typing the boot command
    ==> proxmox-iso.ubuntu-2204: Error running boot command:
        Put "https://10.9.1.0:8006/api2/json/nodes/pm0/qemu/104/sendkey": EOF


drew-viles:
Hi @tbabej could you sign the CLA please? We can process this change then - seems sensible to me though!

mboersma:
@tbabej could you sign the Kubernetes CLA through the link in the comment above? This looks like a simple change we can merge once that is done. Thanks!

mboersma:
/close

Please re-open this if it’s still an issue and you’ve signed the Kubernetes CLA.

PR #1486, opened 2024-06-06
cloud-init: improve logging to stream logs to the cloud-init-output.log file instead of dumping at the end

chrischdi:
## Change description

This PR changes the logging command for cloud-init-output.log to stream the logs instead of dumping after finishing the command.

When debugging, the old command's python script is written to finish reading from stdin until it starts to write to stdout.
This makes it hard to debug failures while they occur because it is not possible to see easily what cloud-init is currently doing.

Example command to run for the old behaviour:

```sh
$ (echo "foo"; sleep 1; echo "bar"; sleep 1; echo "foobar") | python3 -c 'import sys,time;sys.stdout.write("".join(( " ".join((time.strftime("[%Y-%m-%d %H:%M:%S]", time.localtime()), line)) for line in sys.stdin )))'
[2024-06-06 12:02:07] foo
[2024-06-06 12:02:08] bar
[2024-06-06 12:02:09] foobar
```

Using the same echo and the new python code shows it immedately outputs the log lines as they get into the pipe:

```sh
$ (echo "foo"; sleep 1; echo "bar"; sleep 1; echo "foobar") | python3 -c 'exec("import sys,time\nfor line in sys.stdin:\n  sys.stdout.write(\" \".join((time.strftime(\"[%Y-%m-%d %H:%M:%S]\", time.localtime()), line)))")'
[2024-06-06 12:03:37] foo
[2024-06-06 12:03:38] bar
[2024-06-06 12:03:39] foobar
```

So the output is still the same, but what changes is the time when it gets streamed to the actual output file.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
+++ b/images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
@@ -64,4 +64,4 @@ log_cfgs:
 # this tells cloud-init to redirect its stdout and stderr to
 # 'tee -a /var/log/cloud-init-output.log' so the user can see output
 # there without needing to look on the console.
-output: {all: '| python3 -c ''import sys,time;sys.stdout.write("".join(( " ".join((time.strftime("[%Y-%m-%d %H:%M:%S]", time.localtime()), line)) for line in sys.stdin )))'' | tee -a /var/log/cloud-init-output.log'}
+output: {all: '| python3 -c ''exec("import sys,time\nfor line in sys.stdin:\n  sys.stdout.write(\" \".join((time.strftime(\"[%Y-%m-%d %H:%M:%S]\", time.localtime()), line)))")'' | tee -a /var/log/cloud-init-output.log'}


mnaser (changes requested):
Why don't we just use what the cloud images do out of the box?

```
# This tells cloud-init to redirect its stdout and stderr to
# 'tee -a /var/log/cloud-init-output.log' so the user can see output
# there without needing to look on the console.
output: {all: '| tee -a /var/log/cloud-init-output.log'}
```

This pipes everything automatically.

chrischdi:
> Why don't we just use what the cloud images do out of the box?
> 
> ```
> # This tells cloud-init to redirect its stdout and stderr to
> # 'tee -a /var/log/cloud-init-output.log' so the user can see output
> # there without needing to look on the console.
> output: {all: '| tee -a /var/log/cloud-init-output.log'}
> ```
> 
> This pipes everything automatically.

Because we will loose timestamp information then.

images/capi/ansible/roles/providers/files/etc/cloud/cloud.cfg.d/05_logging.cfg
```diff
@@ -64,4 +64,4 @@ log_cfgs:
 # this tells cloud-init to redirect its stdout and stderr to
 # 'tee -a /var/log/cloud-init-output.log' so the user can see output
 # there without needing to look on the console.
-output: {all: '| python3 -c ''import sys,time;sys.stdout.write("".join(( " ".join((time.strftime("[%Y-%m-%d %H:%M:%S]", time.localtime()), line)) for line in sys.stdin )))'' | tee -a /var/log/cloud-init-output.log'}
+output: {all: '| python3 -c ''exec("import sys,time\nfor line in sys.stdin:\n  sys.stdout.write(\" \".join((time.strftime(\"[%Y-%m-%d %H:%M:%S]\", time.localtime()), line)))")'' | tee -a /var/log/cloud-init-output.log'}
```
mboersma, line 67: This does seem like an improvement. Why does it need to be wrapped in `exec()` now?

chrischdi: Because inside the exec we can actually do a normal for loop, because it interprets it. Without exec in `python3 -c` we can't define a normal for loop, which is why the old code was written as is.

PR #1488, opened 2024-06-10
Updating photon gpg keys should not enable disabled repo

### LINKED ISSUE #1482: `disable_public_repos` not respected for Photon OS
## Environment

- **Make target:** `make build-node-ova-vsphere-photon-5`
- **Run using container image? (Y/N): No**
- **Vars file:**
```
disable_public_repos=true
extra_repos=repos/xyz/artifactory-photon.repo
```

Run image builder with `disable_public_repos` set to `true` and `extra_repos` value set. This will help downstream systems to pull packages from internal repo rather than relying on public repos. However, [changeset](https://github.com/kubernetes-sigs/image-builder/commit/fb06139edfd4de3b776b9df9511d2c0211bb63dc) introduced fetch the latest gpg keys, may have broken this behaviour.

By the time `setup` role task `Update the repos package to import the recent gpg keys` gets called, the disabling (renaming of existing repo files) and addition of new repos is already complete. However, the task then seems to bring back the disabled repos  `tdnf update -y photon-repos --enablerepo=photon --refresh` due to which rather than packages getting fetched from extra_repos, they will get fetched from the original (previously disabled) repos. This is not the desired intent.

Any disabled repo should not get enabled till actual enabling of the repos in sysprep role.

---

/kind bug

sriramandev:
## Change description
Refreshing the gpg keys for photon upstream repo causes disabled repos(when disable_public_repos is enabled) to get enabled. As a result if there are extra_repos configured, instead of using the extra_repos to pull packages, packages are pulled from public repos. This breaks the functionality of `disable_public_repos` flag

## Related issues
- Fixes #1482

### DIFF
--- a/images/capi/ansible/roles/setup/tasks/photon.yml
+++ b/images/capi/ansible/roles/setup/tasks/photon.yml
@@ -26,6 +26,7 @@
   ansible.builtin.command: tdnf update -y photon-repos --enablerepo=photon --refresh
   register: distro
   changed_when: '"Nothing to do" not in distro.stderr'
+  when: not disable_public_repos|default(false)|bool
 
 - name: Perform a tdnf distro-sync
   ansible.builtin.command: tdnf distro-sync -y --refresh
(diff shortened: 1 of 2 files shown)


sriramandev:
> /ok-to-test

Thank you

chrischdi (commented):
/lgtm

PR #1489, opened 2024-06-13
Support overriding the gallert-image-definition name completely for CAPZ

AverageMarcus:
## Change description

Support completely overriding the name used for the SIG image definition when creating a CAPZ image. This is configurable with the env var `SIG_IMAGE_DEFINITION` and if not set the current default behaviour will be used.

## Related issues

- N/A


mboersma (approved):
/lgtm
/approve

Thanks @AverageMarcus!

PR #1494, opened 2024-06-18
Optionally source /etc/default/kubelet in kubelet.service

toabctl:
On Debian/Ubuntu systems, /etc/default/kubelet is installed instead of /etc/sysconfig/kubelet [0]. So this file also needs to be sources in case it exists.

[0]
https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/kubernetes/tasks/main.yml#L48-L54

## Change description
This PR does optionally source the file `/etc/default/kubelet` for the systemd `kubelet.service`.
That is required because on Debian based systems, `/etc/default/kubelet` is used instead of `/etc/sysconfig/kubelet`.


mboersma (approved):
/lgtm
/approve

Just had to reassure myself that multiple `EnvironmentFile` entries work as expected: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd 👍🏻

PR #1496, opened 2024-06-26
goss validation removing kubectl --short flag

praveen4g0:
## Related issues

- Fixes #


mboersma:
@praveen4g0 thanks for this! It needs to be rebased and you need to sign the Kubernetes CLA, then we can merge this.

drew-viles:
It looks like this may in fact no longer be required: 

https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/goss/goss-command.yaml#L66

/cleanup

mboersma:
/close 

> It looks like this may in fact no longer be required

You're right @drew-viles, I guess this line was refactored away since this PR opened. 👍🏻

PR #1497, opened 2024-06-26
Implement hardening recommendations from kube-bench

heytrav:
Remediations for some things Kube-bench reported on a cluster.

## Change description

Changes permissions of kubeadm config, kubelet config and etcd directory as recommended by [kube bench](https://github.com/aquasecurity/kube-bench).

```
[FAIL] 4.1.1 Ensure that the kubelet service file permissions are set to 600 or more restrictive (Automated)
[FAIL] 4.1.9 If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive (Automated)
[FAIL] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd
```

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/roles/kubernetes/tasks/url.yml
+++ b/images/capi/ansible/roles/kubernetes/tasks/url.yml
@@ -122,7 +122,23 @@
     dest: "{{ systemd_prefix }}/system/kubelet.service.d/10-kubeadm.conf"
     owner: root
     group: root
-    mode: "0644"
+    mode: "0600"
+
+- name: Set more secure permissions on kubelet config.yaml
+  ansible.builtin.file:
+    path: /var/lib/kubelet/config.yaml
+    owner: root
+    group: root
+    mode: "0600"
+
+- name: Set correct ownership of etcd
+  ansible.builtin.file:
+    path: /var/lib/etcd/member
+    owner: etcd
+    group: etcd
+    mode: "0600"
+    recurse: true
+    state: directory
 
 - name: Create kubelet systemd file
   ansible.builtin.template:


heytrav:
I need to test this a bit. I see kubelet/config.yaml is created later by kubeadm so don't need to set that.

images/capi/ansible/roles/kubernetes/tasks/url.yml
```diff
@@ -122,7 +122,23 @@
...
     group: root
-    mode: "0644"
+    mode: "0600"
+
+- name: Set more secure permissions on kubelet config.yaml
+  ansible.builtin.file:
+    path: /var/lib/kubelet/config.yaml
+    owner: root
+    group: root
+    mode: "0600"
```
neolit123, line 132: this file is not publically readable due to the parent folder perission being 700
```
$ sudo stat --format '%a' /var/lib/kubelet/config.yaml
644
$ sudo stat --format '%a' /var/lib/kubelet/
700
```

images/capi/ansible/roles/kubernetes/tasks/url.yml
```diff
@@ -122,7 +122,23 @@
...
+
+- name: Set correct ownership of etcd
+  ansible.builtin.file:
+    path: /var/lib/etcd/member
+    owner: etcd
+    group: etcd
+    mode: "0600"
+    recurse: true
+    state: directory
```
neolit123, line 141: this is also not publically readable:
```
$ sudo stat --format '%a' /var/lib/etcd
700
$ sudo stat --format '%a' /var/lib/etcd/member
700
```

neolit123 (commented):
kube-bench does not consider parent dir permissions, which makes it not so accurate to determine file/dir security.

might be better to send these changes for kubernetes/kubernetes/cmd/kubeadm instead.

both of these files/dirs are created on kubeadm runtime so i would assume they will cause some errors on the ansible side?

images/capi/ansible/roles/kubernetes/tasks/url.yml
heytrav: Yes probably better off trying these with the kubeadm side of things instead.

PR #1498, opened 2024-06-27
fix(packer/openstack): allow availability_zone

simonostendorf:
## Change description

This PR adds support for packer openstack `availability_zone` and `volume_availability_zone` arguments.

This is needed because currently creating the OpenStack Instance fails with `Invalid volume: Instance <instance-id> and volume <volume-id> are not in the same availability_zone. Instance is in None. Volume is in dc1.` and setting `availability_zone` and `volume_availability_zone` has no effect because they are not recognised.

## Related issues

 
none

## Additional context

none

### DIFF
--- a/images/capi/packer/openstack/packer.json
+++ b/images/capi/packer/openstack/packer.json
@@ -1,6 +1,7 @@
 {
   "builders": [
     {
+      "availability_zone": "{{user `availability_zone`}}",
       "config_drive": "{{user `attach_config_drive`}}",
       "flavor": "{{user `flavor`}}",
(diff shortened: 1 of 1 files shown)


mboersma:
@simonostendorf sorry for the slow responses this week.

This looks reasonable to me. If you run `make -C images/capi json-sort` and commit and squash, that should pass the linter. Thanks for the contribution!

drew-viles:
Hi @simonostendorf 
Good spot on that one. I'm using everything in the same availaility zone so hadn't hit this scenario (yet - I'm sure I would have soon!)

Looks good though. Sorry for the delayed response (again). It's been a hectic month! 

/approve

PR #1501, opened 2024-07-01
Add containerd configuration for hosts.toml paths

dalees:
## Change description

This configuration addition allows hosts.toml files to be located when pulling CRI images.

This makes it possible to provide registry hosts config for private registries using self-signed CAs.

This falls back to Docker's Certificate file pattern where PEM certificates can be placed into `/etc/containerd/certs.d/<registry>/registry.crt`

https://github.com/containerd/containerd/blob/main/docs/hosts.md#cri

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -151,6 +151,11 @@
     path: /etc/containerd
     state: directory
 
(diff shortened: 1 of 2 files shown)


drew-viles (commented):
Seems like a sensible addition to me, thanks!

/lgtm
/ok-to-test

mboersma (approved):
/lgtm
/approve

PR #1504, opened 2024-07-07
Support configuring VLAN and MTU for proxmox builds

Tra5is:
## Change description
The VM created in Proxmox to build the template was missing the ability have the VLAN, or MTU set. The packer proxmox integration supports it, so this change is just the addition of the settings and environment variables.

Packer docs are [here](https://developer.hashicorp.com/packer/integrations/hashicorp/proxmox/latest/components/builder/iso#network-adapters)

If the value is not specified via the new environment variables then the vlan and MTU on the VM is not set and everything behaves as it did before.

### DIFF
--- a/images/capi/packer/proxmox/packer.json
+++ b/images/capi/packer/proxmox/packer.json
@@ -27,7 +27,9 @@
       "name": "{{user `build_name`}}",
       "network_adapters": [
         {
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm

PR #1505, opened 2024-07-07
Add Proxmox docs regarding env vars and permissions setup

Tra5is:
Some updates to the proxmox readme to make it clear which values can be modified for a build. Also, how to configure Proxmox and specify the connection details.

### DIFF
--- a/docs/book/src/capi/providers/proxmox.md
+++ b/docs/book/src/capi/providers/proxmox.md
@@ -28,11 +28,45 @@ the different operating systems.
 |--------------------|-----------------------------------------|
 | `ubuntu-2204.json` | The settings for the Ubuntu 22.04 image |
 
+The full list of available environment vars can be found in the `variables` section of `images/capi/packer/proxmox/packer.json`.
+
+Each variable in this section can also be overridden via the `PACKER_FLAGS` environment var.
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/book/src/capi/providers/proxmox.md
```diff
@@ -28,11 +28,45 @@ the different operating systems.
...
+
+Then the user (not token) must be given the following permissions on the path `/` and propagated:
+
+* Datastore.*
+* SDN.*
+* Sys.AccessNetwork
+* Sys.Audit
+* VM.*
+
+*Suggest creating a new role since no built-in PVE role covers just these.*
```
mboersma, line 63: *We suggest creating a new role, since no built-in PVE roles covers just these.*

Tra5is: Thanks for taking a look! Updated.

justinas-b:
Hey @Tra5is , i have been playing around with this, and it took me a while to get proxmox build working. I found multiple requirements:
- There should be a DHCP server available in the network where new template is built, otherwise new VM will get stuck during initial bootstrap. It needs network configuration so that it could download required packages and access cloud-init documents
- Server where packer runs should be in same network, so that new template could access it to download cloud-init documents

Would it make sense adding those requirements also to your pull-request? It took me few days to figure that out.

Tra5is:
@justinas-b I appreciate you taking the time to test this out. Let me know if PR #1506 would help address your issues.

Issue #1508, opened 2024-07-11
Keep mandatory presubmit ci working after 1st of august


chrischdi:
Hey folks,

the presubmit prowjob `pull-ova-all` is backed by infrastructure which is not owned by the community.
Starting by the 1st of August this job will not work anymore. (see [this mail](https://groups.google.com/u/1/a/kubernetes.io/g/dev/c/p6PAML90ZOU/m/Oy2bR918BAAJ) for more information)

We are working on it but do not yet have any estimate yet when we will have community owned infrastructure to run this job after 1st of August (also see [this mail](https://groups.google.com/u/1/g/kubernetes-sig-cluster-lifecycle/c/Qu5vo9JM8-s)).

This prowjob currently is currently used as a signal to merge PRs and is mandatory to be green for merging, when the following paths [have changes in a PR](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/image-builder/image-builder-ova-presubmits.yaml#L11C1-L12C1):

* `images/capi/(Makefile|ansible\.cfg)`
* `images/capi/ansible/.*`
* `images/capi/scripts/ci-ova\.sh`
* `images/capi/packer/(config|goss|ova)/.*`
* `images/capi/hack/ensure-(ansible|packer|goss).*`

Before hitting 1st of august we propably should:

* either make the `pull-ova-all` prowjob  optional or remove it (as long as there is no new backing infrastructure yet available)
* maybe make `pr-pull-image-builder-gcp-all` or `pr-azure-vhds` not optional anymore to cover most of image-builder on every PR.
  * sidenote: `pr-pull-image-builder-gcp-all` is currently failing

chrischdi:
Adding all currently listed maintainers for awareness :-) :

@AverageMarcus, @jsturtevant, @kkeshavamurthy, @mboersma, @drew-viles

cc @fabriziopandini @sbueringer

mboersma:
Thanks for the heads up @chrischdi. I'm working on the changes for the `pull-azure-vhds` and `pull-azure-sigs` to migrate them to community infra and should be able to land that before August 1, so we will at least have some test signal against basic regressions.

It would be great if we could rehabilitate `pr-pull-image-builder-gcp-all` as well. I can take a look after I finish this work, but if anyone wants to volunteer before that, that would be excellent.

drew-viles:
[...]

PR #1514, opened 2024-07-17
Enable use_azure_cli_auth in Packer plugin

mboersma:
## Change description

Allows Azure builds to enable the [`use_azure_cli_auth`](https://github.com/hashicorp/packer-plugin-azure/blob/v1.4.5/docs/builders/index.mdx#azure-cli) setting in the azure-arm Packer plugin by setting the `USE_AZURE_CLI_AUTH` environment variable.

This overrides all other auth settings and uses a previously successful `az login` as its existing auth context. It is enabled in the workload ID / no secrets codepath if `AZURE_FEDERATED_TOKEN_FILE` is defined.

## Related issues

- Refs #1508


mboersma:
/test pull-azure-vhds-wi

(Note that this is a temporary CI job we're using to test workload ID auth. Once this merges, I will delete it and move the appropriate settings to the existing Azure jobs.)

nawazkh:
Looks good to me :) 
/lgtm

PR #1519, opened 2024-07-22
Use UTC+0 for system clock in Windows VMware OVA

### LINKED ISSUE #1518: VMware Guest Tools fails to launch and breaks the build when building Windows OVA
- **Make target:** `make windows-2022-efi`
- **Run using container image? (Y/N): Y**
- **Environment vars:**

VMware Guest Tools failed to launch when building Windows OVA in certain VMware environment.
Packer was not able to connect as the IP address of the build machine was not reported,
which failed the build eventually.

This is caused by the system clock may jumps behind during the VMware Tools installation.
ESXi sets the local time in UTC+0 to the guest VM in BIOS/EFI.
Windows uses local time other than UTC+0 by default so the actual system clock recognized by Windows may be in different time zone.
Windows syncs the time from NTP server during the VMware Tools installation after the first boot,
which causes the system clock jumping forward or backward and breaks the VMware Tools installation.
Building process will fail when VMware Tools is not working.

---

/kind bug

laozc:
## Change description
Set the system clock to UTC+0 to avoid time change during VMware Tools installation. 

## Related issues

- Fixes #1518

### DIFF
--- a/images/capi/packer/ova/windows/windows-2019-efi/autounattend.xml
+++ b/images/capi/packer/ova/windows/windows-2019-efi/autounattend.xml
@@ -127,8 +127,13 @@ Installation Notes:
             <RunSynchronous>
                 <RunSynchronousCommand wcm:action="add">
                     <WillReboot>Always</WillReboot>
-                    <Path>e:\setup64.exe /s /v "/qb REBOOT=R ADDLOCAL=ALL"</Path>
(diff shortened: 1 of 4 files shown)


chrischdi:
@laozc : would be good if you could assign someone with context to review it.

Change looks ok to me, but I have no idea of windows these days :-)

cc @randomvariable

PR #1521, opened 2024-07-22
Update Azure e2e test regions and log in each SIG build

### LINKED ISSUE #1520: `pull-azure-sigs` job failing with "need az login"
Although the first two runs on the new community infra were green, for some reason the `pull-azure-sigs` job is now failing all builds because it doesn't see the logged-in session. This isn't happening to the `pull-azure-vhds` job although they run from the same script.

See https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_image-builder/1515/pull-azure-sigs/1815381131336355840 for an example.

/kind bug

mboersma:
## Change description

Updates `get_random_region` to use those Azure locations which have increased quota for the new community subscription. Also adds an `az login` before each SIG target build to work around the "need to log in" errors.

## Related issues

- Fixes #1520

### DIFF
--- a/images/capi/packer/azure/scripts/init-sig.sh
+++ b/images/capi/packer/azure/scripts/init-sig.sh
@@ -13,7 +13,7 @@ az account set -s ${AZURE_SUBSCRIPTION_ID} >/dev/null 2>&1
 eval "$tracestate"
 
 export RESOURCE_GROUP_NAME="${RESOURCE_GROUP_NAME:-cluster-api-images}"
-export AZURE_LOCATION="${AZURE_LOCATION:-southcentralus}"
+export AZURE_LOCATION="${AZURE_LOCATION:-northcentralus}"
 if ! az group show -n ${RESOURCE_GROUP_NAME} -o none 2>/dev/null; then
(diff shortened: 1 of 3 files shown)


nawazkh:
I like the solution, looks good to me!
/lgtm

nawazkh:
Does logging in multiple times affect the generated token in anyway? I am curious if there are any "gotchas" we need to be aware about with multiple login attempts

jsturtevant:
Is it possible we hit the az login issue randomly since we are spinning off multiple processes?  I don't know the exact flow around the token generation. This seems to work, if we find it isn't we can adjust as we go.

/approve

mboersma:
@nawazkh @jsturtevant you both ask good questions about "why is the redundant login needed?" I admit I'm not sure, so this is just a workaround.

The make target ends up calling `init-sig.sh` which also does `az login`, so that must be where the context is somehow lost. But I haven't been able to reproduce this locally, and it did pass twice before it began to fail, so there seems to be an element of non-determinism here. I will monitor it and try to think of a cleaner fix if I can pinpoint what happened.

nawazkh:
> The make target ends up calling `init-sig.sh` which also does `az login`, so that must be where the context is somehow lost.

Thank you for sharing the context, sounds good to me!

PR #1450, opened 2024-04-23
Bump default Kubernetes to 1.28.9

mboersma:
What this PR does / why we need it:

Keeps with the project's precedent of having the default build version be (latest - 2).

Which issue(s) this PR fixes: 

N/A, but see #1265 for prior art.

**Additional context**

PR #1533, opened 2024-08-06
Bump default Kubernetes to 1.29.7

mloiseleur:
## Change description

Update to latest v1.29.

Latest PR on this subject: https://github.com/kubernetes-sigs/image-builder/pull/1450


AverageMarcus:
I'm not sure we should skip major version when bumping. I think at the very least we should do the latest v1.29 first and then create a release of that then do another bumping to v1.30.

It's also worth being clear that v1.31 isn't officially released yet. As of 4 hours ago the second release candidate was released so it should be soon but it's still considered a pre-release and not a production release yet.

@mboersma @jsturtevant @drew-viles would be good to get your thoughts too. I don't think we've ever had anything officially wrote down for what version we target but I think we've informally followed the oldest still in support.

drew-viles:
Yeah iirc the 13th or 14th is the target date for release of 1.31 so we can't bump to that just yet.

I'd be happy with an eventual bump to 1.30, as in I'm not against jumping a couple of versions but for the sake of having a sensible path of migration, I'd say we need to at least have one release with a 1.29 bump first. That way should anyone experience an issue with a 1.30 release, they have a rollback option.

mloiseleur:
Okay, then I will update this PR to target latest v1.29

mboersma:
/retest

Looks like a flake on Ubuntu 22.04.

mboersma:
(BTW, the pull-ova-all job was supposed to fail starting August 1, but it seems Google gave an additional week grace period before actually pulling the plug. So test ‘em while ya got ‘em.)

mboersma:
One of the parts of the “reset temp space” task seems to be failing, only for Ubuntu 22.04. This seems unlikely to be related to this PR, but we should investigate.

mloiseleur:
I took some times to understand what is this test and what is happening.

The ansible code for _Reset temp space_ is [here](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/sysprep/tasks/main.yml#L133).

I'm not sure to understand why, but it tries to delete all files in `/tmp` (and avoiding `/tmp/.ansible`)

The log says it cannot on `/tmp/snap.rootfs_6Priaa` with _errno 16 : Device or resource busy_

So far, my guess is that there is a snapshot triggered or already pre-existing, since this snapshot file is used, it cannot be deleted.

Removing the code in ansible trying to remove some tmp files would probably turn this test green, but would this be ok to remove it ?

mboersma:
> would it be ok to remove it?

I think we shouldn't, because the intent is to clean the disk image. We could maybe make it a "best effort" thing for Ubuntu 22.04 / vsphere if we can't figure out a better workaround.

I wonder if adding something like this would work:

```yaml
ignore_errors: ansible_distribution_version is version('22.04', '>=')
```

Regardless, the `pull-ova-all`  job has been retired now as of kubernetes/test-infra#33245, so we can't verify any changes there until there is new infrastructure behind it (see #1508).

/lgtm

Issue #1283, opened 2023-08-28
Automated dependency updates


AverageMarcus:
We've had several PRs lately to bump versions of various dependencies that we've noticed have become several versions out of date. We tend to do this on an ad-hoc basis when we remember to or someone needs a new feature from a dependency.

Because of this, we often find out quite a bit later about breaking changes in dependencies that we then need to also resolve at the same time as trying to update everything.

To avoid this I think it would be useful to introduce automated dependency updates on the repo so that we get new PRs every time a new version of something is available and we can then test change version bump in isolation from any other changes.

I recommend using [Renovate](https://docs.renovatebot.com/) to handle this as its free for public repos and is extremely configurable. There are also alternatives such as Dependabot but I'm less familiar with those so couldn't say for sure if they meet all our requirements.

We have versions configured in several different places so we'd need a solution that could handle the following:
* Within the [Dockerfile](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/Dockerfile)
* Within [script files](https://github.com/kubernetes-sigs/image-builder/blob/bbdd3dbacb8c3ad71ef9a266dcd8941a8e3866a1/images/capi/hack/ensure-ansible.sh#L25) as variables
* Within [packer config files](https://github.com/kubernetes-sigs/image-builder/blob/bbdd3dbacb8c3ad71ef9a266dcd8941a8e3866a1/images/capi/packer/config/containerd.json#L6)
* In [Packer plugin configs](https://github.com/giantswarm/image-builder/blob/b7f8d4c3a2cad49497ee29d2b409c2e708b9c137/images/capi/packer/nutanix/config.pkr.hcl#L4)

Ideally, we'd also need a solution that supported updating the associated SHA256 values as well as the version but as of now I'm not aware of any tools that can handle that. [[Renovate feature request](https://github.com/renovatebot/renovate/discussions/22183)]

mboersma:
Dependabot would only help with the Dockerfile, and perhaps with tools such as Packer and Ansible if we were to put those in a `requirements.txt`. I don't think there's a way to write a custom "ecosystem provider" either.

AverageMarcus:
An idea that came from [this Slack thread](https://kubernetes.slack.com/archives/C01E0Q35A8J/p1709280514329139) is to have a periodic job that fetches the latest URLs and also fetches the checksums and generates a PR with the changes.

AverageMarcus:
I'd still like this eventually

/remove-lifecycle rotten

PR #1535, opened 2024-08-08
add dependabot to image builder repo

nawazkh:
## Change description

- This PR adds dependabot to image-builder repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -41,3 +41,22 @@ updates:
       - "ok-to-test"
       - "kind/cleanup"
       - "release-note-none"
+
+  # pip
+  - package-ecosystem: "pip"
+    directories:
+        - "/"
+        - "images/capi"
+    schedule:
+      interval: "weekly"
+      time: "09:00"
+      # Use America/New_York Standard Time (UTC -05:00)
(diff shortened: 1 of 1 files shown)


nawazkh:
Not sure if I am missing any context on why dependabot was not added to this repo, so creating this PR to add it.

AverageMarcus:
There's some background context in this issue: https://github.com/kubernetes-sigs/image-builder/issues/1283

Could you please include examples of what updates dependabot will generate for us? What versions will be covered, what wont?

nawazkh:
Thank you for sharing the context @AverageMarcus. It is an interesting challenge to update various hardcoded semver-versions along with dependabot version updates. 

I have an idea, not sure how viable, but here it is: (I will mark this PR draft to avoid notification churn)
- Configure dependabot to update versions in package-manager. In our case, say Dependabot updates a dependency in Dockerfile.
- Have a script that scans all the dirs for below files and find dependencies variables, replace them with latest version in the same Dependabot's PR(this can be achieved using GitHub actions). 
  - `*.json`
  - `.hcl`
  - `*.sh`


Marking this PR WIP and push changes soon. Thanks!

PR #1538, opened 2024-08-13
Allow specifying Flatcar version for base AMI

AverageMarcus:
## Change description

This PR updates the `ami_filter_name` used for Flatcar when building AMIs so that it supports the `FLATCAR_VERSION` environment variable being set like other providers do.

## Related issues

- Fixes #


drew-viles (commented):
/lgmt
/ok-to-test

Issue #1333, opened 2023-10-25
Cloud-init fails for ubuntu 20.04 base AMI and Cloud-init version '23.3.1-0ubuntu1~20.04.1' 


supershal:
**What steps did you take and what happened:**

The latest cloud-init version `23.3.1-0ubuntu1~20.04.1` that is shipped with base AMI for Ubuntu 20.04 is unable to run boothook https://cloudinit.readthedocs.io/en/latest/explanation/format.html#cloud-boothook provided by CAPA, https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0bf78b04b305a77aec37a68c107102231faa7a16/pkg/cloud/services/secretsmanager/secret_fetch_script.go#L20
As a result the CAPA VMs are not initializing as expected.

**Steps to reproduce:**
1. create an AMI using image-builder
 ```
make build-ami-ubuntu-2004
 ```

2. Create CAPA cluster using the AMI created in step 1 using instructions at: https://cluster-api-aws.sigs.k8s.io/getting-started.html

3. Check logs at `/var/log/cloud-init-output.log`

**What did you expect to happen:**
Cloud-init run successfully on the VM

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Log from cloud-init.
```
2023-10-24 18:53:21] 2023-10-24 18:53:21,892 - util.py[WARNING]: failed stage init
[2023-10-24 18:53:21] failed run of stage init
[2023-10-24 18:53:21] ------------------------------------------------------------
[2023-10-24 18:53:21] Traceback (most recent call last):
[2023-10-24 18:53:21]   File "/usr/lib/python3/dist-packages/cloudinit/url_helper.py", line 78, in read_file_or_url
[2023-10-24 18:53:21]     with open(file_path, "rb") as fp:
[2023-10-24 18:53:21] FileNotFoundError: [Errno 2] No such file or directory: '/etc/secret-userdata.txt'
[2023-10-24 18:53:21]
[2023-10-24 18:53:21] The above exception was the direct cause of the following exception:
[2023-10-24 18:53:21]
[2023-10-24 18:53:21] Traceback (most recent call last):
[2023-10-24 18:53:21]   File "/usr/lib/python3/dist-packages/cloudinit/user_data.py", line 238, in _do_include
[2023-10-24 18:53:21]     resp = read_file_or_url(
[...]

PR #1546, opened 2024-08-27
Patch feature flags directly into cloud-init code since newer versions do not load `feature_overrides.py` anymore

AndiDog:
## Change description

This ports the workaround https://github.com/kubernetes-sigs/image-builder/pull/406 to newer versions of cloud-init which [removed the feature flag support and instead prefer monkeypatching](https://github.com/canonical/cloud-init/pull/4228).

Without this change, the `ERROR_ON_USER_DATA_FAILURE = False` flag doesn't get used by cloud-init, resulting in errors loading the user data:

```
RuntimeError: [Errno 2] No such file or directory: '/etc/secret-userdata.txt' 
```

Such errors mean that a Kubernetes node running this image wouldn't join the cluster (or a cluster cannot even be created if it's the first control plane node). I was able to create a working Ubuntu 24.04 + Kubernetes 1.25.16 cluster again after this fix. I copied the block for RHEL 9 as well, but didn't test it myself. RHEL 8, which is EOL, was left out.

## Related issues

- https://github.com/kubernetes-sigs/image-builder/issues/1333


AverageMarcus (commented):
/lgtm
/ok-to-test
/assign @drew-viles

drew-viles:
Seems sensible to me!!

/approve

PR #1547, opened 2024-08-27
Install AWS CLI with --update flag

ohayak:
## Change description

When using an AMI with aws CLI already the process fails due to this error:
```
Found preexisting AWS CLI installation: /usr/local/aws-cli/v2/current. Please rerun install script with --update flag.
```
I propose to install the aws cli with the --upgrade option in order to overcome these situations


mboersma (commented):
/ok-to-test

mboersma (approved):
/lgtm

I can't easily test this and AWS isn't covered in our e2e tests, but it looks sane.

PR #1552, opened 2024-08-29
Pin base container image to Ubuntu 24.04

AverageMarcus:
## Change description

The `ubuntu:latest` tag has now been updated to point to 24.04 which caused some surprises when I was trying to build a new image built on top of image-builder as some packages have changed and the ubuntu user has also changed.

This PR specifically pins the base image to ubuntu:24.04 so we can release it as an explicit image-builder release that can have the change called out in the changelog.

## Related issues

N/A

### DIFF
--- a/images/capi/Dockerfile
+++ b/images/capi/Dockerfile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


drew-viles (commented):
/approve
/lgtm

AverageMarcus:
I'm going to try kicking this PR. I think Prow might be struggling 😅 

/hold

PR #1564, opened 2024-09-08
add disk size and type to gcs/packer.json

ohayak:
Google image builder is missing disk customization:
disk_size (int64) - The size of the disk in GB. This defaults to 20, which is 20GB.
disk_type (string) - Type of disk used to back your instance, like pd-ssd or pd-standard. Defaults to pd-standard.

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -2,6 +2,8 @@
   "builders": [
     {
       "disable_default_service_account": "{{ user `disable_default_service_account` }}",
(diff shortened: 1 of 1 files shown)


mboersma (commented):
This looks good to me, but image-builder doesn't have GCP covered in e2e tests and I'm not an expert there.

/cc @cpanato @dims @richardcase

mboersma (approved):
/lgtm

AverageMarcus (commented):
/lgtm 
/approve

PR #1566, opened 2024-09-19
Add vnc_bind_address var to QEMU builder

andybotting:
It is useful when building on a remote host to be able to use this variable to gain access to the VNC console directly.

This variable is used in the OVA builder, so this commit is essentially a copy from that.

The default packer value of 127.0.0.1 is defined here to retain the default behaviour.


mboersma (approved):
/lgtm
/assign @drew-viles

PR #1567, opened 2024-09-19
Update Ubuntu 22.04 to latest point release and checksum

andybotting:
Looks like Ubuntu have *just* pushed the new 22.04.5 point release to the mirrors and removed 22.04.4, which was being used here. At time of submitting this, it 22.04.4 isn't even available on https://old-releases.ubuntu.com/releases/

This commit updates the new release and updates the checksums in all the places.


mboersma (approved):
/lgtm
/assign @AverageMarcus

PR #1572, opened 2024-09-27
cleanup: Use artifacts.k8s.io instead of storage.googleapis.com for ECR credential provider

thefirstofthe300:
## Change description
The `k8s-artifacts-prod` is currently fronted by a proxy at `artifacts.k8s.io`. Assuming that there's the potential for S3 or Azure Blob Storage to ever be used as a backend in addition to/as a replacement for GCS, we should be good citizens and use the proxy to access our objects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .vscode/
 .idea/
+.venv/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .vscode/
 .idea/
+.venv/
```
thefirstofthe300, line 3: I installed ansible in a venv for testing.

AverageMarcus (commented):
/lgtm
/ok-to-test
/assign @drew-viles

mboersma (approved):
/lgtm
/approve

PR #1574, opened 2024-10-02
outscale: user can define vm_type

pli01:
## Change description

Using outscale provider, the user can override `vm_type` variables

## Related issues

- Fixes #


mboersma (approved):
/lgtm
/approve

PR #1584, opened 2024-10-10
GCE: use n2-standard-2 machine type

aslafy-z:
## Change description

N1 machines are less and less available. I could not build my image in multiple zones of european regions.
N2 seams to be more available.

### DIFF
--- a/images/capi/packer/gce/packer.json
+++ b/images/capi/packer/gce/packer.json
@@ -114,7 +114,7 @@
     "kubernetes_semver": null,
     "kubernetes_series": null,
     "kubernetes_source_type": null,
-    "machine_type": "n1-standard-1",
+    "machine_type": "n2-standard-2",
     "project_id": "{{env `GCP_PROJECT_ID`}}",
     "service_account_email": "",
     "source_image_family": "{{user `source_image_family`}}",


drew-viles:
Hi,

I'm not seeing anything from Google to suggest there is a removal of N1 instances happening anywhere and it's still listed in their workload types. The N1 machines are _generally_ cheaper to run which is why this option was set originally I suspect.

You can override this via your own configuration file too so I'd recommend taking that approach for now:
https://github.com/kubernetes-sigs/image-builder/blob/main/docs/book/src/capi/capi.md

```
PACKER_VAR_FILES=var_file_1.json make ...
```

AverageMarcus:
Yeah I think I agree with @drew-viles on this unless there is actual indication that the `n1` series is being phased out.

The price difference is quite a lot and could be a surprise for users that use the defaults and builds a lot of images.
`n1-standard-1` - $0.055
`n2-standard-2` - $0.133382

As Drew points out, `machine_type` can be provided in the user vars to override this value.

Maybe we could get away with just updating the GCP docs to point this out specifically?

aslafy-z:
I opened https://github.com/kubernetes-sigs/image-builder/pull/1608 to update the documentation. Closing this PR.

PR #1589, opened 2024-10-11
Add flatcar build for Proxmox

mcbenjemaa:
## Change description

This PR adds:

* Flatcar build for Proxmox

## Related issues

- Fixes #


mcbenjemaa:
The build is working, but let me double check for any findings.

mcbenjemaa:
WIP in CAPMOX
https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/308

but I will provide a link to how this works in the provider once it is released.

AverageMarcus:
@mboersma @drew-viles What do y'all think about this currently requiring the use of the Alpha Flatcar builds? I'm kinda inclined to not include this in image-builder until it's available in the Stable channel but I'm open to being convinced otherwise.

mcbenjemaa:
The PROXMOX VE image, was tested really well with a lot of parties.
And is being added to the alpha channel as that the release process.

On the other hand The image builder has the ability to set Flatcar Channel,
Well in that case if you only accept stable Channel, there is no need for that variable.
i.e. I tried to use the features provided.

AverageMarcus:
I'm more concerned about it _defaulting_ to Alpha channel. 

If user chose to use non-stable channels that's fine but I think the out-of-the-box settings should encourage stability. (Especially true as we're not in a position to provide any support or testing for Proxmox and very limited for Flatcar so if anyone has any issues they are often on their own)

mcbenjemaa:
capmox will have the support soon.
https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/308

mboersma (approved):
/lgtm

> Currently, Proxmox support is only available in the `alpha` channel of Flatcar.

I don't have a problem introducing this build target as an alpha feature, so long as the goal is to move this to a stable channel by default as @AverageMarcus suggested. Having said that, we don't have a way to test this in CI or locally, so all I can say is it looks good to me.

mboersma:
@drew-viles @jsturtevant do either of you have time to look at this as well?

(@AverageMarcus is away until 2025, so I may ask for more review help, or just lgtm+approve a bunch of things myself, mea culpa.)

images/capi/Makefile
```diff
@@ -866,6 +866,7 @@ build-hcloud-rockylinux-9: ## Builds the Hetzner Cloud Rocky Linux 9 image
 build-hcloud-flatcar: ## Builds the Hetzner Cloud Flatcar image
 build-hcloud-all: $(HCLOUD_BUILD_TARGETS) ## Builds all Hetzner Cloud image
 
+build-proxmox-flatcar: ## Builds Flatcar Proxmox image
```
jsturtevant, line 869: What happens if the user doesn't set `export FLATCAR_CHANNEL=alpha` and runs this command?

drew-viles:
In terms of running the alpha as the default, I have no major opinion on it as I don't use flatcar so can't really comment on the stability of the alpha. It seems a bit brave but if it's been tested and we can be sure it stays stable enough I'm happy to go with the crowd.

I'm back from vacation now but will be out again in 2 weeks until Jan so get all the PRs you can in over those two weeks 🤣

I can approved once the squash has happened

images/capi/Makefile
mcbenjemaa: will fail, for not finding the proper OEM_ID

drew-viles:
If everyone is happy now we can unhold and merge this.

mcbenjemaa:
Documentation updated

mcbenjemaa:
> /hold
> 
> @mcbenjemaa could you squash the commits?

@mboersma 
Remove hold label and we are good to go

mboersma (approved):
/hold cancel
/lgtm
/approve

PR #1583, opened 2024-10-10
chore: changes for building CAPA amis

richardcase:
## Related issues

- Fixes #

PR #1590, opened 2024-10-11
fix: pin cloud-init for ubuntu 22.04 image

richardcase:
## Change description

This pins the cloud-init version in the CAPA AMI for ubuntu 22.04 to an older version. This is required as newer versions of cloud-init don't work with the way CAPA gets the user-data onto the instance.

This is a temporary solution until a more long-term solution is found (which is currently being worked on, see #1583 ).

## Related issues

- Relates: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5115
- Relates:https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4745


AverageMarcus (commented):
/lgtm
/approve

PR #1593, opened 2024-10-13
Get Rocky Linux 9 building on Proxmox

manetherenio:
## Change description

This does a handful of things to get Rocky Linux 9 building under Proxmox.

First, in the Ansible playbook, I restrict networkd-dispatcher to Ubuntu since it doesn't appear to exist as a package under RHEL9/Rocky 9.

Next, Goss was looking for cloud-utils as an installed package, but the correct package is cloud-utils-growpart. In addition to that, Goss was verifying open-vm-tools was installed, but that's a hypervisor integration for vSphere, not QEMU/Proxmox. So I replaced that with qemu-guest-agent.

After that, I switched the disk format to raw. QCOW2 isn't supported on ZFS, and the Packer default is raw anyway:

https://developer.hashicorp.com/packer/integrations/hashicorp/proxmox/latest/components/builder/iso#disks

Raw will work on all filesystems supported by Proxmox, so it's a safe default.

I also switched the disk interface to virtio. With the scsi interface, Rocky wasn't detecting any disks. I can switch this to sata if needed, but both Ubuntu 24.04 and 22.04 built fine with virtio and IIRC, it should get slightly better performance than the scsi interface.

Finally, I set the CPU type to Westmere because RHEL9 (and thus Rocky) requires a minimum of x86-64-v2 to boot. Intel Westmere is the oldest microarchitecture that qualifies for this distinction. The default CPU type (kvm64) gives a kernel panic on installer startup without this. Please note that the genericized CPU type (x86-64-v2-AES) requires Proxmox 8.0+, so setting this to Westmere allows older versions of Proxmox to use the image-builder.

https://pve.proxmox.com/wiki/Roadmap#Proxmox_VE_8.0

I've personally tested this against my Proxmox server running the latest (8.2.2) and have built images for Rocky Linux 9, Ubuntu 22.04, and Ubuntu 24.04, so it should be regression free.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/proxmox.yml
+++ b/images/capi/ansible/roles/providers/tasks/proxmox.yml
@@ -60,8 +60,10 @@
   ansible.builtin.systemd:
     name: networkd-dispatcher
     state: started
+  when: ansible_os_family == "Debian"
 
 - name: Ensure networkd-dispatcher is enabled
   ansible.builtin.systemd:
     name: networkd-dispatcher
     enabled: true
+  when: ansible_os_family == "Debian"
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -282,9 +282,9 @@ rockylinux:
         <<: *rh9_rpms
   qemu:
     package:
-      open-vm-tools:
+      qemu-guest-agent:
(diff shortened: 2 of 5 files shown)


mboersma (commented):
/ok-to-test

PR #1596, opened 2024-10-14
Set random ssh password for builder

AverageMarcus:
For providers that don't have a built-in support for random ssh keys this change makes sure that the default builder password used during the image build process is a randomly generated value.

Fixes kubernetes/kubernetes#128007


drew-viles (approved):
/lgtm
/approve

mboersma (approved):
/lgtm++

Issue #1603, opened 2024-10-15
Azure windows-2025 builds failing: timeout during goss tests


mboersma:
The [`pull-azure-sigs`](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-azure-sigs) job has failed often since #1527 was merged, due to a timeout during the goss test stage:

```
[0;32m    azure-arm.sig-windows-2025-containerd: }[0m
[0;32m    azure-arm.sig-windows-2025-containerd: Error: timeout of 3m0s reached before tests entered a passing state[0m
[1;32m==> azure-arm.sig-windows-2025-containerd: Provisioning step had errors: Running the cleanup provisioner, if present...[0m
[1;32m==> azure-arm.sig-windows-2025-containerd: 
```

This is blocking CI and keeping PRs from merging, so we may want to skip this test for now unless a solution is iminent.

## Environment

- **Make target:** `make build-azure-vhd-windows-2025`
---

/kind bug

mboersma:
The specific two goss failures are these:

```
azure-arm.sig-windows-2025-containerd:         {
azure-arm.sig-windows-2025-containerd:             "duration": 30009575200,
azure-arm.sig-windows-2025-containerd:             "err": {},
azure-arm.sig-windows-2025-containerd:             "expected": null,
azure-arm.sig-windows-2025-containerd:             "found": null,
azure-arm.sig-windows-2025-containerd:             "human": "",
azure-arm.sig-windows-2025-containerd:             "meta": null,
azure-arm.sig-windows-2025-containerd:             "property": "exit-status",
azure-arm.sig-windows-2025-containerd:             "resource-id": "Windows Feature - Hyper-V-PowerShell",
azure-arm.sig-windows-2025-containerd:             "resource-type": "Command",
azure-arm.sig-windows-2025-containerd:             "result": 1,
azure-arm.sig-windows-2025-containerd:             "successful": false,
azure-arm.sig-windows-2025-containerd:             "summary-line": "Windows Feature - Hyper-V-PowerShell: exit-status: Error: Command execution timed out (30s)",
azure-arm.sig-windows-2025-containerd:             "test-type": 1,
azure-arm.sig-windows-2025-containerd:             "title": ""
azure-arm.sig-windows-2025-containerd:         },
azure-arm.sig-windows-2025-containerd:         {
azure-arm.sig-windows-2025-containerd:             "duration": 0,
[...]

Issue #1602, opened 2024-10-15
Azure windows-2019 builds failing: can't start sshd


mboersma:
The [`pull-azure-vhds`](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-azure-vhds) and [`pull-azure-sigs`](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-azure-sigs) jobs have begun to fail consistently because of the windows-2019 targets:

```
[0;32m    azure-arm.sig-windows-2019-containerd: TASK [systemprep : Enable ssh login without a password] ************************[0m
[0;32m    azure-arm.sig-windows-2019-containerd: changed: [default][0m
[0;32m    azure-arm.sig-windows-2019-containerd:[0m
[0;32m    azure-arm.sig-windows-2019-containerd: TASK [systemprep : Set SSH service startup mode to auto and ensure it is started] ***[0m
[0;32m    azure-arm.sig-windows-2019-containerd: An exception occurred during task execution. To see the full traceback, use -vvv. The error was: at <ScriptBlock>, <No file>: line 1064[0m
[0;32m    azure-arm.sig-windows-2019-containerd: fatal: [default]: FAILED! => {"changed": false, "msg": "Unhandled exception while executing module: Service 'OpenSSH SSH Server (sshd)' cannot be started due to the following error: Cannot start service sshd on computer '.'."}[0m
[0;32m    azure-arm.sig-windows-2019-containerd:[0m
[0;32m    azure-arm.sig-windows-2019-containerd: PLAY RECAP *********************************************************************[0m
[0;32m    azure-arm.sig-windows-2019-containerd: default                    : ok=19   changed=15   unreachable=0    failed=1    skipped=9    rescued=0    ignored=0[0m
```

This is blocking CI and keeping PRs from merging, so we may want to skip this test for now unless a solution is iminent.

## Environment

- **Make target:** `make build-azure-vhd-windows-2019` or `make build-azure-sig-windows-2019`

---

/kind bug

PR #1604, opened 2024-10-15
Remove windows-2019 and -2025 targets from CI

mboersma:
## Change description

Removes the Azure windows-2019 and windows-2025 targets from CI runs, since they are both failing and blocking unrelated PRs from merging.

## Related issues

- Refs #1602
- Refs #1603


AverageMarcus:
@mboersma Are these targets actually broken or just in CI? As in, do we want to drop support for building these completely?

mboersma:
/test pull-azure-sigs

@AverageMarcus, as far as I can tell, the windows-2019 targets are broken and maybe we should remove them altogether, although @jsturtevant should weigh in. The windows-2025 target supposedly only fails in CI, and needs to be fixed and supported.

mboersma:
/test pull-azure-sigs

Two unrelated flakes and a timeout. :-( Not sure why things are going badly here, I'm running `./images/capi/scripts/ci-azure-e2e.sh` locally to see if I can reproduce.

jsturtevant:
/lgtm
/approve

The time of the failures aligns with patch Tuesday last week. 2019 is technically EOL and only receiving security patches right now so likely ok to remove from CI.

@TinaMor fyi on 2025, hopefully we just need to increase the timeout to make this pass.

PR #1607, opened 2024-10-15
Set SSH password dynamically during auto-install

### LINKED ISSUE #1606: SSH password needs to be generated dynamically during auto-install
## Environment

- **Make target:** Multiple
  - `build-node-ova-vsphere-rhel-8`
  - `build-node-ova-vsphere-ubuntu-2204`
  - `build-nutanix-ubuntu-2204`
  - `build-nutanix-ubuntu-2004`
  - `build-qemu-rhel-8`
  - `build-qemu-rhel-9`
  - `build-raw-rhel-8`
  - `build-raw-rhel-9`
  - `build-raw-ubuntu-2004`
- **Run using container image? (Y/N):** No

We use image-builder to build images for different CAPI providers (VSphere/CloudStack/Nutanix/Raw).

Yesterday, we migrated from image-builder v0.1.37 to v0.1.38 to leverage the CVE fixes for `CVE-2024-9486` and `CVE-2024-9594`. Immediately after doing so, all our builds across all providers started failing (after a 30 minute SSH connection wait) with the following error:
```
Waiting for SSH to become available...
Error waiting for SSH: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain
```
When we temporarily reverted to v0.1.37, we didn't see this behavior and all the builds passed. This pointed to something in the v0.1.38 changelog. The only plausible cause we could think of was #1596, since it affected the SSH password used to connect to VM after boot.

After some digging, we figured out that only the SSH password to be used with `builder` user was changed to be dynamic but the actual password when creating the `builder` user was still hardcoded to `builder` ([example](https://github.com/kubernetes-sigs/image-builder/blob/e59e76be444eba93a4adaf000447fbefea2fd57a/images/capi/packer/ova/linux/ubuntu/http/22.04/user-data#L74)) . This caused a mismatch during the SSH connection because the SSH was using a random generated UUID, while the expected password was `builder`, so the SSH connection would never go through and just timed out.

In terms of a fix, since the UUID thing is internal to Packer, it's never printed or returned to the user. Also the `builder` user and password are created before the `ssh_password` is used by Packer, so anything generated in Packer will be an after-the-fact thing with respect to user creation, since the flow is unidirectional.

So I think we will need to generate it outside of Packer in the Makefile or script and replace in the JSON `ssh_password` field as well as in the user creation step.

/kind bug

abhay-krishna:
## Related issues

- Fixes #1606


drew-viles:
Looking good atm - need to test a few things but we may need to add all of these to the `.gitignore` now as we don't want the generated files being accidentally pushed to the repo. 
```
**/user-data
**/ks.cfg
**/ks.json
**/preseed-efi.cfg
**/preseed.cfg
images/capi/packer/nutanix/packer.json
images/capi/packer/ova/packer-common.json
images/capi/packer/proxmox/packer.json
images/capi/packer/qemu/packer.json
images/capi/packer/raw/linux/rhel/http/9/ks-efi.cfg
images/capi/packer/raw/packer.json
```

images/capi/Makefile
```diff
@@ -59,6 +59,7 @@ deps-common:
 	hack/ensure-python.sh
 	hack/ensure-ansible.sh
 	hack/ensure-packer.sh
+	hack/set-ssh-password.sh
```
AverageMarcus, line 62: The ssh password isn't required by all providers (e.g. AWS makes use of dynamically generated ssh keys rather than a password) so maybe we want to have this more focussed instead of in common?

images/capi/hack/set-ssh-password.sh
```diff
@@ -0,0 +1,43 @@
...
+  exit 1
+fi
+
+export SSH_PASSWORD=$(tr -dc A-Za-z0-9 </dev/urandom | head -c 16; echo)
+SALT=$(tr -dc A-Za-z0-9 </dev/urandom | head -c 16; echo)
+export ENCRYPTED_SSH_PASSWORD=$(openssl passwd -6 -salt $SALT -stdin <<< $SSH_PASSWORD)
+
+for file in $(find $PACKER_DIR -type f -name "*.tmpl"); do
+  sed -e "s|\$SSH_PASSWORD|$SSH_PASSWORD|g" -e "s|\$ENCRYPTED_SSH_PASSWORD|$ENCRYPTED_SSH_PASSWORD|g" $file | tee ${file%.*}
+done
```
AverageMarcus, line 43: Am I reading this right that it mutates the config files? Would this be a one-time operation? Do we have anything to reset it for the next run? What about when running in a container environment with a read-only filesystem?

abhay-krishna: The existing config files are converted into template files and the `sed` command operates on the template files and regenerates the config files. This was done for the purpose of re-using them since the template files will never be overwritten.

AverageMarcus: 👍  Gotcha! 

So just a restriction for read-only env but I actually think we already have that issue anyway so not a concern! 😄

drew-viles:
ok, so Openstack and qemu worked fine for me so I'm happy to say it
/lgtm

drew-viles:
The only thing I'd like to see is those files in the .gitignore we mentioned before :-)

abhay-krishna:
> The only thing I'd like to see is those files in the .gitignore we mentioned before :-)

So instead of placing them in the root .gitignore, I have created gitignores in the individual provider directories
```
images/capi/packer/nutanix/.gitignore
images/capi/packer/ova/.gitignore
images/capi/packer/proxmox/.gitignore
images/capi/packer/qemu/.gitignore
images/capi/packer/raw/.gitignore
```
And de-duplicated the generated files list since gitignore matches just by the file name without requiring the entire path.

drew-viles:
aaah perfect, I must have missed that during my testing :-D Thanks!

drew-viles (commented):
I'm going to approve it as i'm happy with what I'm seeing and what I've tested.
I'm also going to hold it so it's not just me approving and shipping it :shipit: .

Once others have approved we can unhold and create a release.

Thanks @abhay-krishna !!

/hold
/approve

drew-viles:
If @mboersma or @AverageMarcus could take a look to eyeball it too for approval that'd be greatly appreciated. Testing side I've ran a few that I can successfully as noted above and @abhay-krishna has done a decent amount on their side too so I'm confident of this fix.

mboersma (approved):
/lgtm
/hold cancel

PR #1609, opened 2024-10-15
outscale: fix goss error: <.Vars.OS_VERSION>: map has no entry for key "OS_VERSION"

pli01:
## Change description

Trying to build on outscale, goss complains with error:
```
fix goss error: <.Vars.OS_VERSION>: map has no entry for key "OS_VERSION"
```
This patch fix the error

## Related issues

- Fixes #


mboersma (approved):
/ok-to-test
/lgtm

PR #1616, opened 2024-10-21
outscale: user can make OMI publicly accessible or not

pli01:
## Change description

On outscale, user can choose to make OMI publicly accessible or not
Default: publicly accessible

## Related issues

- Fixes #

### DIFF
--- a/images/capi/packer/outscale/packer.json
+++ b/images/capi/packer/outscale/packer.json
@@ -2,7 +2,7 @@
   "builders": [
     {
       "access_key": "{{ user `access_key` }}",
-      "global_permission": true,
+      "global_permission": "{{ user `global_permission` }}",
       "omi_account_ids": [
         "{{ user `account_id` }}"
       ],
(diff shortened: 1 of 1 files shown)


AverageMarcus:
Just so I'm clear - what does "publicly accessible" mean in this context? I'm wondering if `true` is still a sensible default or not?

pli01:
> Just so I'm clear - what does "publicly accessible" mean in this context? I'm wondering if `true` is still a sensible default or not?
> 
> Especially as the property being updated is called `global_permission`.

I agree with you

This option should be "false" by default, to not publish the image publicly outside the "tenant"
Sharing the image publicly should be enabled voluntarily.

On Outscale, an image can have:
- a private visibility to the "tenant" only
- or "publicly available" and visible to all cloud users, to all tenants

https://developer.hashicorp.com/packer/integrations/outscale/outscale/latest/components/builder/bsu

AverageMarcus:
@mboersma What do you think about changing this default to `false`?

/cc @vincentBaer for your input too as you did a lot of the previous work on this provider I think :)

outscale-hmi:
Hello 
thanks @pli01 for the contribution
@AverageMarcus I agree with these changes, as they provide the flexibility to make images public based on the client's input, while still defaulting to true to ensure accessibility when no specific value is provided which is our initial default value chosen.

AverageMarcus (commented):
/lgtm
/approve

PR #1576, opened 2024-10-07
Fix gMSA setup on Windows

TinaMor:
## PR Description
This PR resilves the following issues caused when trying to import registry keys from registerplugin.reg:
1. Error opening the file. There may be a disk or file system error.
2. Failed to open registry key

**TL;DR**

1. The gMSA setup on WS2025 fails with the error: `Error opening the file. There may be a disk or file system error.`

    https://github.com/kubernetes-sigs/image-builder/blob/e224d7c470797cf0458530105f755cea39140ab5/images/capi/ansible/windows/roles/gmsa/files/install-gmsa-keyvault-plugin.ps1#L119

    To resolve the issue, we now use an absolute path to the registerplugin.reg file since it is copied to the same location where the install-gmsa-keyvault-plugin.ps1 script is executed.

    https://github.com/kubernetes-sigs/image-builder/blob/e224d7c470797cf0458530105f755cea39140ab5/images/capi/ansible/windows/roles/gmsa/tasks/gmsa_keyvault.yml#L43

4. `reg.exe import` fails with "Failed to open registry key"
This happens when the registry keys already exist. To resolve this, we parse the registerplugin.reg file (downloaded from [kubernetesartifacts.azureedge.net/ccgakvplugin](https://kubernetesartifacts.azureedge.net/ccgakvplugin/v1.1.4/binaries/windows-gmsa-ccgakvplugin-v1.1.4.zip)) to identify all the registry keys that will be updated by importing registerplugin.reg, we transfer ownership to the user and then restore it to the original owner.

PR #1619, opened 2024-10-22
Skip Get-WindowsFeature goss spec for windows-2025

### LINKED ISSUE #1603: Azure windows-2025 builds failing: timeout during goss tests

mboersma:
## Change description

Tries the `Get-WindowsFeature` goss test spec twice to work around an issue with windows-2025. Also restores the Azure windows-2025 build to CI runs.

## Related issues

- Fixes #1603

### DIFF
--- a/images/capi/packer/goss/goss-package.yaml
+++ b/images/capi/packer/goss/goss-package.yaml
@@ -76,9 +76,11 @@ package:
 
 #################################################################################
 ######## TODO: [SEPTEMBER 2024] Revert timeout to 60000ms after image GA ########
-# The WS2025 preview image build is failing due to timeout issues when running 
-# PowerShell command to Get Windows Features.
+# The WS2025 preview image build is failing due to timeout issues when running
+# PowerShell command to Get Windows Features, although the features are enabled.
+# Skip this test on WS2025 for now.
 {{ if eq .Vars.OS "windows"}} # Windows
+{{ if ne .Vars.distribution_version "2025" }}
 # Workaround until windows features are added to goss
 command:
 {{range $name, $vers := index .Vars .Vars.OS "common-windows-features"}}
@@ -89,5 +91,6 @@ command:
     - {{.}}
     timeout: 30000 # Give it enough time to retry if it fails
     {{end}}
+{{end}}
 {{end}}
 {{end}}
--- a/images/capi/azure_targets.sh
+++ b/images/capi/azure_targets.sh
@@ -1,7 +1,7 @@
 VHD_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 mariner-2 azurelinux-3 rhel-8 windows-2019-containerd windows-2022-containerd"
 VHD_CI_TARGETS="ubuntu-2204 ubuntu-2404 azurelinux-3 windows-2022-containerd"
 SIG_TARGETS="ubuntu-2004 ubuntu-2204 ubuntu-2404 centos-7 mariner-2 azurelinux-3 rhel-8 windows-2019-containerd windows-2022-containerd windows-2025-containerd flatcar"
-SIG_CI_TARGETS="ubuntu-2204 ubuntu-2404 azurelinux-3 windows-2022-containerd flatcar"
+SIG_CI_TARGETS="ubuntu-2204 ubuntu-2404 azurelinux-3 windows-2022-containerd windows-2025-containerd flatcar"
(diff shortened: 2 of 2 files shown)


images/capi/packer/goss/goss-package.yaml
```diff
@@ -76,14 +76,14 @@ package:
...
-# The WS2025 preview image build is failing due to timeout issues when running 
+# The WS2025 preview image build is failing due to timeout issues when running
 # PowerShell command to Get Windows Features.
 {{ if eq .Vars.OS "windows"}} # Windows
 # Workaround until windows features are added to goss
 command:
 {{range $name, $vers := index .Vars .Vars.OS "common-windows-features"}}
   "Windows Feature - {{ $name }}":
-    exec: powershell -command "(Get-WindowsFeature {{ $name }} | select *)"
+    exec: powershell -command "(for ($i = 0; $i -lt 2; $i++) {Get-WindowsFeature {{ $name }} | select *})"
```
TinaMor, line 86: @mboersma Not sure if this will work in the 30sec. Instead, we are using goss to retry. If it fails in 30sec, we retry until the max time out in [goss-args.json](https://github.com/kubernetes-sigs/image-builder/pull/1576/files#diff-f67dff651a08a166195ed31a7737d62fa8605f9ccac20870d18afca153a9db31)

mboersma: I had also previously tried adding some ansible debug statements right after we enable the features, using `community.windows.win_feature_info`. It showed both features were installed correctly, but it also seemed to have the side effect of making the existing goss test pass. 🤷🏻

TinaMor: Here is a sample run:
[artifacts/azure-sigs/windows-2025-containerd.log](https://[storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_image-builder/1576/pull-azure-sigs/1848753512725352448/artifacts/azure-sigs/windows-2025-containerd.log](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_image-builder/1576/pull-azure-sigs/1848753512725352448/artifacts/azure-sigs/windows-2025-containerd.log))

TinaMor: > adding some ansible debug statements right after we enable the features,

Interesting. This may work. The issue is with `Get-WindowsFeature` cmdlet. It fails on the first try on a fresh install, but passes the when executed again. Perhaps `community.windows.win_feature_info` already calls the cmdlet so when the test is run by Goss, it doesn't fail. Will investigate.

Should we skip the test for 2025 for now?

mboersma: We are skipping the windows-2025 SIG test in CI currently as of #1604.

As for the debug statement side effect, I only ran it twice so I'm not that confident, but it did fail 100% of the time before that.

TinaMor: We could add back the WS2025 CI. The issues with 2025 are gMSA (addressed by #1576) and timeout when querying Windows features (investigating).

If we skip the Windows feature test, the tests should run successfully.

mboersma:
/retitle Skip Get-WindowsFeature goss spec for windows-2025

images/capi/packer/goss/goss-package.yaml
mboersma: I redid this PR just to skip the `Get-WindowsFeature` test on windows-2025 for now.

jsturtevant:
Unfortunate that the powershell command doesn't work properly on Windows 2025.  

/lgtm
/approve

PR #1624, opened 2024-10-29
add rhel image info and cleanup

cpanato:
- add rhel image info and cleanup

/hold for the gce job

/assign @mboersma @richardcase @damdo


mboersma (approved):
/lgtm
/approve

PR #1634, opened 2024-11-26
[Fix] - Removing Config Specification tcp_congestion_control 

BrennenMM7:
## Change description

- This PR will remove the `tcp_congestion_control` specification of BBR from the image-builder. Cloud providers default's are typically the preferred and most optimal. AKS, EKS and GKE are all set to `cubic` by default. 

- I've also noticed that when TCP congestion does occur (at least in Azure) BBR performs poorly and packet drops occur at a much higher rate. 

## Related issues

- Fixes #
N/A (Will look for relevant issues if any).

### DIFF
--- a/images/capi/ansible/roles/node/tasks/main.yml
+++ b/images/capi/ansible/roles/node/tasks/main.yml
@@ -58,7 +58,6 @@
     - { param: net.ipv4.ip_forward, val: 1 }
     - { param: net.ipv6.conf.all.forwarding, val: 1 }
     - { param: net.ipv6.conf.all.disable_ipv6, val: 0 }
-    - { param: net.ipv4.tcp_congestion_control, val: bbr }
     - { param: vm.overcommit_memory, val: 1 }
     - { param: kernel.panic, val: 10 }
     - { param: kernel.panic_on_oops, val: 1 }


kkeshavamurthy:
Congrats on your first PR here :)
Looks like it was added to reduce latency https://github.com/kubernetes-sigs/image-builder/commit/0e74c18e83829992cd1155e6969dc50f6f3293fa

Maybe you can make the change configurable so that users can turn it off if needed?

BrennenMM7:
> Congrats on your first PR here :) Looks like it was added to reduce latency [0e74c18](https://github.com/kubernetes-sigs/image-builder/commit/0e74c18e83829992cd1155e6969dc50f6f3293fa)
> 
> Maybe you can make the change configurable so that users can turn it off if needed?

I don't think that it's actually achieving any gains in latency, any provider I test in latency stays generally the same while throughput suffers. Granted the only way to test this is by performing extreme benchmarks to trigger TCP congestion. In my iperf3 tests though it seems to showcase the cloud providers defaults are better. 

The previous change seems to be adding for their specific use case rather than general consensus? 

I'm unsure where I'd place this item to make it configurable, perhaps it should be default first with an option to override?

PR #1637, opened 2024-11-27
add HTTPCallbackAddress, needed when running within a kubernetes cluster

### LINKED ISSUE #1618: Ability to specify callback ip or fqdn & port of HTTP server used by image-builder
## Summary

This feature request is just asking that instead of assuming the callback ip will be the same as the ip where the server is running, (which used to makes sense in the days before kubernetes), is to instead have a variable for the ip where the server is running and another variable for the ip/fqdn to use when accessing the http server... and to let me as a user set that second variable.

e.g. callback_url="http://1.2.3.4"

## Use case

I'm running image builder via a gitlab runner pipeline (which is running in kubernetes) and need to setup configuration for the callback HTTP server.  Using the same IP the server is running on won't work.

## Is your feature request related to a problem? Please describe.
When a pod runs in kubernetes its using a local clusterip, that's not reachable by the outside world.  Instead, it needs to be exposed via a service.  The gitlab runner helm chart has the ability to expose the runner via a service.  So far I do not see where I can specify the IP in image-builder.  Even better, it would be nice to use an ingress and an FQDN as the FQDN would remain consistent (the gitlab runner helm chart has the ability to specify an ingress).

## Describe the solution you'd like
With the gitlab runner running in a pod it can use whatever port it wants, such as port 80.  There's no need to have a random port anymore.  1. Need to be able to specify the port to use so it isn't random.  Now that the port is predictable it can be exposed via a service.  I can take care of that and request a loadbalancer ip.  2. need to be able to specify the ip for image-builder to use in its callback.  Preferably, instead of a loadbalancer ip, I'd rather use an ingress.  I can take care of creating the ingress.  3. need ability to specify an ingress for image-builder to use in its callback.

## Potential configuration
If this ability doesn't already exist, it would be good to add it and configure things via environment variables.  Maybe something like:
```
(top level?)
export CALLBACK_URL="http://10.4.0.71:8080"
export CALLBACK_PORT="8080"
or
export PROXMOX_URL="https://10.0.0.21:8006/api2/json"
export PROXMOX_USERNAME="capmox@pve!capi"
export PROXMOX_CALLBACK_URL="https://image-builder.k.home.net"
export PROXMOX_CALLBACK_PORT="443"
or
export PROXMOX_CALLBACK_URL="http://10.4.0.71:8080"
export PROXMOX_CALLBACK_PORT="8080"
```

---

/kind feature

lknite:
## Change description

When running within a kubernetes workflow/pipelines, the http server used by the vm to communicate back completion cannot be reached directly. Instead, with kubernetes you 'expose' a pod via a LoadBalancer ip or ingress. The exposed ip is available but must be passed in somehow in order for it to be used.

This pull request adds a variable PROXMOX_HTTP_CALLBACK_ADDRESS, which if provided, will be used by the proxmox provider to pass the needed ip to the vm.

Two additional pull requests have been submitted to packer related repos to make this work:
- packer-plugin-sdk: https://github.com/hashicorp/packer-plugin-sdk/pull/268
- packer-plugin-proxmox: https://github.com/hashicorp/packer-plugin-proxmox/pull/305

## Related issues

- Fixes #1618

### DIFF
--- a/images/capi/packer/proxmox/packer.json.tmpl
+++ b/images/capi/packer/proxmox/packer.json.tmpl
@@ -175,6 +175,7 @@
     "crictl_version": null,
     "disk_size": "20G",
     "existing_ansible_ssh_args": "{{env `ANSIBLE_SSH_ARGS`}}",
+    "http_callback_address": "{{env `PROXMOX_HTTP_CALLBACK_ADDRESS`}}",
     "http_directory": "./packer/proxmox/linux/{{user `distro_name`}}/http/",
     "iso_storage_pool": "{{env `PROXMOX_ISO_POOL`}}",
     "kubernetes_cni_deb_version": null,


lknite:
@AverageMarcus of course, ... I like to do documentation right, which usually means a lot of time and effort.  I'd like to first see the pull requests on the 2 packer repos get some traction / added in before I start down that path.

lknite:
@AverageMarcus I'll add a line to the sample clusterctl.yaml on this page (https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/docs/Usage.md#quick-start) with a description; and also add a line for VLAN while I'm in there.  Would it be ok if I were to add the configuration line in a separate pull request or does it need to be added to this one for it to be approved?

AverageMarcus:
It'd be best to have it included in this PR to make sure documentation is in place when a release is made. Unless you have a solid reason for keeping them separate?

---

One thing I've just realised though - this change is reliant on upstream Packer changes being released, correct?

Are those changes only in plugins and not Packer _itself_? Because we're currently pinned to an old version of Packer and cannot upgrade.

Also, can we mark this as Draft until the upstream changes are available to use?

lknite:
@AverageMarcus Only plugins, packer-plugin-sdk & packer-plugin-proxmox, if that works.  Is the packer-plugin-sdk version pinned as well?

I have a few reasons to want to do the documentation as a pull request later (mainly that i'm not a git master and there's a good chance i'll break this pull request trying to add to it and squash it) but it doesn't matter, I can throw something together sooner than later.  We might have some time anyways, I haven't seen any comments on the two packer related pull requests, not sure what the turn-around time is on those projects, or if I need to do something else like attend a meeting or submit something additional to a review board.

mboersma:
image-builder only pins Packer itself, so I think your upstream plugin changes will be available to image-builder (eventually).

I agree with Marcus; we should include doc changes in this PR if possible.

lknite:
keeping alive

PR #1648, opened 2024-12-05
Enable cleanup for /var/log/messages on Photon for vSphere Provider

### LINKED ISSUE #1647: Enable cleanup for /var/log/messages on Photon for vSphere Provider
Enable cleanup for /var/log/messages on Photon for vSphere Provider

- Create cluster and let it run for few days.
- Run applications which pushes output to /var/log/messages (like GPU operators).
- Since no log cleanup or rotation is enabled for /var/log/messages, disk will get filled up and can lead to disk pressure

/var/log/messages be rotated or cleaned up on regular interval

---

/kind bug

sriramandev:
## Change description

This PR adds logrotate binary and corresponding configurations to enable log rotation and cleanup for /var/log/messages in Photon for vSphere Provider

## Related issues

- Fixes #1647

### DIFF
--- a/images/capi/ansible/roles/providers/defaults/main.yml
+++ b/images/capi/ansible/roles/providers/defaults/main.yml
@@ -15,3 +15,4 @@
 networkd_dispatcher_download_url: https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.gz
 packer_builder_type: ""
 build_target: virt
+var_log_messages_maxsize: 100M
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -84,16 +84,19 @@ photon_3_rpms:
   - python-netifaces
   - python-requests
   - jq
+  - logrotate
 
 # Creating photon_4_rpms for adding future packages if needed.
 # Since empty list errors out, jq is added.
 photon_4_rpms:
   - jq
+  - logrotate
 
 # Creating photon_5_rpms for adding future packages if needed.
 # Since empty list errors out, jq is added.
 photon_5_rpms:
   - jq
+  - logrotate
 
 common_virt_rpms:
   - open-vm-tools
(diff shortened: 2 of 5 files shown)


mboersma (commented):
This looks reasonable to me. Could you squash the commits?

sriramandev:
> This looks reasonable to me. Could you squash the commits?

Squashed the commits.

images/capi/ansible/roles/sysprep/defaults/main.yml
```diff
@@ -16,3 +16,4 @@ extra_repos: ""
 pip_conf_file: ""
 remove_extra_repos: false
 flatcar_disable_autologin: false
+journal_log_max_size: 1G
```
mboersma, line 19: Where is this used, as opposed to the `var_log_messages_maxsize` value above?

images/capi/ansible/roles/providers/defaults/main.yml
```diff
@@ -15,3 +15,4 @@
 networkd_dispatcher_download_url: https://gitlab.com/craftyguy/networkd-dispatcher/-/archive/2.1/networkd-dispatcher-2.1.tar.gz
 packer_builder_type: ""
 build_target: virt
+var_log_messages_maxsize: 100M
```
mboersma, line 18: Should we make this configurable?

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -260,4 +260,4 @@
   ansible.builtin.systemd:
     name: ssh
     enabled: true
-  when: ansible_os_family == "Debian"
+  when: ansible_os_family == "Debian"
```
mboersma, line 263: This is a nitpick, but we prefer all files in the repo to end with a newline.

images/capi/ansible/roles/sysprep/defaults/main.yml
sriramandev: This shouldn't be there.... its part of next fix which I want to get in...

mboersma (approved):
/lgtm
/approve

PR #1663, opened 2025-01-08
fix: goss-command.yml gathers kube image names

0ekk:
## Change description

When we override `{{ kubernetes_container_registry }}` to a mirror like `registry.cn-hangzhou.aliyuncs.com/google_containers` which has multi-path, it will cause incorrect image names.
We just need to get to the last part of the image address.

## Related issues

- Fixes #


AverageMarcus:
(Sorry, accidentally clicked close)

/lgtm
/assign @mboersma

mboersma (approved):
/lgtm
/approve

PR #1636, opened 2024-11-26
change proxmox ve image output format to raw from qcow2 

stenwt:
## Change description

Changes proxmox image output format from qcow2, which proxmox does not accept, to raw, which it does

## Related issues

- Fixes #1635 

## Additional context

log output after change: 

```
proxmox-iso.ubuntu-2404: output will be in this color.

==> proxmox-iso.ubuntu-2404: Retrieving ISO
==> proxmox-iso.ubuntu-2404: Trying https://releases.ubuntu.com/noble/ubuntu-24.04.1-live-server-amd64.iso
==> proxmox-iso.ubuntu-2404: Trying https://releases.ubuntu.com/noble/ubuntu-24.04.1-live-server-amd64.iso?checksum=sha256%3Ae240e4b801f7bb68c20d1356b60968ad0c33a41d00d828e74ceb3364a0317be9
==> proxmox-iso.ubuntu-2404: https://releases.ubuntu.com/noble/ubuntu-24.04.1-live-server-amd64.iso?checksum=sha256%3Ae240e4b801f7bb68c20d1356b60968ad0c33a41d00d828e74ceb3364a0317be9 => /home/sten/.cache/packer/85d1bf86e5e0ecdd6e91515a63cc10bdab146dca.iso
==> proxmox-iso.ubuntu-2404: Creating VM
==> proxmox-iso.ubuntu-2404: No VM ID given, getting next free from Proxmox
==> proxmox-iso.ubuntu-2404: Starting VM
==> proxmox-iso.ubuntu-2404: Starting HTTP server on port 8149
==> proxmox-iso.ubuntu-2404: Waiting 10s for boot

```

PR #1665, opened 2025-01-12
Proxmox add config variables

hoppla20:
## Change description

This PR adds two new configuration variables to the Proxmox packer configuration:

- `disk_format`:
  - Depending on the storage backend configured qcow2 may not be available (e.g. ZFS storage pools only support raw disk images)
  - There is already a similar PR, which wants to change the disk format to raw permanently
    - #1636
    - In contrast, this PR only makes the disk format configurable while not changing the default value (qcow2)
- `numa`:
  - Depending on the hardware, enabling numa on the template may be preferred

### DIFF
--- a/docs/book/src/capi/providers/proxmox.md
+++ b/docs/book/src/capi/providers/proxmox.md
@@ -73,7 +73,7 @@ Then the user (not token) must be given the following permissions on the path `/
 
 *We suggest creating a new role, since no built-in PVE roles covers just these.*
 
-Note that you might have to change disk format from `qcow2` to `raw` in `images/capi/packer/proxmox/packer.json.tmpl` if the storage for the VM disk is block only.
+Note that you might have to change disk format from `qcow2` to `raw` using the `disk_format` packer flag if the storage for the VM disk is block only (see the note below).
 
 ### Building images for Flatcar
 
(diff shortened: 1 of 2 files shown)


AverageMarcus:
/ok-to-test
/lgtm
/assign @mboersma 

Just for my understanding - what situations might one want to use numa? Is there some guidance we can add to the docs or would those that need it already know?

hoppla20:
NUMA can make sense on hardware with more than one CPU: https://pve.proxmox.com/wiki/NUMA

> If you enable this feature, your system will try to arrange the resources such that a VM does have all its vCPUs on the same physical socket and only uses memory that is physically connected to that socket.

Because this is an advanced feature requiring knowledge to setup correctly, I don't think it is necessary to document this option. I think those who need it already know.

hoppla20:
But maybe it could be beneficial to document the 'disk_format' option. I don't think that the ZFS storage backend is that uncommon (I don't know whether other storage backends also have the limitation of only supporting raw disk images).

I can add documentation to this PR later, I am currently on mobil.

AverageMarcus:
Yeah that's a good point actually! No rush! Thank you :)

hoppla20:
I have added documentation on how to change the storage type with a brief description in what situation this could be used.

hoppla20:
I rebased the branch onto main due to merge conflicts. Please have a look at the documentation, I'm not sure whether the structure could be improved. I'm open to suggestions.

AverageMarcus:
🤦  Damn! I thought I could fix the merge conflict through GitHub UI but doesn't look like I can. 😞

hoppla20:
Rebased again

hoppla20:
Thank you for making this possible 😄

PR #1668, opened 2025-01-15
Fixes the Windows update categories are not correctly set

### LINKED ISSUE #1667: Windows update does not work when windows_updates_categories specified
```
windows_updates_categories: "CriticalUpdates SecurityUpdates UpdateRollups"
```
No windows update is installed.

## Relevant log output
```
vsphere-iso.vsphere: TASK [systemprep : Install Windows updates based on Categories] ****************
vsphere-iso.vsphere: ok: [default]
vsphere-iso.vsphere:
```
Confirmed that Update did not run.

---

/kind bug

laozc:
## Change description
Fixes the Windows update categories are not correctly set

win_updates module expects a list while string is passed to the module.

## Related issues

- Fixes #1667

### DIFF
--- a/images/capi/ansible/windows/roles/systemprep/defaults/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/defaults/main.yml
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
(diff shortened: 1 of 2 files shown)


AverageMarcus:
/assign @jsturtevant 

LGTM but I'm not aware enough about Windows images to be sure.

mboersma (approved):
/approve

Seems to agree with the ansible docs: https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_updates_module.html#parameter-accept_list

Thanks @laozc!

PR #1670, opened 2025-01-15
Almalinux 9 image on Outscale.

etchevef:
## Change description
This PR adds the necessaries to build an image based on almalinux 9 on Outscale. It also bump the outscale packer module from 1.2.0 to 1.3.0 (latest).

## Additional context
The resulting image has been tested and works for me (No reason why not, it uses redhat based playbooks from other cloud provider) but won't be supported by outscale (I suppose). Proposing the PR just in case.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -375,7 +375,7 @@ DO_BUILD_NAMES 				?=	do-centos-7 do-ubuntu-2004 do-ubuntu-2204 do-ubuntu-2404
 
 OPENSTACK_BUILD_NAMES		?=	openstack-ubuntu-2004 openstack-ubuntu-2204 openstack-ubuntu-2404 openstack-flatcar
 
-OSC_BUILD_NAMES 			?=	osc-ubuntu-2004 osc-ubuntu-2204 osc-almaLinux-9-2024
+OSC_BUILD_NAMES 			?=	osc-ubuntu-2004 osc-ubuntu-2204 osc-almaLinux-9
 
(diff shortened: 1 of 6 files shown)


AverageMarcus:
Hmm... this would introduce yet another OS we'd need to support in image-builder (and we already struggle with the current list of OS's 😅)
I'm not gonna hard-block this but I'd like to get the thoughts from the other maintainers before we proceed. 
/cc @mboersma @jsturtevant @drew-viles

drew-viles:
I'm personally happy for it to go in but it'd have to be on the understanding we can't test or guarantee it. It'd be on someone from/using Outscale to test this with each release as we don't have the capacity to support it. As Marcus said, the list of Os's is already large and we struggle to support them all as it stands.

mboersma:
I'm on the fence: I love seeing new platforms and OS distros supported, but our test coverage is meager and many of the make targets are probably not working. Ideally everything would be tested by CI, or at least we'd have active reviewers in each area.

Is there someone else who works with Outscale we could run this by? I'd like to allow it to merge and mark it as "unsupported" in docs, but TBH that's a slippery slope because in practice that's true for most targets.

andrewlukoshko:
@etchevef 
`osc-almaLinux-9-2024` should be renamed to `osc-almalinux-9`, AlmaLinux doesn't have year in its version.

Hi @mboersma @jsturtevant @drew-viles @AverageMarcus 
AlmaLinux OS Architect and Engineering Steering Committee member here.
We'd love to help with AlmaLinux support in this repository. Not only for Outscale but also for many other targets.
Please tell me how we can help with this. Thanks.

AverageMarcus:
> @etchevef `osc-almaLinux-9-2024` should be renamed to `osc-almalinux-9`, AlmaLinux doesn't have year in its version.
> 
> Hi @mboersma @jsturtevant @drew-viles @AverageMarcus AlmaLinux OS Architect and Engineering Steering Committee member here. We'd love to help with AlmaLinux support in this repository. Not only for Outscale but also for many other targets. Please tell me how we can help with this. Thanks.

Please reach out in the [image-builder Slack channel](https://kubernetes.slack.com/messages/image-builder) where it will be easier to discuss :)

PR #1673, opened 2025-01-16
Support customize Windows admin password when building OVA

### LINKED ISSUE #1672: Allow user to specify password other than a hardcoded one when building Windows OVA
## Is your feature request related to a problem? Please describe.
Current when building the Windows OVA, a hardcoded password is set in the auto-unattended answer file
which may not be a safe practice due to company policies.

## Describe the solution you'd like
Allow user to specify password other than a hardcoded one when building Windows OVA

## Additional context

---

/kind feature

laozc:
## Change description
Allow user to specify password other than a hardcoded one when building Windows

## Related issues

- Fixes #1672

## Additional context
User may set admin password in the variable JSON

```
{
   "windows_admin_password": "mypa33w0rd"
}
```

### DIFF
--- a/images/capi/hack/windows-ova-unattend.py
+++ b/images/capi/hack/windows-ova-unattend.py
@@ -64,16 +64,22 @@ def main():
     
     root = unattend.getroot()
 
-    if data["unattend_timezone"] != "":
+    if data.get("unattend_timezone"):
       modified=1
       setting = set_xmlstring(root, ".//*[@pass='oobeSystem']/*[@name='Microsoft-Windows-Shell-Setup']",'{urn:schemas-microsoft-com:unattend}TimeZone', data["unattend_timezone"])
       print("windows-ova-unattend: Setting Timezone to %s" % data["unattend_timezone"])
     
+    admin_password = data.get("admin_password")
+    if admin_password:
+      modified=1
+      set_xmlstring(root, ".//*[@pass='oobeSystem']/*[@name='Microsoft-Windows-Shell-Setup']/{*}UserAccounts/{*}AdministratorPassword",'{urn:schemas-microsoft-com:unattend}Value', admin_password)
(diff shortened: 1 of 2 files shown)


images/capi/hack/windows-ova-unattend.py
```diff
@@ -64,17 +64,23 @@ def main():
     
     root = unattend.getroot()
 
-    if data["unattend_timezone"] != "":
+    if data.get("unattend_timezone"):
       modified=1
       setting = set_xmlstring(root, ".//*[@pass='oobeSystem']/*[@name='Microsoft-Windows-Shell-Setup']",'{urn:schemas-microsoft-com:unattend}TimeZone', data["unattend_timezone"])
       print("windows-ova-unattend: Setting Timezone to %s" % data["unattend_timezone"])
```
drew-viles, line 70: I'll prefix this with "I'm no python but" - but if you're changing the conditional to use data.get, would it not make sense to also repeat pattern where `data["unattend_timezone"]` is used in the rest of the block on line 69 and 70?

fwiw: Not a requirement for approval, just a question before I do.

mboersma: I think this is legit python (at least, I'd write it this way). Once you're inside the `.get()` block you can assume the key exists and just reference it.

mboersma:
/retest

URL is accessible from here, so maybe a transient network flake.

randomvariable:
Happened twice?

 "https://stable.release.flatcar-linux.net/amd64-usr/4152.2.0/flatcar_production_iso_image.iso.DIGESTS.asc":
dial tcp [64:ff9b::8890:3a71]:443: connect: network is unreachable in
"file:https://stable.release.flatcar-linux.net/amd64-usr/4152.2.0/flatcar_production_iso_image.iso.DIGESTS.asc"
 

not sure what's up? definitely ipv6 capable in prow?

randomvariable:
i'll do

/retest

in the meantime

PR #1684, opened 2025-02-13
Set proxmox oem-id for flatcar

vxav:
## Change description

At the moment, `export OEM_ID=proxmoxve` is required for flatcar to look for metadata in the right place. Otherwise it is set to `qemu` by default and ignition cannot find metadata. The documentation for this is buried in [packer/proxmox](https://github.com/kubernetes-sigs/image-builder/blob/826a2d7e05288b72870ae45797e9d64efc101d07/images/capi/packer/proxmox/README.md#flatcar-for-proxmox).

Unless there is a good reason for this, it would be simpler to have it set statically for `flatcar-proxmox` overall.

Tested without `export OEM_ID=proxmoxve` and it works.

---

✅ This is not a breaking change if:

- You use `export OEM_ID=proxmoxve` in your workflows. In this case you can remove it.

⚠️ This is a breaking change if:

- You don't use `export OEM_ID=proxmoxve` in your workflows, meaning you use `qemu` oem (default before this PR). In this case, set `export PACKER_FLAGS="--var 'oem_id=qemu'"`
- You use `export OEM_ID=<oem>` in your workflows where `<oem>` is anything else than `proxmoxve`. In this case, set `export PACKER_FLAGS="--var 'oem_id=<oem>'"` and remove `export OEM_ID=<oem>`.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/packer/proxmox/flatcar.json
+++ b/images/capi/packer/proxmox/flatcar.json
@@ -16,7 +16,7 @@
   "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
   "kubernetes_cni_source_type": "http",
   "kubernetes_source_type": "http",
-  "oem_id": "{{env `OEM_ID`}}",
+  "oem_id": "proxmoxve",
   "os_display_name": "Flatcar Container Linux ({{env `FLATCAR_CHANNEL`}} channel release {{env `FLATCAR_VERSION`}})",
   "python_path": "/opt/bin/builder-env/site-packages",
   "release_version": "{{env `FLATCAR_VERSION`}}",
--- a/images/capi/packer/proxmox/packer.json.tmpl
+++ b/images/capi/packer/proxmox/packer.json.tmpl
@@ -207,6 +207,7 @@
     "memory": "2048",
     "mtu": "{{env `PROXMOX_MTU`}}",
     "node": "{{env `PROXMOX_NODE`}}",
+    "oem_id": {{ user `oem_id` }},
     "proxmox_url": "{{env `PROXMOX_URL`}}",
     "sockets": "2",
     "ssh_password": "$SSH_PASSWORD",
--- a/docs/book/src/capi/providers/proxmox.md
+++ b/docs/book/src/capi/providers/proxmox.md
@@ -73,6 +73,16 @@ Then the user (not token) must be given the following permissions on the path `/
 
 *We suggest creating a new role, since no built-in PVE roles covers just these.*
 
+Note that you might have to change disk format from `qcow2` to `raw` in `images/capi/packer/proxmox/packer.json.tmpl` if the storage for the VM disk is block only.
+
+### Building images for Flatcar
+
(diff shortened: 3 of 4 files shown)


vxav:
Ping @mcbenjemaa who added Flatcar support.

images/capi/packer/proxmox/flatcar.json
```diff
@@ -16,7 +16,7 @@
   "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
   "kubernetes_cni_source_type": "http",
   "kubernetes_source_type": "http",
-  "oem_id": "{{env `OEM_ID`}}",
+  "oem_id": "proxmoxve",
```
mcbenjemaa, line 19: I'm not sure about this,
cause maybe some people wanted to use another OEM while building for proxmox.
I know However, they could use QEMU build.

mcbenjemaa: @mboersma what do you think?

vxav: The other option would be to leave it configurable but default it to `proxmoxve`.

AverageMarcus: I suggest adding 
```
"oem_id": "proxmoxve"
```
to the [user variables](https://github.com/kubernetes-sigs/image-builder/blob/826a2d7e05288b72870ae45797e9d64efc101d07/images/capi/packer/proxmox/packer.json.tmpl#L170) and then update this line to be 
```
"oem_id": "{{ user `oem_id` }}"
```
then rather than using environment variables to override the default `proxmoxve` the packer vars can be used inline with other values that can be set. This also makes the property "discoverable" by looking at the defined user variables rather than requiring reading all the docs.

AverageMarcus: This would need to be called out in the changelog though so make sure the PR mentions this is technically a breaking change.

vxav: Addressed in https://github.com/kubernetes-sigs/image-builder/pull/1684/commits/7e82ba7683083ea35e7c118c0102b1db17b0e947
I updated the description with details about the breaking nature of the change.

images/capi/packer/proxmox/packer.json.tmpl
```diff
@@ -207,6 +207,7 @@
     "memory": "2048",
     "mtu": "{{env `PROXMOX_MTU`}}",
     "node": "{{env `PROXMOX_NODE`}}",
+    "oem_id": "proxmoxve",
```
mcbenjemaa, line 210: I think you need to do it the other way around. 
```suggestion
    "oem_id": {{ user `oem_id` }},
```

images/capi/packer/proxmox/flatcar.json
```diff
@@ -16,7 +16,7 @@
   "iso_url": "https://{{env `FLATCAR_CHANNEL`}}.release.flatcar-linux.net/amd64-usr/{{env `FLATCAR_VERSION`}}/flatcar_production_iso_image.iso",
   "kubernetes_cni_source_type": "http",
   "kubernetes_source_type": "http",
-  "oem_id": "{{env `OEM_ID`}}",
+  "oem_id": "{{ user `oem_id` }}",
```
mcbenjemaa, line 19:
```suggestion
  "oem_id": "proxmoxve",
```

vxav: Thanks

mcbenjemaa (approved):

vxav:
done

PR #1687, opened 2025-02-17
feat(outscale): parametrize image owner and use timestamp as suffix

DvdChe:
## Change description

two changes : 
- possibility to specify owner of image by using `owner` var
- replaced suffix name isotime by timestamp so you're able to build image several times a day

## Related issues

- Fixes #


mboersma (approved):
/lgtm

PR #1705, opened 2025-03-11
Move boot iso configuration into boot_iso object

jseely:
## Change description
* Migrates deprecated `iso_*` configurations into `boot_iso` configuration object
* Adds feature flag `iso_download_pve` which causes the proxmox instance to download the ISO (and triggers reuse of the image if it's already been downloaded)
  * Tested with flatcar image creation
  * Ubuntu image creation failed so I left it as false by default

## Additional context
* https://developer.hashicorp.com/packer/integrations/hashicorp/proxmox/latest/components/builder/iso

### DIFF
--- a/images/capi/packer/proxmox/packer.json.tmpl
+++ b/images/capi/packer/proxmox/packer.json.tmpl
@@ -11,6 +11,15 @@
       "communicator": "ssh",
       "cores": "{{user `cores`}}",
       "cpu_type": "{{user `cpu_type`}}",
+      "boot_iso": {
+        "type": "scsi",
+        "iso_file": "{{user `iso_file`}}",
+        "iso_checksum": "{{user `iso_checksum_type`}}:{{user `iso_checksum`}}",
+        "iso_storage_pool": "{{user `iso_storage_pool`}}",
+        "iso_url": "{{user `iso_url`}}",
+        "iso_download_pve": false,
+        "unmount": "{{user `unmount_iso`}}"
+      },
       "disks": [
         {
           "disk_size": "{{user `disk_size`}}",
@@ -28,10 +37,6 @@
       "scsi_controller": "{{user `scsi_controller`}}",
       "http_directory": "{{user `http_directory`}}",
       "insecure_skip_tls_verify": true,
-      "iso_checksum": "{{user `iso_checksum_type`}}:{{user `iso_checksum`}}",
-      "iso_file": "{{user `iso_file`}}",
-      "iso_storage_pool": "{{user `iso_storage_pool`}}",
-      "iso_url": "{{user `iso_url`}}",
       "memory": "{{user `memory`}}",
       "name": "{{user `build_name`}}",
       "network_adapters": [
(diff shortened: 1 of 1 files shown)


drew-viles:
Since this (in theory) should work, we should probably consider setting the `"iso_download_pve": false,` to a user var as well as you have with the other vars so that people can override it if they so wish.

Other than that, all looks good to me!

drew-viles:
@jseely Sorry for the delay on this, We're very short on maintainers at the moment. We'll get to this soon :-) I'm just testing now.

jseely:
No worries, appreciate your service! 😉

drew-viles:
I'm having a heck of a time trying to get this to build. But also when running on main too. Being extremely tired + not being a proxmox user probably isn't helping :-D 

So every time I try and run a build I'm getting
```
1 error(s) occurred:

* one of iso_file, iso_url, or a combination of cd_files and cd_content must be
specified for boot_iso
```
I'm passing in the right params from what I can tell for Ubuntu builds - for example
```
{
        "kubernetes_rpm_version": "1.33.2",
        "kubernetes_semver": "v1.33.2",
        "kubernetes_series": "v1.33",
        "kubernetes_deb_version": "1.33.2-1.1",
        "iso_url": "https://releases.ubuntu.com/noble/ubuntu-24.04.2-live-server-amd64.iso",
        "iso_checksum": "d6dab0c3a657988501b4bd76f1297c053df710e06e0c3aece60dead24f270b4d",
        "iso_storage_pool": "local",
        "storage_pool": "local-zfs"
}
```

You said you'd managed to get this to build on flatcar but had issues with the `iso_download_pve` parameter, did you manage to get a Ubuntu one to build too with that set to false?

I can get it to build if I manually hardcode the values in the `packer.json` file so it does, in theory, work. But I'm unable to get it working with any user vars.

Were you?

drew-viles:
You know what. Ignore me. I had a rouge `ISO_FILE` env var still set. Tiredness it was :-D - build is going now - all being well I'll lgtm it and assign it

drew-viles:
OK so builds are working, however the problem I'm seeing is that I'm having to manually remove either `iso_file` OR `iso_url` from the `packer.json.tmpl` depending on which one I'm using. It seems if they are both supplied, the plugin is none too happy!

This seems to be expected behaviour
https://github.com/hashicorp/packer-plugin-proxmox/blob/8d7f07ffa3e85d730a0726a17e38ddc9687fa211/builder/proxmox/common/config.go#L251-L255

https://github.com/hashicorp/packer-plugin-proxmox/blob/8d7f07ffa3e85d730a0726a17e38ddc9687fa211/builder/proxmox/common/config.go#L680-L698

Problem is, because we're using JSON I'm a little stuck on how to proceed with this.
There are probably "dirty" ways we could achieve it but I don't want to head down that route.


I'll have a think but in the meantime if yourself or anyone else can point out either something I'm missing (again, none proxmox expert here) or another way of achiving this, I'm open to it.

PR #1712, opened 2025-04-08
[windows] Use setupcomplete2 and Nocloud provisioner in azure

jsturtevant:
## Change description

In the last few weeks we've begun to see issues with the provisioner in azure where VM's never complete in the Azure portal.  More details can be found in https://github.com/kubernetes-sigs/windows-testing/issues/496

This sets up image builder to use [Cloudbase's nocloud configuration](https://cloudbase-init.readthedocs.io/en/latest/services.html#nocloud-configuration-drive) to do the same configuration we do today but allow Azure's provisioner to do the heavy lifting.  This prevents us from getting into states where the Azure provisioning process doesn't complete or changes (which is opaque to us, in that we can't know why/if things change).  

This does require a VM to add a data disk that can be formated and loaded with the correct information but luckily this is pretty simple to do with CAPZ: https://capz.sigs.k8s.io/self-managed/data-disks#example.

## Related issues

- Fixes https://github.com/kubernetes-sigs/windows-testing/issues/496


images/capi/packer/azure/scripts/sysprep.ps1
```diff
@@ -22,18 +22,15 @@ Write-Output '>>> Waiting for GA Service (WindowsAzureGuestAgent) to start ...'
...
 $unattendedXml = "$ENV:ProgramFiles\Cloudbase Solutions\Cloudbase-Init\conf\Unattend.xml"
 $FileExists = Test-Path $unattendedXml
-If ($FileExists -eq $True) {
-  # Use the Cloudbase-init provided unattend file during install
-  Write-Output "Using cloudbase-init unattend file for sysprep: $unattendedXml"
-  & $Env:SystemRoot\System32\Sysprep\Sysprep.exe /oobe /generalize /mode:vm /quit /quiet /unattend:$unattendedXml
-}else {
-  & $Env:SystemRoot\System32\Sysprep\Sysprep.exe /oobe /generalize /mode:vm /quit /quiet
-}
+
```
zylxjtu, line 31: why we do not use any unattend.xml here?

jsturtevant: If we do, then Azures default unattend and provision process will not run.  The objective here is to get out of the way of Azure's provisioning process so we don't need to stay up to date with changes they make.

PR #1720, opened 2025-04-25
Add LimitNOFILE configuration for Flatcar OS

njuettner:
When running containerd on Flatcar OS without proper file descriptor limits, 
services like Squid proxy or MySQL exhibit extremely long startup times. 

This occurs because containerd's default LimitNOFILE is set to infinity, which can lead to 
resource management issues and negatively impact application startup performance.

While this change is expected at some point in a future Flatcar release, 
the timeline for when that will happen remains unclear. 

This PR provides the necessary fix in the meantime to ensure consistent behaviour across our supported platforms.


mboersma (approved):
/lgtm
/assign @AverageMarcus

PR #1713, opened 2025-04-09
Fix Flatcar ansible_python_interpreter

ubombi:
## Change description
Drop the old workaround, which overwrites Python path to OS python, while Ansible depends on Python 3.6, installed by as pypy in python role. The proper path is defined in Packer variables, as `/opt/pypy/bin/pypy`

## Related issues
- Fixes #1709 

## Additional context

_Copied from issue_

Flatcar build fails for hcloud _(maybe for others)_ with
> ... SyntaxError: future feature annotations is not defined ... "The following modules failed to execute: ansible.legacy.setup

This happens because ansible uses OS Python (3.9), instead of PyPy (3.6) provisioned by the python role.
The "Set the ansible_python_interpreter fact" overwrites proper `/opt/pypy/bin/pypy` path with `{{ external_binary_path }}/python`

The removed step forces Ansible to use system Python, by overwriting the `ansible_python_interpreter` fact, and causing the error.

Alternatively to deletion, we can hardcode a new `/opt/pypy/bin/pypy` path.

PR #1727, opened 2025-05-06
feat(hcloud): add Flatcar arm64 support

hajowieland:
## Change description

Add support to build Flatcar arm64 images for Hetzner Cloud.

## Additional context

- Added new make targets for hcloud-flatcar-arm64
- Added `-o hetzner` flag to flatcar-install (see [Flatcar - Hetzner docs](https://www.flatcar.org/docs/latest/installing/cloud/hetzner/#template)), otherwise Nodes won't work correctly and no SSH possible
- Replaced hardcoded `containerd_sha256` checksum with `containerd_url` + `.sha256sum` (Ansible supports using a URL for the checksum in its [get_url](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-checksum) module)
  - Kept the old hardcoded checksum as fallback, but IMHO it can be removed if not needed anymore
- Included the required change from #1702 (otherwise goss tests are failing)
- Set the `ARCH` variable to be provided by the user and use this var to generate the correct URLs for crictl and containerd
- Added a new flatcar-arm64 Packer json, which sets all required CPU architecture related variables correctly to arm64
- Included the fix for `ansible_python_interpreter` suggested by @simonostendorf mentioned in https://github.com/kubernetes-sigs/image-builder/pull/1713#issuecomment-2849377883


simonostendorf:
Really cool, thanks @hajowieland. I was also preparing a pull request but I think this PR covers all things that are needed.

images/capi/packer/config/containerd.json
```diff
@@ -3,7 +3,7 @@
   "containerd_cri_socket": "/var/run/containerd/containerd.sock",
   "containerd_gvisor_runtime": "false",
   "containerd_gvisor_version": "latest",
-  "containerd_sha256": "041fa3cfd4e6689d37516e4c7752741df0974e7985d97258c1009b20f25f33c7",
+  "containerd_sha256": "{{ if user `containerd_url` }}{{ user `containerd_url` }}.sha256sum{{ else }}041fa3cfd4e6689d37516e4c7752741df0974e7985d97258c1009b20f25f33c7{{ end }}",
```
AverageMarcus, line 6: I like this. I would prefer it as it's own PR though so it can get called out in the release notes.

Would you mind creating a PR with just this change (and also include the Windows value below?) ? 🙏 

Otherwise I don't mind pulling this out into its own PR first.

images/capi/packer/hcloud/flatcar.json
```diff
@@ -1,5 +1,6 @@
 {
-  "ansible_extra_vars": "ansible_python_interpreter=/opt/pypy/bin/pypy",
+  "ARCH": "amd64",
```
AverageMarcus, line 2: Just want to double-check as I'm not sure - it's correct that this is all upper case, yes?

AverageMarcus: Ah, it's a user variable that's used later. 

Do you know if it's already referenced anywhere in the codebase? Otherwise I think I'd prefer to have it lowercase to match the other vars.

AverageMarcus:
Thanks so much @hajowieland !

I've left a couple small comments.

I'd also like to try and get some parity across the providers and see how much work it'd be to get Flatcar arm64 added to at least the most used providers (ami, azure, gce, qemu). I will try and find some time myself to do this unless you think you would be able to tackle that too?

/ok-to-test

images/capi/packer/hcloud/flatcar.json
simonostendorf: I think `os_arch` is a variable that is already used, maybe we could use this instead of a new variable?

images/capi/packer/hcloud/flatcar-arm64.json
```diff
@@ -0,0 +1,22 @@
+{
+  "ARCH": "arm64",
+  "ansible_extra_vars": "ansible_python_interpreter=/opt/bin/python",
+  "build_name": "flatcar",
+  "crictl_arch": "arm64",
```
simonostendorf, line 5: Is `crictl_arch` used later?

images/capi/packer/hcloud/flatcar.json
AverageMarcus: I've actually seen just `arch` used but also `ARCH` when passed inline to `goos` which I think requires it in that format. 

I think for the user vars I'd prefer to stick with lowercase `arch` so it is consistent.

images/capi/packer/hcloud/flatcar-arm64.json
AverageMarcus: Here I think - https://github.com/kubernetes-sigs/image-builder/blob/f324366fffe17ca124685fd36b7ff9ee135cf6e6/images/capi/packer/config/kubernetes.json#L5

AverageMarcus: I _think_ these vars layer on top of each other but I haven't checked.

images/capi/packer/config/containerd.json
hajowieland: Sure I will create a separate PR for this change. Should we keep the conditional with the hardcoded checksums as fallback? Or do we want to just set `{{ user `containerd_url` }}.sha256sum` for Linux & Windows ?

images/capi/packer/hcloud/flatcar.json
hajowieland: Maybe we use `goss_arch` for setting the inline `ARCH` (as we already do for setting the `arch` in the goss section of the packer JSON) ?

hajowieland:
> I'd also like to try and get some parity across the providers and see how much work it'd be to get Flatcar arm64 added to at least the most used providers (ami, azure, gce, qemu). I will try and find some time myself to do this unless you think you would be able to tackle that too?

@AverageMarcus I’ll create PRs to add Flatcar arm64 support for the other providers over the next few days.

AverageMarcus:
That would be amazing! Thank you so much!

images/capi/packer/hcloud/flatcar.json
hajowieland: I've added a commit so that we now use the lowercase variant `arch` to set the `ARCH` variable as well as in the `containerd_url`.

AverageMarcus:
👍  Looks good @hajowieland! I just want to get https://github.com/kubernetes-sigs/image-builder/pull/1730 merged in first then we can rebase this on top of it to handle the sha256 change :)

images/capi/packer/hcloud/flatcar-arm64.json
hajowieland: I set `crictl_url` to `null` in the hcloud packer JSON files, as we already have the correct URL in packer/config/kubernetes.json
Thanks @AverageMarcus !

AverageMarcus (commented):
/lgtm

drew-viles (commented):
/approve

PR #1731, opened 2025-05-09
docs: Fix ami variable descriptions and add missing default value

hajowieland:
## Change description

In the AWS docs "Common AWS options" table the description of `ami_users` and `snapshot_users` is having the same description as their `*_groups` equivalent.

Also the default value for `snapshot_groups` is empty which is confusing as in the "Building private AMIs" section further down its stating that the default value is `all` (which is correct).


AverageMarcus (commented):
/lgtm
/approve

💙  I always love a docs update! Thank you!

PR #1752, opened 2025-05-16
Make kubelet starting as a windows service

zylxjtu:
## Change description

To support features such as windows graceful shutdown, we will need to make kubelet (itself) running as a windows service. this means we will have to remove the dependency on nssm which will hijack the call SCM made to the running executable. However, the benefit of using nssm is that it can load the envs/config changes dynamically so that without updating the service itself, through the control of config files, the service will run differently. This was used in the CAPZ set up for kublet to dynamically adopt the changes passed from kubeadm. With this PR, we still use nssm to start kubelet in the beginning to apply dynamic changes made through kubeadm. but we will stop/delete the service and re-install/restart the kubelet with the same parameters through SCM, in this way, the kubelet itself will be running as windows service. that's the reason we will need to stop and delete the original service and re-install a new one because the service will be different (except the same service name of "kubelet")

Ideally, we should totally remove the dependency of nssm, maybe this is still the future direction, but I have tried quite a few times and had difficulties to make kubelet (started from SC) fit well with the cloud-init and join the CAPZ cluster. I believe it is achievable, but it probably will make things trickier so maybe this could be a future direction that we can target with.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/windows/roles/kubernetes/tasks/sc.yml
+++ b/images/capi/ansible/windows/roles/kubernetes/tasks/sc.yml
@@ -18,14 +18,20 @@
 - name: Install kubelet as service
   ansible.windows.win_service:
     name: kubelet
-    start_mode: manual
+    start_mode: auto
     path: >
-      {{ kubernetes_install_path }}\kubelet.exe --windows-service
+      "{{ kubernetes_install_path }}\kube-log-runner.exe" --log-file={{ systemdrive.stdout | trim }}/var/log/kubelet/kubelet.log
+      {{kubernetes_install_path }}\kubelet.exe --windows-service
       --cert-dir={{ systemdrive.stdout | trim }}/var/lib/kubelet/pki
       --config={{ systemdrive.stdout | trim }}/var/lib/kubelet/config.yaml
       --bootstrap-kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/bootstrap-kubelet.conf
       --kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/kubelet.conf
-      --hostname-override=$(hostname) --pod-infra-container-image=`"{{ pause_image }}`"
-      --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`"
-      --resolv-conf=`"`" --log-dir={{ systemdrive.stdout | trim }}/var/log/kubelet
-      --logtostderr=false {{ additional_kubelet_args if additional_kubelet_args is defined }}
+      --pod-infra-container-image="{{ pause_image }}"
+      --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=""
+      --container-runtime-endpoint="npipe:////./pipe/containerd-containerd"
+      --resolv-conf=""
+
+- name: Create file to restart kubelet as a windows service
+  ansible.windows.win_template:
+    src: templates/RestartKubelet.ps1
+    dest: "{{ kubernetes_install_path }}\\RestartKubelet.ps1"
--- a/images/capi/ansible/windows/roles/kubernetes/templates/RestartKubelet.ps1
+++ b/images/capi/ansible/windows/roles/kubernetes/templates/RestartKubelet.ps1
@@ -1,0 +1,54 @@
+# Copyright 2025 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Need to keep sync with StartKubelet.ps1
+$FileContent = Get-Content -Path "$env:SYSTEMDRIVE/var/lib/kubelet/kubeadm-flags.env"
+$kubeAdmArgs = $FileContent.TrimStart('KUBELET_KUBEADM_ARGS=').Trim('"')
+
+$args = "--cert-dir=$env:SYSTEMDRIVE/var/lib/kubelet/pki",
+        "--config=$env:SYSTEMDRIVE/var/lib/kubelet/config.yaml",
+        "--bootstrap-kubeconfig=$env:SYSTEMDRIVE/etc/kubernetes/bootstrap-kubelet.conf",
+        "--kubeconfig=$env:SYSTEMDRIVE/etc/kubernetes/kubelet.conf",
+        "--hostname-override=$(hostname)",
+        "--pod-infra-container-image=`"{{ pause_image }}`"",
+        "--enable-debugging-handlers",
+        "--cgroups-per-qos=false",
+        "--enforce-node-allocatable=`"`"",
+        "--resolv-conf=`"`"",
+        "--windows-service"
+
+$KubeletArgListStr = ($args -join " ") + " $kubeAdmArgs"
+$KubeletArgListStr = $KubeletArgListStr.Replace("`"", "\`"")
+# Used by sc.exe to create the service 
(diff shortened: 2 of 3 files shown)


sriramandev:
@laozc Can you please take at view of this PR?

sriramandev:
Do we have a issue related to this PR?

marosset:
/lgtm 
This is related to https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/4802-windows-node-shutdown but I don't see us discussing logging in that KEP.
@zylxjtu do you remember where we had those discussions?

jsturtevant (commented):
Why does the service need to deleted and restarted in the script?

zylxjtu:
> /lgtm This is related to https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/4802-windows-node-shutdown but I don't see us discussing logging in that KEP. @zylxjtu do you remember where we had those discussions?

I remembered that we had some discussions at that time, but maybe it is landed in being removed from the KEP. The idea is to utilize the kube-log-runner (from the k8s repo) to redirect the log to stdout/stderr

zylxjtu:
> Why does the service need to deleted and restarted in the script?
This is the solution I have after a few try of other options. let me try to explain a bit here so that more people may have a better idea on what I'm trying to achieve here. To support features such as windows graceful shutdown, we will need to make kubelet (itself) running as a windows service. this means we will have to remove the dependency on nssm which will hijack the call SCM made to the running executable.  However, the benefit of using nssm is that it can load the envs/config changes dynamically so that without updating the service itself, through the control of config files, the service will run differently. This was used in the CAPZ set up for kublet to dynamically adopt the changes passed from kubeadm. With this change, we still use nssm to start kubelet in the beginning to apply dynamic changes made through kubeadm. but we will stop/delete the service and re-install/restart the kubelet with the same parameters through SCM. that's the reason we will need to stop and delete the original service and re-install a new one because the service will be different (except the same service name of "kubelet")

Ideally, we should totally remove the dependency of nssm, maybe this is still the future direction, but I have tried quite a few times and had difficulties to make kubelet (started from SC) fit well with the cloud-init and join the CAPZ cluster. I believe it is achievable but it probably will make things trickier so maybe this could be a future direction that we can target with.

images/capi/ansible/windows/roles/kubernetes/tasks/nssm.yml
```diff
@@ -26,6 +26,11 @@
     src: templates/StartKubelet.ps1
     dest: "{{ kubernetes_install_path }}\\StartKubelet.ps1"
 
+- name: Create file to restart kubelet as a windows service
+  ansible.windows.win_template:
+    src: templates/RestartKubelet.ps1
+    dest: "{{ kubernetes_install_path }}\\RestartKubelet.ps1"
+
```
jsturtevant, line 33: are you modify the default install to get sc? https://github.com/kubernetes-sigs/image-builder/blob/81195c751067238dd4cad651f5d17c0baa522d00/images/capi/packer/config/windows/common.json#L13  I would have thought this change here wouldn't be needed since the file is updated in the SC.yaml template which restarts and loads the file.

We should also update the document to say that Service manager now supports kubeadm.
https://image-builder.sigs.k8s.io/capi/windows/windows#service-manager

images/capi/ansible/windows/roles/kubernetes/templates/RestartKubelet.ps1
```diff
@@ -0,0 +1,56 @@
...
+        "--enforce-node-allocatable=`"`"",
+        "--resolv-conf=`"`"",
+        "--windows-service"
+
+$KubeletArgListStr = ($args -join " ") + " $kubeAdmArgs"
+$KubeletArgListStr = $KubeletArgListStr.Replace("`"", "\`"")
+# Used by sc.exe to create the service 
+$KubeletCommandLine =  "`"" + "\`"" + "$env:SYSTEMDRIVE\k\kube-log-runner.exe" + "\`" " + "--log-file=/var/log/kubelet/kubelet.log " + "$env:SYSTEMDRIVE\k\kubelet.exe " + $KubeletArgListStr + "`""
+
+# Write-Output $kubeletCommandLine
```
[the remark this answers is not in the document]
zylxjtu: prefer to keep it here as the parameters (kubeletcommdnline) send to sc.exe tend to be tricky, and we can use this to check that kind of commandline is used to start the kubelet service, when we need to do some debugging

images/capi/ansible/windows/roles/kubernetes/tasks/sc.yml
```diff
@@ -14,18 +14,7 @@
...
-      {{ kubernetes_install_path }}\kubelet.exe --windows-service
-      --cert-dir={{ systemdrive.stdout | trim }}/var/lib/kubelet/pki
-      --config={{ systemdrive.stdout | trim }}/var/lib/kubelet/config.yaml
-      --bootstrap-kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/bootstrap-kubelet.conf
-      --kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/kubelet.conf
-      --hostname-override=$(hostname) --pod-infra-container-image=`"{{ pause_image }}`"
-      --enable-debugging-handlers --cgroups-per-qos=false --enforce-node-allocatable=`"`"
-      --resolv-conf=`"`" --log-dir={{ systemdrive.stdout | trim }}/var/log/kubelet
-      --logtostderr=false {{ additional_kubelet_args if additional_kubelet_args is defined }}
+- name: Create file to restart kubelet as a windows service
```
msft-linliu, line 17: Why add this file in both sc.yml and nssm.yml? Only one is needed?

zylxjtu: I will revert the change in sc.yaml as this is not the one we will use

zylxjtu:
> Do we have a issue related to this PR? Also can you please elaborate the description. Looking at trying to understand whats the core issue thats being addressed here.

Added more explanation/clarification

images/capi/ansible/windows/roles/kubernetes/templates/RestartKubelet.ps1
```diff
@@ -0,0 +1,54 @@
...
+for ($i = 0; $i -lt 10; $i++) {
+	$service = Get-Service -Name kubelet -ErrorAction SilentlyContinue
+	if ($null -eq $service) {
+		Write-Host "kubelet service deleted successfully, restarting"
+		sc.exe create kubelet binPath= $KubeletCommandLine start= auto depend= containerd
+		sc.exe start kubelet
+		return
+	}
+	else {
+		Write-Host "Waiting for service to be fully deleted... (attempt $($i + 1)/$Retries)"
```
rzlink, line 48: The variable `$Retries` is undefined here.

images/capi/ansible/windows/roles/kubernetes/tasks/nssm.yml
jsturtevant: Doesn't this section in `restartkubelet` do the same thing when it is installed as a service via Service manager?  Otherwise I don't think it would work?

```
$FileContent = Get-Content -Path "$env:SYSTEMDRIVE/var/lib/kubelet/kubeadm-flags.env"
$kubeAdmArgs = $FileContent.TrimStart('KUBELET_KUBEADM_ARGS=').Trim('"')
```

zylxjtu: yes, startkubelet.ps1 and restartkubelet.ps1are essentially they are the same, expect some special escape required by sc.exe. The different is that. nssm will register the powershell startkubelet.ps1 as windows service, which mean during boot time, it will read the env file and reload the parameters. This unfortunately is not the case for sc.exe, which mean after we register the kubelet.exe as a windows service, its parameters will be static (it can not read parameters from env file).

I had also tried to manipulate kublet to read from config files such as (kubelet.json etc), but looks that not all kubelet commandline parameter will have equivalent support in the json/yaml config file.

zylxjtu: Yes, this can definitely be improved, although I would expect a good amount of efforts probably needed.

In this PR, basically, we just added one more script, which will be pretty safe. people can still choose to use it or stick with the old way if they want.

images/capi/packer/config/windows/common.json
```diff
@@ -10,7 +10,7 @@
   "prepull": "true",
   "runtime": "containerd",
   "ssh_source_url": "",
-  "windows_service_manager": "nssm",
+  "windows_service_manager": "windows_service",
```
zylxjtu, line 13: or maybe I should make it applied to azure only in the begining?

jsturtevant (approved):

Issue #1743, opened 2025-05-13
Publish Ansible Galaxy Collection


Lirt:
## Is your feature request related to a problem? Please describe.

I would like to use ansible roles from image-builder repository using ansible-galaxy. This would enable users to use collections (or perhaps individual roles) outside of this repository (in their CI). Ansible roles are already well written and generalized. It is possible to use them also outside this image building process.

## Describe the solution you'd like

Add ansible-galaxy meta files and CI commands to build and publish collection of roles for Linux/Unix like systems and Windows. There would be 2 collections. I don't need individual roles to be published. Roles can be published on git tag (eg. next version - tag 0.1.43 would publish collection with semver 0.1.43).

## Describe alternatives you've considered

I was using `git clone` to get the roles into my pipeline. There already exists `ansible` native approach (galaxy), so I am opening this feature request to see if community is also interested to avoid "hacky" approach of git cloning and use proper way of packaging the collection using `ansible-galaxy`.

## Additional context

I will open PR with change proposal. This however has to be set up by some of the maintainers as it requires to create some organization/namespace at https://galaxy.ansible.com. 

---
/kind feature

PR #1756, opened 2025-05-20
Publish roles and playbooks as Ansible galaxy collection

Lirt:
## Publish roles and playbooks as Ansible galaxy collection

This PR sets up CI that publishes ansible roles and playbooks as Ansible Galaxy Collection.

By pushing image-builder roles into galaxy we allow users to use them without cloning full git repository.

## Related issues

- Fixes https://github.com/kubernetes-sigs/image-builder/issues/1743

## Additional context

- To publish also playbooks into collection, playbooks now must live under `./playbooks` in the galaxy dir.
- First version of the galaxy collection can be v0.2.0 to avoid breaking contract with users that clone this repository to set up CI from it.
- Galaxy repository `kubernetes_sigs` has to be set up by a maintainer with access to `kubernetes_sigs` group. Possibly other namespace names could be used. 
- Additional README files had to be created to pass galaxy lint and LICENSEs added as symlinks.


drew-viles:
Hey! Thanks for the PR.

Whilst this looks fine on the face of it, we are pretty low on maintainers and will potentially be losing another one soon.  The only problem I have is that by adding in the Galaxy support, it will introduce another area of expectation around support of the feature. I personally wouldn't have the time to keep on top of this as well as all of the providers we currently support too. It's also only moving a subset of the roles over too which then brings in questions around what should be considered a Galaxy collection and what should be considered native to the repo.

I'll have to see how other maintainers feel about this as obviously this is one persons opinion 😄 but right now I'm not sure we could merge this in and provide it the support it would need moving forward and I don't want to see it just start to rot as a result.

AverageMarcus:
Adding my thoughts to this...

We generally want to only encourage a single way of using image-builder to ensure we can support it as much as possible. Officially this is via the container image we provide but we also understand that a lot of people run it directly by checking out the repo and running Make so we also support that.

We have had similar discussions in the past about other ways of using image-builder (such as from within EKS-Anywhere) and we've said we can't support it as we don't have the skills, time or resources to do so.

If we had more test coverage across the different providers and configurations to the point where we had more confidence in changes this might be a different discussion but right now it's too much of a burden on the maintainers and contributors unfortunately.

Lirt:
Hi,

thanks for looking at the PR.

This change shouldn't add extra maintenance burden. It only packages the ansible part into an ansible-native artifact. Any new role or playbook that is added under galaxy repository `images/capi/ansible/` or `images/capi/ansible/windows/`) is automatically included into a collection. This change doesn't affect build processes - only ansible roles. Ansible artifacts are released automatically by tagging the repo (in the CI I added) (this repository is already being tagged).

> It's also only moving a subset of the roles over too which then brings in questions around what should be considered a Galaxy collection and what should be considered native to the repo.
All of the roles are included in the collection (no subset). You can check it in my fork:
1. https://galaxy.ansible.com/ui/repo/published/lirt/image_builder/content/?showing=role
2. https://galaxy.ansible.com/ui/repo/published/lirt/image_builder_windows/content/

The MR required some renaming changes (generally playbooks must use absolute paths) and some file moves, but it was very small. People would need to get used to reference roles in playbooks with absolute paths.

Let me know what do you think.

Lirt:
@drew-viles, @AverageMarcus 

Any reconsideration after my [previous comment](https://github.com/kubernetes-sigs/image-builder/pull/1756#issuecomment-2912121597)?

drew-viles:
Sorry we're pretty thin on the ground for maintainers right now so I just cannot test this. Feel free to keep it open but right now I won't be able to make any progress on this.

PR #1766, opened 2025-06-09
ova: make the root partition xfs for Photon 5, Ubuntu 22.04, and Ubuntu 24.04

palnabarun:
Signed-off-by: Nabarun Pal <pal.nabarun95@gmail.com>

## Change description

This PR updates the root filesystem to `xfs` from `ext4`

## Related issues

None

## Additional context

None


drew-viles:
This seems like a sensible approach and one I've been considering throwing into the qemu/raw approach too, but you're right in your issue, it'd be nice if we could template it out before making such a change.

/approve

PR #1767, opened 2025-06-10
capi: adjust ova script for community infrastructure and fixes

chrischdi:
## Change description

Adjusts changes to make the `pull-ova-all` job work again

xref: https://github.com/kubernetes/test-infra/pull/34959

## Related issues

- Fixes #

PR #1773, opened 2025-06-16
[WIP] OVA E2E: also build photon-5

chrischdi:
## Change description

Adds `photon-5` as target for vSphere based CI

## Related issues

Follow-up from:
- #1767

### DIFF
--- a/docs/book/src/capi/providers/vsphere.md
+++ b/docs/book/src/capi/providers/vsphere.md
@@ -43,7 +43,7 @@ This file must have the following format (`cluster` can be replace by `host`):
     "network": "network_attached_to_template",
     "insecure_connection": "false",
     "template": "base_template_used_by_clone_builder",
-    "create_snbapshot": "creates a snaphot on base OVA after building",
+    "create_snapshot": "creates a snaphot on base OVA after building",
     "linked_clone": "Uses link cloning in vsphere-clone builder: true, by default"
 }
 ```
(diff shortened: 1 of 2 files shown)


chrischdi:
cc @sriramandev

AverageMarcus (commented):
/lgtm
/ok-to-test
/assign @drew-viles

chrischdi:
/retest

(maybe there were ci infra changes during the jobs)

chrischdi:
Hit the timeout of the prowjob

/retest

sriramandev:
Will trigger re-test. Seems to be a flake. ssl handshake timeout when fetching the shasum for containerd.

chrischdi:
Not sure why that is. The env requires mtu to be set correctly (which is handed out via DHCP). Maybe that's the issue. Obviously not an issue for Ubuntu or so.

sriramandev:
> Not sure why that is. The env requires mtu to be set correctly (which is handed out via DHCP). Maybe that's the issue. Obviously not an issue for Ubuntu or so.

I am assuming mtu if is a problem should be consistent with Ubuntu as well right? This does not seem to be an issue locally which I just confirmed.

AverageMarcus:
(Debugging an issue seen in another PR...)

/test pull-ova-all

chrischdi:
/retest

(trying some mtu change on test-infra)

chrischdi:
/retest

(a gcve issue got fixed, internet traffic routing changed. Maybe this resolves it)

chrischdi:
Nice, photon-5 was green, weird issue on Ubuntu, maybe something in the makefile which is not compatible to run in parallel

PR #1782, opened 2025-06-25
Fix/re enable ec2kubernetes datasource

sl1pm4t:
This resets default cloud-init datasource to `Ec2Kubernetes` which is required for Cluster API on AWS images.


sl1pm4t:
Sorry wrong, repo.

PR #1787, opened 2025-06-27
Remove Photon-3 OS as end of life March 2024

AverageMarcus:
## Change description

Photon-3 has been unsupported / end of life since March 2024. As is no longer a supported OS we also shouldn't support it as a built-in option for image-builder.

## Related issues

- N/A

## Additional context

- https://endoflife.date/photon


drew-viles (commented):
/lgtm
/approve

PR #1797, opened 2025-07-01
Introduce hack to avoid rm failing with not found

AverageMarcus:
## Change description

We seen some flakey test failures on PRs where the `rm` command seems to fail due to the file not existing.

For example:
```
rm: cannot remove '/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/packer/ova/linux/rhel/http/7/ks.cfg': No such file or directory
```
Taken from https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_image-builder/1783/pull-ova-all/1939613496043900928/artifacts/flatcar.log

It's not clear how this occurs as the if statement should ensure the file exists before attempting to remove it. 

As this file being missing isn't critical enough to fail the tests (as we're deleting it anyway) this hack ignored failures deleting the file and allows the test to continue as normal.

## Related issues

- Fixes #


drew-viles (commented):
It's what the hack directory is for!

/lgtm
/approve

PR #1798, opened 2025-07-01
Work around 2022 4B image

zylxjtu:
## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
+++ b/images/capi/ansible/windows/roles/systemprep/tasks/main.yml
@@ -125,6 +125,15 @@
 
 # Apply HNS flags for fixes that need to be enabled via Registry
 # these eventually get turned on automatically and can be removed in future releases
+- name: Apply the fix for 2025-4B patch of windows 2022
+  ansible.windows.win_regedit:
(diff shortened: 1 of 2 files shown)


images/capi/ansible/windows/roles/systemprep/tasks/main.yml
```diff
@@ -125,6 +125,15 @@
 
 # Apply HNS flags for fixes that need to be enabled via Registry
 # these eventually get turned on automatically and can be removed in future releases
+- name: Apply the fix for 2025-4B image
```
mboersma, line 128:
```suggestion
- name: Apply the fix for 2022-4B image
```

mboersma (commented):
Could you add any detail about what this fixes or how it works? It's hard to tell what setting `1114842764` does to this registry key.

images/capi/ansible/windows/roles/systemprep/tasks/main.yml
zylxjtu: thanks for the catching, will update

mboersma (commented):
/retest

@zylxjtu could you explain what the value `1114842764` does or provide a reference? It's just a magic number and it would help readers of the code (or even just this PR) if we could explain what it means.

mboersma (approved):
/lgtm
/approve

PR #1802, opened 2025-07-04
Updated Azure CI tests to automatically detect targets

AverageMarcus:
## Change description

Updates the Azure test script to dynamically detect target OSs we support and run tests for those.

We also filter out the following:

* Any RHEL targets (because of subscription requirements)

In addition to the targets current being tested this will also add Ubuntu 20.04 and Windows 2019.

### Before:

VHD:
Screenshot 2025-07-04 at 10 18 47 AM

SIG:
Screenshot 2025-07-04 at 10 20 00 AM

### After:

VHD:
Screenshot 2025-07-04 at 10 18 59 AM

SIG:
Screenshot 2025-07-04 at 11 58 57 AM

## Related issues

- Towards improving https://github.com/kubernetes-sigs/image-builder/issues/1605

Issue #1605, opened 2024-10-15
Testing of all providers and distros


AverageMarcus:
## Is your feature request related to a problem? Please describe.

We've had a few instances now where changes introduced issues into existing targets, e.g.

* https://github.com/kubernetes-sigs/image-builder/issues/881
* https://github.com/kubernetes-sigs/image-builder/issues/879
* https://github.com/kubernetes-sigs/image-builder/issues/828

and most recently https://github.com/kubernetes-sigs/image-builder/pull/1596 caused an issue for all OS's that made use of init data that wasn't caught during manual testing due to focussing on Flatcar.

## Describe the solution you'd like

Where possible we should have automated tests in place to ensure that all currently supported providers and operating systems can build successfully with default values (this might be tricky in situations where there are required values).

This issue is aimed at tracking the progress of this effort and should be updated as progress is made.

Ideally, we should implement provider-specific tests that then build all OSs for that provider (e.g. using the `make build-raw-all` that should trigger all supported OSs).

## Progress 

| Provider | Implemented | Comments |
| --- | --- | --- |
| ami | ❌ | I think the CAPA project had something that they (used to?) use for testing image builds. We should reach out to them and see if its something we can adopt. |
| azure | Partial | We currently have two Prow jobs for Azure - `pull-azure-vhds` & `pull-azure-sigs`. These make use of [`ci-azure-e2e.sh`](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/scripts/ci-azure-e2e.sh) but uses a pre-defined list of target ([`azure_targets.sh`](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/azure_targets.sh)). Ideally we should try to update this to dynamically load all targets we support. |
| digitalocean | ❌ | |
| gce | Partial | We currently have the Prow job ` pull-image-builder-gcp-all` that uses [`ci-gce.sh`](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/scripts/ci-gce.sh) but this is currently configured as an optional test and not automatically run on any PRs. We should update this to at least trigger when changes to GCE files are made, similar to the Azure ones. |
| hcloud | ❌ | |
| nutanix | ❌ | |
| oci | ❌ | |
| openstack | ❌ | |
| outscale | ❌ | |
| ova | ❌ | We recently had to remove this and need a new environment to run the tests in, see https://github.com/kubernetes-sigs/image-builder/issues/1508 |
[...]

PR #1805, opened 2025-07-07
Added provider/os matrix table to Readme

AverageMarcus:
## Change description

I needed a way to better track what we support and don't so I could get an overview of what is tested. This adds a table to the main readme which lists all providers and the OSs they each have built-in to image-builder. The ones currently (or in the case of Azure once https://github.com/kubernetes-sigs/image-builder/pull/1802 is merged) covered by CI tests are highlighted in the table.

## Related issues

- Towards https://github.com/kubernetes-sigs/image-builder/issues/1605

## Additional context


drew-viles (commented):
Very nice addition!

/lgtm
/approve

PR #1806, opened 2025-07-07
Fix gvisor installation and add goss test

ffais:
## Change description
It seems that the gvisor installation does not work properly because the _containerd_gvisor_runtime_ and _containerd_gvisor_version_ variables are not passed to ansible. I added the variables in the ansible-args.json file and this solved the problem.

I took the opportunity to add GOSS tests to to verify that gvisor is installed and configured properly.


AverageMarcus (commented):
/lgtm

PR #1812, opened 2025-07-15
Exclude grub packages from apt-mark hold

kreeuwijk:
## Change description

Fixes the problem of MaaS not being able to update the `grub` packages during deployment, as documented [here](https://discourse.maas.io/t/curtin-command-fails-due-to-grub-efi-amd64-pkg-conflicts/7475/3)

- Is this change including a new Provider or a new OS? (y/n) __n__

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -20,10 +20,19 @@
 - name: Pin all installed packages with apt-mark
   ansible.builtin.shell: |
     set -o pipefail
-    dpkg-query -f '${binary:Package}\n' -W | grep -v grub | xargs apt-mark hold
+    dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
 
   args:
     executable: /bin/bash
+
+- name: Unpin grub packages with apt-mark for MaaS
+  ansible.builtin.shell: |
+    set -o pipefail
+    dpkg-query -f '${binary:Package}\n' -W | grep grub | xargs apt-mark unhold
+
+  args:
+    executable: /bin/bash
+  when: provider == "maas"
 
 - name: Remove extra repos
   ansible.builtin.file:
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -20,7 +20,7 @@
 - name: Pin all installed packages with apt-mark
   ansible.builtin.shell: |
     set -o pipefail
-    dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
+    dpkg-query -f '${binary:Package}\n' -W | grep -v grub | xargs apt-mark hold
 
   args:
     executable: /bin/bash
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -32,7 +32,7 @@
 
   args:
     executable: /bin/bash
-  when: provider == "maas"
+  when: provider is defined and provider is search('maas')
 
 - name: Remove extra repos
   ansible.builtin.file:


mboersma (commented):
This is a specific exception for a use case that hasn't been reported as an image-builder issue. It's also not immediately clear to me how the old [MaaS bug report](https://discourse.maas.io/t/curtin-command-fails-due-to-grub-efi-amd64-pkg-conflicts/7475/3) involves image-builder: they just discuss Packer.

Can MaaS users just run `apt-mark unhold grub` before running `curtin` as a workaround?

kreeuwijk:
I ran into the issue myself when Canonical released an updated version of grub for Ubuntu 22.04. It was working fine before that point but afterwards MaaS deployments would fail as MaaS needs to align grub with the system architecture (legacy BIOS or UEFI), which requires adjusting the grub packages. When the grub packages are held, the process fails.

Curtin isn't helpful here, as you can only control `early_commands` and `late_commands` through curtin_userdata in MaaS. The `early_commands` run even before the image is applied to the disk and `late_commands` runs after grub has already failed.

Hence it would be better if the fix was in the image-builder. What we could do though, is split the original ansible command into 2 variants, and only run the variant that excludes grub packages when it runs for a MaaS build. There would need to be a way to identify that it's running for a MaaS build to do so. Can we set a variable in `images/capi/packer/maas/packer.json` that could be queried from the ansible task?

AverageMarcus:
Could we leverage the `node_custom_roles_post_sysprep` var [used here](https://github.com/kubernetes-sigs/image-builder/blob/15c3fe58e139fb074aee274bdef594e937aa6602/images/capi/ansible/node.yml#L51-L56) to define specific role to be used in the MaaS [packer.json.tmpl](https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/maas/packer.json.tmpl). 

I think this should work but haven't tested it.

kreeuwijk:
Yes that should work, I'll give it a try a update the PR with those changes

kreeuwijk:
I found a cleaner way to do it that doesn't require adding a custom role.
```
    qemu: TASK [sysprep : Pin all installed packages with apt-mark] **********************
    qemu: changed: [default]
    qemu:
    qemu: TASK [sysprep : Unpin grub packages with apt-mark for MaaS] ********************
    qemu: changed: [default]
```

AverageMarcus:
Oh nice! I didn't know we had a provider var 🤣 Do you know where that gets set?

/ok-to-test
/lgtm

AverageMarcus:
Answering my own question - the provider is specifically set in the makefile for the MaaS targets. That's nice. Maybe we should have that as a standard thing 🤔

kreeuwijk:
Yeah setting the `provider` variable would probably be useful to handle this sort of stuff in the cleanest way. Otherwise you end up with endless custom roles. Calling custom roles also doesn't work in the `packer.json.tmpl` but has to be done from the OS json files (such as `maas-ubuntu-2404-efi.json`). That's not ideal either so a conditional task works better.

AverageMarcus:
Looks like the `provider` var needs to also be checked for existing

```
    googlecompute.ubuntu-2204: fatal: [default]: FAILED! => {"msg": "The conditional check 'provider == \"maas\"' failed. The error was: error while evaluating conditional (provider == \"maas\"): 'provider' is undefined. 'provider' is undefined\n\nThe error appears to be in '/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/ansible/roles/sysprep/tasks/debian.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Unpin grub packages with apt-mark for MaaS\n  ^ here\n"}
```

AverageMarcus (commented):
/lgtm
/assign @mboersma @AndiDog 
/cc @glitchcrab

mboersma (approved):
/lgtm
/approve

Thanks @kreeuwijk!

Issue #1817, opened 2025-07-21
[OVA] Better disk packing by shrinking removed data to provide better disk compression numbers


sriramandev:
Currently we do not do any shrinking or byte optimization for any cleaned up data during our node generation /build. Writing zeroes to unused bytes can provide lower disk utilization thereby reducing the exported VMDK disk image size.

sriramandev:
Changes merged as part of [PR](https://github.com/kubernetes-sigs/image-builder/pull/1818)

PR #1818, opened 2025-07-21
Optimize vmdk packing by writing zeroes to unused bytes

sriramandev:
## Change description
This PR writes zeroes to unused bytes and clears them there by reducing the actual disk utilization. 

## Related issues
- Fixes # [1817](https://github.com/kubernetes-sigs/image-builder/issues/1817)

## Additional context
Log output reference:
```

```


AverageMarcus:
I'm not too familiar with this practice. Is this also beneficial to other platforms besides OVA? Should / could it be a default for all?

sriramandev:
> I'm not too familiar with this practice. Is this also beneficial to other platforms besides OVA? Should / could it be a default for all?

This is not specific to OVA or VMDK. When we delete or remove content/files, if not mistaken, only the pointer to the disk location is cleared and not the actual content. What we do here is writing zeroes across and then removing the content. This will help in actually shrinking the disk.

AverageMarcus:
Does this need to be done with Packer or can we achieve it with Ansible? I'm wondering if this could be an optional Ansible role that we could turn on as wanted.

sriramandev:
> Does this need to be done with Packer or can we achieve it with Ansible? I'm wondering if this could be an optional Ansible role that we could turn on as wanted.

This is nothing to do with packer. Its just that its easier to call the script from there as a provisioner script. We can do that from Ansible as well.

AverageMarcus:
I think if possible I'd prefer Ansible to handle it. Less reliant on Packer is better as we still have the pinned version problem.

We also have built in capabilities of user-selected additional roles which means this could be leveraged in other providers without copy&paste :) 

What do you think?

sriramandev:
> I think if possible I'd prefer Ansible to handle it. Less reliant on Packer is better as we still have the pinned version problem.
> 
> We also have built in capabilities of user-selected additional roles which means this could be leveraged in other providers without copy&paste :)
> 
> What do you think?

One of the think I missed mentioning is the fact that this should be run as the last step. There should not be any disk writing to happen after this. Hence this is part of the packer provisioner which runs at the end just before the goss validations.

AverageMarcus:
Ah yeah, that makes sense. Thank you 

/lgtm
/assign @mboersma @drew-viles @jsturtevant @AndiDog

ohauer:
Another thought, 
in packer/ova/packer-common.json.tmpl the following is defined
```yaml
"disk_thin_provisioned": "true"
```
which is circumvented by the dd command because the virtual disk on the host is expanded to the full size

PR #1821, opened 2025-07-25
Fix pypy installer on Flatcar Azure 

natitomattis:
## Change description

On ansible python installation, it is using by default python 3.6 what is deprecated, I've updated it to use python 3.9. 

Also I've hit issues when building Azure images which doesn't have python 3.9. It seems like the URL used to install pypy is not maintained([last version released was 3.6](https://github.com/squeaky-pl/portable-pypy/releases/tag/pypy3.6-7.2.0)). Thats why I've replaced it with the official downloads URL.


natitomattis:
@AverageMarcus can I have some help merging this PR ? Not sure what are the next steps to remove the label `:do-not-merge/contains-merge-commits`

AverageMarcus:
You need to [squash your commits](https://www.k8s.dev/docs/guide/github-workflow/#squash-commits) so that this PR only contains a single commit.

PR #1827, opened 2025-08-06
Make kubelet log redirection optional for windows

sriramandev:
## Change description
Changes made as part of [PR](https://github.com/kubernetes-sigs/image-builder/pull/1752), will break any older releases or newer ones which do not have `kube-log-runner` made available. So to avoid this situation, making the changeset optional using `enable_auto_kubelet_service_restart` parameter whose default is set to `true` which means it should follow the flow introduced in [PR](https://github.com/kubernetes-sigs/image-builder/pull/1752) and setting the value to `false` (which is the case in case of OVA), will take the older flow that was available before [PR](https://github.com/kubernetes-sigs/image-builder/pull/1752).

- If yes, has the Provider/OS matrix been updated in the readme? NA
- If adding a new provider, are you a representative of that provider? NA

## Related issues
- Fixes # https://github.com/kubernetes-sigs/image-builder/issues/1829


AverageMarcus:
@sriramandev Could you please run `make json-sort` to fix the JSON ordering?

sriramandev:
> @sriramandev Could you please run `make json-sort` to fix the JSON ordering?

Fixed it. Thanks for pointing out

mboersma (approved):
/lgtm
/approve

zylxjtu:
I do not quite understand which scenario will this https://github.com/kubernetes-sigs/image-builder/pull/1752 break? would you please elaborate @sriramandev? why we would not have "kube-log-runner" available, it has been there for long time?

PR #1828, opened 2025-08-06
Disable hardcoded MaaS curtin scripts

kreeuwijk:
## Change description

The MaaS support added to image-builder contained hardcoded hooks and scripts for curtin. However, MaaS has since evolved and now included all those steps (and more) already builtin. The builtin steps are more powerful & flexible, and should be used instead. For example, the builtin properly reconfigures the UEFI boot order (to keep PXE at the top) and handles situations where the MaaS proxy should be used for APT package updates.

Hence, this PR comments out the copying of the hardcoded curtin files for MaaS, so that the standard curtin hooks from MaaS can be used instead.

- Is this change including a new Provider or a new OS? (y/n) __n__

## Related issues

- Fixes issues with the UEFI boot order not getting repaired, and MaaS proxy scenarios not working for custom MaaS images generated by this builder.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/maas-ubuntu.yml
+++ b/images/capi/ansible/roles/providers/tasks/maas-ubuntu.yml
@@ -4,12 +4,13 @@
     state: directory
     mode: "0775"
 
-- name: Copy curtin scripts to /curtin
-  ansible.builtin.copy:
-    src: "files/maas/curtin/{{ item }}"
-    dest: "/curtin/{{ item }}"
-    mode: "0750"
-  loop:
-    - curtin-hooks
-    - install-custom-packages
-    - setup-bootloader
+# Don't copy in hard-coded scripts as modern MaaS already has builtin tasks that are more flexible and powerful
+# - name: Copy curtin scripts to /curtin
+#   ansible.builtin.copy:
+#     src: "files/maas/curtin/{{ item }}"
+#     dest: "/curtin/{{ item }}"
+#     mode: "0750"
+#   loop:
+#     - curtin-hooks
+#     - install-custom-packages
+#     - setup-bootloader


images/capi/ansible/roles/providers/tasks/maas-ubuntu.yml
```diff
@@ -4,12 +4,13 @@
...
-- name: Copy curtin scripts to /curtin
-  ansible.builtin.copy:
-    src: "files/maas/curtin/{{ item }}"
-    dest: "/curtin/{{ item }}"
-    mode: "0750"
-  loop:
-    - curtin-hooks
-    - install-custom-packages
-    - setup-bootloader
+# Don't copy in hard-coded scripts as modern MaaS already has builtin tasks that are more flexible and powerful
```
mboersma, line 7: If this really isn't needed now, I'd prefer just to delete the lines rather than comment it out.

kreeuwijk: By commenting it out, I figured if anyone *wanted* to customize the MAAS curtin hooks, they would at least have to framework available to do it without needing to consult other sources to recreate it from scratch.

If you really wanted it gone, we should also remove the files in `images/capi/ansible/roles/providers/files/maas/curtin`, as this is the only task that references it.

mboersma: That makes sense. Maybe we should change the first comment?

```suggestion
# Uncomment this ansible task and customize the curtin scripts to replace the builtin MaaS curtin hooks
```

moondev:
@kreeuwijk Do you know the specific MaaS version that started including these curtin hooks by default?

> However, MaaS has since evolved and now includes all those steps (and more) already builtin.

For all MaaS versions previous to this new change, it would seem leaving the hooks out would build a non bootable image. Perhaps a `maas_version` variable could conditionally add the hooks, or at the least call it out in the README.

The image-builder release that first included support for MaaS was July 4 https://github.com/kubernetes-sigs/image-builder/releases/tag/v0.1.45 

Have the new hooks only landed in MaaS 3.7.0~beta1?

kreeuwijk:
@moondev MaaS doesn't ship its own curtin hooks, it just installs the [python3-curtin](https://github.com/canonical/maas/blob/master/required-packages/base#L28) package. They are then present on the server as:
```
/snap/maas/current/usr/lib/python3/dist-packages/curtin/commands/curthooks.py
```

So that means there's no dependency on the MaaS version; any version will install the latest `python3-curtin` package, which has all the hooks you could wish for.

moondev:
> @moondev MaaS doesn't ship its own curtin hooks, it just installs the [python3-curtin](https://github.com/canonical/maas/blob/master/required-packages/base#L28) package. They are then present on the server as:
> ```
> /snap/maas/current/usr/lib/python3/dist-packages/curtin/commands/curthooks.py
> ```
> 
> So that means there's no dependency on the MaaS version; any version will install the latest `python3-curtin` package, which has all the hooks you could wish for.

Thanks for the clarification and great to hear images won't be fragmented by their destination MaaS version. Cheers.

drew-viles:
So I'm happy with this in theory but I'm not a user of MAAS so cannot confirm anythign from my side. I was hoping @vasartori would be able to check this out as they are the original contributor of the MAAS code.

If they don't come back soon and we're all happy I'm happy to approve this.

vasartori:
Folks, sorry for the delay. I’ve only just seen this PR.

I’ll take a closer look and get back to you with an update. I didn’t see the GitHub notifications.

Please bear with me while I review this carefully. Once again, apologies for the delay.

vasartori:
@kreeuwijk I have tested this using the Spectro Cloud CAPI provider 0.6.1 with MAAS 3.7.0, and it works.
I’ll run some tests on MAAS 3.5.10.
I believe it will work on MAAS 3.5 as well.

To provide a bit of context on why that curtin script exists: I based part of the work on Canonical’s custom image packer. In their packer, there are curtin scripts that are explicitly copied into the image. I replicated the same behavior.
I learned something new today :)

If everything works correctly (I still want to test this on MAAS 3.5), I would remove that Ansible code entirely. Anyone who needs to provide a custom curtin script can use a “node custom role” to copy the curtin script.

vasartori:
@drew-viles Could you please approve it?

kreeuwijk:
Great, thank you @vasartori !

PR #1830, opened 2025-08-07
Make sure all references to builder user is deleted - OVA

### LINKED ISSUE #1826: ubuntu OVA leaves sudoer config for builder user
## Environment

- **Make target:** `make _______`
- **Run using container image? (Y/N):**
- **Environment vars:**
```

```
- **Vars file:**
```

```

make 

`builder ALL=(ALL) NOPASSWD:ALL` can still be seen in `/etc/sudoers.d/90-cloud-init-users`
This leaves some unintended rule which failed at security audit.

```
# Created by cloud-init v. 25.1.2-0ubuntu0~22.04.2 on Thu, 12 Jun 2025 08:28:09 +0000

# User rules for builder
builder ALL=(ALL) NOPASSWD:ALL
```

`builder ALL=(ALL) NOPASSWD:ALL` should be removed from the generated OVA.

## Relevant log output

<details>
<summary>Log Output</summary>

```
<!--
If you have any relevant build logs that could help debug this issue please include them here
but MAKE SURE ANY SENSITIVE INFO IS REMOVED!
-->
```
</details>

---

/kind bug

sriramandev:
## Change description
There is residue entry in /etc/sudoers.d/90-cloud-init-users for builder builer user. This can be flagged by security audits since this is not required and by default give passwordless sudo access to a user if a new user with username builder is created.

- If yes, has the Provider/OS matrix been updated in the readme? NA
- If adding a new provider, are you a representative of that provider? NA

## Related issues
- Fixes #1826

## Local validation
Validated the changes by locally booting the image and making sure the remnants are not present

```
root@capi:/home/capv# ls /etc/sudoers.d/
90-cloud-init-users  README
root@capi:/home/capv# cat /etc/sudoers.d/90-cloud-init-users
# Created by cloud-init v. 25.1.4-0ubuntu0~24.04.1 on Mon, 11 Aug 2025 05:04:31 +0000

# User rules for capv
capv ALL=(ALL) NOPASSWD:ALL
```


AverageMarcus:
Is this also something that should be added to all providers? If so can we make this common for all providers please?

sriramandev:
> Is this also something that should be added to all providers? If so can we make this common for all providers please?

This change is something that needs to be done in packer configuration which is per provider specific. The change is not something which can be done on a common level. I do see `maas`, `nutanix`  ,`qemu` and `raw`  may encounter similar issue.

AverageMarcus:
Anything stopping us from adding it to each provider?

ffais:
I can confirm that the same thing happens for RAW and QEMU. I also noticed that for RAW and QEMU the user isn't deleted, but only locked. In my opinion, the best solution is to delete the user in RAW and QEMU, just like it happens with OVA.

sriramandev:
For raw, qemu, nutanix and maas, I see that the user is not deleted but locked like @ffais mentioned. Hence we cannot delete the sudoers content.

ffais:
If necessary I can work on a PR to introduce the same cleanup mechanisms that are present in OVA in the RAW and QEMU targets.

sriramandev:
> If necessary I can work on a PR to introduce the same cleanup mechanisms that are present in OVA in the RAW and QEMU targets.

I can add those changes into a separate PR. But the reason I refrained from doing that was, I thought there might be specific reason why the user was kept in locked state rather than deleted(like in ova). Possibly to enable later debuging for build failures?

AverageMarcus:
> I thought there might be specific reason why the user was kept in locked state rather than deleted(like in ova).

Nothing that I'm aware of. I think it might just be an inconsistency. 

I'm in favour of having it removed everywhere. :)

sriramandev:
> > I thought there might be specific reason why the user was kept in locked state rather than deleted(like in ova).
> 
> Nothing that I'm aware of. I think it might just be an inconsistency.
> 
> I'm in favour of having it removed everywhere. :)

My suggestion is to go ahead with this one in current shape. Will raise another PR to address it for the other providers. What do you suggest @AverageMarcus

AverageMarcus:
Works for me :) 

/ok-to-test
/lgtm
/assign @drew-viles @AndiDog

drew-viles (commented):
/approve

PR #1835, opened 2025-08-11
Feat add azure arm64 support

natitomattis:
## Change description
Add support for Azure Arm64 builds:
- Added variables for architecture where necessary.
- Removed support for deprecated azure managed instances https://learn.microsoft.com/en-us/azure/virtual-machines/migration/migration-managed-image-to-compute-gallery?tabs=portal
- Upgraded azure packer plugin

## Related issues

- Fixes #


AverageMarcus:
> Removed support for deprecated azure managed instances [learn.microsoft.com/en-us/azure/virtual-machines/migration/migration-managed-image-to-compute-gallery?tabs=portal](https://learn.microsoft.com/en-us/azure/virtual-machines/migration/migration-managed-image-to-compute-gallery?tabs=portal)

Is this change required for this PR? If not I think it might be best to have it in its own PR so we can specifically call this out in the changelog.

Would you mind squashing your commits too please?

/ok-to-test

AverageMarcus:
😞  Looks like we're blocked by this: https://github.com/hashicorp/packer-plugin-azure/issues/526

natitomattis:
On my environment I've followed the steps described [here](https://developer.hashicorp.com/packer/docs/plugins/install#manually-install-plugins-using-the-cli) and I could install it. Probably I'll need to update the installation on this repo as well

natitomattis:
@AverageMarcus Sorry I just saw this comment:
> Is this change required for this PR? If not I think it might be best to have it in its own PR so we can specifically call this out in the changelog.

Yes it is required, since the [base ARM image](https://www.flatcar.org/docs/latest/installing/cloud/azure/#arm64) doesn't require the pan information. If we add the plan details it fails.

AverageMarcus:
FYI - we're discussing the issue with the plugin install change on Slack here: https://kubernetes.slack.com/archives/C01E0Q35A8J/p1754927876822259

PR #1838, opened 2025-08-13
fix: builder user remenants should not be present in node images

### LINKED ISSUE #1837: Cleanup builder user residue - raw, qemu, nutanix and maas
Create node images for `raw`, `qemu`, `maas` or `nutanix`. The node images contains residue of `builder` user that gets created as part of the initial node image generation

When the node image build is complete, `builder` user is completely cleaned.

---

/kind bug

sriramandev:
## Change description
Currently the builder user is not deleted, just locked, which could pose security risks for qemu, raw, nutanix and maas. So based on the suggestions, cleaning up the user just like in the case of ova.

## Related issues
- Fixes #1837 

## Additional context
This changeset has been introduced on the basis of the discussions that happened as part of [PR](https://github.com/kubernetes-sigs/image-builder/pull/1830).

### DIFF
--- a/images/capi/packer/maas/packer.json.tmpl
+++ b/images/capi/packer/maas/packer.json.tmpl
@@ -29,7 +29,7 @@
       "net_device": "virtio-net",
       "output_directory": "{{user `output_directory`}}",
       "qemu_binary": "{{user `qemu_binary`}}",
-      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'usermod -L {{user `ssh_username`}} && {{user `shutdown_command`}}'",
+      "shutdown_command": "echo '{{user `ssh_password`}}' | sudo -S -E sh -c 'userdel -f -r {{user `ssh_username`}} && rm -f /etc/sudoers.d/{{user `ssh_username` }} && rm -f /etc/sudoers.d/90-cloud-init-users && {{user `shutdown_command`}}'",
       "ssh_password": "{{user `ssh_password`}}",
       "ssh_timeout": "2h",
       "ssh_username": "{{user `ssh_username`}}",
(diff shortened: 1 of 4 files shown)


sriramandev:
@AverageMarcus @ffais Based on the discussions in [PR](https://github.com/kubernetes-sigs/image-builder/pull/1830) creating this PR. Kindly have a look and provide your comments on the same.

ffais:
@sriramandev I tested your changes on the raw target, the build completed successfully, the user and sudoers files were deleted, but I received this error:

```
==> qemu: userdel: user builder is currently used by process 774
==> qemu: userdel: builder mail spool (/var/mail/builder) not found
```

Does the same thing happen to you with OVA?

sriramandev:
> @sriramandev I tested your changes on the raw target, the build completed successfully, the user and sudoers files were deleted, but I received this error:
> 
> ```
> ==> qemu: userdel: user builder is currently used by process 774
> ==> qemu: userdel: builder mail spool (/var/mail/builder) not found
> ```
> 
> Does the same thing happen to you with OVA?

Interesting, no, do not see this in OVA.

ffais:
I ran further tests, and the user is being deleted correctly. At this point, I think it's just a warning we can ignore.

drew-viles:
Could you test it with `pkill -u builder` before trying the remove? This may solve the error. I suspect because the pipeline is still technically running and tasks are being executed by the user, this is where it's coming from. I'm slightly reluctant to add it with the error is all because people may wonder where it's coming from. 

The pkill, may just kill the pipeline ofc making it a non-viable option. Just a thought. :-) 

If you don't have time I can look at it later but I'm slammed atm.

ffais:
@drew-viles I ran some tests using pkill, but unfortunately the build times out. My suspicion is that pkill terminates the ssh session, preventing the pipeline from completing.

However, I can't figure out why this error doesn't appear on OVA.

drew-viles:
Thanks for confirming. 

If you're sure it's removing the user I'm happy to proceed personally unless anyone else can interject with another idea ofc.

/ok-to-test

ffais:
Yes, I can confirm that builder user is not present anymore, on all tests with RAW target, despite the warning message.

PR #1859, opened 2025-10-03
fix: Ensure correct architecture containerd/runc is downloaded

thefirstofthe300:
## Change description
The current arm64 implementation for Flatcar is half-baked, mainly due to the containerd and runc binaries being downloaded being the wrong architecture.

This PR makes an attempt to better automatically detect the OS and architecture of the machine during the installation of those two binaries.


AndiDog:
Very nice, thank you!

/lgtm

PR #1862, opened 2025-10-10
Add Ubuntu 24.04 LTS for Vultr

mondragonfx:
## Change description

- Is this change including a new Provider or a new OS? (y/n) _y__
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) _y__


AndiDog:
(Wrong role, should be:)

/lgtm

PR #1866, opened 2025-10-14
Remove deprecated pod-infra-container-image flag

### LINKED ISSUE #1864: Remove deprecated pod-infra-container-image flag from kubelet commandline
This is related with the change in kubelet: https://github.com/kubernetes/kubernetes/pull/133779

mboersma:
## Change description

Removes the deprecated `--pod-infra-container-image` flag to kubelet. See kubernetes/kubernetes#133779

## Related issues

- Fixes #1864


zylxjtu:
does this need to be kublet version specific?

mboersma:
/hold

> does this need to be kublet version specific?

Yes, I'll update the PR.

mboersma:
Actually, based on my reading of kubernetes/kubernetes#106893, `--pod-infra-container-images` has been deprecated in kubelet for many releases and has been a no-op since k8s 1.30...I think. But I'm relatively confident that it can be ignored for all supported versions of Kubernetes now, and it would be far easier not to make this change conditional on the k8s version.

@zylxjtu what do you think?

zylxjtu:
> Actually, based on my reading of [kubernetes/kubernetes#106893](https://github.com/kubernetes/kubernetes/issues/106893), `--pod-infra-container-images` has been deprecated in kubelet for many releases and has been a no-op since k8s 1.30...I think. But I'm relatively confident that it can be ignored for all supported versions of Kubernetes now, and it would be far easier not to make this change conditional on the k8s version.
> 
> @zylxjtu what do you think?

if it will not break kubelet 1.35, then I will totally agree. Based on my reading of https://github.com/kubernetes/kubernetes/pull/133779, this does not seem to be the case, let me double check with @carlory and  get back

images/capi/ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -43,6 +43,6 @@ kubernetes_cni_http_checksum: sha256:{{ kubernetes_cni_http_source }}/{{ kuberne
 
 kubeadm_template: etc/kubeadm.yml
 
-kubelet_extra_args: --pod-infra-container-image={{ pause_image }}
```
zylxjtu: it looks to me that we can revert the change here, as the new kubeadm will be able to remove it https://github.com/kubernetes/kubernetes/pull/133778.

images/capi/ansible/windows/roles/kubernetes/tasks/sc.yml
```diff
@@ -26,7 +26,6 @@
       --config={{ systemdrive.stdout | trim }}/var/lib/kubelet/config.yaml
       --bootstrap-kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/bootstrap-kubelet.conf
       --kubeconfig={{ systemdrive.stdout | trim }}/etc/kubernetes/kubelet.conf
-      --pod-infra-container-image="{{ pause_image }}"
```
zylxjtu: I think maybe it is safe for us to remove this parameter in this file and the following two powershell files, as this parameter has already been in the kubeadm env files (for old version) and it will be able to be loaded anyway

neolit123:
> Actually, based on my reading of [kubernetes/kubernetes#106893](https://github.com/kubernetes/kubernetes/issues/106893), `--pod-infra-container-images` has been deprecated in kubelet for many releases and has been a no-op since k8s 1.30...I think. But I'm relatively confident that it can be ignored for all supported versions of Kubernetes now, and it would be far easier not to make this change conditional on the k8s version.
> 
> @zylxjtu what do you think?

it's NOOP since 1.32:
https://github.com/kubernetes/kubernetes/pull/133774#issuecomment-3236017436

and another comment mentioned that the GC logic relevant to the flag was removed in 1.29:
https://github.com/kubernetes/kubernetes/pull/133774#issuecomment-3236055945

carlory:
/test pull-ova-all
LGTM

mboersma (commented):
/hold cancel
/approve

images/capi/ansible/roles/kubernetes/defaults/main.yml
zylxjtu: In the CAPZ linux node, it is failing  "Process: 47900 ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS (code=exited, status=1/FAILURE)", looks that the kubelet_extra_args was not controlled from kubeadm so it still showed as the kubelet command parameter and cause the new kubelet(force error on this) failing @mboersma

mboersma: @zylxjtu should we open a new PR to remove this arg value then?

neolit123: kubeadm will stop setting the flag on `kubeadm init/join`. on `kubeadm upgrade` it will try to clean the flag from the `/var/lib/kubelet/kubeadm-flags.env` file. but users can continue passing the flag from "extraArgs" if they wish.

PR #1868, opened 2025-10-16
feat: flatcar disable usb

simonostendorf:
## Change description

This PR disables usb devices for Flatcar as described in their [hardening guide](https://www.flatcar.org/docs/latest/setup/security/hardening-guide/#disable-usb).

Default value is false to keep the current behavior. But a user can disable usb devices in flatcar by setting `disable_flatcar_usb` to true.

- If yes, has the Provider/OS matrix been updated in the readme? (y/n) -
- If adding a new provider, are you a representative of that provider? (y/n) -

## Related issues

none

## Additional context

none

### DIFF
--- a/docs/book/src/capi/capi.md
+++ b/docs/book/src/capi/capi.md
@@ -231,3 +231,15 @@ Put the Ansible role files in the `ansible/roles` directory.
 ```
 
 Note, for backwards compatibility reasons, the variable `custom_role_names` is still accepted as an alternative to `node_custom_roles_post`, and they are functionally equivalent.
+
+##### Reenabling Flatcar USB devices
+
+Flatcar usb devices are disabled by default for security reasons.
+See [flatcar documentation](https://www.flatcar.org/docs/latest/setup/security/hardening-guide/#disable-usb) for more information.
+To reenable them, set the following variable:
+
+```json
+{
+  "ansible_user_vars": "disable_flatcar_usb=false"
+}
+```
--- a/images/capi/ansible/roles/node/defaults/main.yml
+++ b/images/capi/ansible/roles/node/defaults/main.yml
@@ -131,3 +131,7 @@ external_binary_path: "{{ '/opt/bin' if ansible_os_family == 'Flatcar' else '/us
 
 # Enable containerd trace audit in auditd, default: false.
 enable_containerd_audit: false
+
+# Disable flatcar usb devices, default: true
+# See hardening guide: https://www.flatcar.org/docs/latest/setup/security/hardening-guide/#disable-usb
+disable_flatcar_usb: true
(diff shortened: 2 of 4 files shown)


AndiDog:
@mboersma What do you think of disabling USB by default (breaking change for very few users or none, I assume)?

/ok-to-test
/lgtm

mboersma:
> What do you think of disabling USB by default

It's safer to leave it enabled by default of course. But disabling it by default is *probably* safe to do: it seems unlikely that USB resources are required in a Kubernetes environment.

@simonostendorf the linter has these complaints:
> WARNING  Listing 2 violation(s) that are fatal
command-instead-of-module: sed used in place of template, replace or lineinfile module
ansible/roles/node/tasks/flatcar.yml:26 Task/Handler: Set default HOME_MODE in login.defs
no-changed-when: Commands should not change things if nothing needs doing.
ansible/roles/node/tasks/flatcar.yml:26 Task/Handler: Set default HOME_MODE in login.defs

AverageMarcus:
If we're disabling USB by default can we please make sure there are some docs on how to re-enable it. Thinking about the on-prem folks here that might be using Kuberentes in things like factories.

simonostendorf:
> If we're disabling USB by default can we please make sure there are some docs on how to re-enable it. Thinking about the on-prem folks here that might be using Kuberentes in things like factories.

Good point, I can do that.

Can you help me with the ci? What do I have to change to get the pull-ova-all job green? :D

AverageMarcus:
/retest

It might be a flake. Let's see if it works this time. 🙂

mboersma:
> 2025/10/22 13:32:52 packer-plugin-ansible_v1.1.3_x5.0_linux_amd64 plugin: 2025/10/22 13:32:52 [INFO] 0 bytes written for 'stdin'
[0;32m    vsphere-iso.vsphere:[0m
[0;32m    vsphere-iso.vsphere: TASK [node : Blacklist usb-storage module] *************************************[0m
[0;32m    vsphere-iso.vsphere: fatal: [default]: FAILED! => {"changed": false, "checksum": "956ad786624ff9c9d06cfc3e78996595024049dc", "msg": "**Destination directory /etc/modprobe.d does not exist**"}[0m
[0;32m    vsphere-iso.vsphere:[0m

simonostendorf:
I suggest to add the following code to the sed command because it was implemented like this before and flatcar needs the sed command instead of the lineinfile because of the read-only filesystem in the /etc folder:

```yaml
  args:
    warn: false
  tags:
    - skip_ansible_lint
```

What do you think?

mboersma:
@simonostendorf I think that's fine to skip the linting in this case.

simonostendorf:
next try, hopefully now :/

simonostendorf:
linting and pull job works now, I will add docs and I guess then its ready to merge :D

PR #1877, opened 2025-11-05
Expose network tags for GCE builder

anmazzotti:
## Change description

This PR exposes network tags for the GCE builder. See [reference](https://developer.hashicorp.com/packer/integrations/hashicorp/googlecompute/latest/components/builder/googlecompute) for `tags` usage.

This can be consumed from user set variables, with a comma separated list, see [reference](https://developer.hashicorp.com/packer/docs/templates/legacy_json_templates/user-variables#use-array-values).

For example:

```json
{
  "tags": "packer,foo,bar"
}
```

This is a small quality of life change to facilitate network management.

## Related issues

- Fixes #


mboersma:
This is blocked on #1888.

PR #1880, opened 2025-11-06
Fix: runc_version overwrite and bump containerd and runc

arteonprifti:
## Change description

This pr renames the containerd_runc_version back to runc_version so that it can get overwritten if you specify it in packer vars. Currently, this is broken

Also bump containerd to v1.7.29 and runc to v1.2.8 to address https://www.openwall.com/lists/oss-security/2025/11/05/3


drew-viles:
@arteonprifti could you fix the lint errors please? Then we can run a retest and see where it lands.

drew-viles:
This is failing due to #1888 - could override potentially since it doesn't touch Google directly. **But** with it being runc which affects all "things", I'd be inclined to wait for fix to the issue first.

mboersma (approved):
/lgtm

PR #1889, opened 2025-11-19
fix(flatcar): use newer butane instead of ct

mgrzybek:
## Change description

The current Flatcar Linux building procedure uses `ct` to create ignition files. However this tool does not support newer  versions of ignition files.

This PR replaces `ct` with `butane`. As a consequence, current "variants" (1.1.0) and "versions" (flatcar) are supported.  

## Related issues

N.A.

## Additional context

N.A.

### DIFF
--- a/images/capi/hack/ensure-butane.sh
+++ b/images/capi/hack/ensure-butane.sh
@@ -20,7 +20,7 @@ set -o pipefail
 
 [[ -n ${DEBUG:-} ]] && set -o xtrace
 
-_version="v0.25.0"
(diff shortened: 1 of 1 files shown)


AndiDog:
rockylinux-9 failed, let's try again.

/retest

mboersma:
/retest

GitHub 503 error downloading a file.

AndiDog:
Intermittent error again, I presume

> Allocation failed. We do not have sufficient capacity for the requested VM size in this region

/test pull-azure-sigs

PR #1893, opened 2025-12-01
remove AWS_DEFAULT_REGION from example env file

stelucz:
## Change description

Documentation mentions AWS env file with key `AWS_DEFAULT_REGION`. However, it is not used anywhere in image-builder and could be confusing that it will be used for underlaying packer. I myself had it set and I was wondering why packer uses different region.

- Is this change including a new Provider or a new OS? (y/n) N
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) N
- If adding a new provider, are you a representative of that provider? (y/n) N

## Related issues

- Fixes #

## Additional context
```
 grep -ir "AWS_DEFAULT_REGION"
docs/book/src/capi/container-image.md:      AWS_DEFAULT_REGION=xxxxxx
```


drew-viles:
Thanks, good spot!

If we're not using it I'm happy for it to be removed from the docs, a quick check from my side confirms it's not being used anywhere too.

/ok-to-test

PR #1894, opened 2025-12-04
Upgrade ansible to v2.20.0 to support Python 3.14

AndiDog:
## Change description

Otherwise, with newer Python versions, we get the error `Invalid conditional detected: module 'ast' has no attribute 'Str'` in playbook runs.

Python 3.14, released in October 2025, [removes](https://docs.python.org/3.13/whatsnew/3.13.html#pending-removal-in-python-3-14) `ast.Str`. Upgrading Ansible to a newer version ensures this isn't used anymore – I didn't look for the exact changelog item for that since we should anyway upgrade Ansible.

## Related issues

n/a

### DIFF
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Note: ansible-core v2.16 supports Python 3.10-3.12.
-_version_ansible_core="2.16.16"
+_version_ansible_core="2.20.0"
 
 case "${OSTYPE}" in
 linux*)


mboersma:
Our build pipeline for Azure has nodes that still use Python 3.9, so we've kept this pinned here to avoid breaking that. And in the past when we've bumped this, users have complained it broke their builds. So jumping to 2.20 is likely to be problematic--is the `ast.Str` fix in an earlier ansible version?

We should also update the minimum Python version in `ensure-python.sh` to whatever ansible requires:
https://github.com/kubernetes-sigs/image-builder/blob/d8f7a3cb01b8514c4823be9dc0fd9bfd3aa5b5fe/images/capi/hack/ensure-python.sh#L23-L24

I agree it's time to move forward with ansible and Python. I'll investigate how we can support a newer python for Azure builds.

AndiDog:
The test error says that Python 3.12 is required at least, and that version is exactly two years old. One option to support older Python versions, probably with a very loud warning message, is to fall back to a slightly older Ansible version. Just an idea, even if I don't like it that much. @mboersma can you refer me to the Azure-related code that's problematic?

mboersma:
> can you refer me to the Azure-related code that's problematic?

The scripts for the pipelines are in images/capi/packer/azure/.pipelines, with `stages.yaml` as the top-level one. But there isn't any problematic code really, it's just that the actual Azure DevOps pipelines that run these for CAPZ reference images need to update Python on their build nodes.

It's something I can work around for now just by maintaining a fork. We should merge this so image-builder uses a version of Python that is actually supported. 😄 I was just hoping we could move the requirements forward more slowly, but I see 2.20.0 is the first version to support Python 3.14.

Let's do an image-builder release first, so we have an updated release tag that works with Python 3.9 - 3.11.

/hold for a checkpoint release

mboersma:
Also it appears we're using Python 3.11 in the OVA build pipeline. 🫤

mboersma:
@AndiDog we've updated our Azure build agents so they're now using python 3.12. So I remove that objection to this PR. 😄 

If we can figure out how to get this to work with the OVA build pipeline, we can merge it.

AndiDog:
We only have Debian Bookworm with Python 3.11 in the [kubekins](https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/Dockerfile) image that we use in CI. Let's "just wait" until that gets updated in `test-infra`. I can keep the PR open.

PR #1908, opened 2026-01-06
fix: update Azure build VM size to a more modern v2 generation

abenn135:
## Change description

Updates the VM used by the Azure build pipeline to a more modern and widely available v2 generation. B-series (v1) have been deprecated and have restricted availability, resulting in build failure in some regions.

Note that this will increase the cost of performing a build slightly. No exact match for the previous B2ms size is widely available.

## Related issues

n/a

## Additional context

See https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/bv1-series?tabs=sizebasic and https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/retirement/retirement-overview#capacity-limited.


mboersma (approved):
/lgtm
/approve

Thanks @abenn135!

(I'm making a similar change in kubernetes-sigs/cluster-api-provider-azure#6017 but I like your choice of SKU better so I think I'll use it there as well.)

PR #1912, opened 2026-01-14
ARM64 support for MaaS

kreeuwijk:
## Change description

This PR adds the ability to generate Ubuntu 22.04/24.04 ARM64 images.

The work is inspired by [this work](https://github.com/kubernetes-sigs/image-builder/pull/1142) from @ader1990. While that PR was never merged, this PR generalizes the approach and ensures no impact to existing workflows.

- Is this change including a new Provider or a new OS? (y/n) __N__

## Related issues

- Provides a generic solution for #1142, which could be adopted into the qemu builds (this PR only implements it for maas builds)

## Additional context

images/capi/packer/maas/README.md has been updated to detail the build steps fully for both AMD64 and ARM64.

## Acknowledgements
Kudos to @blik616287 for helping out with the Ansible filtering


images/capi/ansible/roles/setup/templates/etc/apt/sources.list.d/ubuntu.sources.j2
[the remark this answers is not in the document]
kreeuwijk: Actually that is already being set here: https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/packer/config/common.json#L23-L24, so I think this doesn't need a change?

kreeuwijk:
`pull-azure-sigs` has no relevance to the PR

vasartori:
I don't have ARM64 hardware available to test this PR. However, based on my review, this PR will not break anything for amd64 builds.
I am starting tests on amd64 right now.

ader1990:
Thanks for spearheading this work!

vasartori:
Tested! Everything seems to work well on amd64.
/lgtm

kreeuwijk:
@vasartori for completeness: I used an AWS Graviton instance (`m7g.metal`) to test this in ARM64. The changes allowed me to generate CAPI images for Ubuntu 22.04 and 24.04. I also tested deploying a CAPI cluster through MaaS (using Spectro Cloud Palette) with the ARM64 Ubuntu 22.04 + Kubernetes 1.33.5 image and that worked.

I did find a bug in MaaS itself though, which has been reported [here](https://bugs.launchpad.net/maas/+bug/2138312)

images/capi/packer/goss/goss-vars.yaml
```diff
@@ -496,7 +496,7 @@ ubuntu:
         enabled: true
         running: true
     package:
-      linux-cloud-tools-virtual:
+      #linux-cloud-tools-virtual:
```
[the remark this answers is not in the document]
kreeuwijk: Yes, unless someone can think of a way to not check for this package when the CPU architecture is aarch64. I think this is an acceptable compromise though as the package isn't required for CAPI.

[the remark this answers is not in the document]
vasartori: A crazy idea I had here maybe we could try to move it forward
What if we create a templating mechanism for the GOOS YAML?

Through environment variables, we could decide what should be rendered in the GOOS configuration, and perhaps this could be extended to other parts of the image generation process.

Using a Jinja template (pre goos) could be an option what do you all think?

vasartori:
@kreeuwijk Would you mind squashing the commits please?

kreeuwijk:
@vasartori Commits squashed

vasartori:
@drew-viles could you help us to merge this PR?

drew-viles:
So in principle I've no issues with merging this. My only concern is it'll cause issues with the azure tests moving forward. This is where my knowledge of the repo and subsequent prow usage falters tbh. I'm going to ask @mboersma if he can chip in on how we can target it in a way that skips the azure checks. I'm sure there is as there are some build that only use Azure for example, and they only run those checks.

kreeuwijk:
@drew-viles @mboersma any chance of getting this merged?

images/capi/packer/maas/maas-ubuntu-2204-arm64.json
```diff
@@ -0,0 +1,21 @@
+{
+  "arch": "arm64",
+  "boot_command_prefix": "c<wait>linux /casper/vmlinuz console=ttyAMA0 autoinstall ds='nocloud-net;s=http://{{ .HTTPIP }}:{{ .HTTPPort }}/' --- <enter><wait>initrd /casper/initrd<enter><wait>boot<enter>",
+  "boot_type": "uefi",
+  "build_name": "ubuntu-2204-arm64",
+  "cpus": "16",
```
mboersma, line 6: The `maas-ubuntu-2404-arm64-json` file sets `"distribution_version": "2404"` here but this file doesn't. Could you add `"distribution_version": "2204"` for symmetry?

images/capi/packer/maas/packer-arm64.json.tmpl
```diff
@@ -0,0 +1,235 @@
+{
```
mboersma, line 1: This is a big file that's 90% identical to `packer.json.tmpl`. It would be great if we could use a single file and parameterize the keys like `qemu_binary`,  `machine_type: virt`, etc., or at least have a comment noting that they need to stay in sync with each other.

Not a blocker, just noting this.

images/capi/packer/maas/README.md
```diff
@@ -1,16 +1,60 @@
...
+# Uploading to MaaS
 
-Upload images to MaaS
+To upload the generates images to MaaS, run the following command.
 
+For AMD64:
+```
+maas <PROFILE> boot-resources create name=<IMAGE NAME> architecture=amd64/generic title=<IMAGE NAME> base_image=ubuntu/<SEE NOTES> content@=./<FILE>.tar.gz
 ```
-maas <PROFILE> boot-resources create name=custom/<IMAGE NAME> architecture=amd64/generic title=<IMAGE NAME> subarches=generic base_image=ubuntu/<SEE NOTES> content@=./<FILE>.tar.gz
```
mboersma, line 12: We dropped `subarches=generic` from the upload example here (as well as for arm64). I just want to confirm that this is intentional and doesn't break anything.

mboersma (commented):
Looking good! I just had some minor comments.

images/capi/packer/maas/packer-arm64.json.tmpl
kreeuwijk: I agree, but is such parameterization supported here?

images/capi/packer/maas/README.md
kreeuwijk: Yes, this parameter is indeed not necessary.

images/capi/packer/maas/packer-arm64.json.tmpl
mboersma: It is--we already have several parametrized values. But it might be tricky to do so for the `qemuargs` block. It wouldn't be trivial, so let's not.

You could add a note in packer/maas/README.md telling developers to keep the files in sync, or a `"_comment": "keep in sync with <other file>"` top-level key in the json.tmpl itself.

kreeuwijk: The main issue here is that the differences in the `builders:` section are significant and don't seem to have a way to be parameterized anywhere else:
* Added `cdrom_interface`, `efi_boot`, `efi_firmware_code`, `efi_firmware_vars`, `iso_target_path`, `machine_type` and `qemuargs` parameters that were previously undefined

I remember looking for ways to handle this outside of this file, but Packer did not support handling it any other way,

mboersma (commented):
Retracting—I misread the file. No typo, please disregard.

mboersma (approved):
/lgtm
/approve

PR #1920, opened 2026-01-29
outscale: update base images to v2026-01-12

jfbus:
## Change description

Update Outscale base images to v2026-01-12


jfbus:
thx !

PR #1921, opened 2026-01-29
Remove windows-2019 from default Azure e2e

mboersma:
## Change description

Removes Azure -windows-2019 build targets from the default e2e list, since they're long-deprecated. The images can still be built through the `make` targets but it doesn't seem necessary to include them in PR e2e now.

## Related issues

- Fixes #


mboersma:
@marosset @zylxjtu does this seem reasonable to you? The make targets remain but we won't be testing windows-2019 builds by default.

PR #1924, opened 2026-02-02
Make py3_install more flexible

mboersma:
## Change description

Adds fallback behavior for when `pip3 install --user` raises a specific error we've seen on (for example) Azure Linux.

## Related issues

- Fixes #


drew-viles:
Now all the tests are resolved,
/approve

PR #1926, opened 2026-02-02
Revert Azure VM update and sync test regions

mboersma:
## Change description

This reverts commit ce0d0b8406c1f2621c7dc3e7a99f139e28e047c1.

We've been seeing lots of "no SKUs available" errors for the new VM type (and I can't tell if the -cvm builds ever pass now), so let's try rolling back to see if that improves things. The SKU update deadline isn't until 2028.

This also reworks the Azure test locations to match those we've successfully used in CAPZ, and expands the confidential VM locations to those that seem to be available to the test subscription.

### DIFF
--- a/images/capi/scripts/ci-azure-e2e.sh
+++ b/images/capi/scripts/ci-azure-e2e.sh
@@ -66,11 +66,20 @@ fi
 set -o nounset
 
 get_random_region() {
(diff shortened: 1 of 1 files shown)


mboersma:
See kubernetes/test-infra#36338 related to the pull-azure-vhds failure.

mboersma:
/close

This seems to have been less of a problem recently, so let's leave things alone until proven otherwise.

mboersma:
/assign @drew-viles 

This seems to ameliorate the Azure `SKUNotFound` errors we've been seeing on ~50% of PRs. 🤞🏻

PR #1927, opened 2026-02-04
fix: move extra_kernel_boot_params default variable from setup to sysprep role

ffais:
## Change description
Some users are experiencing an undefined variable error caused by the variable "extra_kernel_boot_params " being declared as a default for the wrong ansible role. 
With this PR I moved extra_kernel_boot_params  variable from setup role to sysprep role where this variable is used.

## Additional context
This is the [link](https://github.com/kubernetes-sigs/image-builder/commit/be8da68145f8516571ecf7edbe7da16e089a7aa8#diff-b2a18945cf178a8116d2523dc23d6284dacbe4443e055c522790aded29cb38b8R22) to the commit where the user commented about the issue described above.


mboersma:
/test pull-ova-all

Seems like this job got stuck? And the Azure failure was a 502 downloading a file from GitHub.

PR #1947, opened 2026-03-03
move kubelet --system-reserved deprecated flag to kubelet-config file

ffais:
## Change description
Kubelet  '--system-reserved' has been deprecated. With this PR, the flag has been moved to a specific configuration file in the drop-ins directory. The drop-ins directory specified with '--config-dir' flag allows the user to optionally specify additional configs to overwrite what is provided by default and in the KubeletConfigFile flag.

### DIFF
--- a/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
+++ b/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
@@ -99,7 +99,7 @@ CPU_CORE_RESERVATION_MICROCORES=(
 )
 
 # Calculate the CPU reservation
-cpu_milicores_to_reserve() {
+cpu_millicores_to_reserve() {
   local cpu_microcores_reserved=0
 
   for ((i = 0; i < schedulable_cores_no; i++)); do
@@ -115,5 +115,15 @@ cpu_milicores_to_reserve() {
 }
 
 mkdir -p /var/lib/kubelet/kubelet.conf.d
-echo "$(jq --arg mebibytes_to_reserve "${mebibytes_to_reserve}Mi" --arg cpu_millicores_to_reserve "${cpu_millicores_to_reserve}m" \
-    '. += {systemReserved: {"cpu": $cpu_millicores_to_reserve, "ephemeral-storage": "1Gi", "memory": $mebibytes_to_reserve}}' $KUBELET_CONFIG)" > $KUBELET_CONFIG
+
+# Initialize config file if it doesn't exist
+if [ ! -f "$KUBELET_CONFIG" ]; then
+  echo "{}" > "$KUBELET_CONFIG"
+fi
+
+# Get the computed values from the functions
+memory_reservation_mebibytes=$(memory_reservation_mebibytes)
+cpu_millicores_to_reserve=$(cpu_millicores_to_reserve)
+
+echo "$(jq --arg memory_reservation_mebibytes "${memory_reservation_mebibytes}Mi" --arg cpu_millicores_to_reserve "${cpu_millicores_to_reserve}m" \
+    '. += {"apiVersion": "kubelet.config.k8s.io/v1beta1","kind": "KubeletConfiguration", "systemReserved": {"cpu": $cpu_millicores_to_reserve, "ephemeral-storage": "1Gi", "memory": $memory_reservation_mebibytes}}' "$KUBELET_CONFIG")" > "$KUBELET_CONFIG"
--- a/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
+++ b/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
@@ -5,6 +5,7 @@
 
 #RPM and DEB systems kubelet sysconfig PATH
 KUBELET_SYSCONFIG_FILES=( "/etc/sysconfig/kubelet" "/etc/default/kubelet" )
+KUBELET_CONFIG="/var/lib/kubelet/kubelet.conf.d/kubelet-resource-sizing.json"
 
 for KUBELET_SYSCONFIG in "${KUBELET_SYSCONFIG_FILES[@]}"
 do
@@ -13,7 +14,7 @@ do
     # shellcheck source=/dev/null
     . "${KUBELET_SYSCONFIG}"
     # If system-reserved is already set by user, ignore
-    if grep -q 'KUBELET_EXTRA_ARGS=.*--system-reserved' "${KUBELET_SYSCONFIG}"; then
+    if grep -q 'systemReserved' "${KUBELET_SYSCONFIG}"; then
       exit 0
     fi
   fi
@@ -113,13 +114,6 @@ cpu_milicores_to_reserve() {
   echo "$cpu_microcores_reserved" | awk '{result = $1 / 10; if (result != int(result)) result++; printf "%d\n", result}'
 }
 
-mkdir -p /run/kubelet
-# Check if system-reserved already exists
-if grep '.*--system-reserved' <<< "${KUBELET_EXTRA_ARGS}"; then
-  # If system-reserved is already set by a previous run, replace old value with new one and write to /run/kubelet/extra-args.env
-  system_reserved=$(sed -E "s|--system-reserved=cpu=[0-9]+m,memory=[0-9]+Mi|--system-reserved=cpu=$(cpu_milicores_to_reserve)m,memory=$(memory_reservation_mebibytes)Mi|" <<< "${KUBELET_EXTRA_ARGS}")
-  echo "KUBELET_EXTRA_ARGS=${system_reserved} >/run/kubelet/extra-args.env"
-else
-  # If not append system-reserved to KUBELET_EXTRA_ARGS and write to /run/kubelet/extra-args.env
-  echo "KUBELET_EXTRA_ARGS=${KUBELET_EXTRA_ARGS} --system-reserved=cpu=$(cpu_milicores_to_reserve)m,memory=$(memory_reservation_mebibytes)Mi" >/run/kubelet/extra-args.env
-fi
+mkdir -p /var/lib/kubelet/kubelet.conf.d
+echo "$(jq --arg mebibytes_to_reserve "${mebibytes_to_reserve}Mi" --arg cpu_millicores_to_reserve "${cpu_millicores_to_reserve}m" \
+    '. += {systemReserved: {"cpu": $cpu_millicores_to_reserve, "ephemeral-storage": "1Gi", "memory": $mebibytes_to_reserve}}' $KUBELET_CONFIG)" > $KUBELET_CONFIG
--- a/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
+++ b/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
@@ -3,20 +3,23 @@
 # This script is used to calculate the resource sizing for the kubelet based on values used by GKE and repeated
 # in https://github.com/awslabs/amazon-eks-ami/pull/367/files
 
-#RPM and DEB systems kubelet sysconfig PATH
-KUBELET_SYSCONFIG_FILES=( "/etc/sysconfig/kubelet" "/etc/default/kubelet" )
-KUBELET_CONFIG="/var/lib/kubelet/kubelet.conf.d/kubelet-resource-sizing.json"
 
-for KUBELET_SYSCONFIG in "${KUBELET_SYSCONFIG_FILES[@]}"
-do
-  # Check if the file exists
-  if [ -f "${KUBELET_SYSCONFIG}" ]; then
-    # shellcheck source=/dev/null
-    . "${KUBELET_SYSCONFIG}"
-    # If system-reserved is already set by user, ignore
-    if grep -q 'systemReserved' "${KUBELET_SYSCONFIG}"; then
-      exit 0
-    fi
+# If the user has already configured systemReserved (in the main kubelet
+# config or any other drop-in), don't overwrite their value.
+KUBELET_CONFIG="/var/lib/kubelet/kubelet.conf.d/kubelet-resource-sizing.json"
+USER_KUBELET_CONFIGS=( "/var/lib/kubelet/config.yaml" )
+if [ -d /var/lib/kubelet/kubelet.conf.d ]; then
+  while IFS= read -r -d '' f; do
+    [ "$f" = "$KUBELET_CONFIG" ] && continue
+    USER_KUBELET_CONFIGS+=( "$f" )
+  done < <(find /var/lib/kubelet/kubelet.conf.d -maxdepth 1 -type f -print0)
+fi
+
+for cfg in "${USER_KUBELET_CONFIGS[@]}"; do
+  [ -f "$cfg" ] || continue
+  if grep -Eq '^[[:space:]]*systemReserved[[:space:]]*:' "$cfg" \
+    || grep -q '"systemReserved"' "$cfg"; then
+    exit 0
   fi
 done
 
--- a/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
+++ b/images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
@@ -129,7 +129,6 @@ memory_reservation_mebibytes=$(memory_reservation_mebibytes)
 cpu_millicores_to_reserve=$(cpu_millicores_to_reserve)
 
 tmp=$(mktemp) && \
-echo $tmp && \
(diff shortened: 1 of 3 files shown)


Whisper40:
@ffais Hi, any news about this ?

ffais:
Hi @Whisper40, looks like pull-ova-all test failing is not related to the changes made in this PR.

I'm waiting for a review from one of the maintainers. I know @drew-viles  is very busy these days, @mboersma could take a look?

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
```diff
@@ -113,13 +114,6 @@ cpu_milicores_to_reserve() {
...
-if grep '.*--system-reserved' <<< "${KUBELET_EXTRA_ARGS}"; then
-  # If system-reserved is already set by a previous run, replace old value with new one and write to /run/kubelet/extra-args.env
-  system_reserved=$(sed -E "s|--system-reserved=cpu=[0-9]+m,memory=[0-9]+Mi|--system-reserved=cpu=$(cpu_milicores_to_reserve)m,memory=$(memory_reservation_mebibytes)Mi|" <<< "${KUBELET_EXTRA_ARGS}")
-  echo "KUBELET_EXTRA_ARGS=${system_reserved} >/run/kubelet/extra-args.env"
-else
-  # If not append system-reserved to KUBELET_EXTRA_ARGS and write to /run/kubelet/extra-args.env
-  echo "KUBELET_EXTRA_ARGS=${KUBELET_EXTRA_ARGS} --system-reserved=cpu=$(cpu_milicores_to_reserve)m,memory=$(memory_reservation_mebibytes)Mi" >/run/kubelet/extra-args.env
-fi
+mkdir -p /var/lib/kubelet/kubelet.conf.d
+echo "$(jq --arg mebibytes_to_reserve "${mebibytes_to_reserve}Mi" --arg cpu_millicores_to_reserve "${cpu_millicores_to_reserve}m" \
```
mboersma: I don't see where the env var `$cpu_millicores_to_reserve` or `$mebibytes_to_reserve` is ever set. Previously, these were function calls of `cpu_milicores_to_reserve` (one letter `l`) and `memory_reservation_mebibytes`.

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
```diff
@@ -113,13 +114,6 @@ cpu_milicores_to_reserve() {
...
+    '. += {systemReserved: {"cpu": $cpu_millicores_to_reserve, "ephemeral-storage": "1Gi", "memory": $mebibytes_to_reserve}}' $KUBELET_CONFIG)" > $KUBELET_CONFIG
```
mboersma: `echo "$(jq ... $FILE)" > $FILE` truncates input before reading.
I think you need to write this `jq` output to a file, then `mv` the file to `$KUBELET_CONFIG`. Otherwise the shell will truncate the file pointed to by `$KUBELET_CONFIG` when opening the shell redirect and `jq` will read in a 0-byte file.

mboersma: `kubelet --config-dir` requires each file to be a complete `KubeletConfiguration` or it will reject it with "missing TypeMeta". So I think `jq` needs also to emit `apiVersion: kubelet.config.k8s.io/v1beta1` and `kind: KubeletConfiguration`.

images/capi/ansible/roles/kubernetes/templates/usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -1,7 +1,11 @@
 # Note: This dropin only works with kubeadm and kubelet v1.11+
 [Service]
 Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf"
-Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml"
+Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml {% if kubernetes_enable_automatic_resource_sizing | bool %}--config-dir /var/lib/kubelet/kubelet.conf.d{% endif %}"
```
mboersma: `--config-dir` itself was added in Kubernetes 1.28, so this will fail on older versions of Kubernetes. Maybe we're not concerned about out-of-support versions, but it's safer to wrap this with a version check as we do a couple lines below.

mboersma (changes requested):

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
ffais: Hi @mboersma thank you for the review, yes you're right, I fixed it with the latest commit.

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
ffais: I fixed it with the latest commit.

images/capi/ansible/roles/kubernetes/templates/usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf
ffais: I added the version check both in this file and in the main.yml file, now --config-dir is added only if the kubernetes version is higher than 1.28.

drew-viles:
/override pull-ova-all

Known photon-5 build failure.

drew-viles (commented):
with the changed been made.

/lgtm

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
```diff
@@ -13,7 +14,7 @@ do
     # shellcheck source=/dev/null
     . "${KUBELET_SYSCONFIG}"
     # If system-reserved is already set by user, ignore
-    if grep -q 'KUBELET_EXTRA_ARGS=.*--system-reserved' "${KUBELET_SYSCONFIG}"; then
+    if grep -q 'systemReserved' "${KUBELET_SYSCONFIG}"; then
```
mboersma: This check no longer does what the comment above it claims. `${KUBELET_SYSCONFIG}` is `/etc/sysconfig/kubelet` or `/etc/default/kubelet` — a shell-style env file consumed by the kubelet unit, not a `KubeletConfiguration`. A user who wants to pre-set `systemReserved` would do it in `/var/lib/kubelet/config.yaml` or in a drop-in under `/var/lib/kubelet/kubelet.conf.d/`, so `systemReserved` will essentially never appear in sysconfig and this guard will never trip.

Since this PR moves `system-reserved` out of `KUBELET_EXTRA_ARGS` and into a drop-in, I think the intent should be: "if the user has already provided a `systemReserved` value via the main kubelet config or any other drop-in, don't overwrite it." Something like:

```sh
# If the user has already configured systemReserved (in the main kubelet
# config or any other drop-in), don't overwrite their value.
USER_KUBELET_CONFIGS=( "/var/lib/kubelet/config.yaml" )
if [ -d /var/lib/kubelet/kubelet.conf.d ]; then
  while IFS= read -r -d '' f; do
    [ "$f" = "$KUBELET_CONFIG" ] && continue
    USER_KUBELET_CONFIGS+=( "$f" )
  done < <(find /var/lib/kubelet/kubelet.conf.d -maxdepth 1 -type f -print0)
fi

for cfg in "${USER_KUBELET_CONFIGS[@]}"; do
  [ -f "$cfg" ] || continue
  if grep -Eq '^[[:space:]]*systemReserved[[:space:]]*:' "$cfg" \
     || grep -q '"systemReserved"' "$cfg"; then
    exit 0
  fi
done
```

The outer loop over `KUBELET_SYSCONFIG_FILES` can probably be dropped entirely now — sourcing it was only useful for the old `KUBELET_EXTRA_ARGS` path. It's worth a sanity check that nothing downstream still relies on that.

mboersma (commented):
One follow-up on the `systemReserved`-already-set guard:

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
ffais: Hi @mboersma, I've incorporated the changes you suggested and run some tests. Everything seems to be working fine.

mboersma (changes requested):
Thanks for sticking with this! One blocker I missed first time round:

Kubelet's `--config-dir` only reads files whose extension is exactly `.conf` — from [`cmd/kubelet/app/server.go`](https://github.com/kubernetes/kubernetes/blob/release-1.31/cmd/kubelet/app/server.go):

```go
const dropinFileExtension = ".conf"
...
if !info.IsDir() && filepath.Ext(info.Name()) == dropinFileExtension {
```

Anything else gets silently skipped. The content can stay JSON (the decoder accepts either), but `KUBELET_CONFIG` needs to be renamed from `kubelet-resource-sizing.json` → `kubelet-resource-sizing.conf`, otherwise the drop-in is a no-op at runtime.

Two nits while you're in there:

- The `echo "$(jq ... "$FILE")" > "$FILE"` line is still there. It happens to work in bash, but if `jq` ever fails the file gets truncated to empty. `tmp=$(mktemp) && jq ... "$KUBELET_CONFIG" > "$tmp" && mv "$tmp" "$KUBELET_CONFIG"` is safer.
- `--config-dir` is now plumbed via both the Jinja `{% if %}` in `10-kubeadm.conf` and the `lineinfile` in `main.yml`. Same output, just two places to keep in sync — happy to land as-is, just flagging.

ffais:
Thanks for the review!

Regarding the `.conf` extension ,apologies for the oversight in the PR. In my local environment I had the correct `.conf` extension all along; the `.json` name slipped in when I cleaned up the branch for the PR. Fixed now.

The `echo "$(jq ... "$FILE")" > "$FILE"` pattern has been replaced with the safer approach, as suggested.

I also notice that there was a minor inconsistency in the template, `KUBELET_CONFIG_DROPIN_DIR_ALPHA` is added for all v1.28.x http installs regardless of `kubernetes_enable_automatic_resource_sizing`, but the `lineinfile` task at line 94 is gated by that variable. I removed the flags from the template, now the `lineinfile` tasks in `main.yml` remain the single source of truth for both flags, covering all source types.

I apologize again for the oversight and thank you for the review.

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -84,6 +84,19 @@
...
+    path: "{{ systemd_prefix }}/system/kubelet.service.d/10-kubeadm.conf"
+    regexp: '^Environment="KUBELET_CONFIG_ARGS='
+    line: Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml --config-dir /var/lib/kubelet/kubelet.conf.d"
+  when: kubernetes_semver is version('v1.28.0', '>=') and kubernetes_enable_automatic_resource_sizing | bool
+
+- name: Add KUBELET_CONFIG_DROPIN_DIR_ALPHA environment variable for Kubernetes v1.29 and below
+  ansible.builtin.lineinfile:
+    path: "{{ systemd_prefix }}/system/kubelet.service.d/10-kubeadm.conf"
+    line: Environment="KUBELET_CONFIG_DROPIN_DIR_ALPHA"
+  when: kubernetes_semver is version('v1.28.0', '>=') and kubernetes_semver is version('v1.29.0', '<') and kubernetes_enable_automatic_resource_sizing | bool
```
mboersma: This is the one that'll actually bite us. `--config-dir` is alpha in *both* 1.28 and 1.29 (it only goes beta/on-by-default in 1.30), so 1.29 needs this env var too. As written, a 1.29 node gets `--config-dir` from the task above but not the alpha gate here, and kubelet will refuse to start. Bump the upper bound to `v1.30.0`.

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -84,6 +84,19 @@
...
+  ansible.builtin.lineinfile:
+    path: "{{ systemd_prefix }}/system/kubelet.service.d/10-kubeadm.conf"
+    regexp: '^Environment="KUBELET_CONFIG_ARGS='
+    line: Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml --config-dir /var/lib/kubelet/kubelet.conf.d"
+  when: kubernetes_semver is version('v1.28.0', '>=') and kubernetes_enable_automatic_resource_sizing | bool
+
+- name: Add KUBELET_CONFIG_DROPIN_DIR_ALPHA environment variable for Kubernetes v1.29 and below
+  ansible.builtin.lineinfile:
+    path: "{{ systemd_prefix }}/system/kubelet.service.d/10-kubeadm.conf"
+    line: Environment="KUBELET_CONFIG_DROPIN_DIR_ALPHA"
```
mboersma, line 97: Minor: `Environment="KUBELET_CONFIG_DROPIN_DIR_ALPHA"` sets it to empty string. Worth confirming kubelet checks for presence and not a non-empty value — if it's the latter, make it `="true"` to be safe.

mboersma (commented):
Thanks for tackling this — moving off the deprecated `--system-reserved` is the right call, and the drop-in approach is clean. A couple things to sort out before this can go in (left inline). The big one is the v1.29 gate; the rest are quick cleanups. `pull-ova-all` is red but that's just vSphere infra flaking, not your change.

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
```diff
@@ -113,13 +117,19 @@ cpu_milicores_to_reserve() {
...
+if [ ! -f "$KUBELET_CONFIG" ]; then
+  echo "{}" > "$KUBELET_CONFIG"
 fi
+
+# Get the computed values from the functions
+memory_reservation_mebibytes=$(memory_reservation_mebibytes)
+cpu_millicores_to_reserve=$(cpu_millicores_to_reserve)
+
+tmp=$(mktemp) && \
+echo $tmp && \
```
mboersma: Leftover debug `echo` — let's drop it. It spams the journal every boot, and unquoted `$tmp` will likely trip shellcheck (SC2086) in `make lint`.

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
```diff
@@ -113,13 +117,19 @@ cpu_milicores_to_reserve() {
...
+jq --arg memory_reservation_mebibytes "${memory_reservation_mebibytes}Mi" --arg cpu_millicores_to_reserve "${cpu_millicores_to_reserve}m" \
+    '. += {"apiVersion": "kubelet.config.k8s.io/v1beta1","kind": "KubeletConfiguration", "systemReserved": {"cpu": $cpu_millicores_to_reserve, "ephemeral-storage": "1Gi", "memory": $memory_reservation_mebibytes}}' "$KUBELET_CONFIG" > "$tmp" && \
```
mboersma: Heads up: `ephemeral-storage: "1Gi"` is new — the old `--system-reserved` path only reserved cpu+memory. That changes node allocatable, so let's call it out in the PR description (or make it configurable) rather than sneaking it in.

images/capi/ansible/roles/kubernetes/tasks/main.yml
ffais: From the [documentation](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#kubelet-conf-d):
> For Kubernetes v1.28 to v1.29, you can only specify --config-dir if you also set the environment variable KUBELET_CONFIG_DROPIN_DIR_ALPHA for the kubelet process (the value of that variable does not matter).

If it's ok for you I'll leave it as it is.

images/capi/ansible/roles/kubernetes/tasks/main.yml
ffais: Yes, you're right, I didn't notice that 1.29 was included.

images/capi/ansible/roles/kubernetes/files/usr/libexec/kubernetes/kubelet-resource-sizing.sh
ffais: And this is another one.

mboersma (approved):
/lgtm
/retest

I think this is good now, the Azure CI failure isn't related. I'll see if I can fix that in a separate PR.

mboersma (approved):
/approve

mboersma:
Trying to fix the azure-sigs failures in #2030 and #2032.

PR #1953, opened 2026-03-16
Fix Windows failure caused by WinRM connection reset

mboersma:
## Change description

In the Azure GH workflow, Windows SIG image builds fail during the Packer provisioning phase with "connection reset by peer".

Probable root cause: the ansible_winrm.ps1 script calls `net stop winrm` / `net start winrm` synchronously, which severs the WinRM session that Packer is using to execute the script. With Packer 1.9.5, the built-in WinRM communicator does not recover from this mid-script connection loss.

This change replaces the synchronous net stop/net start with a scheduled task that restarts the WinRM service 5 seconds after the script completes. This allows the script to return cleanly to Packer before the service cycles. The existing pause_before: "15s" on the subsequent Ansible provisioner provides additional time for WinRM to come back up with the new configuration.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible/windows/ansible_winrm.ps1
+++ b/images/capi/ansible/windows/ansible_winrm.ps1
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This file is from packer documentation: 
+# This file is from packer documentation:
 # https://www.packer.io/docs/provisioners/ansible.html#winrm-communicator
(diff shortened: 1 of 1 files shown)


mboersma:
This does seem to get past the WinRM connection error in the pipeline, but the task
`TASK [systemprep : Update Windows Defender signatures]` is now failing, although apparently it's not considered fatal. See https://github.com/kubernetes-sigs/image-builder/actions/runs/23207750655/job/67448601797.

mboersma:
/close

See #1956

PR #1968, opened 2026-03-27
Proxmox: make variables user-configurable

minimalmattt:
## Change description
This PR adds the options to customize a few packer variables for Proxmox templates

## Additional context

These changes were tested with the ubuntu 24.04 image against Proxmox 9.1


mboersma:
Thanks for this! I pulled it down and validated it locally with our pinned Packer 1.9.5 and the Proxmox plugin, and it works as intended. Behavior is preserved: `disk_type` defaults to `scsi` matching the old hardcoded value, and the other four resolve to empty when unset, which matches their prior "field omitted" state so the plugin falls back to the same defaults. `Prepare()` passed cleanly with empty `io_thread`, `model`, `pool`, and `http_interface`.

One small thing: only `disk_type` gets a declared default in the variables block. I'd suggest declaring `io_thread`, `http_interface`, `model`, and `pool` there too, the way `disk_type` is, so the defaults are documented and discoverable. There's also some inconsistent spacing (`"model" : `, `{{ user `pool` }}`) versus the file's usual `{{user `x`}}` style. Neither is a blocker.

minimalmattt:
Hi @mboersma, thanks for the review.

I didn't add the defaults initially to keep them the same as upstream packer's defaults, I am however happy to add them since having the defaults specified would skip having to go to the docs altogether.

Regarding spacing, some keys (`node`,`numa`, `mtu` for example) have different spacing, Would you like me to format them as well? Or should that be a separate PR?

Thanks again

mboersma:
Thanks for the thoughtful questions!

On the defaults: yes, please go ahead and add them. The nice thing is that declaring them doesn't have to change any behavior. If you set `http_interface`, `model`, and `pool` to empty strings in the variables block, that's exactly what the template already resolves to today when they're undeclared, so it's the same behavior we validated, just documented and discoverable. For `io_thread` an empty string works too since it decodes to false, though `"false"` is fine if you'd rather make it explicit for a boolean. Either way the defaults stay behavior-preserving.

On the spacing: let's keep this PR focused. You're right that the file is already inconsistent, but that predates your change. The dominant style is the no-space `{{user `x`}}` form, so I'd just match that on the lines you're adding (`"model":`, `{{user `pool`}}`, etc.) and leave the existing lines as-is. If you want to tidy the rest, a separate PR would be very welcome, but no need to widen the scope here.

Thanks again for working on this!

minimalmattt:
Hi again, after a little fight with an old .gitconfig (which I hope to have won 🫪), everything should be in order.

I tested  this config on my Proxmox cluster and it completed correctly FWIW.

Let me know if it looks good and if I need to squash the commits afterwards.

Thanks again

mboersma (approved):
/lgtm
/approve

@minimalmattt thanks! Sorry I let this dangle so long.

minimalmattt:
@mboersma thank you for taking the time to review!  The PR for cleaning up the template will come shortly

Thanks again

PR #1925, opened 2026-02-02
Bump ansible to v2.16.16

mboersma:
## Change description

Updates ansible-core to [v2.16.16](https://pypi.org/project/ansible-core/2.16.16/), supporting Python 3.10-3.12.

## Related issues

See #1653 for prior art.

## Additional context

See also #1894. I think we can get there, but maybe in baby steps...

PR #1972, opened 2026-04-01
Bump ansible to v2.16.18

mboersma:
## Change description

Updates ansible-core to [v2.16.18](https://pypi.org/project/ansible-core/2.16.18/), supporting Python 3.10-3.12.

## Related issues

See #1925 for prior art.

## Additional context

See also #1894. I think we can get there, but maybe in baby steps...

### DIFF
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Note: ansible-core v2.16 supports Python 3.10-3.12.
-_version_ansible_core="2.16.16"
+_version_ansible_core="2.16.18"
 
 case "${OSTYPE}" in
 linux*)


mboersma:
/hold

I realize ansible-core 2.17 also supports Python 3.10-3.12. I'll open another PR and see if we can bump up to that.

mboersma:
Note that ansible-core's [support matrix](https://docs.ansible.com/projects/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) claims broader support for Python versions, but the packages available at pypi are much more constrained, and that's what determines whether they can actually be installed with a given version of Python.

mboersma:
/hold cancel

This is a safe change. Let's revisit #1973 and #1974 after this merges.

mboersma:
/retest

Transient error.

drew-viles (commented):
/lgtm
/approve

PR #1975, opened 2026-04-06
set default ubuntu repos for setup role

joshfrench:
## Change description
Adds default values to the Ansible vars added in #1961

## Additional context
These are only available to Ansible because the templates pass them in via ansible_common_vars. Set defaults so this role doesn't depend on the specifics of the build pipeline.

We maintain custom build targets that reuse some of the existing Ansible roles. Our templates are HCL based though, and can't make use of the JSON-based common var files (which are how these values are currently passed to Ansible.) This PR just makes the setup role fully self-contained and agnostic of how it's executed.


mboersma (commented):
/ok-to-test

mboersma:
/lgtm
/approve
/retest

`The zone 'projects/k8s-infra-e2e-boskos-030/zones/us-central1-a' does not have enough resources available to fulfill the request.  '(resource type:compute)'.`

PR #1980, opened 2026-04-15
Fix more ansible lint violations

mboersma:
## Change description

Fixes five types of `ansible-lint` warnings and removes them from the ignore file.

## Related issues

- Fixes #

## Additional context

I stuck with changes that seemed entirely safe; we don't want to break behavior just to placate the linter.

I think the remaining violations are trickier and will require more careful thought, if indeed we want to fix them.


mboersma:
cc: @ffais

images/capi/ansible/roles/providers/tasks/awscliv2.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: Upgrade pip to latest
+- name: Upgrade pip to latest  # noqa: package-latest
```
mboersma, line 2: This doesn't really change anything, but it seemed better to make the linter exception more visible by doing it inline rather than in .ansible-lint-ignore.

ffais:
Looks good to me. 
There are still some rule violations with ansible-linter version 25.7.0, but we can deal with them as soon as we bump it.

drew-viles (commented):
/lgtm
/approve

PR #1981, opened 2026-04-20
Migrate bare ansible_* facts to ansible_facts[] dict form

mboersma:
## Change description

This PR sets `inject_facts_as_vars=False` in ansible.cfg and converts all bare ansible_* fact references (e.g. ansible_os_family) to the dict form (ansible_facts['os_family']) across playbooks, roles, and templates.

This eliminates the INJECT_FACTS_AS_VARS deprecation warnings emitted by ansible-core >=2.18. The current default (True) will flip to False in ansible-core 2.24, so this change future-proofs the codebase.

The Flatcar OS-family override tasks retain bare ansible_os_family as the set_fact key, which Ansible syncs back into ansible_facts.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/ansible.cfg
+++ b/images/capi/ansible.cfg
@@ -15,6 +15,7 @@
 [defaults]
 remote_tmp = /tmp/.ansible
 display_skipped_hosts = False
+inject_facts_as_vars = False
 
 [ssh_connection]
 pipelining = False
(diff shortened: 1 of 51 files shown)


drew-viles (commented):
/lgtm
/approve

PR #1982, opened 2026-04-20
Add westus2 to default replicated regions for Azure SIG images

mboersma:
## Change description

Adds "westus2" to the default set of regions for replicating Azure SIG images.

## Related issues

- Fixes #


willie-yao (approved):
/lgtm

PR #1994, opened 2026-05-01
feat(containerd): adding ability to override the default image_pull_progress_timeout for containerd. Defaults to 5m

drew-viles:
## Change description
Adding the ability to override the image_pull_progress_timeout value for containerd.

It currently defaults to 5m which is fine for the most part but some people may want to bump that number if they are pulling large images - such as ones containing LLMs.

I've run this on OpenStack to test it with an override:

It's been a while since I did a "covers all the providers" change like this so remind me if I've missed anything 🤣


images/capi/packer/proxmox/packer.json.tmpl
```diff
@@ -1,3 +1,4 @@
+    "ssh_proxy_user": null,
```
mboersma: This line seems like it accidentally crept in. It makes the file no longer valid JSON.

images/capi/packer/proxmox/packer.json.tmpl
```diff
@@ -47,6 +48,7 @@
       "sockets": "{{user `sockets`}}",
       "ssh_password": "{{user `ssh_password`}}",
       "ssh_timeout": "2h",
+      "ssh_timeout": "2h",
```
mboersma: This line is a duplicate of the one above.

images/capi/.gitignore
```diff
@@ -11,3 +11,4 @@ manifest.json
 # Goss test droppings
 debug-goss-spec.yaml
 goss-spec.yaml
+downloaded_iso_path
```
mboersma: This doesn't seem related to the rest of the PR. If it's needed, could we make it a separate commit?

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
```diff
@@ -10,6 +10,8 @@ imports = ["/etc/containerd/conf.d/*.toml"]
 
 [plugins]
 {% if containerd_version is version('2.0.0', '>=') %}
+  [plugins.'io.containerd.cri.v1.images']
+    image_pull_progress_timeout = "{{containerd_image_pull_progress_timeout}}"
```
mboersma:
```suggestion
    image_pull_progress_timeout = "{{ containerd_image_pull_progress_timeout }}"
```

Maybe add spaces for consistency with the rest of the template.

mboersma: Also, should we add `containerd_image_pull_progress_timeout` to the docs somewhere if it's user-overridable? And maybe a goss assertion that the rendered /etc/containerd/config.toml actually contains the configured value?

images/capi/packer/proxmox/packer.json.tmpl
drew-viles: aaah dang it, I thought I'd removed that and the download_iso bit. Just working on some proxmox stuff in the background which is why they snuck in. Good catch. I'll remove

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
drew-viles: Yeah fair shout, I'll get that in.

images/capi/packer/proxmox/packer.json.tmpl
drew-viles: I don't even have an explanation for how that got there. 🫢

docs/book/src/capi/containerd/customizing-containerd.md
```diff
@@ -73,3 +73,15 @@ root@sandboxed-container:/# dmesg
 ```
 
 You are running a sandboxed container.
+
+# Additional Customizations
+
+Containerd can be further customized in a couple of ways. One option that is directly inserted into the containerd
+[
```
mboersma: Accidental newline here.

docs/book/src/capi/containerd/customizing-containerd.md
```diff
@@ -73,3 +73,15 @@ root@sandboxed-container:/# dmesg
 ```
 
 You are running a sandboxed container.
+
+# Additional Customizations
```
mboersma:
```suggestion
## Additional Customizations
```

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
```diff
@@ -10,6 +10,8 @@ imports = ["/etc/containerd/conf.d/*.toml"]
 
 [plugins]
 {% if containerd_version is version('2.0.0', '>=') %}
+  [plugins.'io.containerd.cri.v1.images']
+    image_pull_progress_timeout = "{{ containerd_image_pull_progress_timeout }}"
```
mboersma, line 15:
```suggestion
{% if containerd_image_pull_progress_timeout | default('') | length > 0 %}
    image_pull_progress_timeout = "{{ containerd_image_pull_progress_timeout }}"
{% endif %}
```

This would guard any provider not having explicitly set this value in their JSON config.

mboersma (changes requested):

images/capi/ansible/roles/containerd/templates/etc/containerd/config.toml
drew-viles: Yeah I'm happy to add this. So, I set the default for `5m` in the `default.yaml` to set it to the default containerd value, but actually this makes more sense in case a provider has their own defaults. It prevents overwriting theres with ours I guess. Although unlikely, as we write our own containerd config anyway, it doesn't hurt to provide that protection.

docs/book/src/capi/containerd/customizing-containerd.md
drew-viles: That'll be my IDE "helping" with wrap around 🤦

images/capi/ansible/roles/containerd/defaults/main.yml
```diff
@@ -21,3 +21,4 @@ containerd_url: "{{ containerd_baseurl }}/{{ containerd_filename }}"
 containerd_runc_url: "https://github.com/opencontainers/runc/releases/download/v{{ runc_version }}/runc.{{ arch }}"
 runc_version: "1.3.4"
 containerd_runc_checksum_url: "https://github.com/opencontainers/runc/releases/download/v{{ runc_version }}/runc.sha256sum"
+containerd_image_pull_progress_timeout: "5m"
```
AndiDog: We shouldn't bake in a fixed default but use the one from containerd. With the PR, nothing should render differently from before unless the timeout is overridden. Therefore I think it should be like this:
```suggestion
containerd_image_pull_progress_timeout: ""
```

This is repeated in many locations.

goss matching likely has to be adapted, too.

images/capi/ansible/windows/roles/runtimes/templates/config.toml
```diff
@@ -25,6 +25,8 @@ imports = ["{{ containerd_conf_dir }}\\conf.d\\*.toml"]
 
 [plugins]
 {% if containerd_version is version('2.0.0', '>=') %}
+  [plugins.'io.containerd.cri.v1.images']
+    image_pull_progress_timeout = "{{ containerd_image_pull_progress_timeout }}"
```
AndiDog, line 30: This needs `{% if [...]` as well.

images/capi/ansible/roles/containerd/defaults/main.yml
drew-viles: That's a fair argument

drew-viles:
I'll just work through the validate errors now

drew-viles:
ok, should be all sorted now.

mboersma (approved):
/lgtm
/approve

mboersma:
/override pull-ova-all

Known photon-5 build failure.

PR #2000, opened 2026-05-08
Remove dead python2 references

### LINKED ISSUE #578: Remove python2 from CAPi Photon images
The Photon 3 CAPI images install both Python2 and Python3. AFAICT cloud-init is OK with Python3, so we should be able to remove Python2 and save a bit of disk space.

mboersma:
## Change description

Cleanup of dead `python2` references that target RHEL/CentOS 7 and 8 — neither of which are build targets in image-builder anymore (the README only lists `rhel-9`, `rockylinux-9`, `almalinux-9`, `centos-9`, and `Makefile` has no `*-7`/`*-8` entries).

Removes:

- `roles/providers/tasks/vmware-redhat.yml`: the `Install python2 pip` task (gated on `distribution_major_version <= 8`) and the `Upgrade pip` task (gated on `== 7`). Both conditions are now unreachable.
- `packer/goss/goss-vars.yaml`: four `python2-pip:` entries (two ungated under `centos:`, two under `distro_version: "8"` blocks).

`make lint` passes cleanly.

## Related issues

- Closes #578 (no longer applicable — Photon 3 has been removed entirely; current Photon 4/5 paths already use python3 only — but this PR cleans up the unrelated python2 cruft that lingered elsewhere)

## Additional context

No behavior change for any currently-built distro. Easy revert if desired.

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/vmware-redhat.yml
+++ b/images/capi/ansible/roles/providers/tasks/vmware-redhat.yml
@@ -30,23 +30,6 @@
   ansible.builtin.set_fact:
     cloud_init_version: "{{ ansible_facts.packages['cloud-init'][0].version }}"
 
-- name: Install python2 pip
-  ansible.builtin.dnf:
-    name: "{{ packages }}"
-    state: present
-  vars:
-    packages:
-      - python2-pip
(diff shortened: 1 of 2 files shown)


drew-viles (commented):
/lgtm
/approve

PR #2001, opened 2026-05-08
Remove dead RHEL 7/8 ansible code paths and orphaned nutanix cloud-init files

mboersma:
## Change description

Cleanup of dead code and orphaned files for RHEL/CentOS 7 and 8 paths that are no longer build targets. The Makefile and README matrix have only `*-9` variants for RHEL/CentOS/Rocky/AlmaLinux.

Removes:

- `roles/providers/tasks/main.yml`: the `Set cloudinit feature flags for redhat 8` task (gated on `distribution_major_version == "8"`).
- `roles/sysprep/tasks/main.yml`: the `<= 7` guard around the `Rotate journalctl` task (always-true now).
- `packer/nutanix/linux/cloud-init/{rhel,rockylinux}/8/user-data.tmpl`: orphaned — there's no `rhel-8.json` or `rockylinux-8.json` in `packer/nutanix/` that would consume them.

`make lint` passes cleanly.

- Is this change including a new Provider or a new OS? (y/n) **n**

## Related issues

- Follow-up cleanup discovered while looking at #578 (closed). A second small PR will follow for the `rh8_rpms` blocks in `packer/goss/goss-vars.yaml`.

## Additional context

No behavior change for any currently-built distro.

PR #2002, opened 2026-05-08
Remove dead RHEL/CentOS 8 entries from goss-vars

mboersma:
## Change description

Removes dead `distro_version: "8"` blocks from `images/capi/packer/goss/goss-vars.yaml`. These goss test entries are unreachable: image-builder no longer has any `*-8` build targets — `Makefile` and the README matrix only reference `rhel-9`, `rockylinux-9`, `almalinux-9`, and `centos-9`.

Also drops a few empty `os_version:` keys left behind after removing the only child entry, and deletes a couple of `python2-pip:` entries that lived inside the removed blocks (also covered by #2000 in the lines outside these blocks).

The `rh8_rpms` YAML alias itself is **kept** because the `oracle linux:oci` block still references it; all five packages it lists (`curl`, `yum-utils`, `nftables`, `python3-netifaces`, `python3-requests`) are valid on Oracle Linux 9, so goss behavior for `oci-oracle-linux-9` is unchanged.

## Related issues

- Follow-up cleanup discovered while looking at #578 (closed). Companion to #2001.

## Additional context

YAML structure validated locally with `yaml.safe_load`. No behavior change for any currently-built distro.

### DIFF
--- a/images/capi/packer/goss/goss-vars.yaml
+++ b/images/capi/packer/goss/goss-vars.yaml
@@ -176,9 +176,6 @@ almalinux:
     package:
       open-vm-tools:
     os_version:
-    - distro_version: "8"
-      package:
-        <<: *rh8_rpms
     - distro_version: "9"
       package:
         <<: *rh9_rpms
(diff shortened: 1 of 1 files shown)


drew-viles (commented):
/override pull-ova-all

Known photon-5 build failure wrt iptables.

/lgtm
/approve

drew-viles (commented):
/lgtm
/approve

PR #2009, opened 2026-05-11
Windows: improve StartKubelet.ps1 kubelet bootstrap

mboersma:
## Change description

Three small robustness fixes to the Windows `StartKubelet.ps1` template:

1. Replace `String.TrimStart('KUBELET_KUBEADM_ARGS=')` (char-set strip, fragile) with a regex substring strip.
2. Replace `Invoke-Expression` with `& $kubelet @argList` so PowerShell doesn't re-parse flag values containing `=`, `:`, `$`, etc.
3. Log the final kubelet command line to `…\var\log\kubelet\start-kubelet.log` before exec, so future bootstrap issues are diagnosable from a single file.

No behavioral change intended for the current kubeadm output format; this is purely defensive against future kubeadm/kubelet flag changes (e.g. recent activity around `--cloud-provider` and `--register-with-taints`).

## Related issues

- Fixes #

## Additional context

Surfaced while debugging Windows nodes that registered without the `node.cloudprovider.kubernetes.io/uninitialized` taint in the `cloud-provider-azure-ccm-windows-capz` job.

### DIFF
--- a/images/capi/ansible/windows/roles/kubernetes/templates/StartKubelet.ps1
+++ b/images/capi/ansible/windows/roles/kubernetes/templates/StartKubelet.ps1
@@ -13,18 +13,30 @@
 # limitations under the License.
 
 # From https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/kubeadm/scripts/PrepareNode.ps1
-$FileContent = Get-Content -Path "/var/lib/kubelet/kubeadm-flags.env"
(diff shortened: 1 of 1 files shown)


zylxjtu:
do you plan to upload the start-kublet logs?
Agree that this kind of commandline args can go tricky, and will need to test/validate against the image-building-publishing

/lgtm
/approve
/hold for test results

mboersma (commented):
/approve
/hold cancel

PR #2025, opened 2026-05-26
Allow Bring Your Own Packer via PACKER_BIN/PACKER_VERSION env vars

mboersma:

image-builder pins Packer to v1.9.5 — the last MPL-2.0 release before HashiCorp relicensed Packer under the BUSL in v1.10.0. `hack/ensure-packer.sh` enforces that pin by actively *replacing* any other Packer the user has on `PATH` with 1.9.5 in `.local/bin`, and the Makefile gives no way to point at a different binary without editing sources.

This PR introduces a **Bring Your Own Packer** capability while keeping the existing pinned behavior as the default.

## The new knobs

| Env var | Default | Effect |
|---|---|---|
| `PACKER_BIN` | unset | Explicit path to a Packer binary. `ensure-packer.sh` validates it and exits 0; the Makefile uses it for every `packer init/build/validate`. |
| `PACKER_VERSION` | `1.9.5` | Version that `ensure-packer.sh` downloads/verifies when managing Packer itself. |
| `IB_ALLOW_ANY_PACKER` | `0` | Opt-in: accept an existing Packer on PATH whose version differs from `PACKER_VERSION` instead of silently downgrading it. Emits a license warning if the detected version is ≥ 1.10.0 (BUSL). |

`PACKER` is also now overridable on the `make` command line (`make PACKER=/path/to/packer build-...`) thanks to `?=`.

`PACKER_VERSION` and `IB_ALLOW_ANY_PACKER` are also exposed as `Dockerfile` build args so container-image consumers can rebuild without forking.

## Commits

1. **Allow bring-your-own Packer via PACKER_BIN/PACKER_VERSION env vars** — Makefile + `hack/ensure-packer.sh` (plus drive-by `shellcheck` SC1035 fix and a fall-through bug fix in the 'unexpected packer found' branch).
2. **Plumb PACKER_VERSION/IB_ALLOW_ANY_PACKER through Dockerfile** — new `ARG`s and `ENV`s; defaults preserve current behavior.
3. **Document Packer pin and Bring Your Own Packer** — new `docs/book/src/capi/packer.md` page, added to `SUMMARY.md`, and `capi.md` prerequisites updated to point at it (replacing the stale '>= 1.6.0' constraint).

## Default behavior is unchanged

With no env vars set, `ensure-packer.sh` still installs and uses Packer 1.9.5 in `.local/bin`, and the Makefile resolves `PACKER` exactly as before. Verified locally:

```
--- no env vars ---
PACKER=[…/images/capi/.local/bin/packer]
--- PACKER_BIN=/usr/local/bin/foo ---
PACKER=[/usr/local/bin/foo]
--- PACKER=cli-override ---
PACKER=[cli-override]
```

`shellcheck -x hack/ensure-packer.sh` is clean.

## Follow-ups (not in this PR)

* Optionally a non-blocking CI matrix entry against a newer Packer to flag drift early.

None tracked — motivated by the kubernetes-sigs/image-builder concern around Packer's BUSL relicense and the desire to let downstream users adopt newer Packer at their own risk.

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -27,5 +27,6 @@
   - [Testing the Images](./capi/goss/goss.md)
   - [Using Container Images](./capi/container-image.md)
   - [Customizing containerd](./capi/containerd/customizing-containerd.md)
+  - [Packer version and BYO Packer](./capi/packer.md)
   - [Releasing](./capi/releasing.md)
 - [Glossary](./glossary.md)
--- a/docs/book/src/capi/capi.md
+++ b/docs/book/src/capi/capi.md
@@ -6,7 +6,7 @@ The Image Builder can be used to build images intended for use with Kubernetes [
 
 [Packer](https://www.packer.io) and [Ansible](https://github.com/ansible/ansible) are used for building these images. This tooling has been forked and extended from the [Wardroom](https://github.com/heptiolabs/wardroom) project.
 
-- [Packer](https://www.packer.io/intro/getting-started/install.html) version >= 1.6.0
+- [Packer](https://www.packer.io/intro/getting-started/install.html) version `1.9.5` (the default pinned by `hack/ensure-packer.sh`; see [Packer version and Bring Your Own Packer](./packer.md) to use a different version)
 - [Goss plugin for Packer](https://github.com/YaleUniversity/packer-provisioner-goss) version >= 1.2.0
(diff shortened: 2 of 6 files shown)


drew-viles (commented):
I can't believe we didn't originally think of this as the solution to the (almost) age old problem of. "what do with do with Packer now?"

Nice one!!

/lgtm
/approve

mboersma:
/retest

Couple of Azure builds flaked out.

drew-viles (commented):
No probs

/lgtm
/approve

PR #2026, opened 2026-05-29
ansible: add disable-modules role

anders-elastisys:
## Change description

Adds a disable-modules role to be able to disable certain kernel modules when building new images.

## Related issues

- Fixes #

## Additional context
Added default vars to be modules related to recent exploits, but maybe the default should be kept empty and let user override them instead?


images/capi/ansible/roles/disabled_kernel_modules/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+- name: Disable module
+  ansible.builtin.copy:
```
mboersma, line 2: ansible-lint's `risky-file-permissions` flags this since there's no `mode`. Could you add something like `mode: "0644"`?

images/capi/ansible/roles/disable-modules/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+- name: Disable module
+  ansible.builtin.copy:
+    dest: "/etc/modprobe.d/disable-{{ item }}.conf"
+    content: |
+      install {{ item }} /bin/false
+  loop: "{{ disabled_kernel_modules }}"
+
+- name: Blacklist module
+  ansible.builtin.copy:
+    dest: "/etc/modprobe.d/blacklist.conf"
```
mboersma: Two things here. This writes a fixed `blacklist.conf` with templated content inside a loop, so each pass overwrites the file and only the last module (`rxrpc`) ends up blacklisted. The per-module `install ... /bin/false` files above already disable everything, but if you want the blacklist too, I'd use `blockinfile` or per-module filenames like the first task does. Also same `mode` note as above applies to this copy.

images/capi/ansible/roles/disable-modules/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+# currently disable modules to address the following exploits:
+# - https://copy.fail/
+# - https://github.com/V4bel/dirtyfrag
+# - https://github.com/0xdeadbeefnetwork/Copy_Fail2-Electric_Boogaloo
+disabled_kernel_modules:
```
mboersma: I'd put this list in `defaults/main.yml` instead, ideally defaulting to empty so users opt in to what they want disabled.

mboersma (commented):
Thanks for this, disabling risky modules at build time is a nice addition. A few things I'd like to see before we merge:

- The role directory is `disable-modules`, but ansible-lint's `role-name` rule doesn't allow hyphens, so this fails `make lint`. All our other roles use underscores, so could you rename it to `disable_modules`?
- I'd move `disabled_kernel_modules` into `defaults/main.yml` rather than `vars/main.yml`. Every other role keeps its tunables in `defaults/` so folks can override them, and `vars/` precedence makes that awkward. That also answers your own question in the description: I'd lean toward defaulting it empty and letting users opt in to the specific modules.

I left a couple of inline notes too. Heads up that CI hasn't actually run yet since the PR still needs an ok-to-test.

images/capi/ansible/roles/disabled_kernel_modules/tasks/main.yml
anders-elastisys: https://github.com/kubernetes-sigs/image-builder/pull/2026/changes/e89c213b99b7d754c0f4bbdf4d295b42ddd76d3b

images/capi/ansible/roles/disable-modules/tasks/main.yml
anders-elastisys: I decided to remove the blacklist, as I do not really see a strong need for it: https://github.com/kubernetes-sigs/image-builder/pull/2026/changes/53ff0c062f1c540c7bca67f8f22db718b88b1962

images/capi/ansible/roles/disable-modules/vars/main.yml
anders-elastisys: https://github.com/kubernetes-sigs/image-builder/pull/2026/changes/39d392a127aba4f1dd927f6e6b5994a651d03acc and https://github.com/kubernetes-sigs/image-builder/pull/2026/changes/5dfe5b7557b98e94c24fee765e3ad9d60baa1af8

anders-elastisys:
@mboersma thanks for taking your time to review this, I think I managed to address your comments now, PTAL

mboersma (approved):
/lgtm
/approve

mboersma:
@anders-elastisys there's an ansible-lint error to fix, otherwise I think this is good.

anders-elastisys:
@mboersma I addressed the linting issues :)

mboersma (approved):
/lgtm

PR #2027, opened 2026-05-29
Switch Windows kubernetes_base_url to dl.k8s.io

mboersma:
The `kubernetesreleases.blob.core.windows.net/kubernetes/...` path has had intermittent gaps for recent Kubernetes patch releases — for example `v1.35.1` and the entire `v1.36.x` line are currently missing — which blocks Kubernetes version bumps (see #1993).

`dl.k8s.io` is the canonical upstream Kubernetes release CDN, populated automatically by `krel` for every tagged release. I verified complete coverage of both `kubelet.exe` and `kube-log-runner.exe` (and the other Windows binaries) across v1.32.10, v1.33.12, v1.34.8, v1.35.1, v1.35.5, and v1.36.1. The URL shape is identical, so the change is a one-line swap in each provider's `packer-windows.json`.

`dl.k8s.io` is also already the example value documented in `images/capi/ansible/windows/example.vars.yml`.

Users who specifically need the Microsoft-signed Windows binaries previously fetched from `kubernetesreleases` can still override `kubernetes_base_url` via packer vars — it remains a user-configurable variable.

## Related issues

Unblocks #1993 and prevents this class of failure from recurring on future patch releases.

/closes #2041

## Special notes for your reviewer

Files changed (one identical line per file):
- `images/capi/packer/ami/packer-windows.json`
- `images/capi/packer/azure/packer-windows.json`
- `images/capi/packer/nutanix/packer-windows.json`
- `images/capi/packer/oci/packer-windows.json`
- `images/capi/packer/ova/packer-windows.json`

### DIFF
--- a/images/capi/packer/ami/packer-windows.json
+++ b/images/capi/packer/ami/packer-windows.json
@@ -185,7 +185,7 @@
     "ib_version": "{{env `IB_VERSION`}}",
     "image_name": "capa-ami-{{user `build_name`}}-{{user `kubernetes_semver`}}-{{user `build_timestamp`}}",
     "kms_key_id": "",
-    "kubernetes_base_url": "https://kubernetesreleases.blob.core.windows.net/kubernetes/{{user `kubernetes_semver`}}/bin/windows/{{user `kubernetes_goarch`}}",
+    "kubernetes_base_url": "https://dl.k8s.io/release/{{user `kubernetes_semver`}}/bin/windows/{{user `kubernetes_goarch`}}",
     "manifest_output": "manifest.json",
     "nssm_url": null,
     "prepull": null,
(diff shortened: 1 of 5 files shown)


zylxjtu:
just curious, are we going to retire the "kubernetesreleases.blob.core.windows.net/kubernetes/..." path?

mboersma:
> retire the "kubernetesreleases.blob.core.windows.net/kubernetes/..." path?

I don't know, I think that's a SIG-Windows question. Isn't that where those artifacts are created?

I just know that most recent k8s releases don't have Windows artifacts at that location, but that dl.k8s.io does seem to be up to date. So this seemed like a workaround to enable us to build more current k8s images for Windows, and to get some PRs to pass. But we should talk about it. I'm not 100% sure of the historical reasons we've been using the MS-built artifacts.

mboersma (commented):
/approve

(Mea culpa, the other maintainer is on a well-deserved vacation.)

mboersma:
/retest

-gen2 flake failure, Windows was fine.

marosset (approved):
/lgtm
/hold for ci

PR #2029, opened 2026-06-02
Resolve latest Flatcar version from release server version.txt

hajowieland:
## Change description

Resolve the Flatcar version directly from version.txt of the Flatcar release server. Until a few hours ago there was a discrepancy between the version released at release.flatcar-linux.net and flatcar.org - so the latest current stable was not pointing to [stable-4593.2.2](https://github.com/flatcar/scripts/releases/tag/stable-4593.2.2) yet.

To fix this, an image-builder user needed to explicitely set `FLATCAR_CHANNEL` to stable and `FLATCAR_VERSION` to 4593.2.2 - otherwise with the default `FLATCAR_VERSION` set to current, an image with 4593.2.1 was build.

## Additional context

We point to the amd64 URL as the version output for arm64 is the same as well.

### DIFF
--- a/images/capi/hack/image-grok-latest-flatcar-version.sh
+++ b/images/capi/hack/image-grok-latest-flatcar-version.sh
@@ -5,8 +5,6 @@
 channel="$1"
 
 curl -L -s \
-     "https://www.flatcar.org/releases-json/releases-$channel.json" \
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm
/approve

This should be more reliable AFAICT. Thanks @hajowieland!

PR #2030, opened 2026-06-02
Mask apt-daily and unattended-upgrades early to fix dpkg lock race

mboersma:

The `apt-daily`/`apt-daily-upgrade` timers and the `unattended-upgrades` service fire shortly after first boot and intermittently hold the dpkg/apt lock while the `containerd`, `kubernetes`, and `sysprep` roles install packages. This causes flaky Ubuntu (Azure) build failures like:

```
E: Could not get lock /var/cache/apt/archives/lock
```

`sysprep` already disables these units, but that runs at the *end* of the build — too late to protect the install steps that race them. #2024 fixed the sysprep `apt-mark hold` instance, but the same race still bites earlier roles.

This PR stops and **masks** the units at the start of the `node` role, before any package installs, and waits for any in-flight apt process to release the dpkg frontend lock. It complements #2024 by closing the earlier, install-time instances of the same race.

Related to #2024.

## Release note

```release-note
Fix intermittent Ubuntu build failures caused by apt-daily/unattended-upgrades racing the dpkg lock during package installs.
```

### DIFF
--- a/images/capi/ansible/roles/node/tasks/debian.yml
+++ b/images/capi/ansible/roles/node/tasks/debian.yml
@@ -1,0 +1,46 @@
+# Copyright 2026 The Kubernetes Authors.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


drew-viles (commented):
Sorry this week has been mad busy - just FYI I'm on vacation for the next 2 weeks so may be slow to respond again during that.

/lgtm
/approve

PR #2032, opened 2026-06-03
Remove stale Microsoft prod repo on Azure Ubuntu builds

mboersma:

Canonical's Ubuntu 26.04 Azure image ships a packages.microsoft.com prod source that points at the old 18.04/bionic repo. Its signing key isn't on the image, so the first apt update fails with `NO_PUBKEY EB3E94ADBE1229CF` and the azure-sigs `ubuntu-2604` and `ubuntu-2604-gen2` builds break.

We don't use that baked-in repo, so this drops it early in the Azure provider tasks, before any apt update runs. It's scoped to the broken 18.04/prod source only, so 22.04 and 24.04 aren't touched.

## Release note

```release-note
Fix Azure Ubuntu 26.04 image builds failing on a stale preinstalled packages.microsoft.com repo.
```

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/azure.yml
+++ b/images/capi/ansible/roles/providers/tasks/azure.yml
@@ -12,6 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+# Canonical's Ubuntu Azure marketplace images ship a preconfigured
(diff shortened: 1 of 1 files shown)


mboersma:
/close

The 26.04 jobs began passing today on Azure, so perhaps this problem has been corrected at the source.

Issue #2035, opened 2026-06-05
Ubuntu 26.04 OVA build hangs in pull-ova-all (temporarily excluded from CI)


mboersma:
## What happened

The Ubuntu 26.04 (Resolute) OVA build hangs in `pull-ova-all` and times out the whole job, so we temporarily excluded it from CI in `images/capi/scripts/ci-ova.sh` (alongside the existing photon-4 exclusion).

## Details

`ci-ova.sh` runs all vSphere OVA builds in parallel and waits on all of them. The `ubuntu-2604` build boots, gets an IP, but its autoinstall never completes, so the VM never reboots into the installed system. Packer keeps failing password auth against the idle live installer until the 2h prow timeout SIGINTs every build, failing the whole job.

Evidence (same CI run, ubuntu-2404 vs ubuntu-2604):

| | ubuntu-2404 (works) | ubuntu-2604 (hangs) |
|---|---|---|
| auth failures vs live installer | 47 (brief) | 705 (~2h) |
| reboot gap (conn refused / no route) | 1 | 0 |
| `Connected to SSH!` / provisioning | yes | never |

The 26.04 OVA var file and `user-data.tmpl` are byte-identical to 24.04's, so the config itself is fine for the 24.04 installer. The difference is 26.04's newer subiquity not consuming the NoCloud `cidata` seed (likely falling back to the interactive installer, or rejecting the `version: 1` autoinstall schema).

## Next steps

- Run an interactive `FOREGROUND=1 make build-node-ova-vsphere-ubuntu-2604` to capture the subiquity console and confirm whether it's a confirmation prompt, a schema error, or seed-not-detected.
- Fix the 26.04 autoinstall (boot command datasource hint and/or autoinstall schema) so the build completes.
- Re-enable `ubuntu-2604` in `ci-ova.sh` once it builds reliably.

mboersma:
Temporarily excluded via #2036.

PR #2036, opened 2026-06-05
Exclude Ubuntu 26.04 OVA build from CI

mboersma:

The `ubuntu-2604` vSphere OVA build hangs in `pull-ova-all` and times out the whole job. `ci-ova.sh` runs all OVA builds in parallel and waits on all of them, so one stuck build holds the job until the 2h prow timeout and fails every build.

The 26.04 build boots and gets an IP, but its autoinstall never completes, so the VM never reboots into the installed system. Packer keeps failing SSH auth against the idle live installer until the timeout. This isn't a flake, it reproduces on every run and is failing `pull-ova-all` on all PRs.

This excludes `ubuntu-2604` the same way `photon-4` is already excluded, until the autoinstall is fixed.

Tracked in #2035.

## Release note

```release-note
NONE
```

### DIFF
--- a/images/capi/scripts/ci-ova.sh
+++ b/images/capi/scripts/ci-ova.sh
@@ -29,10 +29,12 @@ export ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
 # The following are currently having issues running in the
 # test environment so are specifically excluded for now
 # - Photon-4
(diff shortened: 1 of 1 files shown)


drew-viles (commented):
/lgtm
/approve

PR #2038, opened 2026-06-10
Vultr: use `tags` instead of deprecated `tag` field

mondragonfx:
## Change description

Replaces a deprecated field. `tag` will cause vultr capi builds to fail. 

## Additional context

```
Error: Failed to prepare build: "vultr"

1 error occurred:
        * unknown configuration key: '"tag"'

make: *** [Makefile:665: build-vultr-ubuntu-2604] Error 1
```


mboersma (approved):
/lgtm
/approve

PR #2041, opened 2026-06-11
Switch Windows kubernetes binaries to DALEC packages

mboersma:
The `kubernetesreleases.blob.core.windows.net/kubernetes/...` path has had intermittent gaps for recent Kubernetes patch releases — for example `v1.35.1` and the entire `v1.36.x` line are currently missing — which blocks Kubernetes version bumps (see #1993).

Microsoft now publishes DALEC-built Windows binaries as signed ZIP archives on the AKS package CDN (`packages.aks.azure.com/dalec-packages`), the canonical replacement for the retired blob path. I verified complete coverage of `kubelet`, `kubeadm`, `kubectl`, and `kube-log-runner` across v1.35.0, v1.35.1, v1.35.2, v1.35.5, v1.36.0, and v1.36.1. Each archive contains a single `<binary>.exe` at its root, e.g. `https://packages.aks.azure.com/dalec-packages/kubelet/1.35.2/windows/amd64/kubelet_1.35.2-1_amd64.zip`.

The Windows kubernetes role now downloads and extracts those packages. New vars `kubernetes_package_base_url` and `kubernetes_package_revision` (default `1`) replace `kubernetes_base_url`; the revision is overridable per build via packer vars for the rare release that bumps it.

This is an alternative to #2027, which switches to `dl.k8s.io` instead. Unlike that approach, this keeps Windows on the Microsoft-signed binaries. Reviewers can decide which is preferred.

## Related issues

Unblocks #1993 and prevents this class of failure from recurring on future patch releases.

/closes #2027

## Special notes for your reviewer

Files changed:
- `images/capi/ansible/windows/roles/kubernetes/tasks/url.yml` (download + extract zips)
- `images/capi/ansible/windows/roles/kubernetes/defaults/main.yml` (new vars)
- `images/capi/ansible/windows/example.vars.yml`
- `images/capi/packer/config/windows/ansible-args-windows.json`
- `images/capi/packer/{ami,azure,nutanix,oci,ova}/packer-windows.json`


zylxjtu:
my personal preference would be dl.k8s.io

mboersma:
/hold

I think dl.k8s.io is more consistent for the image-builder repo as well. Let me poll @marosset and the upstream infra team and see what their preferences are.

marosset:
I'm fine with switching to `dl.k8s.io` too

mboersma:
Cool, let's use dl.k8s.io then to get unstuck. We can always revisit this PR if for some reason we need the MS-built binaries.

PR #2049, opened 2026-06-23
capi: allow deferring sysprep SSH cleanup

MaxRink:
## Change description

Allow the CAPI sysprep role to defer SSH host key and authorized key cleanup when the image build needs additional Packer provisioners to connect after sysprep has run.

This keeps the default cleanup behavior unchanged. Builders that set `defer_ssh_cleanup_to_packer=true` can move the final SSH cleanup to a later Packer step instead of losing direct SSH access in the middle of the build.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`

### DIFF
--- a/images/capi/ansible/roles/sysprep/defaults/main.yml
+++ b/images/capi/ansible/roles/sysprep/defaults/main.yml
@@ -17,4 +17,5 @@ pip_conf_file: ""
 remove_extra_repos: false
 flatcar_disable_autologin: false
 sysprep_require_grub_file: true
+defer_ssh_cleanup_to_packer: false
 extra_kernel_boot_params: ""
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -178,29 +178,38 @@
     path: /etc/ssh
     pattern: ssh_host_*
   register: ssh_host_keys
+  when: not (defer_ssh_cleanup_to_packer | bool)
 
 - name: Remove SSH host keys
   ansible.builtin.file:
     state: absent
     path: "{{ item.path }}"
-  loop: "{{ ssh_host_keys.files }}"
+  loop: "{{ ssh_host_keys.files | default([]) }}"
+  when: not (defer_ssh_cleanup_to_packer | bool)
 
 - name: Remove SSH authorized users
   ansible.builtin.file:
     state: absent
     path: "{{ item.path }}"
   loop:
-    - { path: /root/.ssh/authorized_keys }
-    - { path: "/home/{{ ansible_facts['env'].SUDO_USER | default(ansible_facts['user_id']) }}/.ssh/authorized_keys" }
-  when: ansible_facts['os_family'] != "Flatcar"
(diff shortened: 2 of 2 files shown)


mboersma (commented):
This looks good. Could you add `defer_ssh_cleanup_to_packer: false` to images/capi/ansible/roles/sysprep/defaults/main.yml? Right now the flag only exists as an inline `default(false)` , so it's a  bit hard to discover.

MaxRink:
Done in b0d206d: added `defer_ssh_cleanup_to_packer: false` to `images/capi/ansible/roles/sysprep/defaults/main.yml` so the flag is discoverable in role defaults.

Local checks:
- `git diff --check`
- `yamllint images/capi/ansible/roles/sysprep/defaults/main.yml`
- YAML parse/assertion that `defer_ssh_cleanup_to_packer` defaults to `false`

Prow has restarted on the amended commit.

MaxRink:
I checked the new `pull-azure-sigs` failure after the retest. It is not caused by this PR's sysprep SSH cleanup change. Only `sig-ubuntu-2604` failed; all other Azure SIG targets passed.

Root cause from the Prow artifact: apt cache refresh fails against a pre-existing Azure base-image Microsoft repo, `https://packages.microsoft.com/ubuntu/18.04/prod bionic`, because apt cannot see key `EB3E94ADBE1229CF`.

I filed the focused follow-up fix separately as #2060 so this PR stays scoped: https://github.com/kubernetes-sigs/image-builder/pull/2060

MaxRink:
Done in b0d206d9568b4a9eceb1f802667c93e60f806ce4: `defer_ssh_cleanup_to_packer: false` is now declared in `images/capi/ansible/roles/sysprep/defaults/main.yml`.

The remaining red `pull-azure-sigs` job is unrelated to this PR. It fails on `sig-ubuntu-2604` because apt cannot verify the legacy Microsoft repository key (`NO_PUBKEY EB3E94ADBE1229CF` for `packages.microsoft.com/ubuntu/18.04/prod`). That is covered by #2060, which currently has the Azure job green. I will rebase/retest this PR after #2060 lands.

MaxRink:
Addressed the review follow-up more completely in 8a6cbd5df852a2deabde736026bbd88cc747181a and rebased onto current upstream main.

What changed:
- `defer_ssh_cleanup_to_packer: false` stays declared in the sysprep role defaults.
- The SSH cleanup tasks now use that declared role default directly instead of keeping inline `default(false)` filters.
- The affected authorized-key cleanup conditions are split into normal Ansible `when` lists; the skipped-register loop guard remains in place for `ssh_host_keys.files | default([])`.

Local checks:
- `git diff --check`
- `yamllint images/capi/ansible/roles/sysprep/defaults/main.yml`
- YAML parse/assertions for the sysprep SSH cleanup tasks and rendered authorized-key paths

Prow has restarted on the rebased commit. Full `yamllint` on `images/capi/ansible/roles/sysprep/tasks/main.yml` still reports existing style issues outside this PR scope, so I did not include that as a passing check.

mboersma (approved):
/lgtm
/approve

PR #2050, opened 2026-06-23
capi: use deb822 repository for Kubernetes apt repo

MaxRink:
## Change description

Replace the Kubernetes Debian repository setup with `ansible.builtin.deb822_repository`.

Current `main` avoids `apt_key`, but does that by downloading an armored key, running `gpg --dearmor` via `ansible.builtin.shell`, and then configuring the repo with `ansible.builtin.apt_repository`. The native `deb822_repository` module can consume the upstream key URL through `signed_by`, avoids the shell task, and avoids the deprecated repository format.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`

### DIFF
--- a/images/capi/ansible/roles/kubernetes/handlers/main.yml
+++ b/images/capi/ansible/roles/kubernetes/handlers/main.yml
@@ -1,0 +1,4 @@
+---
+- name: Update apt cache
+  ansible.builtin.apt:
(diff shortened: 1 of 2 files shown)


mboersma (approved):
/lgtm
/approve

This is much cleaner, thanks for the improvement!

PR #2051, opened 2026-06-23
capi: dynamically include OS setup tasks

MaxRink:
## Change description

Use `ansible.builtin.include_tasks` for mutually exclusive OS setup task files.

The setup role currently statically imports every OS-family task file and adds a `when` condition to each import. Dynamic includes avoid expanding irrelevant OS task files for the current guest, which keeps skipped OS-specific tasks from being parsed or logged during builds for another family.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`


mboersma (approved):
/lgtm
/approve

PR #2052, opened 2026-06-23
capi: avoid following cri-tools symlinks

MaxRink:
## Change description

Set `follow: false` when the containerd role creates `/usr/bin/{ctr,crictl,critest}` symlinks.

The task creates forced symlinks that can temporarily point at targets that are not visible through `/usr/local/bin` yet. With Ansible's default symlink-follow behavior, applying file attributes can warn about a non-existent symlink target. The module-native fix is to keep `state: link` and explicitly avoid following the link.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -206,6 +206,7 @@
     mode: "0755"
     state: link
     force: true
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm
/approve

PR #2053, opened 2026-06-23
capi: default empty public repo flags before bool

MaxRink:
## Change description

Normalize empty `disable_public_repos` and `reenable_public_repos` values before applying the Ansible `bool` filter.

Some Packer/HCL build paths can pass optional string variables as an empty string. Newer Ansible versions warn when `bool` coerces `""` directly. Using `default(..., true)` preserves the existing false/default behavior for empty values while avoiding the deprecation warning.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`


mboersma (approved):
/lgtm
/approve
/retest

PR #2054, opened 2026-06-23
capi: default additional component flags to booleans

MaxRink:
## Change description

Default the `load_additional_components` boolean switches to `false` instead of an empty string.

The role already treats these values as booleans in task conditions. Setting boolean defaults avoids relying on Ansible coercion of `""` to false, which newer Ansible versions warn about.

## Related issues

None.

## Additional context

Validation:
- `git diff --check`


mboersma (approved):
/lgtm
/approve

PR #2055, opened 2026-06-23
capi: use dpkg selections in Debian sysprep

MaxRink:
## Change description

Replace Debian sysprep `apt-mark` shell pipelines with `ansible.builtin.dpkg_selections`.

The role already gathers package facts before pinning packages. Using `dpkg_selections` keeps the existing package hold/unhold behavior in Ansible modules instead of shelling out to `dpkg-query | xargs apt-mark`. The retry loop remains so transient dpkg lock contention from apt-daily or unattended-upgrades is still handled.

This also adds `sysprep_hold_installed_packages`, defaulting to `true`, so downstream image variants that replace the OS atomically can opt out of the hold-all sysprep step without changing existing image-builder behavior.

## Related issues

None.

## Additional context

Validation:
- `python3` YAML parse of `images/capi/ansible/roles/sysprep/defaults/main.yml` and `images/capi/ansible/roles/sysprep/tasks/debian.yml`
- `git diff --check`
- grep guard confirmed no `apt-mark`, `dpkg-query`, or `ansible.builtin.shell` remains in `images/capi/ansible/roles/sysprep/tasks/debian.yml`

Known lint baseline:
- `ansible-lint images/capi/ansible/roles/sysprep/tasks/debian.yml images/capi/ansible/roles/sysprep/defaults/main.yml` still reports pre-existing sysprep role issues outside this PR, including role-prefix naming findings and the existing `Enable repos` command task.

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -45,32 +45,31 @@
   when: "'unattended-upgrades' in ansible_facts.packages"
 
 # The apt-daily/unattended-upgrades services may still be running from before
-# they were disabled above. Retry the apt-mark calls until the dpkg frontend
+# they were disabled above. Retry package selection changes until the dpkg frontend
 # lock is released to avoid a flaky race with concurrent apt activity.
-- name: Pin all installed packages with apt-mark
-  ansible.builtin.shell: |
-    set -o pipefail
-    dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
-
-  args:
-    executable: /bin/bash
-  register: sysprep_apt_mark_hold
-  until: sysprep_apt_mark_hold.rc == 0
+- name: Pin all installed packages
+  ansible.builtin.dpkg_selections:
+    name: "{{ item }}"
+    selection: hold
+  loop: "{{ ansible_facts.packages.keys() }}"
+  register: sysprep_dpkg_selections_hold
+  until: sysprep_dpkg_selections_hold is not failed
   retries: 30
   delay: 10
+  when: sysprep_hold_installed_packages | default(true) | bool
 
-- name: Unpin grub packages with apt-mark for MaaS
-  ansible.builtin.shell: |
-    set -o pipefail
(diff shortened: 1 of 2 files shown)


MaxRink:
Rebased this PR onto current upstream main. The previous failing `pull-azure-sigs` run failed in Debian sysprep cleanup while `apt-get autoclean` was waiting for `/var/cache/apt/archives/lock`, not because of a merge conflict.\n\nI filed #2059 as the separate module-native fix for that pre-existing apt cleanup race: it stops already-running apt-daily service units and sets `lock_timeout` on the existing `ansible.builtin.apt` cleanup task.

MaxRink:
Rebased onto current upstream main after the sysprep apt-lock retry changes landed. The PR still removes the apt-mark shell pipelines, now preserving the existing retry behavior around ansible.builtin.dpkg_selections. Validation: git diff --check passed; ansible-lint --project-dir images/capi images/capi/ansible/roles/sysprep was run and still reports pre-existing sysprep-role baseline issues unrelated to this change.

MaxRink:
The latest `pull-ova-all` failure looks like a job timeout, not a source failure: the trace reaches `almalinux-9: SUCCESS` and then Prow terminates the job with `Process did not finish before 2h0m0s timeout`. This run was already on current upstream main (`117cceaae7d5816a67620b4d898fffcc8186b8c9`), so I am rerunning the failed lane rather than rebasing.

/test pull-ova-all

images/capi/ansible/roles/sysprep/tasks/debian.yml
```diff
@@ -45,32 +45,31 @@
...
-
-  args:
-    executable: /bin/bash
-  register: sysprep_apt_mark_hold
-  until: sysprep_apt_mark_hold.rc == 0
+- name: Pin all installed packages
+  ansible.builtin.dpkg_selections:
+    name: "{{ item }}"
+    selection: hold
+  loop: "{{ ansible_facts.packages.keys() }}"
```
mboersma, line 54: Potentially this loop is slower than the previous single `apt-mark hold` call, since it makes multiple calls. But in practice it looks like the builds didn't take noticeably longer than before, so I'm ok with it.

mboersma (approved):
/lgtm
/approve

PR #2058, opened 2026-06-24
Fix flaky GitHub-runner to Azure VM SSH that hung Linux SIG builds

mboersma:
Linux "Build Azure SIG Image" runs hung for ~2h and were cancelled, while Windows passed. The flaky leg is the GitHub-hosted runner to Azure VM SSH connection.

Two compounding issues:
- The Linux `ansible` provisioners used Packer's SSH proxy (`use_proxy` defaulted true), whose communicator `ssh_read_write_timeout` defaults to `0` (wait forever), so a dropped connection hung the whole build. Windows uses `use_proxy=false` and was unaffected.
- The provisioner's `ANSIBLE_SSH_ARGS` env var overrides `ansible.cfg`'s `ssh_args`, so keepalive/ControlPersist tuning there was a no-op.

Changes in `packer/azure/packer.json`: set `use_proxy: false` on both Linux ansible provisioners and add SSH keepalives plus ControlMaster/ControlPersist directly in `ANSIBLE_SSH_ARGS`; add `ssh_keep_alive_interval`/`ssh_read_write_timeout` on the builder. `ansible.cfg` enables pipelining, raises `timeout` to 120, and adds `retries`.

Verified with two green end-to-end runs including the Test stage (build plus smoke tests on a VM from the staging gallery image).

### DIFF
--- a/images/capi/ansible.cfg
+++ b/images/capi/ansible.cfg
@@ -19,5 +19,5 @@ inject_facts_as_vars = False
 timeout = 120
 
 [ssh_connection]
-pipelining = True
+pipelining = False
 retries = 5
--- a/images/capi/ansible.cfg
+++ b/images/capi/ansible.cfg
@@ -16,7 +16,8 @@
 remote_tmp = /tmp/.ansible
 display_skipped_hosts = False
 inject_facts_as_vars = False
-timeout = 60
+timeout = 120
 
 [ssh_connection]
-pipelining = False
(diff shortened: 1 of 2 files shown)


drew-viles (commented):
/lgtm
/approve

mboersma:
/override pull-azure-sigs

A transient network error w/ GitHub.com that failed the Flatcar build.

PR #2059, opened 2026-06-24
capi: wait for apt cleanup lock

MaxRink:
## What this fixes

The Debian sysprep cleanup can race with apt-daily/unattended-upgrades activity that is already running. A recent Azure SIG run for #2055 failed in `Remove apt package caches` with:

`E: Could not get lock /var/cache/apt/archives/lock. It is held by process ... (apt)`

The role already disables the apt-daily timers before cleanup, but stopping the timers does not stop service units that are already active.

## Changes

- keep disabling the apt-daily timers
- stop already-running `apt-daily.service` and `apt-daily-upgrade.service` before apt cleanup
- set `lock_timeout: 300` on the existing `ansible.builtin.apt` cleanup task

This intentionally keeps the fix module-native; it does not add shell wrappers arounnd apt.

## Validation

- `git diff --check upstream/main`
- YAML parsed with Python/PyYAML
- `ansible-lint images/capi/ansible/roles/sysprep/tasks/debian.yml` was also run locally; it reports existing upstream-main baseline violations in this file (apt-mark shell tasks and role variable naming), not new violations from this patch.

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -20,7 +20,7 @@
 - name: Get installed packages
   ansible.builtin.package_facts:
 
-- name: Disable apt-daily services
+- name: Disable apt-daily timers
   ansible.builtin.systemd:
     name: "{{ item }}"
     state: stopped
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm
/approve

Thanks @MaxRink, hopefully this fixes some of the Debian sysprep flakes.

PR #2062, opened 2026-06-25
capi: make node Ansible TMPDIR configurable

MaxRink:
## Change description

This adds an opt-in `node_ansible_tmpdir` variable for Linux node builds. When it is set, the node Ansible play exports it as `TMPDIR`; when it is empty, the play environment is unchanged.

This is useful for image variants with custom node or sysprep roles that need Ansible module/runtime temporary files outside the default temporary paths. The default remains `""`, so existing builds do not get a `TMPDIR` override.

## Related issues

- Fixes #

## Additional context

Split from downstream immutable-image validation work. This keeps sysprep cleanup intact by making the temporary path configurable instead of excluding Ansible payload paths from cleanup.

Validation:

- `git diff --check`
- JSON/YAML parse for `images/capi/packer/config/common.json`, `images/capi/packer/config/ansible-args.json`, and `images/capi/ansible/node.yml`
- `cd images/capi && ansible-playbook -i localhost, ansible/node.yml --syntax-check -e python_path=/tmp -e node_custom_roles_post_sysprep= -e node_ansible_tmpdir=/run`
- Ansible render check proving default environment omits `TMPDIR` and `node_ansible_tmpdir=/run` adds `TMPDIR=/run`
- `make validate-gce-ubuntu-2404 PACKER_FLAGS='-syntax-only'`

`ansible-lint images/capi/ansible/node.yml` was also run. It walks the included roles and reports the existing upstream baseline across those roles; no new `node.yml`-specific finding was reported.


mboersma (approved):
/lgtm
/approve

PR #2065, opened 2026-06-25
capi: add final image hygiene goss checks

MaxRink:
What this does
- Adds Linux-only Goss checks for final image hygiene after sysprep.
- Verifies SSH host private keys, machine-id, cloud-init instance identity, crash dumps, and oversized `/var/log` content are absent from captured images.
- Leaves Windows Goss rendering unchanged by guarding the checks on `OS != windows`.

Why
- These are generic final-image correctness checks and mirror cleanup that the sysprep role already performs.
- The checks are provider-neutral and contain no project-specific paths, cache plumbing, or credentials.
Validation
- `git diff --check`
- shell-semantics check proving the SSH-host-key and core-file commands fail when matching files exist
- `docker run --rm --platform linux/arm64/v8 -v "$PWD":/work -w /work/images/capi python:3.12-bookworm bash -lc ./scripts/ci-goss-populate.sh`

### DIFF
--- a/images/capi/packer/goss/goss-image-hygiene.yaml
+++ b/images/capi/packer/goss/goss-image-hygiene.yaml
@@ -1,6 +1,6 @@
 command:
 {{ if ne .Vars.OS "windows" }}
-  find /etc/ssh -xdev -type f -name 'ssh_host_*_key' -print:
+  sh -c '! find /etc/ssh -xdev -type f -name "ssh_host_*_key" -print -quit | grep -q .':
     exit-status: 0
     stdout: []
     stderr: []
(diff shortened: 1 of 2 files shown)


mboersma (commented):
Thanks for this! The intent is great and it's correctly guarded to Linux-only, follows our existing goss conventions, and runs after sysprep so the cleaned-up state is actually present. It won't break builds.

One substantive thing though: the SSH-host-key and core-dump checks don't actually assert anything as written. `find ... -print` exits 0 whether or not it finds matches, and `stdout: []` in goss means "no required stdout patterns", not "output must be empty". So both checks pass unconditionally and would never catch a leftover key or core file. You can see the same `stdout: []` pattern on `crictl ps` in goss-command.yaml, which always prints a header yet passes.

To make them effective, I'd have the command exit non-zero when something is found, something like:

```
sh -c '! find /etc/ssh -xdev -type f -name "ssh_host_*_key" | grep -q .'
```

The machine-id, cloud-init instance-id, and /var/log size checks all use exit-status meaningfully, so those are fine.

A couple of smaller things: the core-dump `find /` scans the whole filesystem with a 10s timeout, which could get slow or flaky on larger images, and once that check is effective the `*.core`/`core.*` patterns could match legitimate files. The 200MB /var/log threshold is also a bit arbitrary. None of those are blockers.

mboersma (approved):
/lgtm
/approve

PR #2066, opened 2026-06-25
docs: document post-sysprep node hook

MaxRink:
## What this PR does

Documents the existing `node_custom_roles_post_sysprep` customization variable in the CAPI image builder docs.

The variable is already wired through `packer/config/common.json` and `packer/config/ansible-args.json`, and the Linux `node.yml` playbook already runs it after the `sysprep` role. This PR only updates the user-facing customization table and example so users can discover the hook.

## Validation

- `git diff --check`


mboersma (approved):
/lgtm
/approve

PR #2067, opened 2026-06-25
capi: remove containerd wasm shim tarballs

MaxRink:
## What this PR does

Fixes the cleanup path for downloaded containerd wasm shim tarballs.

The role downloads and unpacks `/tmp/{{ item }}_wasm_shim.tar.gz`, but the cleanup task currently removes `/tmp/{{ item }}_wasm_shims.tar.gz` with an extra `s`. That leaves the downloaded tarballs in the image when wasm shims are enabled.

## Validation

- `rg -n "wasm_shim" images/capi/ansible/roles/containerd/tasks/main.yml`
- `git diff --check`


mboersma (approved):
/lgtm
/approve

PR #2072, opened 2026-06-26
capi: allow Ubuntu autoinstall mirror overrides

MaxRink:
## What this does

Ubuntu autoinstall for QEMU, KubeVirt, MAAS and Proxmox uses the existing `ubuntu_repo` and `ubuntu_security_repo` Packer variables, including the 24.04 immutable template. The 22.04 templates have no installer apt block; Ansible still applies the mirror settings to the built image.

The renderer receives the same variable arguments as Packer: template defaults, ordered var files, then explicit vars. It includes target-specific settings such as the MAAS ARM64 ports mirrors, rejects unsafe YAML scalar values and unsubstituted password placeholders, and never prints mirror credentials.

Each build/validation recipe renders into a private temporary copy of its HTTP directory and passes that copy to Packer. Targets sharing a template can run concurrently with `make -j`; there is no global `.NOTPARALLEL`. The copy is removed on exit unless `KEEP_RENDERED_AUTOINSTALL=1` is set. A leftover password stash is never restored over the current build password; legacy in-place cleanup removes the stash even when preserving rendered data. Separate simultaneous Make processes need separate checkouts because password/Packer template generation is shared.

## Validation

- Complete QEMU Python suite: 47 tests pass, including variable precedence, stale-password regression, separate HTTP copies and credential handling.
- Two QEMU validations under `make -j2` with a behavioral Packer stub overlap, receive different HTTP directories, and remove both directories on exit.
- `git diff --check` passes.

Full provider image builds were not rerun for this change.

### DIFF
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -67,8 +67,6 @@ GET_UBUNTU_DOTTED_SEMVER=$(strip \
 	$(_UBUNTU_SEMVER) \
 )
 
-PACKER_VAR_FILE_FLAGS=$(foreach f,$(abspath $(PACKER_VAR_FILES)),--extra-var-file "$(f)" )
-
 ## --------------------------------------
 ## Dependencies
 ## --------------------------------------
@@ -232,8 +230,6 @@ verify-containerd-service-template: ## Verifies the bundled containerd.service t
 	hack/update-containerd-service-template.py
 
 .PHONY: set-ssh-password
-set-ssh-password: export PACKER_FLAGS := $(PACKER_FLAGS)
-set-ssh-password: export PACKER_VAR_FILES := $(PACKER_VAR_FILES)
 set-ssh-password:
 	hack/set-ssh-password.sh
 
@@ -372,6 +368,25 @@ PACKER_WINDOWS_NODE_FLAGS := $(foreach f,$(abspath $(COMMON_WINDOWS_VAR_FILES)),
 				$(PACKER_FLAGS)
 PACKER_POWERVS_NODE_FLAGS := $(foreach f,$(abspath $(COMMON_POWERVS_VAR_FILES)),-var-file="$(f)" ) \
 				$(PACKER_FLAGS)
+
+# The Ubuntu autoinstall user-data embeds the apt mirrors held by the
+# ubuntu_repo and ubuntu_security_repo Packer variables, which a target var
+# file, PACKER_VAR_FILES or PACKER_FLAGS may override. It is therefore rendered
+# per build target, from that target's own variables, instead of once for every
+# template: aggregate targets such as build-qemu-all and parallel `make -j` runs
+# must not render one target's user-data with another target's mirrors.
+#
+# Each recipe collects its Packer arguments into a shell array and passes the
+# same array to the renderer and to Packer, so the renderer sees the same
+# -var/-var-file sequence, in the same order, and applies Packer's precedence:
+# template defaults, then -var-file in command line order, then -var.
+#
+# A mirror URL can carry credentials, so the renderer never prints variable
+# values and the substitution is undone when the recipe exits, whether Packer
+# succeeded or not: the user-data set-ssh-password wrote is put back, because a
+# target sharing the same autoinstall directory still needs it. Set
+# KEEP_RENDERED_AUTOINSTALL=1 to keep the rendered file instead.
+RENDER_UBUNTU_AUTOINSTALL := python3 packer/qemu/scripts/render_ubuntu_autoinstall.py
 
 ## --------------------------------------
 ## Platform and version combinations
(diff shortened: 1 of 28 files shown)


MaxRink:
/test pull-azure-sigs

I inspected the failed `pull-azure-sigs` run. The branch does not touch Azure CI or Azure packer files; the `ubuntu-2404` target failed immediately because Packer could not see an Azure CLI default subscription after `init-sig.sh` had already created the resource group/gallery. I opened #2086 to isolate Azure CLI auth per parallel SIG build, since the current CI script shares one Azure CLI config across all background builds.

images/capi/hack/set-ssh-password.sh
```diff
@@ -20,6 +20,89 @@ set -o pipefail
...
+            continue
+        if token.startswith("-var=") or token.startswith("--var="):
+            vars_from_flags.append(token.split("=", 1)[1])
+            index += 1
+            continue
+        index += 1
+    return var_files, vars_from_flags
+
+
+load_var_file(packer_dir / "config" / "common.json")
```
drew-viles, line 87: Could we make the renderer include the target-specific var file passed to Packer? For example, `build-maas-ubuntu-2404-arm64` uses `packer/maas/maas-ubuntu-2404-arm64.json`, where both Ubuntu repositories point to `ports.ubuntu.com`, but this helper only reads `common.json`, `PACKER_VAR_FILES`, and `PACKER_FLAGS`. I reproduced the resolver returning `us.archive.ubuntu.com` for that target, so the generated ARM64 autoinstall data uses different repositories from Packer and can send the installer to the wrong archive.

drew-viles (commented):
Minor note

mboersma (changes requested):
Could we rework the rendering so it is target-specific and follows the exact Packer variable precedence? The shared phony prerequisite currently runs only once, so aggregate or parallel builds can render every template using the wrong target’s mirrors. `PACKER_FLAGS` ordering can also disagree with Packer.

Please include the active 24.04 immutable template and avoid printing or retaining credential-bearing mirror URLs outside the selected build.

MaxRink:
Following up on the parallel/precedence review: current head 3e4f26363ff8e5f80e2627e39bf4a22c748ed36d passes the same per-target variable argument vector to the renderer and Packer, includes the immutable 24.04 template, and restores rendered files on exit. GNU Make 3.81-compatible .NOTPARALLEL serializes recipes even with make -j, including targets sharing a template directory. I corrected the stale PR description, which still described manual serialization. Separate simultaneous Make processes need separate checkouts; this does not claim cross-process locking.

images/capi/packer/qemu/scripts/render_ubuntu_autoinstall.py
```diff
@@ -120,71 +250,164 @@ def data_partition_config(values):
...
+
+def render_user_data(directory, values):
+    template = directory / "user-data.tmpl"
+    user_data = directory / "user-data"
+    stash = directory / "user-data.orig"
+
+    # A run killed between the write below and the restore in clean_user_data
+    # leaves the stash behind. Put it back first so the stash always holds
+    # set-ssh-password's output rather than a mirror-substituted file.
+    if stash.is_file():
```
drew-viles, line 301: This is where the stale password comes from. By the time we get here `set-ssh-password` has already rewritten `user-data` with a fresh password, so anything in the stash is from an older run. Could we delete the stash here instead of restoring it, and have `--clean` remove it even when `KEEP_RENDERED_AUTOINSTALL` is set?

To reproduce: run once with `KEEP_RENDERED_AUTOINSTALL=1`, then run `set-ssh-password` and the renderer again. `packer.json` has the new password, the served `user-data` has the old hash.

images/capi/packer/qemu/scripts/render_ubuntu_autoinstall_test.py
```diff
@@ -35,9 +61,291 @@ def load_renderer():
...
+        (directory / "user-data").write_text(SET_SSH_PASSWORD_OUTPUT, encoding="utf-8")
+
+        for _ in range(2):
+            self.renderer.render_user_data(directory, dict(MIRRORS))
+            rendered = (directory / "user-data").read_text(encoding="utf-8")
+            self.assertIn("passwd: $6$salt$hash", rendered)
+            self.assertIn("uri: http://us.archive.ubuntu.com/ubuntu", rendered)
+            self.renderer.clean_user_data(directory)
+
+    def test_render_recovers_a_stash_left_by_an_interrupted_run(self):
```
drew-viles, line 264: This test locks in the behaviour above. Once the renderer discards the stash, this should assert the opposite: a leftover stash is removed and the current `user-data` is what gets rendered.

images/capi/Makefile
```diff
@@ -504,6 +521,12 @@ MAAS_ARM64_VALIDATE_TARGETS 	:= $(addprefix validate-,$(MAAS_ARM64_BUILD_NAMES))
 SCALEWAY_BUILD_TARGETS		:= $(addprefix build-,$(SCALEWAY_BUILD_NAMES))
 SCALEWAY_VALIDATE_TARGETS	:= $(addprefix validate-,$(SCALEWAY_BUILD_NAMES))
 
+# These recipes render and restore shared autoinstall user-data files. Serialize
+# the invocation so parallel aggregate targets cannot serve another target's
+# mirrors. This is deliberately global because Make 3.81 cannot serialize leaf
+# recipes with a target-specific .NOTPARALLEL prerequisite list.
+.NOTPARALLEL:
```
drew-viles, line 528: This turns off `make -j` for every target in the file, not just the autoinstall ones. I checked with two AMI validate targets under `-j2`: they overlap on main and run one after the other here.

The cleanest way out is to not touch the shared file at all: render into a per-target temp copy of the http directory and pass `-var http_directory=<tmp>` in `packer_args`. The builders already read `http_directory` as a user var. That removes the need for `.NOTPARALLEL`, the stash and restore code, and the bug above in one go. If you would rather keep the current shape, please at least scope this to the targets that render.

drew-viles (commented):
Thanks, the earlier points are all fixed and I was able to prove each of them. Rendering is per target, the var-file and `-var` precedence matches what Packer does, the immutable template is covered, the arm64 maas target now gets `ports.ubuntu.com`, and no mirror values are printed. Good work on that.

Two things stop me approving this round. The first is a new bug that I reproduced, the second is a side effect on every other provider.

1. The renderer restores a leftover `user-data.orig` over the file that `set-ssh-password` has just generated. That stash is always stale, because `set-ssh-password` runs before every render and generates a new random password each time. So one debugging run with `KEEP_RENDERED_AUTOINSTALL=1`, or one killed run, and the next normal build serves the previous password. Packer then times out waiting for SSH with nothing in the log to say why. The PR describes this restore as a feature and a test asserts it, so it needs the test flipped as well.

2. `.NOTPARALLEL:` is global, so `make -j` is now serial for the whole Makefile. The AWS, vSphere and Huawei docs all tell users to build in parallel with `make -j`, and none of those providers touch autoinstall.

Details inline.

PR #2073, opened 2026-06-26
capi: use modules for sysprep log cleanup

MaxRink:
## Change description

Replace the sysprep role's shell-backed `/var/log` cleanup with Ansible module-backed discovery/removal where modules exist:

- use `ansible.builtin.find` to discover remaining `.log` files
- keep truncation as `ansible.builtin.command` with `argv` because ansible-core does not provide a truncate module that preserves existing file metadata
- use `ansible.builtin.find` plus `ansible.builtin.file` to remove logrotated files

This keeps the current non-Flatcar behavior while removing shell `find`, `xargs`, and `rm` usage from these cleanup tasks.

## Related issues

- Fixes #

## Additional context

Validation run locally:

- `ansible-playbook -i localhost, --syntax-check ansible/node.yml`
- `make validate-qemu-ubuntu-2404`
- `git -c core.fsmonitor=false diff --check`

I also ran `make lint` and `ansible-lint --project-dir . ansible/roles/sysprep/tasks/main.yml`; both are still blocked by pre-existing `var-naming[no-role-prefix]` debt in the repository/role. The new registers introduced here use the `sysprep_` prefix and were not among the reported violations.

This is intentionally not a revival of the closed temp-cleanup PRs (#2061/#2069); it only addresses the `/var/log` cleanup tasks.

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/main.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/main.yml
@@ -211,19 +211,48 @@
     - ansible_facts['os_family'] == "Flatcar"
     - not (defer_ssh_cleanup_to_packer | bool)
 
-- name: Truncate all remaining log files in /var/log
-  ansible.builtin.shell: |
-    set -o pipefail
-    find /var/log -type f -iname '*.log' | xargs truncate -s 0
-  args:
-    executable: /bin/bash
+- name: Find remaining log files in /var/log
+  ansible.builtin.find:
+    file_type: file
+    hidden: true
+    paths: /var/log
+    patterns: "(?i).*\\.log$"
+    recurse: true
+    use_regex: true
+  register: sysprep_log_files
+  when: ansible_facts['os_family'] != "Flatcar"
 
+- name: Truncate all remaining log files in /var/log
+  ansible.builtin.command:
+    argv:
+      - truncate
+      - "-s"
+      - "0"
+      - "{{ item.path }}"
+  changed_when: true
+  loop: "{{ sysprep_log_files.files | default([]) }}"
+  loop_control:
+    label: "{{ item.path }}"
+  when: ansible_facts['os_family'] != "Flatcar"
+
+- name: Find logrotated logs in /var/log
+  ansible.builtin.find:
+    file_type: file
+    hidden: true
+    paths: /var/log
+    patterns: ".*[0-9z]$"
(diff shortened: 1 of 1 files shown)


MaxRink:
Root-cause update for the failing Azure job: this PR code path is not what failed. The failed `pull-azure-sigs` run hit two pre-existing Azure provider issues:

- Ubuntu 22.04 failed in `providers : Install iptables persistence` because apt still held `/var/lib/dpkg/lock-frontend`.
- Ubuntu 24.04 failed later in `providers : Install netbase and nfs-common` for the same lock.
- Ubuntu 26.04/gen2 failed on the legacy `packages.microsoft.com` apt source key before Kubernetes package install.

I split fixes into separate PRs instead of changing this PR:

- #2077 adds module-native `lock_timeout` to the affected Azure provider `ansible.builtin.apt` tasks.
- #2078 extends the Microsoft legacy apt key fix to normal Azure provider builds where `debug_tools=false`.

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -211,19 +211,46 @@
     - ansible_facts['os_family'] == "Flatcar"
     - not (defer_ssh_cleanup_to_packer | bool)
 
+- name: Find remaining log files in /var/log
+  ansible.builtin.find:
+    file_type: file
+    paths: /var/log
+    patterns: "(?i).*\\.log$"
+    recurse: true
+    use_regex: true
```
mboersma, line 221: Maybe we should add `hidden: true` here for equivalence. `find` doesn't skip dotfiles and hidden directories, but Ansible's version does by default.

images/capi/ansible/roles/sysprep/tasks/main.yml
```diff
@@ -211,19 +211,46 @@
...
+    label: "{{ item.path }}"
+  when: ansible_facts['os_family'] != "Flatcar"
 
+- name: Find logrotated logs in /var/log
+  ansible.builtin.find:
+    file_type: file
+    paths: /var/log
+    patterns: ".*[0-9z]$"
+    recurse: true
+    use_regex: true
```
mboersma, line 245: Same comment as above: consider adding `find: true`.

mboersma (approved):
/lgtm
/approve

PR #2075, opened 2026-06-26
capi: use current Goss file contents key

MaxRink:

Goss v0.4 uses the `contents` field for file content assertions. The CAPI Goss template still emits `contains`, so these checks are not expressed with the current file schema.

This updates the generated file assertions in `goss-files.yaml` to use `contents` while keeping the existing input variable names unchanged.

Downstream T-CaaS image validation hit the same issue while updating Goss-based image hygiene checks, and this is the generic upstream part of that fix.

None

This is intentionally limited to the rendered Goss assertion key. The values under `goss-vars.yaml` still use `contains` as local template data and are read via `$vers.contains`.

#### Testing done:

```console
cd images/capi
make validate-qemu-ubuntu-2404
```

The validation succeeded locally. The checkout printed existing `fatal: No tags can describe ...` messages while deriving version metadata, but Packer validation completed with `The configuration is valid.`

### DIFF
--- a/images/capi/packer/goss/goss-files.yaml
+++ b/images/capi/packer/goss/goss-files.yaml
@@ -8,7 +8,7 @@ file:
   {{ $name }}:
     exists: {{ $vers.exists }}
     filetype: {{ $vers.filetype }}
-    contains: {{ range $vers.contains }}
+    contents: {{ range $vers.contains }}
     - {{.}}
   {{end}}
 {{end}}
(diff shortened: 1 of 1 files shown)


MaxRink:
Closing this after the required pull-goss-populate job caught the real compatibility issue: upstream currently pins Goss 0.3.23 in images/capi/packer/config/goss-args.json, and that version rejects the file resource key contents. The downstream T-CaaS fix depends on a newer Goss schema, so this is not a valid standalone upstream change against current main. If upstream later bumps Goss, this can be revisited together with that version change.

PR #2074, opened 2026-06-26
capi: add Ubuntu 24.04 QEMU cloud image target

MaxRink:
## Change description

Add an opt-in `qemu-ubuntu-2404-cloudimg` build/validation target that mirrors the existing `qemu-ubuntu-2204-cloudimg` pattern for Ubuntu 24.04.

The new target uses Canonical's Noble current cloud image and SHA256SUMS file:

- `https://cloud-images.ubuntu.com/noble/current/noble-server-cloudimg-amd64.img`
- `https://cloud-images.ubuntu.com/noble/current/SHA256SUMS`

No existing QEMU target behavior changes; this only adds a new build name and matching validate target.

- Is this change including a new Provider or a new OS? (y/n) n
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) n/a
- If adding a new provider, are you a representative of that provider? (y/n) n/a

## Related issues

- Fixes #

## Additional context

Validation run locally:

- `curl -fsSI https://cloud-images.ubuntu.com/noble/current/noble-server-cloudimg-amd64.img` returned 200
- `curl -fsSI https://cloud-images.ubuntu.com/noble/current/SHA256SUMS` returned 200
- `jq . images/capi/packer/qemu/qemu-ubuntu-2404-cloudimg.json >/dev/null`
- `make validate-qemu-ubuntu-2404-cloudimg`
- `git -c core.fsmonitor=false diff --check`

PR #2078, opened 2026-06-26
capi: trust Microsoft key in Azure provider builds

MaxRink:
## What this PR does

Extends the Microsoft legacy apt repository key handling to normal Azure provider builds.

The task is conditional: it only downloads the Microsoft key into apt's trusted key directory when an existing apt source under `/etc/apt/sources.list.d` references `packages.microsoft.com`.

## Why

`pull-azure-sigs` for #2074 failed in `sig-ubuntu-2604` during the Kubernetes apt cache refresh:

`NO_PUBKEY EB3E94ADBE1229CF` for `https://packages.microsoft.com/ubuntu/18.04/prod bionic`.

#2060 added the same class of fix under `azurecli.yml`, but that file is imported only when `debug_tools` is true. The failing SIG build runs with `debug_tools=false`, so normal Azure builds still do not trust the legacy Microsoft source before later apt cache refreshes.

## Validation

- `ansible-playbook --syntax-check ansible/node.yml -e provider=azure`
- `/opt/homebrew/bin/packer validate -syntax-only packer/azure/packer.json`
- `git diff --check`

Note: targeted `ansible-lint --project-dir . ansible/roles/providers/tasks/azure.yml` is currently blocked by pre-existing var-naming findings in `ansible/roles/providers/tasks/azurecli.yml` (`ubuntu_version`, `host_arch`, `azure_cli_codename`), unrelated to this diff.

### DIFF
--- a/images/capi/ansible/roles/setup/tasks/debian.yml
+++ b/images/capi/ansible/roles/setup/tasks/debian.yml
@@ -42,6 +42,24 @@
     mode: "0644"
   when: http_proxy is defined or https_proxy is defined
 
+- name: Trust Microsoft key for legacy Microsoft apt repositories
+  block:
+    - name: Check for legacy Microsoft apt source
+      ansible.builtin.find:
+        paths: /etc/apt/sources.list.d
+        patterns:
+          - "*.list"
+          - "*.sources"
+        contains: ".*packages\\.microsoft\\.com.*"
+      register: setup_ms_legacy_source
+
+    - name: Download Microsoft Package Repository Key for legacy sources
+      ansible.builtin.get_url:
+        url: https://packages.microsoft.com/keys/microsoft.asc
+        dest: /etc/apt/trusted.gpg.d/microsoft.asc
+        mode: "0644"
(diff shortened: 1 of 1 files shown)


MaxRink:
Updated in c4dd15464 after checking the `pull-azure-sigs` failure.

Root cause: `sig-ubuntu-2604` failed during the setup role's first apt cache refresh/dist-upgrade, before `providers/tasks/azure.yml` runs. Installing the Microsoft key in the Azure provider role was therefore too late.

Fix: moved the module-native `ansible.builtin.find` + `ansible.builtin.get_url` key handling into `roles/setup/tasks/debian.yml` before the first apt cache update. No shell workaround was added.

Local checks:
- `git diff --check`
- YAML parse/assertion for `images/capi/ansible/roles/setup/tasks/debian.yml`

`yamllint images/capi/ansible/roles/setup/tasks/debian.yml` still reports pre-existing line-length issues at lines 20, 21, and 36; the new block does not add those findings.

mboersma (approved):
/lgtm
/approve

PR #2080, opened 2026-06-26
capi: add optional QEMU boot smoke helper

MaxRink:
## Change description

Add an opt-in local QEMU boot smoke helper for already-built CAPI QEMU images.

The helper boots a local image through a temporary qcow2 copy-on-write overlay, optionally attaches a generated NoCloud seed ISO, forwards guest SSH to localhost, and verifies a configurable SSH command. The source image is not modified.

The Makefile target is intentionally manual-only and is not wired into required CI.

## Related issues

None.

## Validation
- `bash -n images/capi/hack/qemu-boot-smoke.sh`
- `shellcheck -x images/capi/hack/qemu-boot-smoke.sh`
- `git diff --check`
- `make -C images/capi -n test-qemu-boot-smoke QEMU_BOOT_SMOKE_IMAGE=/tmp/example.qcow2`
- `make -C images/capi help | rg test-qemu-boot-smoke`

A real QEMU boot was not run in this local environment because `qemu-system-x86_64` and `qemu-img` are not installed here.

### DIFF
--- a/images/capi/hack/qemu-boot-smoke.sh
+++ b/images/capi/hack/qemu-boot-smoke.sh
@@ -1,0 +1,315 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+usage() {
+  cat <<'EOF' >&2
+usage: qemu-boot-smoke.sh IMAGE_OR_OUTPUT_DIR [-- QEMU_ARGS...]
+
+Boot a local QEMU image with a copy-on-write overlay and verify that the guest
+accepts SSH on a host-forwarded port. By default, a temporary NoCloud seed ISO
+creates the SSH user, so the source image is not modified.
+
+Environment:
+  QEMU_BINARY              QEMU binary to run. Default: qemu-system-x86_64
+  QEMU_IMG                 qemu-img binary to run. Default: qemu-img
+  QEMU_IMAGE_FORMAT        Backing image format. Default: detected by qemu-img
+  QEMU_ACCELERATOR         QEMU accelerator. Default: kvm on Linux with /dev/kvm,
+                           hvf on macOS, otherwise tcg
+  QEMU_MACHINE             QEMU machine type. Default: pc
+  QEMU_CPUS                vCPU count. Default: 2
+  QEMU_MEMORY              Guest memory. Default: 2048
+  QEMU_SSH_PORT            Host port forwarded to guest port 22. Default: 2222
+  QEMU_SSH_TIMEOUT         Seconds to wait for SSH. Default: 600
+  QEMU_SSH_INTERVAL        Seconds between SSH checks. Default: 5
+  QEMU_SSH_USER            SSH user to verify. Default: capi
+  QEMU_SSH_PRIVATE_KEY     SSH private key. Default: cloudinit/id_rsa.capi
+  QEMU_SSH_PUBLIC_KEY      SSH public key. Default: cloudinit/id_rsa.capi.pub
+  QEMU_SMOKE_COMMAND       Command to run over SSH. Default: true
+  QEMU_SEED                cloud-init or none. Default: cloud-init
+EOF
+}
+
+if [[ $# -lt 1 ]]; then
+  usage
+  exit 1
+fi
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+capi_dir="$(cd "${script_dir}/.." && pwd -P)"
+
+image_arg="${1}"
+shift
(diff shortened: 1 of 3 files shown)


images/capi/hack/qemu-boot-smoke.sh
```diff
@@ -0,0 +1,315 @@
...
+  case "$(uname -s)" in
+  Linux)
+    if [[ -r /dev/kvm && -w /dev/kvm ]]; then
+      echo kvm
+    else
+      echo tcg
+    fi
+    ;;
+  Darwin)
+    echo hvf
```
drew-viles: Could we make this architecture-aware and fall back to `tcg` when Darwin is arm64 but the selected/default binary is `qemu-system-x86_64`? On an Apple Silicon Mac the current defaults select that binary with HVF; I reproduced the combination failing immediately with `-accel hvf: invalid accelerator hvf`. As a result, the documented macOS default currently requires an undocumented `QEMU_ACCELERATOR=tcg` override.

images/capi/hack/qemu-boot-smoke.sh
```diff
@@ -0,0 +1,315 @@
...
+  public_key="$(ssh-keygen -y -f "${ssh_key}")"
+fi
+
+backing_format="${QEMU_IMAGE_FORMAT:-$(detect_image_format "${image}")}"
+runtime_disk="${tmp_dir}/disk.qcow2"
+"${QEMU_IMG}" create -f qcow2 -F "${backing_format}" -b "${image}" "${runtime_disk}" >/dev/null
+
+seed_args=()
+case "${QEMU_SEED}" in
+cloud-init)
```
drew-viles, line 313: Could we either support an Ignition seed for Flatcar or fail early and document this target as unsupported? `qemu-flatcar` is one of the repository’s QEMU targets, but it uses Ignition and cloud-init is installed only on Debian/Red Hat builds. The build also removes `builder` before shutdown, so neither the default cloud-init/`capi` path nor the documented `QEMU_SEED=none`/`builder` example can authenticate to that artifact. Currently this consumes the full SSH timeout even when the image booted successfully.

drew-viles (commented):
Minor notes

mboersma (approved):
/lgtm

mboersma (approved):
/approve

images/capi/hack/qemu-boot-smoke.sh
```diff
@@ -0,0 +1,377 @@
...
+
+  dir="$(dirname "${path}")"
+  base="$(basename "${path}")"
+  echo "$(cd "${dir}" && pwd -P)/${base}"
+}
+
+is_flatcar_image() {
+  local image="${1}"
+
+  [[ "${image,,}" == *flatcar* ]]
```
drew-viles: Could we avoid the Bash 4-only lowercasing here? macOS still ships Bash 3.2, and `${image,,}` fails with `bad substitution`. I reproduced this on Apple Silicon with `/bin/bash 3.2.57`, so the helper now exits for every image before it reaches the accelerator selection. A Bash 3-compatible `case` match would keep the documented macOS path working.

images/capi/hack/qemu-boot-smoke.sh
```diff
@@ -0,0 +1,377 @@
...
+qemu_binary_arch() {
+  case "$(basename "${1}")" in
+  qemu-system-x86_64)
+    echo x86_64
+    ;;
+  qemu-system-aarch64)
+    echo aarch64
+    ;;
+  *)
+    echo ""
```
drew-viles, line 164: Could we fall back to `tcg` when `qemu_binary_arch` cannot identify the target? At the moment only x86_64 and aarch64 are recognised; valid binaries such as `qemu-system-riscv64`, or a wrapped/renamed binary, return an empty value which is then treated as safe for HVF/KVM. That contradicts the help text saying hardware acceleration is selected only when the target and host architectures match.

images/capi/hack/qemu-boot-smoke.sh
```diff
@@ -0,0 +1,377 @@
...
+    kill -9 "${pid}" >/dev/null 2>&1 || true
+  fi
+}
+
+require_command "${QEMU_BINARY}"
+require_command "${QEMU_IMG}"
+require_command ssh
+
+image="$(abs_path "$(resolve_image "${image_arg}")")"
+if is_flatcar_image "${image}"; then
```
drew-viles: Could we avoid inferring Flatcar solely from the absolute path? A renamed Flatcar artifact bypasses this guard and still waits through the full SSH timeout, while an unrelated image under a directory containing `flatcar` is rejected. If reliable image detection is not practical, an explicit input at the target boundary may be less surprising.

images/capi/packer/qemu/README.md
```diff
@@ -140,3 +140,40 @@ should:
 For a local pre-CAPI boot smoke, boot the produced artifact with QEMU and a
 temporary NoCloud seed. The smoke should use the same checks as above, plus a
 reboot/persistence check, before the image is promoted to provider CAPI tests.
+
+## Boot smoke test
+
+After building a QEMU image, you can run an optional local boot smoke test
+against the output image:
```
drew-viles, line 147: Could we also call out here that `qemu-flatcar` is unsupported? This currently reads as applying after building any QEMU image, and `qemu-flatcar` is a first-class build target. The caveat only appears in the script usage text, so it is easy to miss before invoking the Make target.

images/capi/packer/qemu/README.md
```diff
@@ -140,3 +140,44 @@ should:
 For a local pre-CAPI boot smoke, boot the produced artifact with QEMU and a
 temporary NoCloud seed. The smoke should use the same checks as above, plus a
 reboot/persistence check, before the image is promoted to provider CAPI tests.
```
mboersma, line 142: This says the smoke includes reboot/persistence and immutable-image checks, but this helper only verifies SSH and one command. Could we change this to say it is simply an SSH-reachability helper?

mboersma (approved):
/lgtm

PR #2081, opened 2026-06-26
capi: document Ansible Galaxy configuration and forward it in GCE CI

MaxRink:
## Summary

ansible-core already exposes every Galaxy knob an earlier revision of this PR wrapped in custom variables, so this documents the native settings instead of adding a wrapper.

## Changes

A new "Ansible Galaxy configuration" section in the CAPI book covers `images/capi/ansible.cfg`, the `ANSIBLE_GALAXY_*` and `ANSIBLE_COLLECTIONS_PATH` variables, a private Automation Hub example via `server_list`, and the Red Hat SSO variant, each linked to the official ansible-core configuration reference and to the collections guide section on configuring the ansible-galaxy client. The GCE CI scripts forward every `ANSIBLE_*` variable across `su - packer` with `--whitelist-environment`, built from the environment because per-server credentials are named after `ANSIBLE_GALAXY_SERVER_LIST` entries. `GCP_PROJECT_ID`, `GOOGLE_APPLICATION_CREDENTIALS`, `PACKER_VAR_FILES` and `PACKER_FLAGS` move from inline assignments in the `su` command string to exported variables on the same whitelist; the `PATH` handling in the command string is unchanged.

## Validation

The su block simulated under bash 3.2 with `set -euo pipefail`, with zero and multiple `ANSIBLE_*` variables, and with values containing spaces and `$`; shellcheck output unchanged from main; every documented setting checked against `ansible-config list` for ansible-core 2.21.3 and against the linked configuration reference, with each anchor verified to exist on the fetched pages; a real collection install into a scratch `ANSIBLE_COLLECTIONS_PATH` found by `ansible-galaxy collection list` at runtime; the book built with the mdbook version pinned in `docs/book/build.sh`.


images/capi/hack/utils.sh
```diff
@@ -119,6 +119,34 @@ pip3_install() {
...
+    galaxy_args+=(--collections-path "${ANSIBLE_GALAXY_COLLECTIONS_PATH}")
+  fi
+  if [[ "${ANSIBLE_GALAXY_NO_CACHE:-false}" == "true" ]]; then
+    galaxy_args+=(--no-cache)
+  fi
+  if [[ "${ANSIBLE_GALAXY_OFFLINE:-false}" == "true" ]]; then
+    galaxy_args+=(--offline)
+  fi
+
+  ansible-galaxy collection install "${galaxy_args[@]}" "$@"
```
mboersma: On macOS the system bash is still 3.2, and there `"${galaxy_args[@]}"` on an empty array under `set -o nounset` fails with `unbound variable`. Since both callers set `set -o nounset` and the default path leaves `galaxy_args` empty (no `ANSIBLE_GALAXY_*` vars), `make deps` exits before `ansible-galaxy` runs. Linux CI uses bash 4.4+ so it stays green and misses this.

We'd rather not require bash 4, since we don't have much control over the build environment. Could you use the empty-array-safe expansion instead?

```bash
ansible-galaxy collection install ${galaxy_args[@]+"${galaxy_args[@]}"} "$@"
```

I confirmed the current form errors under `/bin/bash` 3.2 with `set -o nounset`, and this guard fixes it while staying safe on newer bash.

drew-viles (commented):
Seems logical to me!

/ok-to-test

images/capi/hack/utils.sh
```diff
@@ -119,6 +119,34 @@ pip3_install() {
...
 }
 
+ansible_galaxy_collection_install() {
+  local -a galaxy_args=()
+
+  if [[ -n "${ANSIBLE_GALAXY_SERVER:-}" ]]; then
+    galaxy_args+=(--server "${ANSIBLE_GALAXY_SERVER}")
+  fi
+  if [[ -n "${ANSIBLE_GALAXY_TOKEN:-}" ]]; then
+    galaxy_args+=(--token "${ANSIBLE_GALAXY_TOKEN}")
```
drew-viles: Could we ensure the Galaxy token is never expanded while xtrace is enabled? Both callers enable `set -o xtrace` when `DEBUG` is set, so `DEBUG=1 ANSIBLE_GALAXY_TOKEN=... make deps` writes the private token into the build log, both while evaluating this condition and in the final `ansible-galaxy --token` command. I reproduced this through the actual `ensure-s3.sh` path. Could we suspend tracing before all token handling and restore its previous state afterward, or use authentication that does not put the token on the command line?

images/capi/hack/utils.sh
```diff
@@ -119,6 +119,34 @@ pip3_install() {
...
+  fi
+  if [[ "${ANSIBLE_GALAXY_IGNORE_CERTS:-false}" == "true" ]]; then
+    galaxy_args+=(--ignore-certs)
+  fi
+  if [[ -n "${ANSIBLE_GALAXY_TIMEOUT:-}" ]]; then
+    galaxy_args+=(--timeout "${ANSIBLE_GALAXY_TIMEOUT}")
+  fi
+  if [[ -n "${ANSIBLE_GALAXY_COLLECTIONS_PATH:-}" ]]; then
+    galaxy_args+=(--collections-path "${ANSIBLE_GALAXY_COLLECTIONS_PATH}")
```
drew-viles, line 160: How is this installation path expected to reach the later Ansible provisioner runs? `ANSIBLE_GALAXY_COLLECTIONS_PATH` is local to this helper, while Ansible discovers collections through `ANSIBLE_COLLECTIONS_PATH`, and most provider templates do not pass the QEMU/OpenStack-only `ansible_collections_path` variable. I reproduced this with `community.general` installed only in a temporary custom directory: it was not found with only this variable set, but was found immediately with `ANSIBLE_COLLECTIONS_PATH`. Could we use the standard variable for both installation and runtime discovery, or wire the custom path into every provisioner?

drew-viles (commented):
Minor notes actually.

mboersma (changes requested):
Could we address a few propagation and security issues before merging?

Galaxy credentials may leak through tracing/process arguments, custom install paths aren’t always added to Ansible’s runtime search path, `make` can evaluate `$()` in environment-provided paths, and the GCE `su - packer` flow drops the new Galaxy variables.

MaxRink:
Thanks for the detailed feedback. I addressed all four propagation/security points in commit `bba786d7912d1d46c91bd80d7dea9d6ba59071bc`:

1. **Galaxy credential handling** — `images/capi/hack/utils.sh` now keeps xtrace disabled for the complete credential lifecycle, writes `ANSIBLE_GALAXY_TOKEN` through stdin to a 0600 temporary token file using a race-safe `mktemp` path, removes the token from child environments, never passes it as `--token`, and cleans up the file.
2. **Runtime collection discovery** — `images/capi/hack/utils.sh` exports `ANSIBLE_COLLECTIONS_PATH` with the custom Galaxy path, and `images/capi/Makefile` reliably prepends it to any existing runtime path.
3. **Literal Make paths** — the Make assignments now use raw `$(value ...)` values with a separate `override`/`export`, so environment-provided `$()` sequences remain literal and cannot be evaluated.
4. **GCE `su - packer` propagation** — both existing GCE scripts retain the complete Galaxy variable whitelist, including `ANSIBLE_COLLECTIONS_PATH` and `ANSIBLE_GALAXY_TOKEN_PATH`; focused tests now assert this for both flows.

Changed files:
- `images/capi/hack/utils.sh`
- `images/capi/Makefile`
- `images/capi/hack/utils_test.py`
- `docs/book/src/capi/capi.md`

Validation:
- `make --no-print-directory -C images/capi test-galaxy`
- `python3 -m unittest discover -s images/capi/packer/qemu/scripts -p '*_test.py'`
- `shellcheck images/capi/hack/utils.sh`
- `ruff check --ignore EXE001 images/capi/hack/utils_test.py`
- `ruff format --check images/capi/hack/utils_test.py`
- Bash syntax checks and `git diff --check`

The broader shellcheck invocation still reports pre-existing warnings in unchanged CI/helper code (`SC1091`, `SC2155`, `SC2086`). Please re-review when convenient.

MaxRink:
Follow-up on EasyCLA: the failure was caused by the `Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>` trailer in the just-added commit. EasyCLA explicitly reported `@Copilot` as missing CLA authorization; PR #2100's passing commits contain only the authorized MaxRink author/committer metadata.

I amended only that commit message (no code/tree changes), preserving Maximilian Rink as author and committer. The final commit is `e30d0cdf3ffcbd4c7b514c156c33259f59a7dc55`, and EasyCLA now reports **pass**. The prior implementation and review reply remain unchanged.

images/capi/hack/utils.sh
```diff
@@ -119,6 +119,94 @@ pip3_install() {
...
+  local -a galaxy_args=()
+  local xtrace_was_on=false
+  local galaxy_token_path=
+  [[ $- == *x* ]] && xtrace_was_on=true
+  # Galaxy credentials must not be exposed in xtrace output or process
+  # arguments. Keep tracing disabled until all credential handling is done.
+  set +o xtrace
+
+  if [[ -n "${ANSIBLE_GALAXY_SERVER:-}" ]]; then
+    galaxy_args+=(--server "${ANSIBLE_GALAXY_SERVER}")
```
drew-viles: Passing `--server` makes ansible-galaxy ignore `ANSIBLE_GALAXY_SERVER_LIST`, which is the only way to configure Automation Hub auth (`auth_url` / `client_id`). That regresses the private hub case this PR is aimed at.

images/capi/hack/utils.sh
```diff
@@ -119,6 +119,94 @@ pip3_install() {
...
+      esac
+    else
+      export ANSIBLE_COLLECTIONS_PATH="${ANSIBLE_GALAXY_COLLECTIONS_PATH}"
+    fi
+  fi
+  if [[ "${ANSIBLE_GALAXY_NO_CACHE:-false}" == "true" ]]; then
+    galaxy_args+=(--no-cache)
+  fi
+  if [[ "${ANSIBLE_GALAXY_OFFLINE:-false}" == "true" ]]; then
+    galaxy_args+=(--offline)
```
drew-viles: `--offline` only works with tarball paths or a requirements file; both callers pass collection names, so this option can never succeed here.

images/capi/hack/utils.sh
```diff
@@ -119,6 +119,94 @@ pip3_install() {
...
+  if [[ -n "${ANSIBLE_GALAXY_SERVER:-}" ]]; then
+    galaxy_args+=(--server "${ANSIBLE_GALAXY_SERVER}")
+  fi
+
+  if [[ -n "${ANSIBLE_GALAXY_TOKEN:-}" ]]; then
+    if ! mkdir -p "${HOME}/.ansible"; then
+      [[ "${xtrace_was_on}" == true ]] && set -o xtrace
+      return 1
+    fi
+    if ! galaxy_token_path="$(umask 077 && mktemp "${HOME}/.ansible/image-builder-galaxy-token-XXXXXX")"; then
```
drew-viles: The token file is only removed on the two inline paths, so a Ctrl-C or timeout during install leaves the plaintext token on disk. Could we add a trap?

images/capi/scripts/ci-gce.sh
```diff
@@ -112,7 +112,11 @@ fi
...
 chown -R packer:packer /home/prow/go/src/sigs.k8s.io/image-builder
 # use the packer user to run the build
-su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS PACKER_VAR_FILES=scripts/ci-disable-goss-inspect.json make deps-gce build-gce-all'" || test_status="${?}"
+galaxy_env_whitelist="ANSIBLE_GALAXY_SERVER,ANSIBLE_GALAXY_TOKEN,ANSIBLE_GALAXY_TOKEN_PATH,ANSIBLE_GALAXY_IGNORE_CERTS,ANSIBLE_GALAXY_TIMEOUT,ANSIBLE_GALAXY_COLLECTIONS_PATH,ANSIBLE_GALAXY_NO_CACHE,ANSIBLE_GALAXY_OFFLINE,ANSIBLE_COLLECTIONS_PATH"
+export GCP_PROJECT_ID="${GCP_PROJECT}"
+export PACKER_VAR_FILES="scripts/ci-disable-goss-inspect.json"
+su --whitelist-environment="${galaxy_env_whitelist},GCP_PROJECT_ID,GOOGLE_APPLICATION_CREDENTIALS,PACKER_VAR_FILES" \
+  - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=\$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin make deps-gce build-gce-all'" || test_status="${?}"
```
drew-viles: `PATH=$PATH` became `PATH=\$PATH`, so root's PATH (including Go from `ensure-go.sh`) no longer reaches the packer user; `su` resets PATH regardless of the whitelist. Could we restore the outer expansion?

images/capi/scripts/ci-gce-nightly.sh
```diff
@@ -39,12 +39,17 @@ gcloud auth list
...
 
 # Build image for each available config
 for f in packer/gce/ci/nightly/*.json
 do
   # using PACKER_FLAGS=-force to overwrite the previous image and keep the same name
-  su - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin GCP_PROJECT_ID=$GCP_PROJECT PACKER_VAR_FILES=${f} PACKER_FLAGS=-force make deps-gce build-gce-all'"
+  export PACKER_VAR_FILES="${f}"
+  export PACKER_FLAGS=-force
+  su --whitelist-environment="${galaxy_env_whitelist},GCP_PROJECT_ID,PACKER_VAR_FILES,PACKER_FLAGS" \
+    - packer -c "bash -c 'cd /home/prow/go/src/sigs.k8s.io/image-builder/images/capi && PATH=\$PATH:~packer/.local/bin:/home/prow/go/src/sigs.k8s.io/image-builder/images/capi/.local/bin make deps-gce build-gce-all'"
```
drew-viles: Same PATH escaping change as `ci-gce.sh`.

images/capi/Makefile
```diff
@@ -624,6 +635,10 @@ $(QEMU_VALIDATE_TARGETS): deps-qemu set-ssh-password
 test-qemu-immutable: ## Runs immutable QEMU helper unit tests
 	python3 -m unittest discover -s packer/qemu/scripts -p '*_test.py'
 
+.PHONY: test-galaxy
+test-galaxy: ## Runs Galaxy authentication and collection path tests
```
drew-viles: Nothing runs `test-galaxy`. Could we hang it off a validate target the way `test-qemu-immutable` is?

drew-viles (commented):
Thanks for the follow-ups, the token handling and bash 3.2 fixes check out. Before going further though: checking against ansible-core 2.21, five of the seven variables already exist natively (`ANSIBLE_GALAXY_SERVER`, `ANSIBLE_GALAXY_IGNORE`, `ANSIBLE_GALAXY_SERVER_TIMEOUT`, `ANSIBLE_GALAXY_TOKEN_PATH`, `ANSIBLE_COLLECTIONS_PATH`), and `ANSIBLE_COLLECTIONS_PATH` already sets both the install and runtime path. Could we document those plus `images/capi/ansible.cfg` instead of wrapping them? Specific issues inline.

docs/book/src/capi/capi.md
```diff
@@ -12,6 +12,65 @@ The Image Builder can be used to build images intended for use with Kubernetes [
...
+  URL to `ansible-galaxy --server` ignores `server_list` too, and drops the
+  per-server credentials with it; passing the name of a `server_list` entry
+  selects that entry with its configuration.
+* `ANSIBLE_COLLECTIONS_PATH` is both the install destination (its first entry)
+  and the runtime search path, so a custom path needs no further wiring.
+* Keep tokens in `ANSIBLE_GALAXY_SERVER_<NAME>_TOKEN` or in the file named by
+  `ANSIBLE_GALAXY_TOKEN_PATH`. `--token` (`--api-key`) would expose them in the
+  process arguments, and it is ignored for servers defined in `server_list`.
+* ansible-core also accepts `ANSIBLE_GALAXY_SERVER_<NAME>_CLIENT_SECRET` for a
+  Keycloak service account, although the guide above does not list it.
```
drew-viles: `client_secret` is not in `GALAXY_SERVER_DEF` on 2.18.18 (`lib/ansible/config/manager.py`), it arrived in 2.19 with Keycloak service account support. On 2.18 the key is silently ignored and auth fails with no hint. Could we either drop this bullet or qualify it as ansible-core 2.19 or later, noting the pinned version is older?

docs/book/src/capi/capi.md
```diff
@@ -12,6 +12,65 @@ The Image Builder can be used to build images intended for use with Kubernetes [
...
+A few things worth knowing:
+
+* Setting `server_list` makes ansible-core ignore `[galaxy] server`. Passing a
+  URL to `ansible-galaxy --server` ignores `server_list` too, and drops the
+  per-server credentials with it; passing the name of a `server_list` entry
+  selects that entry with its configuration.
+* `ANSIBLE_COLLECTIONS_PATH` is both the install destination (its first entry)
+  and the runtime search path, so a custom path needs no further wiring.
+* Keep tokens in `ANSIBLE_GALAXY_SERVER_<NAME>_TOKEN` or in the file named by
+  `ANSIBLE_GALAXY_TOKEN_PATH`. `--token` (`--api-key`) would expose them in the
```
drew-viles: These are not interchangeable. `GALAXY_TOKEN_PATH` is only read for the default server; a `server_list` entry with no `token` key gets `NoTokenSentinel` (`cli/galaxy.py` L662, `token.py` L42 on 2.18) and sends unauthenticated requests. Since the section steers readers to `server_list`, could this say the token file applies only to the default server and `server_list` entries must carry their own `token`?

docs/book/src/capi/capi.md
```diff
@@ -12,6 +12,65 @@ The Image Builder can be used to build images intended for use with Kubernetes [
...
+| [Per-server keys](https://docs.ansible.com/ansible/latest/collections_guide/collections_installing.html#configuring-the-ansible-galaxy-client): `url`, `token`, `username`, `password`, `auth_url`, `client_id`, `validate_certs`, `timeout` | `[galaxy_server.<name>]` | `ANSIBLE_GALAXY_SERVER_<NAME>_URL`, `ANSIBLE_GALAXY_SERVER_<NAME>_TOKEN`, ... |
+| [`GALAXY_IGNORE_CERTS`](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#galaxy-ignore-certs): do not validate TLS certificates | `[galaxy] ignore_certs` | `ANSIBLE_GALAXY_IGNORE` |
+| [`GALAXY_SERVER_TIMEOUT`](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#galaxy-server-timeout): default API timeout in seconds | `[galaxy] server_timeout` | `ANSIBLE_GALAXY_SERVER_TIMEOUT` |
+| [`GALAXY_TOKEN_PATH`](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#galaxy-token-path): local galaxy access token file | `[galaxy] token_path` | `ANSIBLE_GALAXY_TOKEN_PATH` |
```
drew-viles: Same point as below: worth a note that this only applies to the default `server`, not to `server_list` entries.

drew-viles (commented):
one more small revision needed on this one: the scripts check out and all the earlier items are resolved, but two of the documented settings do not behave as described on the ansible-core this repo actually installs. `hack/utils.sh` pins 2.18.18 and `ensure-ansible.sh` installs exactly that, so the docs need to hold for 2.18, not 2.21. Details inline.

docs/book/src/capi/capi.md
MaxRink: Addressed in 61713e7644cc54cba8317708ac1b71bfe8f3a737: the docs now explicitly qualify CLIENT_SECRET as ansible-core 2.19 or later and note that the pinned 2.18 version does not support it.

docs/book/src/capi/capi.md
MaxRink: Addressed in 61713e76: the docs now distinguish the default-server token file from server_list authentication. Named server_list entries must set their own per-server TOKEN; they do not inherit the default token file.

docs/book/src/capi/capi.md
MaxRink: Addressed in 61713e76: the ANSIBLE_GALAXY_TOKEN_PATH documentation now explicitly limits it to the default server, not server_list entries.

mboersma (approved):
/lgtm
/approve

PR #2082, opened 2026-06-26
capi: avoid duplicate phased custom roles

MaxRink:
## What this PR does

Filters roles assigned to phase-specific hooks out of the generic post-node custom-role loop.

Linux now excludes roles from `custom_role_names` when they are already assigned to:

- `node_custom_roles_pre`
- `node_custom_roles_post_sysprep`

Windows now excludes roles from `custom_role_names` when they are already assigned to:

- `node_custom_roles_pre`

Roles explicitly listed in `node_custom_roles_post` still run in the post-node/post-configuration phase.

## Why it is needed

Downstream validation found that some build flows need a role listed in `custom_role_names` for custom-role wiring, while also assigning that same role to a phase-specific hook. With the current node playbooks, that role is then executed again by the generic post-node loop.

That is surprising for hook users and can run setup or validation roles in the wrong phase. The new filtering keeps the existing `custom_role_names` behavior for normal post-node roles, while avoiding duplicate execution for roles that have a more specific phase assignment.

## Validation

- `git diff --check`
- `ansible-playbook --syntax-check images/capi/ansible/node.yml -i localhost,`
- `ansible-playbook --syntax-check images/capi/ansible/windows/node_windows.yml -i localhost,`

### DIFF
--- a/images/capi/ansible/node.yml
+++ b/images/capi/ansible/node.yml
@@ -54,7 +54,15 @@
     - name: Include post-node custom roles
       ansible.builtin.include_role:
         name: "{{ role }}"
-      loop: "{{ custom_role_names.split() + node_custom_roles_post.split() }}"
+      loop: >-
+        {{
+          (
+            custom_role_names.split()
+            | reject('in', node_custom_roles_pre.split() + node_custom_roles_post_sysprep.split())
+            | list
+          )
+          + node_custom_roles_post.split()
+        }}
       loop_control:
         loop_var: role
       when: custom_role_names != "" or node_custom_roles_post != ""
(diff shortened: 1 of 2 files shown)


MaxRink:
The current `pull-azure-sigs` failure looks unrelated to this PR's hook filtering change.

Failed target: `sig-ubuntu-2604-gen2`.

The per-target log fails during apt cache update because the Microsoft Ubuntu 18.04 prod repository key is missing:

`NO_PUBKEY EB3E94ADBE1229CF`

Log: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_image-builder/2082/pull-azure-sigs/2070511038284959744/artifacts/azure-sigs/ubuntu-2604-gen2.log

I am not retesting this until the Microsoft repo key handling is fixed separately (for example by #2078 or equivalent).

mboersma (approved):
/lgtm
/approve

PR #2084, opened 2026-06-26
capi: resolve Kubernetes Debian package revisions

MaxRink:
## What this does

- Resolves the Kubernetes Debian package revision from `pkgs.k8s.io` in the `configure-k8s-version` composite action instead of assuming every release uses `${version}-1.1`.
- Keeps the existing RPM behavior unchanged.
- Adds an action-local Python helper so the workflow fails early if kubeadm, kubectl, and kubelet do not share an available Debian package version for the requested Kubernetes patch release.

## Why

Kubernetes `1.36.2` is published in the Debian repo as `1.36.2-2.1`, not `1.36.2-1.1`. The current action would generate a bad `kubernetes_deb_version` for that input.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -m py_compile .github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py`
- `PYTHONDONTWRITEBYTECODE=1 python3 .github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py 1.36.2` -> `1.36.2-2.1`
- `PYTHONDONTWRITEBYTECODE=1 python3 .github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py 1.36.1` -> `1.36.1-1.1`
- `python3 - <<'PY' ... yaml.safe_load(...) ... PY` for `.github/actions/configure-k8s-version/action.yaml`
- `git diff --check`

Duplicate check: searched open PRs/issues for `1.36.2 2.1 kubernetes_deb_version`, `resolve Kubernetes Debian package revisions`, and `configure-k8s-version kubernetes_deb_version`; no matching PR/issue found.

### DIFF
--- a/.github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py
+++ b/.github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Resolve the Debian package revision for a Kubernetes release."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+import time
+from urllib.error import URLError
+from urllib.request import urlopen
+
+
+REQUIRED_PACKAGES = ("kubeadm", "kubectl", "kubelet")
+PACKAGES_URL = "https://pkgs.k8s.io/core:/stable:/v{series}/deb/Packages"
+
+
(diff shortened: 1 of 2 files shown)


.github/actions/configure-k8s-version/resolve-kubernetes-deb-version.py
```diff
@@ -0,0 +1,123 @@
...
+from urllib.request import urlopen
+
+
+REQUIRED_PACKAGES = ("kubeadm", "kubectl", "kubelet")
+PACKAGES_URL = "https://pkgs.k8s.io/core:/stable:/v{series}/deb/Packages"
+
+
+def load_packages(url: str) -> str:
+    try:
+        with urlopen(url, timeout=30) as response:
```
mboersma: This network fetch could now fail builds if there's any transient error talking to pkgs.k8s.io. Could you add a retry to make it more robust? Something like:

```python
 import time

 def load_packages(url: str, attempts: int = 3, backoff: float = 2.0) -> str:
     last_exc: Exception | None = None
     for attempt in range(1, attempts + 1):
         try:
             with urlopen(url, timeout=30) as response:
                 return response.read().decode("utf-8")
         except URLError as exc:
             last_exc = exc
             if attempt < attempts:
                 time.sleep(backoff * attempt)
     raise RuntimeError(f"failed to fetch {url} after {attempts} attempts: {last_exc}")
```

mboersma (changes requested):

mboersma (approved):
/lgtm
/approve

PR #2086, opened 2026-06-28
capi: isolate Azure CI CLI auth per SIG build

MaxRink:
## What this does

The Azure SIG presubmit starts many SIG builds in the background. Today all children share the same Azure CLI config directory while Packer is using Azure CLI auth, which can leave a child without a visible default subscription.

This gives each background SIG build its own `AZURE_CONFIG_DIR`, logs in inside that child, and keeps the selected Azure location scoped to that child. The parent still logs in once for shared setup such as Flatcar version discovery and cleanup.

## Why

This addresses an Azure SIG CI race observed while maintaining the open image-builder PR train. A target failed immediately after `init-sig.sh` had successfully created the resource group/gallery:

```text
error fetching tenantID and subscriptionID from Azure CLI (are you logged on using `az login`?): Unable to find default subscription
```

The same target succeeded in another PR, which points at a CI auth race rather than a target-specific build regression.

## Validation

- `bash -n images/capi/scripts/ci-azure-e2e.sh`
- `git diff --check`
- `shellcheck images/capi/scripts/ci-azure-e2e.sh` still reports pre-existing findings from `main`; the unquoted array expansion errors and the changed `wait` PID handling are fixed by this PR.
- `NO_COLOR=1 make lint` fails locally on unrelated pre-existing Ansible `var-naming[no-role-prefix]` violations under `images/capi/ansible/`; no changed shell-script syntax failure was reported.

### DIFF
--- a/images/capi/scripts/ci-azure-e2e.sh
+++ b/images/capi/scripts/ci-azure-e2e.sh
@@ -80,6 +80,8 @@ get_random_cvm_region() {
 
 export PATH=${PWD}/.local/bin:$PATH
 export PATH=${PYTHON_BIN_DIR:-"/root/.local/bin"}:$PATH
+export AZURE_CONFIG_DIR="${AZURE_CONFIG_DIR:-${ARTIFACTS}/azure-cli/main}"
+mkdir -p "${AZURE_CONFIG_DIR}"
 
 export AZURE_LOCATION="${AZURE_LOCATION:-$(get_random_region)}"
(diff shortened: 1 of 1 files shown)


MaxRink:
/test pull-lint
/test pull-azure-sigs

Requesting the focused lint and Azure SIG presubmit for this CI-auth isolation change.

mboersma (approved):
/lgtm
/approve

PR #2087, opened 2026-06-30
capi: add systemd-sysext image tooling foundation

MaxRink:
## What

This is the first upstreamable slice of optional systemd-sysext image work. It adds only reusable foundation pieces:

- `ansible/node-sysext.yml`, which composes the same roles as `node.yml` minus `containerd`, `kubernetes` and `ecr_credential_provider`. The `node` and `providers` roles stay in, so the base image still gets the common packages, kernel modules, sysctls and guest agents that a system extension cannot supply later, while Kubernetes, containerd and CNI are left to extension images.
- a small `systemd_sysext` Ansible role that checks `systemd-sysext --version`, creates the `/etc/extensions` and `/var/lib/extensions` search paths plus an `/opt/extensions` staging directory (not a search path, so staged images stay inactive), and enables `systemd-sysext.service` only when the version check succeeded and `systemd_sysext_enable_service` is true. The role is skipped inside docker builds.
- `packer/goss/goss-sysext.yaml` for sysext base images: checks the three directories, `systemd-sysext` and its unit, and fails if kubelet, kubeadm, kubectl, containerd or CNI binaries are present.
- `sysext/build-sysext-layer.sh`, `sysext/test-build-sysext-layer.sh` and `sysext/README.md` for building ext4 systemd-sysext layer images.

It intentionally does not add provider build targets yet. Provider target wiring should be reviewed as separate follow-up PRs so the generic sysext foundation can be evaluated independently.

## Layer helper behaviour

- `--os-id` and `--os-version` are required and must match the target's `/usr/lib/os-release`, otherwise systemd-sysext refuses the image at runtime. The extension-release file is generated when the rootfs does not carry one.
- Image contents are always owned by uid 0 / gid 0. The payload is packed into one tar archive with numeric owner 0 and extended attributes, which is handed to `mke2fs -d` directly where mke2fs was built with libarchive. Otherwise the same archive is extracted as root and that tree is passed to `mke2fs`. Running unprivileged without libarchive support fails instead of writing a wrongly owned image. Both paths go through the archive so file capabilities survive; a chown of a plain copy would clear them.
- The image is sized from the payload: `SYSEXT_OVERHEAD_PERCENT` (default 25) is the block headroom and the inode margin, `SYSEXT_MIN_OVERHEAD_KIB` (default 16384) the headroom floor, `SYSEXT_MIN_INODES` (default 1024) the inode floor. The inode count is passed with `-N` because mke2fs otherwise derives it from the image size and runs out on payloads of many small files. The ext4 journal is disabled. The three variables are validated as positive integers.
- mke2fs output goes to stderr so the only stdout line is the image path. INT and TERM remove the work directory before exiting.

## Validation

- `make test-sysext` is new and part of `make validate-all`. The smoke test builds layers from a payload with a non-root owner and asserts with debugfs that files in the image are owned by uid 0 / gid 0, that the extension-release file is present, and that the inode count honours the computed and forced floors. It skips with exit 0 when `mke2fs` or `debugfs` is missing or the host can neither feed a tar stream to mke2fs nor run as root.
- `bash -n` and `shellcheck` on `images/capi/sysext/build-sysext-layer.sh` and `images/capi/sysext/test-build-sysext-layer.sh`
- `yq` parse of `images/capi/ansible/node-sysext.yml`, the `systemd_sysext` role files and the three `images/capi/packer/goss/goss-sysext*.yaml` files
- `cd images/capi && ansible-playbook --syntax-check -i localhost, ansible/node-sysext.yml`
- `docker run --rm -v "$PWD":/work -w /work ubuntu:24.04 bash -lc 'apt-get update >/dev/null && apt-get install -y --no-install-recommends e2fsprogs make >/dev/null && make -C images/capi test-sysext'` passes and exercises the root fallback, since Ubuntu 24.04 ships mke2fs 1.47.0 without tar input; the same test passes on Ubuntu 22.04, Debian 12 and Fedora 42 as root, and on Fedora 42 unprivileged through the tar path
- `git diff --check`

## Provenance

This is a generalized upstream version of image-building functionality developed at Deutsche Telekom. AI assistance was used while removing organization-specific details from the source material.


images/capi/packer/goss/goss-sysext-command.yaml
```diff
@@ -0,0 +1,23 @@
...
+  systemctl cat systemd-sysext.service:
+    exit-status: 0
+    stdout: []
+    stderr: []
+    timeout: 0
+  ? >-
+    test ! -e /usr/bin/kubelet &&
+    test ! -e /usr/bin/kubeadm &&
+    test ! -e /usr/bin/kubectl &&
+    test ! -e /usr/local/bin/containerd &&
```
drew-viles, line 17: Could we make this check independent of the distro-specific installation prefix, or cover all supported paths? Flatcar targets set `sysusr_prefix` to `/opt`, so the existing roles can place kubelet and containerd under `/opt/bin`; distro packages can also install containerd as `/usr/bin/containerd`. I reproduced the predicate succeeding with `/opt/bin/kubelet` and `/opt/bin/containerd` present, even though this suite is intended to ensure those payloads are not baked into the base.

images/capi/sysext/build-sysext-layer.sh
```diff
@@ -0,0 +1,107 @@
...
+
+Builds an ext4 .raw systemd-sysext image from a rootfs containing only usr/ and opt/.
+EOF
+}
+
+name=""
+version=""
+rootfs=""
+output_dir=""
+os_id="ubuntu"
```
drew-viles: Could we make `--os-id` and `--os-version` required instead of defaulting them to Ubuntu 24.04? The helper is described as generic, and this repository also builds Ubuntu 22.04/26.04, RHEL, and Rocky bases. Running the documented command unchanged produces Ubuntu 24.04 compatibility metadata, so the image builds successfully but systemd refuses to merge it on those hosts. Requiring the values would turn that runtime surprise into a build-time error.

images/capi/ansible/node-sysext.yml
```diff
@@ -0,0 +1,69 @@
...
+      loop_control:
+        loop_var: role
+      when: node_custom_roles_pre != ""
+    - name: Include systemd sysext role
+      ansible.builtin.include_role:
+        name: systemd_sysext
+    - name: Include post-node custom roles
+      ansible.builtin.include_role:
+        name: "{{ role }}"
+      loop: "{{ custom_role_names.split() + node_custom_roles_post.split() }}"
```
drew-viles: Could we keep the same filtering used by `ansible/node.yml` here? With `verifySealedImage` in both `custom_role_names` and `node_custom_roles_post_sysprep`, this expression schedules the role before sysprep and the later loop runs it again afterward. The existing node playbook filters that overlap specifically to preserve the requested phase; sharing that expression would keep the behavior consistent.

drew-viles (commented):
Minor notes

MaxRink:
Retiggering the optional OVA job after confirming the failure is outside the PR changes. The failing Rocky Linux job stopped before provisioning at Packer ISO cache handling: `Skipping download ... already exists in the local cache`, followed by `stat /tmp/image-builder-packer-cache-2088327406719864832/rockylinux-9/d06557dd8adaa35716d43c2b8596dbcde8caae31.iso: no such file or directory`. The other six OVA targets completed successfully, and the PR only adds sysext tooling files; no changed file affects ISO download/cache handling.

/test pull-ova-all

MaxRink:
The retriggered `pull-ova-all` job succeeded at https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_image-builder/2087/pull-ova-all/2090564043868737536. Its log reports `rockylinux-9: SUCCESS` along with all other OVA targets. The PR head remains `9e5cf2c118e4d8f4afbbe4ae97c753dc30b77aff`; no code changes or push were needed.

images/capi/ansible/node-sysext.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Include pre-node custom roles
+      ansible.builtin.include_role:
+        name: "{{ role }}"
+      loop: "{{ node_custom_roles_pre.split() }}"
+      loop_control:
+        loop_var: role
+      when: node_custom_roles_pre != ""
+    - name: Include systemd sysext role
+      ansible.builtin.include_role:
+        name: systemd_sysext
```
drew-viles, line 57: This drops the `node` and `providers` roles as well as the payload roles, so the base image gets no common packages, kernel modules, sysctls or guest agents, none of which a sysext can supply later. Could we keep `node` and `providers` and skip only containerd/kubernetes?

images/capi/ansible/node-sysext.yml
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+---
+- name: Provision sysext-capable node base
+  hosts: all
+  become: true
+  vars:
+    node_custom_roles_pre: ""
+    node_custom_roles_post: ""
+    custom_role_names: ""
+    sysprep_remove_swapfiles: false
```
drew-viles: `sysprep_remove_swapfiles` is not read anywhere; the sysprep swapfile task is unconditional.

images/capi/ansible/roles/systemd_sysext/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+systemd_sysext_required: true
+systemd_sysext_directories:
+  - /etc/extensions
+  - /run/extensions
```
drew-viles: `/run` is tmpfs, so this directory does not exist in the shipped image and the goss check for it can only pass during the build. Could we drop it or use a tmpfiles.d entry?

images/capi/ansible/roles/systemd_sysext/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  - /var/lib/extensions
+  - /var/lib/sysext-staging
```
drew-viles: This is not a systemd search path and nothing writes to it. The gpu role already stages sysexts under `/opt/extensions/<name>`; could we follow that convention?

images/capi/ansible/roles/systemd_sysext/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+    state: directory
+    owner: root
+    group: root
+    mode: "0755"
+  loop: "{{ systemd_sysext_directories }}"
+  when:
+    - systemd_sysext_service_manager == "systemd"
+    - systemd_sysext_virtualization_type != "docker"
+
+- name: Enable systemd-sysext at boot
```
drew-viles, line 53: With `systemd_sysext_required: false` on a host without systemd-sysext, this task still fails. Could we gate it on the version check result?

drew-viles (commented):
Thanks for addressing the earlier comments, all three look good now. A few new things inline. The first one on `node-sysext.yml` is the one I would want resolved before this merges.

images/capi/sysext/build-sysext-layer.sh
```diff
@@ -0,0 +1,112 @@
...
+ARCHITECTURE=${sysext_arch}
+SYSEXT_ID=${name}
+SYSEXT_VERSION_ID=${version}
+EOF
+fi
+
+mkdir -p "${output_dir}"
+raw="${output_dir}/${raw_basename}.raw"
+size_kib="$(du -sk "${workdir}" | awk '{print $1}')"
+image_size_kib=$((size_kib + 16384))
```
drew-viles: A fixed 16 MiB headroom will not cover ext4 metadata plus journal for a Kubernetes-sized payload, so `mke2fs -d` fails with ENOSPC. Proportional headroom and `-O ^has_journal` would fix it.

images/capi/sysext/test-build-sysext-layer.sh
```diff
@@ -0,0 +1,53 @@
...
+  --name sysext-test \
+  --version v1.2.3 \
+  --rootfs "${rootfs}" \
+  --output-dir "${workdir}/out" \
+  --os-id ubuntu \
+  --os-version 24.04 \
+  --arch x86_64)"
+
+expected="/usr/lib/extension-release.d/extension-release.sysext-test-v1.2.3-x86-64"
+if ! debugfs -R "stat ${expected}" "${raw}" >/dev/null 2>&1; then
```
drew-viles: `debugfs -R stat` exits 0 whether or not the path exists, so this assertion cannot fail. Grepping the output for `Inode:` would make it real.

images/capi/hack/ensure-sysext-tools.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+missing=()
+for tool in mke2fs find stat; do
```
drew-viles: Nothing references this script or `sysext/templates/basic-layer/layer.yaml`, and the list checks `stat` (unused) while missing `du`, `awk` and `debugfs`. Could we drop both until something consumes them?

images/capi/packer/goss/goss-sysext-command.yaml
MaxRink: You were right: the earlier fix added /opt paths but missed /usr/bin/containerd. Fixed that remaining omission in 2de0261b7. A focused check executes the actual Goss predicate in a temporary filesystem: all payload paths absent passes, and each supported path present fails. The pre-fix predicate reproduces the missed /usr/bin/containerd case. Sorry this was marked resolved before that path was covered.

PR #1986, opened 2026-04-23
feat(outscale): Add Flatcar and NVIDIA GPU support

hajowieland:
## Change description
This change adds support for Flatcar Linux on Outscale with optional NVIDIA GPU support (via [sysext](https://www.flatcar.org/docs/latest/setup/customization/using-nvidia/#prebuilt-sysext-method)).

- Is this change including a new Provider or a new OS? (y/n) n
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) y
- If adding a new provider, are you a representative of that provider? (y/n) n

## Additional context

The change was tested with GPU and without. Also the [Cluster API Provider Outscale](https://github.com/outscale/cluster-api-provider-outscale) needs to be updated to support Ignition.

PR #2090, opened 2026-07-01
fix(outscale): use ansible_facts for os_family with inject_facts_as_vars disabled

sebglon:
**What this PR does / why we need it**:

`images/capi/ansible.cfg` sets `inject_facts_as_vars = False`, so the bare
`ansible_os_family` variable is no longer injected into the play namespace.

The cloud-init task added to `roles/providers/tasks/outscale.yml` in #1986 still
guards on the bare fact:

```yaml
```

With injection disabled this evaluates against an **undefined variable** on
Outscale (OSC) builds. This PR switches it to the `ansible_facts[...]` accessor,
matching the two sibling tasks in the same file (lines 18/24) and the facts
migration from #1981:

```yaml
  when: ansible_facts['os_family'] == "Debian"
```

Scope is OSC-only — this is the sole remaining bare-fact read under
`roles/providers`. `roles/node/tasks/main.yml` only *sets* `ansible_os_family`
(Flatcar override), which is unaffected.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:
No behavioural change intended: `ansible_facts['os_family']` resolves to the
same value the bare fact previously did.

**Release note**:

```release-note
Fix Outscale image builds failing on an undefined `ansible_os_family` variable when `inject_facts_as_vars` is disabled.
```

### DIFF
--- a/images/capi/ansible/roles/providers/tasks/outscale.yml
+++ b/images/capi/ansible/roles/providers/tasks/outscale.yml
@@ -9,7 +9,7 @@
       - cloud-init
       - cloud-guest-utils
       - cloud-initramfs-dyn-netconf
-  when: ansible_os_family == "Debian"
+  when: ansible_facts['os_family'] == "Debian"
 
 - name: Install Debian specific packages
   ansible.builtin.apt:


mboersma:
@sebglon this isn't passing the CLA check because it has a "Co-authored by: Claude" trailer in the commit message. I don't know of any workaround other than to remove that so it just shows you as the sole committer. (It's good to disclose the use of such tools, but at this point I think most of us are also using AI so even without that attribution it's assumed. You can disclose "Co-authored by: Claude" in the PR description if you'd like.)

mboersma:
This looks good and correctly scoped, thanks for the clear writeup.

One heads-up for a follow-up (not blocking this PR): `roles/gpu/tasks/main.yml` still has four bare `ansible_os_family` reads (lines 21, 26, 35, and 39) with the same root cause. Since the Flatcar `set_fact` only defines the bare variable on Flatcar, a non-Flatcar GPU build would hit the same undefined-variable failure once `inject_facts_as_vars` is disabled. Your "sole remaining read under `roles/providers`" statement is accurate since gpu lives elsewhere, but it'd be nice to clean those up too.

mboersma (approved):
/lgtm
/approve

PR #2092, opened 2026-07-01
capi: add qemu Ubuntu immutable target

MaxRink:

Adds an opt-in QEMU Ubuntu 24.04 immutable target for Cluster API images. The
existing Ubuntu targets stay unchanged while the new target adds generic
image-builder plumbing for:

- a separate persistent data partition;
- explicit writable runtime bind mounts for CAPI bootstrap and node runtime state;
- tmpfs scratch paths;
- final read-only-root fstab configuration.

The image root stays writable while Packer and Ansible provision the guest. The
immutable runtime finalizer then creates the data partition, copies the selected
runtime paths while preserving top-level metadata, writes required fstab entries,
adds service ordering drop-ins, remounts the writable paths, and makes `/`
read-only for the next boot before Goss validates the contract.

#### Configuration included

The target is configured through normal Packer variables:

- `immutable_data_partition`
- `immutable_data_partition_fstype`
- `immutable_data_partition_label`
- `immutable_data_partition_mount`
- `immutable_data_partition_mount_options`
- `immutable_persistent_paths`
- `immutable_root_partition_size`
- `immutable_read_only_root`
- `immutable_tmpfs_paths`

The QEMU Ubuntu 24.04 immutable target defaults to:

- a required labeled `CAPI-DATA` partition mounted at `/.capi-data`;
- required persistent bind mounts for `/etc`, `/home`, `/root`, `/opt`, `/srv`, `/usr/local`, and the common mutable `/var` subtrees: `/var/backups`, `/var/cache`, `/var/crash`, `/var/lib`, `/var/local`, `/var/log`, `/var/mail`, `/var/opt`, and `/var/spool`;
- tmpfs mounts for `/tmp` and `/var/tmp`;
- `/` marked read-only in `/etc/fstab` for the next boot;
- systemd drop-ins so cloud-init, containerd, kubelet, and SSH wait for the writable mounts.

The data partition and bind mounts are required for the booted node contract;
this PR intentionally does not use `nofail` for those mounts.

#### Docs included

This PR adds both build config and reviewer-facing docs:

- `images/capi/packer/qemu/qemu-ubuntu-2404-immutable.json`: ready-to-run QEMU Ubuntu 24.04 immutable target.
- `images/capi/packer/qemu/scripts/configure_immutable_runtime.sh`: image finalizer that creates the writable data/runtime layout and makes `/` read-only for next boot.
- `images/capi/packer/qemu/README.md`: build command, output format, immutable variables, sizing notes, and CAPI validation checklist.
- `docs/book/src/capi/immutable-qemu.md`: docs-book page with build command, variable reference, writable path contract, and provider verification checklist.
- `docs/book/src/SUMMARY.md`: docs-book navigation entry.

Example build command from `images/capi`:

```console
PACKER_VAR_FILES="packer/config/kubernetes.json packer/config/cni.json packer/config/containerd.json" \
PACKER_FLAGS="\
  --var 'format=qcow2' \
  --var 'kubernetes_semver=v1.36.1' \
  --var 'kubernetes_series=v1.36' \
  --var 'kubernetes_deb_version=1.36.1-1.1' \
  --var 'immutable_read_only_root=true'" \
  make build-qemu-ubuntu-2404-immutable
```

The docs also cover first-boot checks for cloud-init completion, provider
hostname writes, non-empty machine-id generation, read-only `/`, writable data
partition paths, and reboot persistence for bootstrap/runtime state.

#### Provenance

This is a generalized upstream version of image-building functionality developed at Deutsche Telekom. AI assistance was used while removing organization-specific details from the source material.

None

#### Testing

```console
git diff --check
bash -n images/capi/packer/qemu/scripts/configure_immutable_runtime.sh
shellcheck images/capi/packer/qemu/scripts/configure_immutable_runtime.sh
python3 -m py_compile images/capi/packer/qemu/scripts/configure_immutable_runtime_test.py images/capi/packer/qemu/scripts/render_ubuntu_autoinstall_test.py
make -C images/capi test-qemu-immutable
make -C images/capi validate-qemu-ubuntu-2404-immutable
```


images/capi/packer/qemu/scripts/render_ubuntu_autoinstall.py
```diff
@@ -0,0 +1,174 @@
...
+def render_user_data(var_file, values):
+    profile = values.get("autoinstall_profile")
+    distro = values.get("distro_name")
+    if distro != "ubuntu" or not profile:
+        return
+
+    profile_dir = ROOT / "qemu" / "linux" / "ubuntu" / "http" / str(profile)
+    user_data = profile_dir / "user-data"
+    template = profile_dir / "user-data.tmpl"
+    if template.exists():
```
drew-viles: Could we preserve the password rendering here? `set-ssh-password` has already generated `user-data` and substituted `$ENCRYPTED_SSH_PASSWORD`, but this reads the original `.tmpl` again and later overwrites that output. I reproduced the target render and the resulting file still contains `passwd: $ENCRYPTED_SSH_PASSWORD`, so the installed builder account will not match Packer’s SSH password. Could the immutable substitutions be applied to the already-rendered `user-data`, or incorporated into the original rendering pass?

images/capi/packer/qemu/packer.json.tmpl
```diff
@@ -140,6 +140,23 @@
...
+        "IMMUTABLE_DATA_PARTITION_LABEL={{user `immutable_data_partition_label`}}",
+        "IMMUTABLE_DATA_PARTITION_MOUNT={{user `immutable_data_partition_mount`}}",
+        "IMMUTABLE_DATA_PARTITION_MOUNT_OPTIONS={{user `immutable_data_partition_mount_options`}}",
+        "IMMUTABLE_PERSISTENT_PATHS={{user `immutable_persistent_paths`}}",
+        "IMMUTABLE_READ_ONLY_ROOT={{user `immutable_read_only_root`}}",
+        "IMMUTABLE_TMPFS_PATHS={{user `immutable_tmpfs_paths`}}"
+      ],
+      "inline": [
+        "if [ \"$IMMUTABLE_DATA_PARTITION\" = \"true\" ] || [ \"$IMMUTABLE_READ_ONLY_ROOT\" = \"true\" ] || [ -n \"$IMMUTABLE_PERSISTENT_PATHS\" ]; then sudo systemctl stop kubelet.service containerd.service || true; fi",
+        "bash ./packer/qemu/scripts/configure_immutable_runtime.sh"
```
drew-viles: Could this helper be supplied through the shell provisioner’s `script` field? This provisioner runs inside the guest, while `./packer/qemu/scripts/configure_immutable_runtime.sh` exists only in the host checkout and nothing uploads it. Packer uploads the generated inline script, not arbitrary local files referenced by it, so this command fails with `No such file or directory`. The existing Flatcar provisioner above demonstrates the required `script` pattern.

images/capi/packer/qemu/scripts/configure_immutable_runtime.sh
```diff
@@ -0,0 +1,346 @@
...
+  local source_parent
+
+  validate_absolute_mount_path "${path}" "IMMUTABLE_PERSISTENT_PATHS"
+  source_parent="$(dirname "${source}")"
+  run_privileged install -d -m 0755 "${source_parent}"
+  run_privileged install -d -m 0755 "${source}"
+  if [ ! -d "${path}" ]; then
+    run_privileged install -d -m 0755 "${path}"
+  fi
+  if [ -z "$(find "${source}" -mindepth 1 -print -quit 2>/dev/null)" ] &&
```
drew-viles: Should these probes run through `run_privileged` as well? The helper runs as `builder`, and the default path list includes `/root`, which is normally mode `0700`. An unprivileged `find /root` produces no output here with the permission error suppressed, so the copy is skipped and the later empty bind mount hides the original contents. Running the checks with the same privilege as the copy would preserve restricted paths correctly.

images/capi/packer/qemu/scripts/render_ubuntu_autoinstall.py
```diff
@@ -0,0 +1,174 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("var_file", type=pathlib.Path)
+    parser.add_argument("--extra-var-file", action="append", default=[], type=pathlib.Path)
+    args = parser.parse_args()
+
+    values = dict(IMMUTABLE_DEFAULTS)
+    values.update(load_json(args.var_file))
+
+    packer_values, packer_var_files = parse_packer_flags(os.environ.get("PACKER_FLAGS", ""))
+    for var_file in [pathlib.Path(path) for path in packer_var_files] + args.extra_var_file:
```
drew-viles: Could we preserve the same var-file precedence as the Packer command? The Make recipe places var files from `PACKER_FLAGS` before the target file, but this renderer loads them afterward. I reproduced a case where the renderer formatted the partition as `RUNTIME-DATA` while Packer resolved the later target value as `CAPI-DATA`, causing the runtime mount to fail. Could we mirror the exact CLI ordering or reject `--var-file` through `PACKER_FLAGS`?

images/capi/packer/qemu/packer.json.tmpl
```diff
@@ -140,6 +140,23 @@
...
+        "IMMUTABLE_DATA_PARTITION_FSTYPE={{user `immutable_data_partition_fstype`}}",
+        "IMMUTABLE_DATA_PARTITION_LABEL={{user `immutable_data_partition_label`}}",
+        "IMMUTABLE_DATA_PARTITION_MOUNT={{user `immutable_data_partition_mount`}}",
+        "IMMUTABLE_DATA_PARTITION_MOUNT_OPTIONS={{user `immutable_data_partition_mount_options`}}",
+        "IMMUTABLE_PERSISTENT_PATHS={{user `immutable_persistent_paths`}}",
+        "IMMUTABLE_READ_ONLY_ROOT={{user `immutable_read_only_root`}}",
+        "IMMUTABLE_TMPFS_PATHS={{user `immutable_tmpfs_paths`}}"
+      ],
+      "inline": [
+        "if [ \"$IMMUTABLE_DATA_PARTITION\" = \"true\" ] || [ \"$IMMUTABLE_READ_ONLY_ROOT\" = \"true\" ] || [ -n \"$IMMUTABLE_PERSISTENT_PATHS\" ]; then sudo systemctl stop kubelet.service containerd.service || true; fi",
```
drew-viles: Could we avoid swallowing failures while stopping these services? Both are installed before this immutable target copies and bind-mounts `/var/lib`. A real stop failure currently returns success because of `|| true`, allowing the copy to proceed against live containerd or kubelet state. If the intent is only to tolerate absent units on other variants, could we handle that case explicitly and fail on other errors?

drew-viles (commented):
Minor notes

mboersma (approved):
/approve

PR #2093, opened 2026-07-03
capi: avoid OVA CI SSH password races

MaxRink:
What this does
- Runs OVA CI shared prerequisites once before starting parallel image builds.
- Generates the SSH password files once, then skips the phony dependency reruns inside each background make target.
- Quotes the OVA target loop and checks child process status directly.

Why
- `ci-ova.sh` fans out several vSphere OVA builds in parallel. Each Make target also depends on `set-ssh-password`, which rewrites shared Packer and installer data. Parallel rewrites can make Packer authenticate with credentials that do not match the installer data used by a VM.
- This was observed as an OVA job timeout with a target repeatedly reaching SSH but failing password authentication while other OVA targets succeeded.

Validation
- `git diff --check`
- `shellcheck images/capi/scripts/ci-ova.sh` (remaining warnings are pre-existing outside the changed loop)
- `make -n -o deps-ova -o set-ssh-password build-node-ova-vsphere-ubuntu-2204 PACKER_LOG=0` shows the background target goes directly to the Packer build without rerunning `deps-ova` or `set-ssh-password`


mboersma (approved):
/lgtm
/approve

PR #2095, opened 2026-07-05
capi: remove new subiquity cloud-init networking disable file

### LINKED ISSUE #2057: Enabling cloud-init networking with ubuntu 26.04 fails
## Environment

- **Make target:** `make build-proxmox-ubuntu-2604-efi`
- **Run using container image? (Y/N): N
- **Environment vars:**
  - PROXMOX_URL="https://redacted:8006/api2/json"
  - PROXMOX_USERNAME="root@pam"
  - PROXMOX_PASSWORD="redacted"
  - PROXMOX_NODE="redacted"
  - PROXMOX_ISO_POOL="redacted-cephfs1"
  - PROXMOX_BRIDGE="vmbr0"
  - PROXMOX_STORAGE_POOL="redacted-rbd1"
  - PROXMOX_MAC="redacted"
  - PROXMOX_VLAN="redacted"

I ran image-builder and the image was created successfully. I am using (cluster-api) with InClusterIPPool and am assigning a static IP. When the image boots - it does DHCP and gets a DHCP-Pool-Address (which it shouldn't). I verified in the cloud-init cd-rom image that the IP is set correctly.

I expected that the VM sets a static IP and boots.

I think I understand the problem and I also found a solution.

The problem:
In the ansible task https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/sysprep/tasks/debian.yml - the task item "Removing subiquity disable cloud-init networking config" is supposed to delete the file: /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg

Due to this change in ubuntu (https://github.com/canonical/subiquity/commit/f058687342ee9149ced1769a9b464a31b8b8c9b6) - the file name has been changed to "00-subiquity-disable-cloudinit-networking.cfg" - therefore the deletion fails and the network initialization is not working as supposed.

One possible solution:
My research indicates that the change was applied for the ubuntue releases 25.10 and 26.04. I implemented a quick fix in the ansible task by adding the task item:

```
- name: Removing subiquity disable cloud-init networking config (00- prefix)
  ansible.builtin.file:
    path: /etc/cloud/cloud.cfg.d/00-subiquity-disable-cloudinit-networking.cfg
    state: absent
  when: ansible_facts['distribution_version'] is version('22.04', '>=')
```

I guess the "when" for the old file should be something like: ```when: ansible_facts['distribution_version'] is version('22.04', '>=') and version('25.10', '<')``` while the when for the file prefixed with "00-" should be something like: ```when: ansible_facts['distribution_version'] is version('25.10', '>=')```

I would like to write a pull request with a proper solution. Can you guide me on the process for doing so?

Kind regards
Dirk

---

/kind bug

MaxRink:

Removes both known Subiquity cloud-init networking disable filenames during
Debian sysprep:

- `/etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg`
- `/etc/cloud/cloud.cfg.d/00-subiquity-disable-cloudinit-networking.cfg`

Subiquity changed the filename for newer Ubuntu releases. If the new filename is
left behind, first-boot cloud-init networking can stay disabled and provider
static network data is ignored.

This keeps the existing module-native `ansible.builtin.file` cleanup and extends
it with a loop over both filenames.

Fixes #2057

#### Testing

```console
git diff --check
python3 - <<'PY'
from pathlib import Path
import yaml
p = Path('images/capi/ansible/roles/sysprep/tasks/debian.yml')
with p.open() as f:
    data = yaml.safe_load(f)
task = next(t for t in data if t.get('name') == 'Removing subiquity disable cloud-init networking configs')
assert task['ansible.builtin.file']['path'] == '{{ item }}'
assert '/etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg' in task['loop']
assert '/etc/cloud/cloud.cfg.d/00-subiquity-disable-cloudinit-networking.cfg' in task['loop']
print('ok')
PY
cd images/capi && ansible-playbook --syntax-check ansible/node.yml
```

### DIFF
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -151,10 +151,13 @@
     - _stat_update_grub.stat.exists
     - sysprep_require_grub_file|default(true)|bool
 
-- name: Removing subiquity disable cloud-init networking config
+- name: Removing subiquity disable cloud-init networking configs
   ansible.builtin.file:
-    path: /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg
+    path: "{{ item }}"
     state: absent
+  loop:
+    - /etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg
+    - /etc/cloud/cloud.cfg.d/00-subiquity-disable-cloudinit-networking.cfg
   when: ansible_facts['distribution_version'] is version('22.04', '>=')
 
 - name: Removing 99-installer.cfg which sets the cloud-init datasource to None


mboersma (approved):
/lgtm
/approve

mboersma (approved):
/lgtm

PR #2098, opened 2026-07-05
capi: decouple CNI plugin installation from the Kubernetes source type

### LINKED ISSUE #1861: Ensure image-builder continues to work once kubernetes-cni is discontinued
There is an ongoing discussion about discontinuing the `kubernetes-cni` package (https://github.com/kubernetes/release/issues/4156).

Related discussion here: https://github.com/kubernetes/release/issues/4100

IIRC image-builder has a dependency on it, i.e. it always tries to install it (not sure if it is *actually* required to install the package)

MaxRink:
## Summary

`kubernetes_cni_source_type` becomes a real, independent choice of where the CNI plugins come from, and CNI installation moves out of the Kubernetes install path into its own task files.

Before this change the CNI tarball was installed from `url.yml`, so it was tied to `kubernetes_source_type=http` and could not be selected separately. There was no way to ask for tarball plugins on a package Kubernetes build, or for package plugins on an http Kubernetes build.

## Behaviour

`kubernetes_cni_source_type` is empty by default, which means the CNI plugins follow `kubernetes_source_type`. Since `kubernetes_source_type` defaults to `pkg`, the default image is unchanged: the plugins come from the distro `kubernetes-cni` package. Flatcar sets both to `http` and keeps installing the tarball.

Setting `kubernetes_cni_source_type` explicitly decouples the two:

* `http` installs the upstream containernetworking/plugins tarball even when Kubernetes comes from packages.
* `pkg` installs the `kubernetes-cni` package even when Kubernetes comes from `kubernetes_http_source`.

Extra vars have top precedence in Ansible, so the incoming variable is not overwritten. The role resolves the pair into `kubernetes_cni_source_type_effective` in `defaults/main.yml` and selects `cni-pkg.yml` or `cni-url.yml` on that. The Goss templates compute the same fallback into `$cni_source`, because the Packer templates pass the raw user variable straight through and legacy JSON templates have no conditionals.

`kubernetes_cni_semver` is bumped from `v1.2.0` to `v1.9.1` for the tarball path. This changes the plugin version in every build whose effective CNI source is `http`: Flatcar, which sets `http` explicitly, http Kubernetes builds that leave the CNI source empty, and builds that opt into `http` CNI. Package builds are unaffected.

## Review follow-ups

* CNI installation is removed from `url.yml` entirely. It lives only in `cni-url.yml` and `cni-pkg.yml`, so `http`/`http` no longer installs the plugins twice and `http`/`pkg` no longer overlays a package install on tarball files.
* Both source types are validated before any install task runs. `kubernetes_source_type` must be `http` or `pkg`; `kubernetes_cni_source_type` must be `http`, `pkg`, or empty to follow.
* Package CNI is rejected explicitly on Flatcar, on the resolved value, so it fails whether it was asked for directly or reached by following `kubernetes_source_type`.
* The Goss image checks no longer depend on the CNI source. The three `crictl images` checks key on `kubernetes_source_type` alone; previously an http Kubernetes build with package CNI, or a package Kubernetes build with http CNI, checked no images at all. CNI checks key on the CNI source alone.
* The tarball download keeps the retry and until semantics it had in `url.yml`, five retries with a three second delay.
* The docs describe what happens if you choose `http` CNI on the package Kubernetes path: kubelet's hard dependency on `kubernetes-cni (>= 1.2.0)` still installs the package, the tarball then overwrites the same paths under `/opt/cni/bin`, `dpkg --verify` and `rpm -V kubernetes-cni` report the binaries as modified, and any later upgrade, reinstall or removal of the package silently replaces or deletes them. pkgs.k8s.io does move `kubernetes-cni` within a single Kubernetes minor, so this is reachable without changing minor. The `sysprep` role already holds or excludes every installed package at the end of a build, so the built image is safe until an operator removes that pin.

## Validation

Verified locally on macOS with ansible-core 2.21 and packer 1.16:

* `ansible-playbook --syntax-check` on `ansible/node.yml`.
* `ansible-lint --project-dir . ansible/` reports no findings that are not already present on `main`.
* A temporary playbook running the role's validation tasks against localhost, covering twelve combinations of the two source types, undefined and empty values, invalid values, and Flatcar. Each passes or fails as intended.
* The Goss templates rendered with Go `text/template` over `goss-vars.yaml` for twelve combinations of Kubernetes source, CNI source including empty, and `kubernetes_load_additional_imgs`. Every render is valid YAML and each check appears only under the intended source type.
* `shellcheck` on the shell embedded in the new Goss checks.
* `make validate-node-ova-local-ubuntu-2404`, `make validate-node-ova-local-flatcar` and `make validate-ami-ubuntu-2404`.
* `hack/generate-goss-specs.py --dry-run` for the ova and amazon providers.


drew-viles (commented):
Looks good. It's a heck of a jump on the CNI version but I am also using the latest on mine so I think it's something that's just fallen behind. Unless there is a reason it's currently locked to that of which I am unaware.

I'll prod @mboersma as a follow up to confirm my thinking.

/ok-to-test

MaxRink:
@drew-viles As far as i can see 1.2.0 never was the effective version that ended up in an Image.

Ive updated the PR description above to indicate the actual versions that ended up on images

MaxRink:
https://github.com/kubernetes-sigs/image-builder/pull/2106 would be a port of how we handle this downstream

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -30,7 +30,15 @@
...
-  when: kubernetes_source_type == "http" and kubernetes_cni_source_type == "http"
+  when: kubernetes_source_type == "http"
+
+- name: Import CNI URL tasks
+  ansible.builtin.import_tasks: cni-url.yml
+  when: kubernetes_cni_source_type | default("pkg") == "http"
+
+- name: Import CNI package validation tasks
+  ansible.builtin.import_tasks: cni-pkg.yml
+  when: kubernetes_cni_source_type | default("pkg") == "pkg"
```
drew-viles, line 45: Could we also install `kubernetes-cni` when Kubernetes itself comes from HTTP? With `kubernetes_source_type=http` and `kubernetes_cni_source_type=pkg`, the distro task files above are skipped because they are gated on the Kubernetes source. On a clean image we then jump directly into validation, where `/opt/cni/bin/bridge` is absent and the build fails. Moving CNI package installation into its own distro-specific path would make these source settings genuinely independent.

images/capi/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -30,7 +30,15 @@
 
 - name: Import URL Kubernetes tasks
   ansible.builtin.import_tasks: url.yml
-  when: kubernetes_source_type == "http" and kubernetes_cni_source_type == "http"
+  when: kubernetes_source_type == "http"
+
+- name: Import CNI URL tasks
+  ansible.builtin.import_tasks: cni-url.yml
+  when: kubernetes_cni_source_type | default("pkg") == "http"
```
drew-viles, line 49: Could we assert that `kubernetes_cni_source_type` is one of `http` or `pkg` before these imports? An override typo such as `htpp` makes both conditions false, so a Flatcar build skips CNI installation entirely. The Goss template uses the same equality checks and consequently omits all CNI validation as well, allowing the misconfigured image to pass. Failing early here would make unsupported values explicit.

drew-viles (commented):
Minor notes

mboersma:
Really nice work, the `cni-url.yml`/`cni-pkg.yml` split is clean and bumping `kubernetes_cni_semver` off the bogus `v1.2.0` lower-bound is a good fix on its own.

One thing to work through before this lands, since it changes every provider's default image. The default is `kubernetes_source_type=pkg`, and `kubelet` hard-depends on `kubernetes-cni (>= 1.2.0)` (I checked v1.32 and v1.36 repo metadata). So on the default path we still pull the `kubernetes-cni` package via kubelet, then overlay it with the v1.9.1 tarball. Two side effects:

- The package stays installed but its files no longer match on disk, so `dpkg --verify` / `rpm -V kubernetes-cni` will flag the plugin binaries.
- CNI is now decoupled from the k8s minor: every image ships v1.9.1 (a 1.32 image gets 1.9.1, not the package's 1.6.0). Compatibility is fine, but it's a real behavior change worth calling out.

Ordering is safe, the tarball always wins. Should we document this explicitly, or scope the http-CNI default to the http-Kubernetes path where kubelet isn't a package?

mboersma (changes requested):
Could we finish separating Kubernetes and CNI installation before merging? `url.yml` still installs the CNI tarball whenever Kubernetes uses HTTP, so `http/http` installs it twice and `http/pkg` overlays a package installation on top of tarball files. Please remove the CNI work from `url.yml` and install it only through `cni-url.yml` or `cni-pkg.yml`.

It would also be good to validate both source-type variables before installation starts, explicitly reject package CNI on Flatcar, make the HTTP Goss checks independent of the CNI source, and retain the existing download retries.

MaxRink:
Following up on the package-overwrite/default question: current head bf7a2cbb48abdd2f8d521bd54e4f7c1b93f90aff makes the CNI source default follow the Kubernetes source, so the default pkg/pkg path retains packaged CNI. HTTP CNI on packaged Kubernetes is opt-in. docs/book/src/capi/capi.md now explains the shared tarball version across targets, package file verification mismatches, and package upgrades potentially replacing those files. This addresses the default-path concern without silently changing every image to tarball CNI.

Issue #1334, opened 2023-10-27
Handle latest iso's not being immutable


AverageMarcus:
**Is your feature request related to a problem? Please describe.**

For our providers that build based on top of an iso image we've seen a problem several times now with Ubuntu (maybe others?) where the latest release is only available as a non-specific URL that is updated as new releases are made.

For example, Ubuntu 23.04 is currently (at time of writing) only available at the following location:

```
https://releases.ubuntu.com/23.04/ubuntu-23.04-live-server-amd64.iso
```

When a new release of Ubuntu 23.04 is made, this image is replaced but the URL remains the same. There isn't a location to get the specific patch version of the release that we've been able to find.

This causes the following problems:
* checksums become out of date so we need to keep updating them even though the URL isn't changing
* breaking changes could possibly be introduced into a new release of Ubuntu that could break for users of `image-builder` without there being any changes made to `image-builder` or the provided variables. This means image-builder can't guarentee reproducible builds.

**Describe the solution you'd like**

Ideally there would be a location where we can get the patch versions of Ubuntu images but I'm not aware of such a place existing.

**Describe alternatives you've considered**

We could fetch the checksum on the fly from the https://releases.ubuntu.com/23.04/SHA256SUMS file but this doesn't solve the reproducibility issue.

**Additional context**

For old releases, they become available at a different URL (e.g. `https://old-releases.ubuntu.com/releases/jammy/ubuntu-22.04.2-live-server-amd64.iso`) but the same doesn't seem to be possible with the latest patch release.

/kind feature

AverageMarcus:
A little more investigation...

The current latest Ubuntu release is available at:
```
https://releases.ubuntu.com/22.04.4/ubuntu-22.04.4-live-server-amd64.iso
```

With curl:
```
curl https://releases.ubuntu.com/22.04.4/ubuntu-22.04.4-live-server-amd64.iso
HTTP/1.1 200 OK
Date: Fri, 05 Apr 2024 07:23:52 GMT
Server: Apache/2.4.29 (Ubuntu)
Last-Modified: Fri, 16 Feb 2024 23:52:36 GMT
ETag: "7d6eb800-61188703a4abd"
Accept-Ranges: bytes
Content-Length: 2104408064
Content-Type: application/x-iso9660-image

+-----------------------------------------+
| NOTE: binary data not shown in terminal |
+-----------------------------------------+
```

The previous Ubuntu release can be fetched using:
```
[...]

PR #2100, opened 2026-07-05
Use Ubuntu point-release ISO URLs

MaxRink:
## Change description

Use Ubuntu 24.04 point-release download directories for the live-server ISO configs that already pin the 24.04.4 checksum, including `qemu-ubuntu-2404-immutable.json`, and update the MAAS arm64 24.04 config from 24.04.3 to 24.04.4.

This avoids depending on mutable series or codename aliases for Ubuntu 24.04 ISO downloads where Ubuntu already publishes point-release artifacts. A unit test in `render_ubuntu_autoinstall_test.py` asserts that the immutable QEMU config keeps a point-release URL.

26.04.1 has since been published, so the 26.04 configs are updated the same way: `https://releases.ubuntu.com/26.04.1/ubuntu-26.04.1-live-server-amd64.iso` replaces the `/resolute/` and `/releases/26.04/` paths, which Canonical rewrites in place on every point release. That change is the output of running the new `make update-ubuntu-iso-checksums` target and is kept in its own commit. All nine 26.04 ISO configs are amd64; there is no 26.04 arm64 ISO config.

The update scripts now fail loudly instead of writing bad data: a download or HTTP failure aborts with curl's own error, an empty checksum match is an error instead of an empty `iso_checksum`, a missing series directory is a hard error, the `jq` write is no longer masked by `&& mv`, and the temporary file is removed on exit.

This also adds `update-ubuntu-iso-urls` and wires it into `update-ubuntu-iso-checksums`, so a scheduled/manual update workflow can discover newer Ubuntu point-release ISO URLs and checksums. Dependabot was checked, but it cannot track these arbitrary Packer JSON URL strings because Dependabot updates are driven by supported package ecosystems/manifests; the Make target is the maintainable path here.

## Related issues

Related to #1334

## Validation

- `git diff --check`
- `jq empty` for all changed JSON configs
- Verified Ubuntu SHA256SUMS entries for 24.04.4 amd64 and arm64
- Verified HTTP availability for the new 24.04.4 amd64 and arm64 ISO URLs
- `bash -n images/capi/hack/update-ubuntu-iso-urls.sh images/capi/hack/update-iso-checksums.sh`
- `shellcheck images/capi/hack/update-ubuntu-iso-urls.sh images/capi/hack/update-iso-checksums.sh`
- `cd images/capi/packer/qemu/scripts && python3 -m unittest render_ubuntu_autoinstall_test`
- `cd images/capi && make update-ubuntu-iso-checksums`
- `make validate-qemu-ubuntu-2404 validate-qemu-ubuntu-2404-efi validate-qemu-ubuntu-2604 validate-qemu-ubuntu-2604-efi`
- `make validate-raw-ubuntu-2404 validate-raw-ubuntu-2404-efi validate-raw-ubuntu-2604 validate-raw-ubuntu-2604-efi`
- `make validate-maas-ubuntu-2404-efi validate-maas-ubuntu-2404-arm64 validate-maas-ubuntu-2604-efi`
- `make validate-node-ova-local-ubuntu-2404 validate-node-ova-local-ubuntu-2604 validate-node-ova-local-ubuntu-2604-efi`
- `env PROXMOX_URL=https://example.invalid:8006/api2/json PROXMOX_USERNAME=root@pam PROXMOX_TOKEN=dummy PROXMOX_NODE=pve1 PROXMOX_ISO_POOL=local PROXMOX_BRIDGE=vmbr0 PROXMOX_STORAGE_POOL=local-lvm make validate-proxmox-ubuntu-2404 validate-proxmox-ubuntu-2404-efi validate-proxmox-ubuntu-2604 validate-proxmox-ubuntu-2604-efi`
- Failure paths exercised on scratch copies: unresolvable host, HTTP 404, no matching SHA256SUMS line and an empty checksum column all exit non-zero with distinct errors and leave the config untouched
- A second `make update-ubuntu-iso-checksums` run is a no-op


drew-viles (commented):
/ok-to-test

mboersma (approved):
/lgtm
/assign @drew-viles 

This looks good to me, but let's get more eyes on it if possible.

images/capi/hack/update-ubuntu-iso-urls.sh
```diff
@@ -0,0 +1,141 @@
...
+          .iso_url = $iso_url
+          | .iso_checksum = $iso_checksum
+          | .iso_checksum_type = (.iso_checksum_type // "sha256")
+          | if (.iso_target_path? | type) == "string" then
+              .iso_target_path |= sub("ubuntu-[0-9]+\\.[0-9]+(\\.[0-9]+)?-live-server-(amd64|arm64)\\.iso$"; $iso_file)
+            else
+              .
+            end
+        ' "${file}" > "${tmp}" &&
+        mv "${tmp}" "${file}"
```
drew-viles: Could we split the `jq` and `mv` commands instead of joining them with `&&`? Because `jq` is on the left side of an `&&` list, Bash’s errexit behavior won’t terminate the script if it fails. The script can then print `Updated ...` even though the file wasn’t changed.

Something like this would allow failures to propagate correctly:
```
jq ... "${file}" > "${tmp}"
mv "${tmp}" "${file}"
echo "Updated ${file} to ${latest_iso_file}"
```

images/capi/hack/update-ubuntu-iso-urls.sh
```diff
@@ -0,0 +1,141 @@
...
+    fi
+
+    if [[ -e "${cache_dir}/${cache_key}.skip" ]]; then
+        return 1
+    fi
+
+    if ! checksum_data="$(curl -fsSL "${checksum_url}")"; then
+        echo "WARNING: unable to fetch ${checksum_url}; skipping Ubuntu ${series} ${arch}" >&2
+        touch "${cache_dir}/${cache_key}.skip"
+        return 1
```
drew-viles: Could we fail the target when fetching SHA256SUMS fails, rather than warning and continuing? As written, make update-ubuntu-iso-urls can exit successfully without checking or updating every configuration, which could make automation report a successful refresh when it was incomplete.

I think “this Ubuntu series doesn’t have a point release yet” should remain a valid skip, but any potential network or HTTP failures should return non-zero.

drew-viles (changes requested):

images/capi/packer/qemu/scripts/render_ubuntu_autoinstall_test.py
```diff
@@ -204,6 +204,17 @@ def test_immutable_template_runs_cleanup_in_target(self):
...
         self.assertNotIn("    - swapoff -a\n", template)
 
+    def test_immutable_qemu_target_uses_point_release_iso_url(self):
+        target = (
+            pathlib.Path(__file__).resolve().parents[1]
+            / "qemu-ubuntu-2404-immutable.json"
+        )
+
+        self.assertEqual(
+            "https://releases.ubuntu.com/24.04.4/ubuntu-24.04.4-live-server-amd64.iso",
```
mboersma: Could we avoid hard-coding `24.04.4` here? The URL updater will move this config to the next point release without updating this test, so the normal update flow would make the test fail. Could the test check that the URL uses a consistent point-release directory and filename instead?

mboersma (changes requested):

images/capi/packer/qemu/scripts/render_ubuntu_autoinstall_test.py
MaxRink: Updated the test to accept any Ubuntu 24.04 point release while requiring the directory and filename versions to match. Targeted tests (8) and git diff --check pass.

MaxRink: Addressed in 8ab571d: the assertion now accepts any Ubuntu 24.04 point release while requiring the directory and filename versions to match. The commit author and committer are MaxRink, and the unauthorized Copilot co-author metadata was removed so EasyCLA can identify the authorized contributor. Please re-review.

images/capi/hack/update-iso-checksums.sh
```diff
@@ -24,22 +24,29 @@ _checksum_file=${3:-}
...
-        iso_file_name=$(basename $iso_url)
-        sha256sums_url="$(dirname $iso_url)/${_checksum_file}"
+        sha256sums_url="$(dirname "${iso_url}")/${_checksum_file}"
         _checksum_search_pattern=${4/iso_file_name/$iso_file_name}
-        iso_checksum=$(curl -SsL $sha256sums_url | grep "${_checksum_search_pattern}" | awk -v col=$_checksum_position '{print $col}')
+        iso_checksum=$(curl -SsL "${sha256sums_url}" | grep "${_checksum_search_pattern}" | awk -v col="${_checksum_position}" '{print $col}')
     fi
     tmp=$(mktemp)
-    jq --arg iso_checksum $iso_checksum --arg iso_checksum_type ${iso_checksum_type} '.iso_checksum = $iso_checksum | .iso_checksum_type = $iso_checksum_type' $file > "$tmp" && mv "$tmp" $file
+    jq --arg iso_checksum "${iso_checksum}" --arg iso_checksum_type "${iso_checksum_type}" '.iso_checksum = $iso_checksum | .iso_checksum_type = $iso_checksum_type' "${file}" > "${tmp}" && mv "${tmp}" "${file}"
```
drew-viles: Same `jq ... && mv` pattern as the one split in the new script. Could we split it here too so a jq failure propagates?

images/capi/hack/update-iso-checksums.sh
```diff
@@ -24,22 +24,29 @@ _checksum_file=${3:-}
...
     if [[ "${_compute_checksum}" = "true" ]]; then
-        iso_checksum=$(curl -SsL  $iso_url | ${iso_checksum_type}sum | awk '{print $1}')
+        iso_checksum=$(curl -SsL "${iso_url}" | "${iso_checksum_type}"sum | awk '{print $1}')
     else
-        iso_file_name=$(basename $iso_url)
-        sha256sums_url="$(dirname $iso_url)/${_checksum_file}"
+        sha256sums_url="$(dirname "${iso_url}")/${_checksum_file}"
         _checksum_search_pattern=${4/iso_file_name/$iso_file_name}
-        iso_checksum=$(curl -SsL $sha256sums_url | grep "${_checksum_search_pattern}" | awk -v col=$_checksum_position '{print $col}')
+        iso_checksum=$(curl -SsL "${sha256sums_url}" | grep "${_checksum_search_pattern}" | awk -v col="${_checksum_position}" '{print $col}')
```
drew-viles: If grep finds no matching line this writes `"iso_checksum": ""` and exits 0. Could we fail on an empty checksum?

drew-viles (commented):
Re-checked after the new commits: all three earlier comments are addressed and every checksum and URL matches upstream SHA256SUMS. Three small hardening points inline. One note for the description: 26.04.1 now exists, so the "26.04 intentionally unchanged" paragraph is stale and the first update run will bump those configs too.

images/capi/hack/update-ubuntu-iso-urls.sh
```diff
@@ -0,0 +1,156 @@
...
+    fi
+
+    checksum_data_file="${cache_dir}/${cache_key}.raw"
+    if ! http_status="$(curl -sSL -o "${checksum_data_file}" -w '%{http_code}' "${checksum_url}")"; then
+        echo "ERROR: network error fetching ${checksum_url}" >&2
+        exit 1
+    fi
+
+    if [[ "${http_status}" == "404" ]]; then
```
drew-viles: A 404 on the series directory means the series is gone or the URL is wrong, not "no point release yet" (the no-match branch below already handles that case). Could this be a hard error?

MaxRink: It is now

images/capi/hack/update-iso-checksums.sh
MaxRink: Done

images/capi/hack/update-iso-checksums.sh
MaxRink: done

PR #2102, opened 2026-07-06
capi: render containerd service unit locally

MaxRink:
## What
- Renders `containerd.service` from a bundled Ansible template by default.
- Keeps `containerd_service_url` as an opt-in override for custom unit files.
- Removes the default provider dependency on `raw.githubusercontent.com` for the containerd unit.
- Adds `make update-containerd-service-template` and `make verify-containerd-service-template` for keeping the bundled unit aligned with the pinned containerd version.
- Adds a Dependabot-only maintenance workflow that refreshes the bundled unit on same-repository Dependabot PRs that update the Linux containerd version pins.
- Documents the `containerd_service_url` override and the maintenance workflow.

## Validation
- `git diff --check`
- YAML parse of `.github/workflows/update-containerd-service-template.yaml`
- Extracted workflow shell blocks and checked them with `bash -n`
- `python3 -m py_compile images/capi/hack/update-containerd-service-template.py`
- `make -C images/capi verify-containerd-service-template`
- `make -C images/capi update-containerd-service-template`
- `make -C images/capi validate-gce-ubuntu-2404 PACKER_FLAGS=-syntax-only`

Credential-free syntax-only validation passed for the GCE template; a full GCE build still requires cloud configuration.


mboersma:
Really nice work, this removes a `raw.githubusercontent.com` dependency from every provider build, which is a solid reliability win. I verified the bundled unit matches the canonical upstream `containerd.service` and the currently pinned v2.3.2, and the Dependabot workflow guards look right (actor + same-repo head + file-scope check).

One request before I lgtm: `verify-containerd-service-template` isn't wired into any presubmit, and the auto-update workflow only fires for `dependabot[bot]`. So if someone bumps `containerd_version` manually, the bundled unit silently drifts with nothing to catch it. Rather than a new test-infra job, the low-friction option is to call `make verify-containerd-service-template` from `images/capi/scripts/ci-packer-validate.sh`. That script backs the `pull-packer-validate` presubmit, whose `run_if_changed` already covers `images/capi/packer/config/containerd.json`, so a manual version bump would fail fast. (The verify step fetches the unit from GitHub, but only in CI, never during an image build.) I'm happy to help wire that up.

One minor, non-blocking note: since the bundled unit has no template variables, `ansible.builtin.copy` would be equivalent to `ansible.builtin.template` and avoids Jinja interpreting the file if upstream ever adds `{{`/`{%`. Feel free to leave it as-is.

images/capi/ansible/roles/containerd/tasks/main.yml
```diff
@@ -33,11 +33,19 @@
...
 
-- name: Download containerd.service
+- name: Download containerd.service from URL
   ansible.builtin.get_url:
     url: "{{ containerd_service_url }}"
     dest: /tmp/containerd.service
     mode: "0600"
+  when: containerd_service_url | default("") | length > 0
+
+- name: Create containerd.service from template
```
drew-viles, line 60: Could we keep the service unit tied to the effective `containerd_version`? With `containerd_service_url` now empty, overriding only the version installs the bundled v2.3.2 unit regardless of the selected binary. That changes the previous behavior, where the URL followed the overridden version. For example, upstream v1.7.29 includes `local-fs.target` ordering and `LimitNOFILE=infinity`, while this bundled unit does not. Could we reject a non-pinned version without a matching service URL, or retain a version-matched fallback?

images/capi/hack/update-containerd-service-template.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    version = containerd_version(CONTAINERD_CONFIG)
+    ppc64le_version = containerd_version(PPC64LE_CONTAINERD_CONFIG)
+    if version != ppc64le_version:
```
drew-viles: Could we avoid requiring the architecture pins to be identical? They have legitimately diverged in this repository—for example, commit `452f98e` updated the generic pin to v1.7.29 while ppc64le remained on v1.7.25. In that state this verifier, and therefore the update workflow, exits before it can refresh anything. Could we support a version-specific template/fallback or explicitly verify that the upstream units are equivalent when the versions differ?

drew-viles (commented):
Minor notes

mboersma (changes requested):
Could we fix two updater consistency issues before merging?

The update script now modifies both the service template and `defaults/main.yml`, but the workflow only permits and commits the template.

Also, the updater accepts equivalent architecture-specific version differences but provisioning rejects anything that differs from the generic version pin. These paths should use consistent version rules.

mboersma (approved):
/lgtm
/approve

PR #2104, opened 2026-07-06
capi: initialize GCE CI test status

MaxRink:
## What
- Initializes `test_status` before the GCE CI cleanup trap can read it.
- Preserves the `su - packer` build exit status under `errexit` instead of exiting before `test_status` is assigned.
- Guards cleanup when `GCP_PROJECT`, `BOSKOS_HOST`, or `RESOURCE_NAME` are unavailable.
- Skips post-build image listing after failed builds so cleanup exits with the original build failure.

## Validation
- `git diff --check`
- `bash -n images/capi/scripts/ci-gce.sh`
- `shellcheck images/capi/scripts/ci-gce.sh`

Plain ShellCheck still reports existing source/include and export-assignment warnings unrelated to this change.

### DIFF
--- a/images/capi/scripts/ci-gce.sh
+++ b/images/capi/scripts/ci-gce.sh
@@ -28,6 +28,8 @@ set -o pipefail
 CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
+test_status=0
+
 # shellcheck source=ensure-go.sh
 source "./hack/ensure-go.sh"
 # shellcheck source=ensure-boskosctl.sh
--- a/images/capi/scripts/ci-gce.sh
+++ b/images/capi/scripts/ci-gce.sh
@@ -43,6 +43,9 @@ function boskosctlwrapper() {
 }
 
 cleanup() {
+  # Capture the status that triggered this EXIT trap before running any other
+  # command, otherwise it gets clobbered by the cleanup steps below.
+  local trap_status=$?
   echo "Cleaning up image"
   if [ -n "${GCP_PROJECT:-}" ]; then
     filter="name~cluster-api-ubuntu-*"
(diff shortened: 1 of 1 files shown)


images/capi/scripts/ci-gce.sh
```diff
@@ -28,6 +28,8 @@ set -o pipefail
 CAPI_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${CAPI_ROOT}" || exit 1
 
+test_status=0
```
drew-viles, line 31: Could we preserve the status that triggered the `EXIT` trap? With `test_status` initialized to `0`, `cleanup` exits successfully for failures outside the guarded `su` command. For example, an expired GCP key or a failing post-build image lookup reaches `exit "${test_status}"` and turns the failure into exit 0. I reproduced the same Bash behavior with an `EXIT` trap around `false`. Capturing `$?` as the first statement in `cleanup` would preserve those failures while still allowing the guarded build status to be returned.

images/capi/scripts/ci-gce.sh
```diff
@@ -42,21 +44,23 @@ function boskosctlwrapper() {
...
+      --no-standard-images --format="table[no-heading](name)" --filter="${filter}" \
+      | awk '{print "gcloud compute images delete --quiet --project '"$GCP_PROJECT"' "$1" " "\n"}' \
+      | bash ) || true
+  fi
 
   # stop boskos heartbeat
-  if [ -n "${BOSKOS_HOST:-}" ]; then
-    boskosctlwrapper release --name "${RESOURCE_NAME}" --target-state dirty
+  if [ -n "${BOSKOS_HOST:-}" ] && [ -n "${RESOURCE_NAME:-}" ]; then
+    boskosctlwrapper release --name "${RESOURCE_NAME}" --target-state dirty || true
```
drew-viles: Could we ignore a release failure only when we are already preserving a non-zero job status? If the build succeeds but Boskos rejects or times out the release, `|| true` plus the final `exit 0` makes the job green and leaves the project busy until external recovery. It would be useful to retain the original job failure when one exists, but return the release error after an otherwise successful run.

drew-viles (commented):
Minor notes

drew-viles (commented):
/lgtm
/assign @mboersma

mboersma (approved):
/approve

PR #2105, opened 2026-07-06
capi: isolate OVA CI packer caches

MaxRink:
## What this does
- gives each parallel OVA CI target its own local `PACKER_CACHE_DIR`
- keeps the cache under `/tmp` by default so it is not uploaded as a job artifact
- scopes `PACKER_VAR_FILES` and `PACKER_LOG` to each target subprocess

## Why
`pull-ova-all` can run targets in parallel. A failed run showed Flatcar selecting a cached ISO path and then immediately failing with `stat /root/.cache/packer/...iso: no such file or directory`, while the other targets completed. Isolating target caches avoids cross-target or stale global cache state.

## Validation
- `shellcheck images/capi/scripts/ci-ova.sh` reports existing warnings in unchanged lines


mboersma (approved):
/approve

PR #2106, opened 2026-07-06
capi: add Kubernetes minor version matrix

MaxRink:
## Change description

Adds a canonical per-Kubernetes-minor dependency matrix for CAPI builds:

- release-pinned entries in `images/capi/packer/config/kubernetes-version-matrix.yaml`
- a rolling latest entry in `images/capi/packer/config/kubernetes-version-latest.yaml`
- `images/capi/hack/kubernetes-version-matrix.py` to render entries as Packer var JSON, verify the matrix, refresh Kubernetes/package-manager pins from upstream release metadata, and sync the matrix from generated Dependabot tracking manifests
- generated Go module manifests under `images/capi/packer/config/kubernetes-version-dependencies/` so Dependabot can track Kubernetes, CNI plugins, containerd, runc, and cri-tools versions
- a scoped `pull_request_target` workflow that only handles same-repository Dependabot PRs changing those generated tracking manifests, then regenerates the YAML matrix on the PR branch
- docs showing how to consume the rendered JSON with `PACKER_VAR_FILES` and how the Dependabot tracking flow works

This is additive and does not change the default `kubernetes.json`, `cni.json`, or `containerd.json` behavior. Kubernetes release tracking uses `k8s.io/client-go` module tags and maps those `v0.x.y` module versions back to Kubernetes `v1.x.y` matrix values.

## Related issues

- None

## Validation

- `python3 -m py_compile images/capi/hack/kubernetes-version-matrix.py`
- `python3 images/capi/hack/kubernetes-version-matrix.py verify`
- `python3 images/capi/hack/kubernetes-version-matrix.py sync-tracking`
- `python3 images/capi/hack/kubernetes-version-matrix.py render 1.31 | jq -e ...`
- `python3 images/capi/hack/kubernetes-version-matrix.py render latest | jq -e ...`
- `ruby -e 'require "yaml"; YAML.load_file(".github/dependabot.yml"); YAML.load_file(".github/workflows/update-kubernetes-version-matrix.yaml")'`
- temporary-copy `go list -mod=mod -m all` for each generated tracking `go.mod`
- `git diff --check`


images/capi/packer/config/kubernetes-version-matrix.yaml
```diff
@@ -0,0 +1,102 @@
...
+#   images/capi/hack/kubernetes-version-matrix.py render 1.36
+releasePins:
+  "1.31":
+    containerd_version: "2.3.1"
+    crictl_version: "1.31.1"
+    kubernetes_cni_deb_version: "1.5.1-1.1"
+    kubernetes_cni_http_source: "https://github.com/containernetworking/plugins/releases/download"
+    kubernetes_cni_rpm_version: "1.5.1"
+    kubernetes_cni_semver: "v1.5.1"
+    kubernetes_cni_source_type: "pkg"
```
drew-viles: Could we keep the source-type fields out of this shared version var file, or otherwise preserve target-specific values? `PACKER_VAR_FILES` is appended after the target var file, so for a real target such as `build-ami-flatcar`, the rendered matrix overrides `flatcar.json`’s `http` values back to `pkg`. Flatcar matches none of the package-install imports, while URL installation requires both values to be `http`, so Kubernetes and CNI are not installed. I reproduced the effective values from the real Flatcar target plus the rendered `1.31` matrix, and both resolve to `pkg`.

images/capi/hack/kubernetes-version-matrix.py
```diff
@@ -0,0 +1,614 @@
...
+def validate_entry(selector: str, entry: dict[str, Any]) -> list[str]:
+    errors = []
+    missing = [key for key in REQUIRED_KEYS if key not in entry]
+    if missing:
+        errors.append(f"{selector}: missing keys: {', '.join(missing)}")
+        return errors
+
+    kubernetes_semver = entry["kubernetes_semver"]
+    kubernetes_version = kubernetes_semver.removeprefix("v")
+    kubernetes_minor = ".".join(kubernetes_version.split(".")[:2])
```
drew-viles, line 505: Could we also check that `kubernetes_minor == selector` for release-pinned entries? The validator currently checks only that the fields agree with one another. I changed the `1.31` entry into a self-consistent `v1.32.13` entry in memory and `validate_entry("1.31", entry)` still returned no errors. Because the tracking manifest is rendered from the same entry, verification can accept a `release-1-31` pin that actually builds Kubernetes 1.32.

drew-viles (commented):
Minor notes

mboersma (changes requested):
This is looking good, but I think it needs to regenerate the pinned versions. For example, it's got Kubernetes 1.36.2 whereas 1.36.3 is now current,  containerd 2.3.1 instead of 2.3.2, and runc 1.4.2 instead of 1.4.3.

mboersma (approved):
/lgtm
/approve

PR #2109, opened 2026-07-07
capi/qemu: skip shell cleanup during reboot handoff

MaxRink:
#### What this PR does
- Leaves the intermediate QEMU reboot as the existing foreground `sudo reboot now` shell provisioner.
- Adds `skip_clean: true` to the reboot provisioner and the first post-reboot provisioner so Packer does not try to remove temporary shell scripts while SSH is dropping or recovering.
- Extends the post-reboot wait and SSH retry window before the existing `systemctl is-system-running --wait` check runs.

#### Why
When the guest reboots, SSH can disappear after the reboot command succeeds but before Packer removes the uploaded shell script. That cleanup failure can make a successful reboot look like a provisioner failure. Skipping cleanup across this reboot handoff keeps reconnect and readiness validation in the post-reboot provisioner.

#### Validation
- `git diff --check`
- `cd images/capi && make validate-qemu-ubuntu-2404-cloudimg`

### DIFF
--- a/images/capi/packer/qemu/packer.json.tmpl
+++ b/images/capi/packer/qemu/packer.json.tmpl
@@ -102,7 +102,7 @@
     {
       "expect_disconnect": true,
       "inline": [
-        "sudo reboot now"
+        "sudo systemd-run --unit=image-builder-reboot --on-active=1s /usr/sbin/reboot"
       ],
       "inline_shebang": "/bin/bash -e",
       "type": "shell"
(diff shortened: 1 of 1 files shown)


mboersma:
The change looks reasonable, but the title and description don't quite match the diff. 

Nothing here actually schedules the reboot through systemd or makes it asynchronous. It's still `sudo reboot now` in the foreground. The real fix is `skip_clean: true`, which stops Packer from hanging while it tries to clean up its temp script over the dying SSH connection, plus the longer `pause_before` and `start_retry_timeout` for the reconnect. 

Could you reword the title and description to reflect that? Otherwise the git history will be a bit misleading.

drew-viles (commented):
/ok-to-test

mboersma (approved):
/lgtm

mboersma (approved):
/approve

PR #2110, opened 2026-07-07
Guard OVA Boskos cleanup before resource acquire

MaxRink:
#### What this PR does
- Makes the OVA CI cleanup trap release Boskos resources only after a resource name was acquired.
- Handles optional Prow job metadata safely under `set -u`.
- Captures Boskos acquire failures explicitly before loading the generated environment file.

#### Why
When OVA CI exits before acquiring a Boskos resource, the cleanup trap can fail on unset resource variables and obscure the original error. The trap should preserve the primary failure while still releasing resources that were actually checked out.

#### Validation
- `git -c core.fsmonitor=false diff --check`
- `shellcheck -e SC2207,SC2155,SC2001 images/capi/scripts/ci-ova.sh`


mboersma (approved):
/lgtm
/approve

PR #2111, opened 2026-07-07
capi: refresh Microsoft apt key after setup packages

MaxRink:
## What this does
- Reuses the legacy Microsoft apt repository key check through an included task file.
- Runs the check before apt updates and again after baseline package setup, before later roles trigger apt cache refreshes.

## Why
Ubuntu Azure images can expose a legacy packages.microsoft.com apt source after early package setup. If the key is not refreshed after that point, later apt cache updates fail with NO_PUBKEY for the Microsoft repository.

## Validation
- git diff --check
- ansible-playbook --syntax-check ansible/node.yml -e packer_build_name=sig-ubuntu-2604-gen2 -e packer_builder_type=azure-arm -e ansible_python_interpreter=/usr/bin/python3
- ansible-playbook --syntax-check ansible/firstboot.yml -e packer_build_name=sig-ubuntu-2604-gen2 -e packer_builder_type=azure-arm -e ansible_python_interpreter=/usr/bin/python3

### DIFF
--- a/images/capi/ansible/roles/setup/tasks/debian.yml
+++ b/images/capi/ansible/roles/setup/tasks/debian.yml
@@ -42,26 +42,8 @@
     mode: "0644"
   when: http_proxy is defined or https_proxy is defined
 
-- name: Trust Microsoft key for legacy Microsoft apt repositories
-  block:
-    - name: Check for legacy Microsoft apt source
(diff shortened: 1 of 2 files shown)


mboersma (approved):
/lgtm
/approve

PR #2112, opened 2026-07-07
capi: normalize crictl archive ownership

MaxRink:

Ensures the crictl binary extracted from the upstream cri-tools archive is installed as `root:root` with executable mode. The archive currently carries its build-time numeric owner, so images can end up with `/usr/local/bin/crictl` owned by that numeric UID/GID.

None

#### Testing

- `git diff --check`
- `ansible-playbook --syntax-check ansible/node.yml -e packer_build_name=qemu -e packer_builder_type=qemu -e ansible_python_interpreter=/usr/bin/python3`


mboersma (approved):
/lgtm
/approve

PR #2117, opened 2026-07-07
Fix Windows Azure OSProvisioningTimedOut on sysprep

mboersma:
Windows Azure images built by image-builder intermittently fail to provision with `OSProvisioningTimedOut`. After `sysprep /generalize`, the built-in Administrator is left flagged "user must change password at next logon", so on first boot the OOBE auto-logon blocks on the LogonUI password-change dialog and cloudbase-init never signals provisioning-ready. Because it is a timing race, the same image sometimes provisions and sometimes times out.

This disables the built-in Administrator (its normal state on Azure; cloudbase-init creates and renames the real admin in the post-OOBE phase) and sets max password age to unlimited so the packer build user credential cannot expire during a slow validation. Verified against Windows Server 2022 and 2025 builds.


drew-viles (commented):
/lgtm
/approve

PR #2119, opened 2026-07-12
capi: fix swap removal in qemu Ubuntu autoinstall templates

hajowieland:
## Change description

qemu autoinstall user-data tries to disable and remove the swap file via late-commands, but they seem to not work correctly.

- no `curtin in-target` wrapper (like for raw)
- remove `/swapfile` but the file to be removed is at `/swap.img`

This change mirrors the config already in place for the raw Ubuntu templates.

## Related issues

--

## Additional context

Was currently masked by Ansible `node` role which also disables swap and `sysprep` role deleting `/swap.img` and `/swapfile`.

### DIFF
--- a/images/capi/packer/qemu/linux/ubuntu/http/22.04.efi.qemu/user-data.tmpl
+++ b/images/capi/packer/qemu/linux/ubuntu/http/22.04.efi.qemu/user-data.tmpl
@@ -101,8 +101,8 @@ autoinstall:
   late-commands:
     - apt install -y efibootmgr
     - efibootmgr -o $(efibootmgr | grep -oP 'Boot[0-9]+[*]\s+ubuntu' | grep -oP '000[0-9]+' | head -n 1)
-    - swapoff -a
-    - rm -f /swapfile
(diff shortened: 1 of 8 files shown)


mboersma:
Nice catch, thanks for cleaning this up. Two small things:

1. The `raw` templates keep a trailing `curtin in-target --target=/target -- rm -rf /var/lib/apt/lists/*` in addition to `apt-get clean`. This PR drops that line, so it's not quite a full mirror, and the numbered comment still says "5. Removes the cached list of packages." Could you either keep that line or update the comment?

2. Since this was already masked by the `node` and `sysprep` roles, the shipped images weren't affected, so this is really a correctness and consistency fix. Have you been able to run a qemu Ubuntu build to confirm it still passes?

Otherwise this looks good to me.

mboersma (changes requested):
Could you update the active `24.04.immutable` template as well? It still removes `/swapfile`, but curtin creates `/swap.img`. 

Also, please either restore the package-list cleanup or update the comments that say those lists are removed.

PR #2004, opened 2026-05-10
feat: adding containerd_enable_limit_no_file as an option to allow users to add this file on demand as required.

drew-viles:
## Change description
I've hit an issue in our OpenStack image builds where we need this enabling but it's currently conditionally enabled for Marina, Flatcar or Azure.

This adds a new OR condition to allow people to enable this on more than those three options. Rather than changing the flow, which would break existing users implementation, adding the OR made more sense to prevent a regression.

The default for this new value is false to keep everything as it is now and making it purely opt-in.

Added Goss tests and udpated the docs too.

PR #2122, opened 2026-07-15
🚑 Add LimitNOFILE=1048576 configuration to all OS families

### LINKED ISSUE #2121: ULimit regression on Ubuntu images

Due to Containerd 2.X upgrade, Ulimit configuration (at least on Ubuntu) has changed and is now 1024 instead of 1048576.

I expect it to be working as before and be configured to 1048576.

This problem is critical and can cause downtime to a lot of applications.

---

/kind bug

Whisper40:
ContainerD V2 removed LimitNOFILE from his service. It is a regression because now the value is defined as 1024.

It is a problem because it breaks a lot of apps like (mysql/.net/kafka/s3)

We remove this "when" condition, so the limit-nofile.conf will be propagated to every OS.

It will not impact other distributions because the goal of this "when" condition was to remove "LimitNOFILE=infinity"

This value "1048576" was already defined in containerd.service upstream.

## Change description

- Is this change including a new Provider or a new OS? (y/n) no
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) no
- If adding a new provider, are you a representative of that provider? (y/n) no

## Related issues

- Fixes #2121 

Linked to : https://github.com/kubernetes-sigs/image-builder/pull/1720

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -154,7 +154,6 @@
     dest: /etc/systemd/system/containerd.service.d/limit-nofile.conf
     src: etc/systemd/system/containerd.service.d/limit-nofile.conf
     mode: "0644"
-  when: ansible_os_family in ["Common Base Linux Mariner", "Flatcar", "Microsoft Azure Linux"]
 
 - name: Create containerd http proxy conf file if needed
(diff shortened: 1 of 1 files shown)


Whisper40:
Oh it seems already fixed in https://github.com/kubernetes-sigs/image-builder/pull/2004 (in another way).
I will close

PR #2127, opened 2026-07-20
Revert "capi: use dpkg selections in Debian sysprep"

### LINKED ISSUE #2126: periodic-image-builder-gcp-all-nightly failing since Jul 3 due to sysprep dpkg_selections loop doubling build times
## Summary

The `periodic-image-builder-gcp-all-nightly` job has been failing consistently since July 3, 2026, hitting the 4h Prow timeout.
Each individual packer build roughly doubled in duration (from about 10 min to about 20 min) after the sysprep package-pinning task was changed to loop over packages individually, pushing the total job time from around 160 min to around 255 min.

## Potential Root Cause

The sysprep role's package-pinning task in `images/capi/ansible/roles/sysprep/tasks/debian.yml` was updated in #2055 to use `ansible.builtin.dpkg_selections` in a loop, replacing the previous batched shell command. The change itself is reasonable (using a proper Ansible module instead of a shell pipe), but it has an unintended performance side effect when applied to the full package list:

```yaml
- name: Pin all installed packages
  ansible.builtin.dpkg_selections:
    name: "{{ item }}"
    selection: hold
  loop: "{{ ansible_facts.packages.keys() }}"
```

On a typical Ubuntu image there are 670 to 697 installed packages. Each iteration requires its own SSH round-trip and `dpkg --set-selections` invocation, adding roughly 8 to 10 minutes per image build compared to the previous single-command approach.

## Evidence

| | Jul 2 (last SUCCESS) | Jul 3 (first FAILURE) |
|---|---|---|
| Commit | `14f7205538fd` | `0dc9c28452d9` |
| Task name | "Pin all installed packages with apt-mark" | "Pin all installed packages" |
| Mechanism | Single shell command (`changed: [default]`) | Per-package loop (`changed: [default] => (item=...)` x 670+ items) |
| Per-image build time | ~10 min | ~19-20 min |
| Total job time (15 builds) | 159 min (SUCCESS) | 255 min (FAILURE, timeout at build 13/15) |

Jul 2 Ansible output for the pin task (single command, one line):
```
TASK [sysprep : Pin all installed packages with apt-mark] **********************
changed: [default]
```

Jul 3 Ansible output for the same step (per-package loop, 670+ lines):
```
TASK [sysprep : Pin all installed packages] ************************************
changed: [default] => (item=adduser)
changed: [default] => (item=apparmor)
changed: [default] => (item=apt)
changed: [default] => (item=apt-transport-https)
changed: [default] => (item=apt-utils)
... (665 more items)
```

Full build logs:
- Jul 2 (SUCCESS): https://storage.googleapis.com/kubernetes-ci-logs/logs/periodic-image-builder-gcp-all-nightly/2072696494179225600/build-log.txt
- Jul 3 (FAILURE): https://storage.googleapis.com/kubernetes-ci-logs/logs/periodic-image-builder-gcp-all-nightly/2073058887694553088/build-log.txt

The job has failed every day since July 3 (17 consecutive failures as of July 20). On July 15 it failed after only 82 min due to a separate issue, but all other failures are timeout-related.

## Impact

- No new nightly GCP images have been successfully published since July 2
- CAPG CI jobs that depend on nightly images are running on stale images
- With `overwrite-1-35.json` now added on `main`, the matrix is 3 OS variants x 5 K8s versions = 15 builds, requiring around 300 min at current per-build times, well beyond the 4h timeout

## Potential Fix

Restore batched package pinning. Some options:

1. **Revert to the shell approach** - the original `dpkg-query | xargs apt-mark hold` was a single command and worked reliably
2. **Batch via `dpkg --set-selections`** - if using dpkg selections is preferred over apt-mark, pipe all packages into a single `dpkg --set-selections` call instead of looping
3. **Bump the timeout** - this only treats the symptom and will not scale as more OS variants or K8s versions are added

## References

- Failing job: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/periodic-image-builder-gcp-all-nightly
- Job definition: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-k8s-infra/trusted/image-builder/image-builder-periodics.yaml
- Related PR: #2055
- Sysprep task: `images/capi/ansible/roles/sysprep/tasks/debian.yml`

/sig cluster-lifecycle
/area provider/gcp

mboersma:
Reverts #2055 (commit 179549f), which doubled GCP nightly build times by looping `dpkg_selections` over every installed package. 

Fixes #2126.

### DIFF
--- a/images/capi/ansible/roles/sysprep/defaults/main.yml
+++ b/images/capi/ansible/roles/sysprep/defaults/main.yml
@@ -16,7 +16,6 @@ extra_repos: ""
 pip_conf_file: ""
 remove_extra_repos: false
 flatcar_disable_autologin: false
-sysprep_hold_installed_packages: true
 sysprep_require_grub_file: true
 defer_ssh_cleanup_to_packer: false
 extra_kernel_boot_params: ""
--- a/images/capi/ansible/roles/sysprep/tasks/debian.yml
+++ b/images/capi/ansible/roles/sysprep/tasks/debian.yml
@@ -45,31 +45,32 @@
   when: "'unattended-upgrades' in ansible_facts.packages"
 
 # The apt-daily/unattended-upgrades services may still be running from before
-# they were disabled above. Retry package selection changes until the dpkg frontend
+# they were disabled above. Retry the apt-mark calls until the dpkg frontend
 # lock is released to avoid a flaky race with concurrent apt activity.
-- name: Pin all installed packages
-  ansible.builtin.dpkg_selections:
-    name: "{{ item }}"
-    selection: hold
-  loop: "{{ ansible_facts.packages.keys() }}"
-  register: sysprep_dpkg_selections_hold
-  until: sysprep_dpkg_selections_hold is not failed
-  retries: 30
-  delay: 10
-  when: sysprep_hold_installed_packages | default(true) | bool
+- name: Pin all installed packages with apt-mark
+  ansible.builtin.shell: |
+    set -o pipefail
+    dpkg-query -f '${binary:Package}\n' -W | xargs apt-mark hold
 
-- name: Unpin grub packages for MaaS
-  ansible.builtin.dpkg_selections:
-    name: "{{ item }}"
-    selection: install
-  loop: "{{ ansible_facts.packages.keys() | select('search', 'grub') | list }}"
-  register: sysprep_dpkg_selections_unhold_grub
-  until: sysprep_dpkg_selections_unhold_grub is not failed
+  args:
+    executable: /bin/bash
+  register: sysprep_apt_mark_hold
+  until: sysprep_apt_mark_hold.rc == 0
   retries: 30
   delay: 10
-  when:
-    - sysprep_hold_installed_packages | default(true) | bool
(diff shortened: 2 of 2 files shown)


mboersma:
/override pull-ova-all

RockyLinux-9 failed due to a (presumably transient) upstream mirror issue:
```
Depsolve Error: package python-unversioned-command-3.9.25-7.el9_8 requires python3 = 3.9.25-7.el9_8...
cannot install both python3-3.9.25-7.el9_8.2.x86_64 (baseos) and python3-3.9.25-7.el9_8.x86_64 (@System)
```

damdo:
Can we somehow run the periodic that was failing to prove this revert fixes it?

mboersma:
> Can we somehow run the periodic that was failing...?

The job can be retriggered on the job page by maintainers, but it has `base_ref: main` so it won't pick up this change until it's merged. But I think your analysis was correct and this ought to fix the periodic gcp job.

damdo:
Ack thanks!

/lgtm

mboersma (commented):
/approve

PR #1946, opened 2026-03-01
Add AlmaLinux 9 OVA build

aibarbetta:
## Change description

Adds support for building AlmaLinux 9 OVA images, following the same patterns used by the existing RockyLinux and RHEL builds.

### How I tested this
I built the AlmaLinux 9 OVA successfully against my own vCenter environment and confirmed the resulting image boots and joins a CAPI-managed cluster.

### This PR introduces a new OS
- Is this change including a new Provider or a new OS? (y/n) y
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) y

## Related issues

None

## Additional context

Hi sig-cluster-lifecycle 👋
I'm Agus, I've been doing small contributions to sig-storage and sig-cli, and I'm currently part of the v1.36 Release Team. 
A couple of months ago, I started using CAPI and image-builder in my day job. We have a few vCenter/vSphere appliances, and we decided to use CAPV and AlmaLinux as our base OS. I forked image-builder and added the AlmaLinux 9 OVA build; this is currently working for our infrastructure.
I'd like to contribute this work, AlmaLinux is a widely adopted RHEL-compatible distribution, and having official OVA support feels like a natural addition alongside the existing RockyLinux builds. 

I'm very new to image-builder, so I don't know if additional changes are needed. I modeled everything closely after the RockyLinux 9 implementation, but I'd appreciate guidance from maintainers on any changes needed to make this mergeable.

Happy to iterate on feedback. Thanks for your time!

/kind feature

PR #2132, opened 2026-07-31
Add AlmaLinux 9 AMI build

aibarbetta:
## Change description

Adds support for building AlmaLinux 9 AMI. I did this previously for OVA at https://github.com/kubernetes-sigs/image-builder/pull/1946

New `packer/ami/almalinux-9.json` sources the AlmaLinux OS Foundation's official images (owner `764336703387`).

AlmaLinux is the first non-Amazon distro to use the `amazon-ebs` builder, which surfaced three places where existing code used "which Packer builder ran" as a proxy for "which OS is this". Each needed a targeted condition fix:

- **`setup/tasks/redhat.yml`**: EPEL tasks were gated on `packer_builder_type.startswith('amazon')`. On AlmaLinux that runs `amazon-linux-extras` (an Amazon Linux 2 tool) and skips the generic `redhat_epel_rpm` path. Now keyed on `ansible_facts['distribution']`. EPEL is required here: `python3-netifaces` from `rh8_rpms` is in neither AlmaLinux 9 BaseOS nor AppStream.
- **`providers/tasks/awscliv2.yml`**: prerequisites were gated on `distribution == "RedHat"` (strictly RHEL), so AlmaLinux skipped them and the following `gpg --import`/unzip steps failed. Widened to `os_family`, and `gnupg` → `gnupg2` (EL8/EL9 have no `gnupg` package). This task was previously unreachable, which is why the wrong package name went unnoticed.
- **`packer/goss/goss-vars.yaml`**: `almalinux` had no `amazon` section, so `index .Vars .Vars.OS .Vars.PROVIDER` returned nil and the goss templates errored after provisioning completed. Added, mirroring `rockylinux.amazon`.

**No behavior change to existing builds.** `amazon-ebs` is the only builder starting with `amazon`, and Amazon Linux is built only under `packer/ami/`, so the old and new conditions are truth-table identical for every current distro/provider combination.

- Is this change including a new Provider or a new OS? (y/n) y
- If yes, has the Provider/OS matrix been updated in the readme? (y/n) y
- If adding a new provider, are you a representative of that provider? (y/n) -

## Additional context

This work was verified with a real build in an AWS account I manage:
- `make validate-ami-almalinux-9` → `The configuration is valid.`
- `make build-ami-almalinux-9` → succeeded
- Ansible: `ok=115 changed=82 unreachable=0 failed=0`
- goss: `Count: 55, Failed: 0, Skipped: 0`

I'm still new to image-builder, and I did this with Claude assistance! I'd appreciate guidance from maintainers on any changes needed to make this mergeable.

Happy to iterate on feedback. Thanks for your time!

/kind feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The table below shows the currently provided operating systems for each provider
 
 | OS                | ami | azure | digitalocean | gce | hcloud | huaweicloud | maas | nutanix | oci | openstack | outscale | ova | oxide | powervs | proxmox | qemu | raw | scaleway | vultr |
 |-------------------|----|----|----|----|----|----|----|----|----|----|---|----|----|----|----|----|----|----|----|
-| AlmaLinux 9       | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| AlmaLinux 9       | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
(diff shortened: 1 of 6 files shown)


mboersma (approved):
/lgtm
/assign @drew-viles

drew-viles (commented):
/lgtm
/approve

PR #2136, opened 2026-08-09
capi: retry shared Debian apt lock contention

MaxRink:
Azure image builds repeatedly fail when background apt activity outlives the module lock timeout.

This adds the existing retry pattern to the Azure provider package installs and sysprep apt cache cleanup, covering the shared provider/sysprep lock failures without changing existing PR branches or weakening apt error handling.

Validation:
- ansible-playbook --syntax-check ansible/node.yml
- ansible-lint --profile min on changed task files
- git diff --check


mboersma (approved):
/lgtm
/approve

PR #2137, opened 2026-08-09
capi: remove stale Microsoft prod apt source

MaxRink:
Ubuntu 26.04 Azure images can contain an obsolete Microsoft Ubuntu 18.04 prod source. Its missing signing key causes apt updates to fail with NO_PUBKEY EB3E94ADBE1229CF.

Remove only matching stale Microsoft 18.04 prod source files before the existing Microsoft key setup, preserving signature verification for current sources.

Validation:
- ansible-playbook --syntax-check ansible/node.yml
- ansible-playbook --syntax-check ansible/firstboot.yml
- ansible-lint --profile min on the changed task
- git diff --check


mboersma (approved):
/lgtm
/approve

PR #2138, opened 2026-08-14
capi: retry upstream artifact downloads

MaxRink:

The build fetches release artifacts from github.com and storage.googleapis.com. One read timeout stops the whole packer build. Unrelated pull requests then show a red check.

Example from a recent `pull-azure-sigs` run:

```
TASK [kubernetes : Download crictl]
fatal: [default]: FAILED! => {"changed": false, "elapsed": 10,
  "msg": "Connection failure: The read operation timed out",
  "url": ".../cri-tools/releases/download/v1.36.0/crictl-v1.36.0-linux-amd64.tar.gz.sha256"}
```

This change adds a retry loop to the `get_url` tasks in the node build path:

- kubernetes role: crictl, CNI tarball, Kubernetes binaries, Kubernetes images
- containerd role: containerd, containerd.service, runc, wasm shims, gvisor runsc

Three tasks already set `retries` without `until`. Ansible sets `retries` to 1 when `until` is absent, so those retries never run. This change adds `until` to them:

- load_additional_components role: additional executables, additional images
- gpu role: Flatcar nvidia-runtime sysext image

The retry pattern matches the one for apt lock contention in `roles/kubernetes/tasks/debian.yml`.

The downloads keep their checksum. A retry cannot hide a corrupt artifact.

## How this was verified

- `ansible-lint` reports 0 failures on the changed files.
- The changed files parse as valid YAML.

The failure appears on #2072, #2080, #2087 and #2098. None of these pull requests change the download tasks.

## Release note

```release-note
Retry upstream artifact downloads so a transient network failure does not stop an image build.
```

### DIFF
--- a/images/capi/ansible/roles/containerd/tasks/main.yml
+++ b/images/capi/ansible/roles/containerd/tasks/main.yml
@@ -32,12 +32,20 @@
     checksum: sha256:{{ containerd_url }}.sha256sum
     dest: /tmp/{{ containerd_filename }}
     mode: "0600"
+  register: containerd_download
+  until: containerd_download is not failed
+  retries: 5
+  delay: 3
 
 - name: Download containerd.service
   ansible.builtin.get_url:
     url: "{{ containerd_service_url }}"
     dest: /tmp/containerd.service
     mode: "0600"
+  register: containerd_service_download
+  until: containerd_service_download is not failed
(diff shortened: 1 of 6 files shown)


mboersma (approved):
/lgtm
/approve

PR #2139, opened 2026-08-16
feat: migrate nutanix linux to HCL

fad3t:
## Change description
Migrate the Nutanix Linux Packer JSON template into HCL.

## Additional context
Following changes have been applied:
- new HCL files replacing the `packer.json.tmpl` file -- using the standard Packer folder structure (variables, locals, ..)
- the HCL variables have been defined in the Nutanix folder for now, even if they're shared across providers. JSON vars don't need to be declared upfront, so other providers don't require this for now. Once all providers have been migrated to HCL, we should be able to simplify this part.
- the old `packer.json.tmpl` file has been removed
- the `.tmpl` is no longer needed as the bash script writes the password into a `pkrvars` file directly
- I kept Windows out of scope for now. I'd personally make a dedicated `windows/` directory containing all the Windows-related HCL files to avoid naming conflicts with Linux variables, but I kept it for later so people can share their opinion/feedback on this.
- The flatcar build name has been simplified, as the channel and version variables seem unused in the Nutanix context. Please correct me if I'm wrong.

Misc:
- I tested a build on a Nutanix setup, image was successfully built.
- Some code has been AI-generated, mostly the repetitive copy-paste and vars descriptions.
- Idea would be to continue the migration effort, however I can only test a few providers.. Let me know if you'd like to contribute.


tuxtof (approved):
LGTM

@mboersma are we OK to start migrating templates to HCL, is it a request for the other providers too ?

adarshanand25 (approved):

mboersma:
>  are we OK to start migrating templates to HCL?

@tuxtof I don't see why not, it's what Packer has recommended for a while. I don't know that we need to convert everything, but I think it's fine to work in that direction.

mboersma (changes requested):
Thanks for working on this migration! I noticed two compatibility issues: the Linux build and validation commands no longer load `packer/nutanix/nutanix.json`, even though that’s the documented configuration path, and `nutanix_port` / `nutanix_insecure` no longer honor `NUTANIX_PORT` / `NUTANIX_INSECURE`. Could we preserve both behaviors so existing Nutanix build workflows continue to work?

tuxtof:
> > are we OK to start migrating templates to HCL?
> 
> @tuxtof I don't see why not, it's what Packer has recommended for a while. I don't know that we need to convert everything, but I think it's fine to work in that direction.

perfect so in this case , let's pursue 👍🏻

fad3t:
> Thanks for working on this migration! I noticed two compatibility issues: the Linux build and validation commands no longer load `packer/nutanix/nutanix.json`, even though that’s the documented configuration path, and `nutanix_port` / `nutanix_insecure` no longer honor `NUTANIX_PORT` / `NUTANIX_INSECURE`. Could we preserve both behaviors so existing Nutanix build workflows continue to work?

Thanks for the review!

For the `packer/nutanix/nutanix.json` you're right, I removed it because it was empty (thought it was a leftover) -- and I didn't know the documentation was explicitly telling users to set their Nutanix config/creds into this file (I'm personally building my own file that I pass to Packer).

Regarding `NUTANIX_PORT` and `NUTANIX_INSECURE` that depends. The migration script initially defined those 2 variables as string, using the environment variable as defaults. However, the Nutanix Packer provider expects number and bool respectively, so I preferred aligning with the expected type, which means we can't read the environment vars anymore..
That's a choice we'll have to make; either we preserve the current behavior and let Packer automatically convert the string into the target type (which it does very well as long as the input is clean) - or we take the opportunity to set the correct variable type (which I personally prefer). I also believe those 2 vars are not often overridden -- or at least the port number. Let me know what you prefer and I'll adapt :)

mboersma:
@fad3t let’s preserve compatibility while keeping the variables strictly typed. Could you fall back to converting `NUTANIX_PORT` and `NUTANIX_INSECURE` when the typed variables are unset, with explicit var-file/`-var`/`PKR_VAR_*` values taking precedence?

Please also restore `nutanix.json` for the Linux build and validation paths. We can document the `PKR_VAR_*` names as preferred going forward.

fad3t:
> @fad3t let’s preserve compatibility while keeping the variables strictly typed. Could you fall back to converting `NUTANIX_PORT` and `NUTANIX_INSECURE` when the typed variables are unset, with explicit var-file/`-var`/`PKR_VAR_*` values taking precedence?
> 
> Please also restore `nutanix.json` for the Linux build and validation paths. We can document the `PKR_VAR_*` names as preferred going forward.

To be sure I fully understood what you mean, you'd like to have something in between (a local) that'd use the typed variable if it is set, and fall back to the string variable (sourced from the env) otherwise?

fad3t:
I've updated as per @mboersma feedback - let me know if anything's missing!

mboersma (approved):
/lgtm
/approve

fad3t:
Are the failed tests related to my changes? Found this in logs.

```
==> azure-arm.sig-flatcar: ERROR: -> MarketplacePurchaseEligibilityFailed : Marketplace purchase eligibilty check returned errors. See inner errors for details. 
==> azure-arm.sig-flatcar: ERROR:   -> BadRequest
==> azure-arm.sig-flatcar: ERROR:   -> InvalidSubscriptionId : Invalid subscription identifier provided.
```

drew-viles:
I don't think it is - looks to be hitting a number of different pipelines now. I'm not actually sure what the fix is for this tbh - @mboersma do you have any ideas on this one? Onering if this is an upstream CI problem that needs reporting?

mboersma:
/override pull-azure-sigs

See #2166

fad3t:
Anything we can do about this failing CI check?

PR #2148, opened 2026-08-22
fix: preserve Dependabot tracking module requirements

MaxRink:
## Summary

Replaces the shared `tools.go` approach from the first revision of this PR. Every dependency the Kubernetes version matrix tracks now gets its own synthetic Go module per selector, so Dependabot can update each pin without the others moving.

## Root cause

dependabot-core's `go_modules` file updater always runs `go get X@v` followed by `go mod tidy`. The generated tracking manifests have no Go source, so tidy removed all five requirements and left behind a go.mod with nothing but a module line and a go directive. `sync-tracking` then failed with "tracking manifest missing modules" on every Dependabot PR (#2141, #2142, #2145, #2146).

Adding one `tools.go` per selector that blank-imported all five dependencies fixed the disappearing requirements but created a new problem, as raised in review: with five dependencies in one module, minimal version selection lets containerd's transitive requirements raise the older release-pinned CNI and client-go versions, and several of the imported packages need a newer go directive than the generated one.

Separately, `update --write` derived `kubernetes_cni_semver` from the kubernetes-cni RPM metadata, conflating the CNI plugins tarball with the distro package. Those version streams are independent, and the conflation makes `sync-tracking` fail outright on a real Dependabot PR: resolving CNI 1.6.2 against the Kubernetes 1.32 package repository raises "no kubernetes-cni RPM package '1.6.2' found", because Kubernetes 1.32 ships kubernetes-cni 1.6.0. The same class of bug affected the rolling entry more broadly: `refresh_matrix` seeded `latest` from the refreshed release pin for the same minor, so `update --write` also overwrote latest's independently tracked `containerd_version` and `runc_version` with the pin's values.

## Fix

The tracking tree becomes one module per dependency per selector, at `kubernetes-version-dependencies/<selector>/<dependency>/`, with `<dependency>` in containerd, cni-plugins, runc, kubernetes and cri-tools. Each go.mod holds exactly one direct requirement, so MVS has nothing to cross-contaminate, and each module's go directive is free to follow its own dependency.

Each module keeps a `tools.go` with a single blank import, which is what survives tidy. The imported package is the lightest one in the tracked module, chosen so that tidy records no indirect requirements at all: containerd/v2/version, plugins/pkg/utils/buildversion, runc/types/features, client-go/util/homedir and cri-tools/pkg/version. Every one of the 35 go.sum files is two lines.

`verify` now parses the manifests instead of comparing them against a rendered template, because Dependabot legitimately rewrites the go directive and may add toolchain and indirect lines. It checks the module path, the direct requirement and its version, both `h1:` checksums for that version in go.sum, that tools.go blank-imports a package under the tracked module, that nothing unexpected sits in the tracking tree, and that the Dependabot gomod directories cover every module and nothing outside it. It needs neither the network beyond a yq binary nor the Go toolchain.

`sync-tracking --write` reads the versions back out of the per-dependency modules and writes only the two YAML files. `update --write` rewrites only what it resolves from upstream release metadata, namely the Kubernetes version fields, the kubernetes-cni package versions and crictl; `containerd_version`, `runc_version` and `kubernetes_cni_semver` are carried over untouched, and the rolling `latest` entry is refreshed from its own values rather than from the release pin for the same minor. It then rewrites the require line of any module whose matrix value moved and tidies it. A new `render-tracking --write` bootstraps the modules for a newly added release selector. Both writing commands check for the `go` command before touching a file, so a missing toolchain cannot leave a go.mod ahead of its go.sum. The in-place require rewrite is block aware and only ever touches the direct requirement. `validate_entry` no longer requires the CNI tarball version to equal the RPM version; it validates the semver shape and still checks that the deb and rpm package versions agree with each other.

The Dependabot config uses globs for the two gomod entries, so a new release selector needs no config change, and adds `allow: dependency-type: direct` so the indirect requirements tools.go could introduce are never bumped on their own. The workflow's scope regexes follow the nested layout and still keep tools.go out of the diff Dependabot is allowed to produce.

## Alternative and live demo

#2163 implements the same fix with plain YAML instead of Go modules, using Dependabot's `pre-commit` ecosystem: seven `.pre-commit-config.yaml` files instead of 105 Go files, no toolchain in the generator, and explicit per-directory version bounds in `dependabot.yml` because `update-types` ignores are inert for that ecosystem in dependabot-core today. The two PRs share the generator fixes and cannot coexist; whichever is merged, the other should be closed. #2163 carries a comparison table.

Both approaches were deployed as `main` of a fresh repository on my account, so the real Dependabot service ran against them with the exact `dependabot.yml` from each branch, and the `update-kubernetes-version-matrix` workflow ran on the resulting pull requests. Nothing was triggered by hand: Dependabot's first pass after the config landed opened the PRs, and every workflow run finished successfully and pushed the `dependabot: update Kubernetes version matrix` commit onto the Dependabot branch.

Go modules (#2148, https://github.com/MaxRink/image-builder-dependabot-gomod):

| Dependabot PR | What it bumped | Regenerated matrix |
| --- | --- | --- |
| [gomod PR 2](https://github.com/MaxRink/image-builder-dependabot-gomod/pull/2) | `latest`: containerd 2.3.2 to 2.3.4, runc 1.4.3 to 1.5.1, Kubernetes 1.36.3 to 1.37.0, cri-tools 1.36.0 to 1.37.0 (four modules, one grouped PR) | [run](https://github.com/MaxRink/image-builder-dependabot-gomod/actions/runs/33920467529) wrote `kubernetes-version-latest.yaml` |
| [gomod PR 3](https://github.com/MaxRink/image-builder-dependabot-gomod/pull/3) | release pins across eleven module directories: containerd 2.3.4 everywhere, CNI plugins 1.6.2 for 1.32 and 1.33, Kubernetes patch bumps for 1.34, 1.35 and 1.36, no minor bumps | [run](https://github.com/MaxRink/image-builder-dependabot-gomod/actions/runs/33920492382) wrote `kubernetes-version-matrix.yaml` |

pre-commit configs (#2163, https://github.com/MaxRink/image-builder-dependabot-precommit):

| Dependabot PR | What it bumped | Regenerated matrix |
| --- | --- | --- |
| [pre-commit PR 7](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/7) | `latest`: the same four revs as gomod PR 2, each a one-line `rev:` change | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920652950) |
| [pre-commit PR 4](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/4) | release-1-32: containerd v2.3.4 and CNI plugins v1.6.2, no Kubernetes, cri-tools or runc update | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920634111) |
| [pre-commit PR 5](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/5) | release-1-31: containerd only | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920640647) |
| [pre-commit PRs 2, 3, 6, 8](https://github.com/MaxRink/image-builder-dependabot-precommit/pulls) | release-1-34, 1-35, 1-36, 1-33: containerd plus the Kubernetes or CNI patch level, one PR per directory | all [runs](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/workflows/update-kubernetes-version-matrix.yaml) green |

The matrix diffs are identical between the two demos for the same selector. The visible operational difference is PR volume: the gomod config groups across directories and produced two matrix PRs, the pre-commit config needs one entry per directory because the version bounds differ, and produced seven.

## Validation

`verify` passes, including with the go toolchain removed from PATH. Dependabot was simulated end to end in a scratch copy of the tree: `go get github.com/opencontainers/runc@v1.5.1`, `go get github.com/containernetworking/plugins@v1.6.2` and `go get github.com/containerd/containerd/v2@v2.3.4`, each followed by `go mod tidy`, in `latest/runc`, `release-1-32/cni-plugins` and `release-1-31/containerd`. Every module kept its direct requirement at the new version, go.sum stayed two lines, and only the files inside the three touched module directories changed. `sync-tracking --write` then produced exactly the expected matrix edits: runc 1.5.1 for latest, containerd 2.3.4 for 1.31, and CNI v1.6.2 for 1.32 with the deb and rpm package versions correctly left at 1.6.0. A grouped bump of containerd across all six release directories yields six matrix edits.

In a second copy, `update --write` moved latest to Kubernetes 1.37.0 and refreshed the 1.34, 1.35 and 1.36 patch levels while preserving every `kubernetes_cni_semver`; a stubbed-network harness confirms that when latest runs ahead of the release pin on containerd, runc and the CNI tarball, all three survive `update --write` while every upstream-resolved field still moves. Running `go mod tidy` in all 35 modules afterwards changed nothing. Adding a synthetic 1.37 release pin and running `render-tracking --write` bootstrapped its five modules, and the existing glob covered them without a config edit.

The missing-toolchain failure mode was reproduced before fixing it: with `go` off PATH, a pin bump left a go.mod at the new version beside a go.sum still holding the old hashes, and `verify` accepted it. Both now fail, the first before any file is written. Negative tests confirm `verify` catches a stripped requirement, a wrong version, a missing module, a tools.go without its blank import, a stray module directory and a Dependabot config that stops covering the tree, while accepting a go.mod written in block form with a toolchain line and indirect requirements. The two workflow regexes match all 70 go.mod and go.sum files, reject all 35 tools.go files, and match nothing else in the repository. Both YAML files parse with yq, all tools.go files are gofmt clean, and `ruff check` reports the same findings as on main with none added.

Once this lands, Dependabot will close the open PRs against the old flat manifests (#2141, #2142, #2143, #2145, #2146) and open fresh ones against the per-dependency modules.


mboersma (changes requested):
Could we reconsider the shared `tools.go` approach before merging? Importing all tracked dependencies into one module activates Go’s minimal version selection, so containerd’s transitive requirements can raise the older release-specific CNI and `client-go` pins. Several imported versions also require a newer Go directive than the generated `go 1.24`.

A safer approach may be separate synthetic tracking modules for each independently pinned dependency, with the generator and Dependabot configuration managing those modules. The generator should also preserve the independently tracked CNI tarball version instead of deriving it from RPM metadata in `update --write`.

MaxRink:
@mboersma Sure, ill give it a try. The general approach was to replicate what we do downstream in our gitlab fork (using renovate) with dependabot as that generally exists for github, but i was woefully unprepared for the different kind of limitations that stack has :S

MaxRink:
Following up on the requested redesign: current head 525e06205cc33a38805db17e1f844d823132d10e uses separate synthetic modules per dependency and selector, with independent Go directives. The generator preserves the independently tracked CNI tarball version instead of deriving it from RPM metadata. The PR description includes the validation and comparison with alternative #2163; those alternatives still need a maintainer selection and cannot both land unchanged. Ready for re-review of these requested changes.

PR #2149, opened 2026-08-22
capi: generate validate target help from build names

### LINKED ISSUE #908: Follow-up concerning validate targets
As @johananl pointed out here https://github.com/kubernetes-sigs/image-builder/pull/834#discussion_r876166796 , currently `validate-*` targets in the Makefile are out of sync with actual build targets. The makefile generates the `validate-*` targets from the build targets, but the separate entries are used when generating the `make help` output.

There are 2 points to tackle here:
- [ ] sync validate targets with build targets (OR generate help string differently)?
- [ ] automatically execute validate targets in CI

There is also the second comment https://github.com/kubernetes-sigs/image-builder/pull/834#discussion_r876165679:
- [ ] harmonize the `$BUILD_NAME == *flatcar*` check to look the same across the codebase.

yindia:
## Change description

The `##@ Validate packer config` section of `images/capi/Makefile` hand-listed a `validate-<image>: ## ...` stub for every per-image validate target. These stubs were maintained separately from the `*_BUILD_NAMES` / `*_VALIDATE_TARGETS` lists they document, so they drifted: new images added to the build lists never got a help entry, and removed ones left stale stubs behind.

This change removes the hand-maintained per-image stubs and derives the help listing from the same variables that generate the real targets:

- Adds `ALL_VALIDATE_TARGETS`, a sorted union of every `*_VALIDATE_TARGETS` list.
- `make help` now prints a generated "Validate packer config (per-image)" section from `ALL_VALIDATE_TARGETS`, so the listing can no longer drift from the build targets it is derived from.
- Keeps the `-all` aggregate targets as-is.

- Removes 16 no-op `validate-node-ova-local-vmx-*` / `-base-*` targets that previously existed only as `make help` stubs and exited 0 as no-ops. They are not derived from any `*_BUILD_NAMES` list and nothing in CI or scripts references them.

## Related issues

- Fixes #908

## Additional context

Verified with:

```
cd images/capi && make help
```

The generated section lists every per-image `validate-*` target sourced from the build-name lists.

### DIFF
--- a/docs/book/src/capi/custom-build-targets.md
+++ b/docs/book/src/capi/custom-build-targets.md
@@ -58,9 +58,12 @@ Builder.
    point.
 2. Add the target name to the provider list in `images/capi/Makefile`, or to the
    shared OS version list used by that provider.
-3. Add static build and validate help entries in the "Document dynamic build
-   targets" and "Document dynamic validate targets" sections of
-   `images/capi/Makefile` so `make help` shows the new target.
+3. Add a static build help entry in the "Document dynamic build targets"
+   section of `images/capi/Makefile` so `make help` shows the new build
+   target. The per-image validate help is generated automatically from
+   `ALL_VALIDATE_TARGETS`, so no manual validate stub is needed. Any variable
+   named `*VALIDATE*_TARGETS` is picked up automatically, so even a new
+   provider family needs nothing added here.
 4. Update the provider documentation when users need new credentials, cloud
    image identifiers, or provider-specific variables.
 5. Run the matching `validate-*` target. Run a real `build-*` target when the
@@ -70,7 +73,8 @@ For example, a new QEMU target named `qemu-example-linux-9` would usually need:
 
 - `images/capi/packer/qemu/qemu-example-linux-9.json`
 - `qemu-example-linux-9` in `QEMU_BUILD_NAMES`
-- `build-qemu-example-linux-9` and `validate-qemu-example-linux-9` help entries
+- a `build-qemu-example-linux-9` help entry (the matching
+  `validate-qemu-example-linux-9` help is generated automatically)
 
 ## Target name reference
 
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -54,6 +54,8 @@ help: ## Display this help
 	@echo '  cleaning artifacts created from building OVAs using a local'
 	@echo '  hypervisor.'
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-35s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+	@printf "\n\033[1mValidate packer config (per-image, generated)\033[0m\n"
+	@for t in $(ALL_VALIDATE_TARGETS); do printf "  \033[36m%-36s\033[0m %s\n" "$$t" "Validates $${t#validate-} packer config"; done
 
 .PHONY: version
 version: ## Display version of image-builder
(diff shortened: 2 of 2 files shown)


yindia:
cc: @jepio @drew-viles @AndiDog

mboersma (changes requested):
Could you also update `custom-build-targets.md` to match this change? It still tells contributors to add manual validate help stubs, including one in the example. Please explain that per-image validate help is generated automatically and note that new provider families must be added to `ALL_VALIDATE_TARGETS`.

yindia:
@mboersma ready for review

images/capi/Makefile
```diff
@@ -504,6 +506,33 @@ MAAS_ARM64_VALIDATE_TARGETS 	:= $(addprefix validate-,$(MAAS_ARM64_BUILD_NAMES))
 SCALEWAY_BUILD_TARGETS		:= $(addprefix build-,$(SCALEWAY_BUILD_NAMES))
 SCALEWAY_VALIDATE_TARGETS	:= $(addprefix validate-,$(SCALEWAY_BUILD_NAMES))
 
+# All generated per-image validate targets, used by `make help` so the listing
+# never drifts from the build targets it is derived from.
+ALL_VALIDATE_TARGETS := $(sort \
```
drew-viles: Could we derive this from `.VARIABLES` instead of hand-listing the families? Something like `$(sort $(foreach v,$(filter %VALIDATE_TARGETS,$(.VARIABLES)),$($(v))))` produces the same 156 targets and removes the last place this can drift.

images/capi/Makefile
```diff
@@ -54,6 +54,8 @@ help: ## Display this help
 	@echo '  cleaning artifacts created from building OVAs using a local'
 	@echo '  hypervisor.'
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-35s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+	@printf "\n\033[1mValidate packer config (per-image, generated)\033[0m\n"
```
drew-viles, line 57: Nit: the generated lines sit one column left of the awk-rendered ones. `%-36s` or a second space before `%s` lines them up.

drew-viles (commented):
Thanks, the docs update covers what @mboersma asked for and I confirmed `make help` now lists every per-image validate target. Two small things inline, plus one note: removing the stubs also removes the 16 `validate-node-ova-local-vmx-*` / `-base-*` targets, which previously exited 0 as no-ops. That is probably the right outcome, but could we mention it in the PR description?

yindia:
Thanks for the reviews @drew-viles @mboersma. Addressed all feedback, PTAL when you get a chance 🙏

docs/book/src/capi/custom-build-targets.md
```diff
@@ -58,9 +58,12 @@ Builder.
...
    shared OS version list used by that provider.
-3. Add static build and validate help entries in the "Document dynamic build
-   targets" and "Document dynamic validate targets" sections of
-   `images/capi/Makefile` so `make help` shows the new target.
+3. Add a static build help entry in the "Document dynamic build targets"
+   section of `images/capi/Makefile` so `make help` shows the new build
+   target. The per-image validate help is generated automatically from
+   `ALL_VALIDATE_TARGETS`, so no manual validate stub is needed. Only a new
+   provider family (a new `*_VALIDATE_TARGETS` list) must be added to
+   `ALL_VALIDATE_TARGETS`.
```
drew-viles: This was true before the `.VARIABLES` change but the Makefile comment now says new families need no change here. Could this say that any `*VALIDATE*_TARGETS` variable is picked up automatically and nothing needs adding?

drew-viles (commented):
sorry one more minor thing I have noticed - the last commit switched `ALL_VALIDATE_TARGETS` to derive from `.VARIABLES`, but `custom-build-targets.md` still tells contributors to add new families to it, so the docs and the Makefile comment now contradict each other. One inline suggestion.

docs/book/src/capi/custom-build-targets.md
yindia: @drew-viles please check now

mboersma (approved):
/lgtm

mboersma (approved):
/approve

drew-viles:
Seeing same failures as in https://github.com/kubernetes-sigs/image-builder/pull/2139

mboersma:
/override pull-azure-sigs

See #2166

PR #2150, opened 2026-08-26
Fix RockyLinux 9 qemu kickstart repo baseurl

Pearl1594:
## Change description

The kickstart repo entry in the qemu RockyLinux 9 template pointed at pub/rocky/9/devel/x86_64/kickstart/, which does not exist on Rocky's mirrors - the devel tree has no kickstart subdirectory.
The PR points it at BaseOS/x86_64/kickstart/ instead, matching the OVA builder's copy of this same kickstart file, which already had the correct path.

## Related issues

- Fixes #

### DIFF
--- a/images/capi/packer/qemu/linux/rockylinux/http/9/ks.cfg.tmpl
+++ b/images/capi/packer/qemu/linux/rockylinux/http/9/ks.cfg.tmpl
@@ -1,6 +1,7 @@
 # Use CDROM installation media
 repo --name="AppStream" --baseurl="http://download.rockylinux.org/pub/rocky/9/AppStream/x86_64/os/"
 repo --name="kickstart" --baseurl="http://download.rockylinux.org/pub/rocky/9/BaseOS/x86_64/kickstart/"
+repo --name="devel" --baseurl="https://download.rockylinux.org/pub/rocky/9/devel/x86_64/os/"
 cdrom
 
 # Use text install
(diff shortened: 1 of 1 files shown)


AndiDog (commented):
/ok-to-test
/lgtm

images/capi/packer/qemu/linux/rockylinux/http/9/ks.cfg.tmpl
```diff
@@ -1,6 +1,6 @@
 # Use CDROM installation media
 repo --name="AppStream" --baseurl="http://download.rockylinux.org/pub/rocky/9/AppStream/x86_64/os/"
-repo --name="kickstart" --baseurl="http://download.rockylinux.org/pub/rocky/9/devel/x86_64/kickstart/"
+repo --name="kickstart" --baseurl="http://download.rockylinux.org/pub/rocky/9/BaseOS/x86_64/kickstart/"
```
mboersma, line 3: Could we keep this pointed at Rocky’s devel repository using its current layout?

`https://download.rockylinux.org/pub/rocky/9/devel/x86_64/os/`

The proposed BaseOS kickstart repository is valid, but it doesn’t contain `cloud-utils`, which this kickstart installs and the QEMU Goss checks require. Maybe we should rename it too?

`--name="devel"`

mboersma (changes requested):

mboersma (approved):
/lgtm
/approve

PR #2152, opened 2026-08-28
fix: bump kubeadm

### LINKED ISSUE #2151: Error building 1.37 OVA
While building OVA for 1.37 I hit an error because the kubeadm pull command is using a config file in version v1beta3, which is now removed

Unfortunately I don't have anymore the error message, but the failing code was

https://github.com/kubernetes-sigs/image-builder/blob/4825f24107f9ab2f5788c96abab93b97b3cfa933/images/capi/ansible/roles/kubernetes/tasks/kubeadmpull.yml#L9

And I got around it by changing the version in https://github.com/search?q=repo%3Akubernetes-sigs%2Fimage-builder+v1beta3&type=code

I'll defer to you if the change if fine in general - e.g. also for building images with older K8s versions (v1beta4 is supported since a couple of release now)

Thanks for the help!

/kind bug

### LINKED ISSUE #2153: kubeadm config uses deprecated v1beta3 API, breaking builds for Kubernetes 1.36+
### What happened?

Building images for Kubernetes 1.36+ fails during the `kubeadm config images pull` step because the kubeadm config template uses `kubeadm.k8s.io/v1beta3`, which is removed starting with kubeadm 1.36.

Error:
```
error: your configuration file uses an old API spec: "kubeadm.k8s.io/v1beta3" (kind: "ClusterConfiguration"). 
Please use kubeadm v1.36 instead and run kubeadm config migrate
```

### What did you expect to happen?

Image builds for Kubernetes 1.36+ should succeed.

### How to reproduce

Build any CAPI image with `kubernetes_semver` >= v1.36.0.

### Proposed fix

Bump the kubeadm config template from `v1beta3` to `v1beta4` in `images/capi/ansible/roles/kubernetes/templates/etc/kubeadm.yml`.

`v1beta4` has been available since kubeadm 1.29, so all currently supported Kubernetes versions (1.29–1.37) accept it. The change is a simple API version bump — no schema changes are needed between v1beta3 and v1beta4 for the fields used in this template.

### Workaround

We are currently working around this by copying a v1beta4 template into the role templates directory before the Ansible run and overriding `kubeadm_template` for versions >= 1.36.

Related: #2151

/kind bug
/area capi

iamfletch:
## Change description
beta3 was deprecated a while ago, beta4 should have been available since 1.31. Bumping now because its been fully removed from 1.37.

more info at kubeadm https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta4/

## Fixes
Fixes https://github.com/kubernetes-sigs/image-builder/issues/2151
Fixes https://github.com/kubernetes-sigs/image-builder/issues/2153

## Additional context
- builds failed during image cache pull, kubeadm refused to run
- build worked after switching to beta4

### DIFF
--- a/images/capi/ansible/roles/kubernetes/templates/etc/kubeadm.yml
+++ b/images/capi/ansible/roles/kubernetes/templates/etc/kubeadm.yml
@@ -1,11 +1,11 @@
-apiVersion: kubeadm.k8s.io/v1beta3
+apiVersion: kubeadm.k8s.io/v1beta4
 kind: ClusterConfiguration
 imageRepository: {{ kubernetes_container_registry }}
 kubernetesVersion: {{ kubernetes_semver }}
 dns:
   imageRepository: {{ kubernetes_container_registry }}/coredns
 ---
-apiVersion: kubeadm.k8s.io/v1beta3
+apiVersion: kubeadm.k8s.io/v1beta4
 kind: InitConfiguration
 nodeRegistration:
   criSocket: {{ containerd_cri_socket }}


drew-viles (commented):
/ok-to-test

mboersma (approved):
/approve

PR #2156, opened 2026-09-04
capi: refresh CentOS Stream 9 and 10 ISO composes and resolve them automatically

MaxRink:
## Change description

The qemu CentOS Stream packer configs pin dated composes. CentOS has since rotated both off the mirror: 9-stream was pinned at 20260706.0 and 10-stream at 20260630.0, and both ISO URLs now return 404 with neither file listed in the published SHA256SUM. The mirror retains only the newest compose alongside the -latest- symlinks, so the pinned files were pruned.

That also breaks the update flow rather than just the builds. `hack/update-iso-checksums.sh` only ever rewrites `iso_checksum`, never `iso_url`, so once the pinned compose is pruned its SHA256SUM entry is gone, the grep matches nothing and the script aborts under pipefail printing nothing at all. Today `make update-centos-9-iso-checksums` and `make update-centos-10-iso-checksums` exit non-zero and refresh nothing, and the only recovery is a manual repin. Because CentOS Stream keeps only the newest compose in the BaseOS iso directory, this recurs every few weeks.

Two commits:

1. Repin both configs to the current composes, 9-stream 20260902.0 and 10-stream 20260901.0, with checksums produced by the repo's own update flow reading the mirror SHA256SUM.

2. Add `hack/update-centos-iso-urls.sh` and run it ahead of the checksum refresh in the two centos targets. It reads SHA256SUM from the directory the config already points at, picks the newest dated `CentOS-Stream-<N>-<date>.<n>-x86_64-dvd1.iso` listed there, and rewrites `iso_url` to it. The dated name is used rather than the `-latest-` symlink so the pin keeps naming the one specific compose the checksum belongs to. Configs without a CentOS Stream DVD URL on mirror.stream.centos.org are reported and skipped; a fetch failure or a directory with no dated compose is a hard error rather than a silent no-op. `json-sort` still runs last.

Together the two steps make the update targets self-healing: a pruned compose is resolved forward and rechecksummed instead of failing.

The powervs CentOS configs are unaffected. They have no `iso_url` and build from a prebuilt OVA in IBM COS, so they carry no compose pin. There are no ova, raw, proxmox, or maas CentOS Stream configs.

## Related issues

None.

## Additional context

Validation:
- `curl -fsSI` returns 200 for both new ISO URLs, and both `iso_checksum` values match the SHA256 entry for their file in the mirror SHA256SUM.
- `shellcheck -x` and `bash -n` are clean on the new script; it was exercised under bash 3.2.
- Resetting `qemu-centos-9.json` to the pruned 20260706.0 compose and running `make update-centos-9-iso-checksums` restores the correct URL and checksum. The same input fails silently on main.
- The script no-ops with a short message when already current, and returns 1 with a clear error and leaves the file untouched when the mirror 404s.
- `make update-centos-9-iso-checksums update-centos-10-iso-checksums` run twice: the second run is a no-op.
- `make validate-qemu-centos-9 validate-qemu-centos-10` both report the configuration is valid.

CentOS Stream dated ISO pins can go stale within days because the mirror retains only the current compose. These pins were refreshed to 20260914.0; rerun `make update-centos-9-iso-checksums update-centos-10-iso-checksums` immediately before merge if the mirror rotates again.

### DIFF
--- a/images/capi/packer/qemu/qemu-centos-9.json
+++ b/images/capi/packer/qemu/qemu-centos-9.json
@@ -10,9 +10,9 @@
   "distro_version": "9",
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-9",
   "guest_os_type": "centos9-64",
-  "iso_checksum": "280f78ca32547eec2f3150090c21526744851d99af006344a07629f88500d9f6",
+  "iso_checksum": "d372351207b3e4ce28a8f93c21c14be68cbb68195568d54f055b935cbed67dea",
   "iso_checksum_type": "sha256",
-  "iso_url": "https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/iso/CentOS-Stream-9-20260706.0-x86_64-dvd1.iso",
+  "iso_url": "https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/iso/CentOS-Stream-9-20260902.0-x86_64-dvd1.iso",
   "os_display_name": "CentOS 9 Stream",
   "redhat_epel_rpm": "https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm",
   "shutdown_command": "/sbin/halt -h -p"
--- a/images/capi/Makefile
+++ b/images/capi/Makefile
@@ -1270,11 +1270,13 @@ update-rockylinux-iso-checksums: ## Updates checksums for Rocky Linux ISOs
 
 .PHONY: update-centos-9-iso-checksums
 update-centos-9-iso-checksums: ## Updates checksums for Centos 9 ISOs
+	hack/update-centos-iso-urls.sh centos-9
 	hack/update-iso-checksums.sh centos-9 false SHA256SUM "SHA256 (iso_file_name)" 4
 	$(MAKE) json-sort
 
 .PHONY: update-centos-10-iso-checksums
 update-centos-10-iso-checksums: ## Updates checksums for Centos 10 ISOs
+	hack/update-centos-iso-urls.sh centos-10
 	hack/update-iso-checksums.sh centos-10 false SHA256SUM "SHA256 (iso_file_name)" 4
 	$(MAKE) json-sort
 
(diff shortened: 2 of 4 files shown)


images/capi/packer/qemu/qemu-centos-9.json
```diff
@@ -10,9 +10,9 @@
   "distro_version": "9",
   "epel_rpm_gpg_key": "https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-9",
   "guest_os_type": "centos9-64",
-  "iso_checksum": "280f78ca32547eec2f3150090c21526744851d99af006344a07629f88500d9f6",
+  "iso_checksum": "d372351207b3e4ce28a8f93c21c14be68cbb68195568d54f055b935cbed67dea",
   "iso_checksum_type": "sha256",
-  "iso_url": "https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/iso/CentOS-Stream-9-20260706.0-x86_64-dvd1.iso",
+  "iso_url": "https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/iso/CentOS-Stream-9-20260902.0-x86_64-dvd1.iso",
```
drew-viles, line 15: This returns 404 today. The mirror now only has `CentOS-Stream-9-20260908.0-x86_64-dvd1.iso`. Same for the centos-10 config.

drew-viles (commented):
The script itself is in good shape. I ran it against a copy and could not break it: HTTP errors, an empty checksum list, a zero-byte SHA256SUM and the `-latest-` input all fail cleanly and leave the file untouched, it is idempotent, and it is clean on bash 3.2 and shellcheck.

The problem is the pins it ships. The mirror only keeps one dated compose and it rotated to `20260908.0` on 8 September, so both `iso_url` values in this PR already return 404. If we merge as-is, `build-qemu-centos-9` and `-10` are broken on main from day one, which is what this PR set out to fix.

Could you rerun `make update-centos-9-iso-checksums update-centos-10-iso-checksums` and push right before this is ready to merge? It will need doing again if merge slips past the next rotation. It would also be worth a sentence in the description saying the dated pin goes stale within days by design, so nobody is surprised when it happens.

PR #2160, opened 2026-09-04
capi: fix autoinstall http_directory wiring for maas and proxmox EFI builds

MaxRink:
## Change description

Several Ubuntu autoinstall builds serve the wrong nocloud-net data directory, so the per-provider templates that were written for them are never used. This PR fixes the wiring and removes the profiles that are now provably unreachable. Four independent commits, so any one can be dropped.

**1. `capi: serve maas autoinstall data from the maas http tree`**

Both maas packer templates defaulted `http_directory` to the qemu tree. The two maas var files that do not override it, `maas-ubuntu-2404-efi.json` and `maas-ubuntu-2604-efi.json`, therefore served `packer/qemu/linux/ubuntu/http/{24.04.efi,26.04.efi}`, and the maas-side directories of the same name were dead. The default now points at `./packer/maas/linux/{{user `distro_name`}}/http/`. The three var files that already set `http_directory` explicitly are untouched. The maas 24.04.efi and 26.04.efi user-data templates are currently byte-identical to their qemu counterparts, so this is behaviour preserving today while making the maas tree the one maas builds actually consume.

**2. `capi: point proxmox EFI builds at the EFI autoinstall profiles`**

`proxmox-ubuntu-2404-efi` and `proxmox-ubuntu-2604-efi` set `"bios": "ovmf"` but their `boot_command` pointed the datasource at the `/24.04/` and `/26.04/` profiles, which lay out a single msdos partition with no ESP. The matching `24.04.efi` and `26.04.efi` profiles, which use a GPT table with a 512 MiB fat32 ESP mounted at `/boot/efi` and set `grub.reorder_uefi: false`, were never served. Both var files now select the `.efi` profile, matching how the qemu and raw EFI var files select theirs. This is the only commit here that changes what a build produces.

**3. `capi: drop unreferenced ubuntu autoinstall profiles`**

Removes `packer/qemu/linux/ubuntu/http/23.04`, `packer/maas/linux/ubuntu/http/23.04` and `packer/qemu/linux/ubuntu/http/22.04.efi`. Ubuntu 23.04 is out of support and was dropped from `UBUNTU_VERSIONS`, so no target resolves either 23.04 profile; `git grep -E '23\.04|2304'` returns nothing at all after this commit. `qemu/22.04.efi` was copied from the ova tree in 4c259828d, when the `packer/qemu/linux` symlink was replaced with a real directory, and no var file has ever referenced it: `qemu-ubuntu-2204-efi.json` has pointed at `22.04.efi.qemu` since it was created. The surviving `22.04.efi` hits belong to the ova and raw trees, which are separate.

**4. `capi: correct the packer variable precedence comment`**

The comment above `PACKER_NODE_FLAGS` claimed "Any existing values from PACKER_FLAGS take precendence over variable files". That holds for `-var` but not for a `-var-file` inside `PACKER_FLAGS`, which is emitted before the per-target var file and `PACKER_VAR_FILES` and so loses to both. Reworded, and the typo is gone. Comment only.

## Related issues

None.

## Additional context

Resolved `(http_directory, profile)` per var file, computed by applying the template default and the `boot_command` profile subpath and checking the joined directory has a `user-data.tmpl`:

| var file | before | after |
| --- | --- | --- |
| maas-ubuntu-2204-arm64 | maas/22.04.arm64 | unchanged |
| maas-ubuntu-2204-efi | maas/22.04.efi.qemu | unchanged |
| maas-ubuntu-2404-arm64 | maas/24.04.arm64 | unchanged |
| maas-ubuntu-2404-efi | qemu/24.04.efi | maas/24.04.efi |
| maas-ubuntu-2604-efi | qemu/26.04.efi | maas/26.04.efi |
| proxmox ubuntu-2204 | proxmox/22.04 | unchanged |
| proxmox ubuntu-2404 | proxmox/24.04 | unchanged |
| proxmox ubuntu-2404-efi | proxmox/24.04 | proxmox/24.04.efi |
| proxmox ubuntu-2604 | proxmox/26.04 | unchanged |
| proxmox ubuntu-2604-efi | proxmox/26.04 | proxmox/26.04.efi |

`packer validate` via the Makefile passes for every touched target on packer 1.16: `validate-maas-ubuntu-{2204,2404,2604}-efi`, `validate-maas-ubuntu-{2204,2404}-arm64`, `validate-proxmox-ubuntu-{2204,2404,2604}`, `validate-proxmox-ubuntu-{2404,2604}-efi`, plus `validate-qemu-ubuntu-{2204,2404,2604}-efi` as a regression check on the directory removals. Proxmox validate needs the `PROXMOX_*` environment set but does not contact the API, so placeholder values are enough.

The precedence claim in commit 4 was checked against packer 1.16 rather than assumed, using a two-var-file legacy JSON template with the resolved value in the output filename: `-var-file=a -var-file=b` gives b, `-var-file=b -var-file=a` gives a, and `-var` wins from either position.

One thing left alone deliberately: after commit 1, `packer/maas/linux/ubuntu/http/{22.04, 22.04.efi, 24.04, 26.04}` are also unreachable, since `MAAS_BUILD_NAMES` and `MAAS_ARM64_BUILD_NAMES` only contain EFI and arm64 targets. They came in with the original maas import as a copy of the qemu tree. Happy to remove them too if preferred, but that felt like a separate call from this PR.


mboersma (approved):
/lgtm

/assign @drew-viles

PR #2101, opened 2026-07-06
capi: add qemu node conformance validation hook

MaxRink:
## What

Adds an opt-in post-build Kubernetes node conformance check for QEMU images. The finished image remains a read-only backing file for a disposable qcow2 overlay; the Packer template is unchanged from main.

## Current design and review follow-ups

- Shared boot/SSH helpers live in `hack/lib/qemu-guest.sh`; the runner copies the guest hook in and logs/JUnit out.
- Removed the unused etcd binary download, kubelet root/cert directory overrides, redundant guest Flatcar guard, and `summary.env` protocol. The SSH exit status determines success, and failure to stop the system kubelet is fatal.
- Forward `NODE_CONFORMANCE_*` settings into the guest; CI uses `QEMU_*` directly with the same defaults as the runner.
- Results use `NODE_CONFORMANCE_OUTPUT_DIR`, then nonempty `ARTIFACTS`, then `images/capi/node-conformance-results`. Each run gets a fresh subdirectory; existing output is not deleted.
- Behavior tests exercise Ginkgo failure propagation through tee, kubelet-stop failure, guest environment forwarding and runner SSH failure. Download checksum/retry handling and disposable overlay cleanup remain.

## Verification (2026-09-19)

38 behavioral script tests pass. Real QEMU execution found and fixed default extensionless-image directory resolution and CNI setup treating `.kubernetes-cni-keep` as a network configuration.

Built Ubuntu 24.04 / Kubernetes v1.36.1 from rebased source `e02416a93c8f2f4e44fa5f2e44a7c29deada70d3`: Packer succeeded and Goss reported **65 checks, 0 failed, 0 skipped**. The final harness is `ed8962bd1f074c0b7cc8acdb7f2740831eb400f3`; Ansible, Packer configuration/template, cloud-image target and Makefile provisioning inputs are identical to that build source. Image SHA256 before and after all disposable-overlay runs: `8f7796d78158717b63812a0619e6bb22e8e87a436e7a9c6c2b7d77baf8080213`.

Runtime used x86_64 QEMU TCG on an arm64 macOS host, `-cpu max`, 4 vCPU and 4096 MiB RAM. Every invocation passed the normal Packer output directory to the runner.

- Default 169-spec selection: interrupted during eager image preparation after containerd's 5-minute no-progress timeout unpacking the unrelated 3.24 GB PyTorch performance image; **0 specs executed**.
[...]

PR #2045, opened 2026-06-16
Raise Windows Feature goss check timeout to 180s

mboersma:
## Change description

The `Windows Feature - <name>` goss checks run `powershell -command "(Get-WindowsFeature <name> | select *)"` with a 60s timeout (`images/capi/packer/goss/goss-package.yaml`). On **Windows Server 2025**, `Get-WindowsFeature` intermittently takes longer than 60s, so the `Containers` and `Hyper-V-PowerShell` checks fail with:

```
Command execution timed out (1m0s)
```

This fails Goss validation and breaks the Azure SIG `windows-2025-containerd` build. It has been showing up as a recurring, content-unrelated failure in `pull-azure-sigs` (e.g. on #1993, the k8s 1.36 bump), where windows-2025 is the only consistently failing target. Windows Server 2022 runs the same 65 checks in ~34s total and passes, so this is a 2025-specific performance issue, not a missing feature (the checks even pass on retry).

Raising the command timeout to **180s** gives `Get-WindowsFeature` enough headroom on Windows Server 2025.

## Related issues

N/A

PR #2161, opened 2026-09-04
capi: give Get-WindowsFeature goss checks time to answer on Windows Server 2025

MaxRink:
## Change description

The `Windows Feature - <name>` goss checks run `powershell -command "(Get-WindowsFeature <name> | select *)"` with a 180s command timeout (`images/capi/packer/goss/goss-package.yaml`, raised from 60s in #2045). On Windows Server 2025 that is still not enough and the Azure SIG `windows-2025-containerd` build keeps failing on unrelated PRs (#2100, #2101 and #2141 today) with:

    Windows Feature - Hyper-V-PowerShell: exit-status: Error: Command execution timed out (3m0s)
    Error: timeout of 3m0s reached before tests entered a passing state

Looking at the `windows-2025-containerd.log` artifact of the last 28 `pull-azure-sigs` runs (Aug 22 to Sep 4), the pattern is the same every time: the first `Get-WindowsFeature` call after boot takes 60 to 140s, the second one about 1s. Which of the two feature checks pays the cost depends on goss ordering. The three failures are the tail of that distribution: in the PR #2141 run the first check hit the 180s timeout and the second one still needed 139s, so the cmdlet took over 300s to answer. Windows Server 2022 pays 17s for the first call and passes.

A timed out check also cannot be rescued by goss retries: the check timeout equals `goss_retry_timeout` (180s, `goss-args.json`), so when the first pass finishes goss has already exceeded its retry budget and exits instead of running a second pass, even though the second pass would succeed.

This raises the command timeout to 600s, well above the observed tail, and documents why it has to stay above `goss_retry_timeout`. Trade-off: if `Get-WindowsFeature` genuinely hangs, the failure surfaces after 10 minutes instead of 3, for the first of the two checks only.

## Related issues

Follow-up to #2045 and #1603.

## Additional context

`make validate-azure-sig-windows-2025-containerd` passes. Windows builds cannot be run locally; the change only affects the goss spec uploaded to the VM.

### DIFF
--- a/images/capi/packer/goss/goss-package.yaml
+++ b/images/capi/packer/goss/goss-package.yaml
@@ -85,6 +85,12 @@ package:
 
 {{ if eq .Vars.OS "windows"}} # Windows
 # Workaround until windows features are added to goss
+# The first Get-WindowsFeature call after boot is slow on Windows Server 2025:
+# in CI it usually takes 60-140s and has been observed to take over 300s,
+# while every later call takes about 1s. Keep this timeout well above that
+# tail. It must also stay larger than goss_retry_timeout (goss-args.json):
+# once a single command times out, goss has already used up its retry budget
+# and does not run a second pass.
 command:
 {{range $name, $vers := index .Vars .Vars.OS "common-windows-features"}}
   "Windows Feature - {{ $name }}":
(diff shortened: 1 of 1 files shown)


mboersma (approved):
/lgtm
/approve

PR #2162, opened 2026-09-04
capi: retry ansible-galaxy collection installs

MaxRink:
## Change description

`make deps-common` calls `ansible-galaxy collection install` exactly once in `images/capi/hack/ensure-ansible.sh` and once in `images/capi/hack/ensure-s3.sh`. When galaxy.ansible.com is briefly degraded the whole CI job dies on the first bad response. `pull-azure-sigs` failed this way on #2156 and #2160 with:

```
ERROR! Error when getting the collection info for ansible.posix from default (https://galaxy.ansible.com/api/) (HTTP Code: 504, Message: Gateway Timeout)
```

This adds a shared `galaxy_collection_install` helper to `hack/utils.sh`, which both scripts already source, and routes both installs through it. The helper retries up to `GALAXY_INSTALL_RETRIES` times (default 5) with exponential backoff starting at `GALAXY_INSTALL_RETRY_DELAY` seconds (default 5) and capped at `GALAXY_INSTALL_RETRY_MAX_DELAY` (default 40), so the default schedule is 5s, 10s, 20s, 40s. It also passes `--timeout` (`GALAXY_INSTALL_TIMEOUT`, default 60, the same value ansible-galaxy uses on its own) so one hung request cannot consume the entire retry budget.

Failures are not swallowed. Each failed attempt prints a line to stderr, the last attempt prints the attempt count and exit code, and the helper returns the original non-zero status so `set -o errexit` still fails the script. The set of collections installed and their version constraints are unchanged, and the Makefile is untouched.

## Related issues

None.

## Additional context

Validated on macOS with /bin/bash 3.2:

- `bash -n` and `shellcheck` (plain and `-x`, as the Makefile runs it) on all three scripts produce findings identical to main, with nothing new.
- With a fake `ansible-galaxy` on PATH that returns the 504 message above twice and then succeeds, both scripts print two retry lines, succeed on the third attempt and exit 0. An xtrace run confirms the surrounding steps such as the `ansible --version` check still run exactly once while the install is retried.
- With a stub that always fails, both scripts stop after 5 attempts, surface the last error, and exit non-zero. With `GALAXY_INSTALL_RETRIES=3` they stop after 3. A stub exiting 2 makes the script exit 2, so the original status is preserved.
- With a stub that succeeds right away, `ansible-galaxy` is invoked exactly once.
- A real run of `hack/ensure-ansible.sh` against galaxy.ansible.com still succeeds with a single invocation.
- `--timeout` was verified against ansible-core 2.18.18, the version pinned in `hack/utils.sh`, both in `--help` and in an actual install.

### DIFF
--- a/images/capi/hack/ensure-ansible.sh
+++ b/images/capi/hack/ensure-ansible.sh
@@ -48,7 +48,7 @@ fi
 
 echo ${ansible_version[*]}
 
-ansible-galaxy collection install \
+galaxy_collection_install \
   'community.general:<=12.0.0' \
   'ansible.posix' \
   'ansible.windows:>=1.7.0' \
--- a/images/capi/hack/ensure-s3.sh
+++ b/images/capi/hack/ensure-s3.sh
@@ -30,4 +30,4 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 export PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_ROOT_USER_ACTION=ignore
 
 # S3 interaction requires the following galaxy collection
-ansible-galaxy collection install amazon.aws:5.5.0
+galaxy_collection_install amazon.aws:5.5.0
--- a/images/capi/hack/utils.sh
+++ b/images/capi/hack/utils.sh
@@ -119,6 +119,44 @@ pip3_install() {
   fi
 }
 
+galaxy_collection_install() {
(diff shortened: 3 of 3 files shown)


mboersma (approved):
/lgtm
/approve

drew-viles:
/override pull-azure-sigs

See https://github.com/kubernetes-sigs/image-builder/issues/2166

PR #2163, opened 2026-09-04
capi: track the Kubernetes version matrix with pre-commit configs

MaxRink:
## Summary

This is an alternative to #2148. Both fix the same bug: the Dependabot tracking manifests under `images/capi/packer/config/kubernetes-version-dependencies/` are source-free Go modules, and Dependabot's gomod updater runs `go mod tidy` after every bump, which drops requirements no package imports. The modules lose their contents on the first update.

#2148 fixes this by making the modules real: one module per dependency per selector, each with a `tools.go` blank import and a `go.sum`, 35 modules and 116 changed files. This PR removes the Go modules instead and tracks the same five upstream repositories through Dependabot's `pre-commit` ecosystem, in 18 changed files, with no Go toolchain anywhere.

## Why pre-commit

Dependabot's `pre-commit` ecosystem tracks any git repository named in a `repos[].repo` entry: the parser reads `repo` and `rev`, the update checker lists the repository's tags and picks the newest allowed one, and the file updater rewrites only the matching `rev` line. It never runs a package manager over the file, so there is no tidy step that can strip a pin.

The manifests are Dependabot tracking files, not pre-commit configuration. The `pre-commit` tool never runs them, the hook lists are deliberately empty (the parser accepts that), and none of the listed repositories has to ship a `.pre-commit-hooks.yaml`.

Kubernetes is tracked through `kubernetes/kubernetes` tags directly rather than `k8s.io/client-go`, so a `rev` is the same `v1.x.y` string the matrix stores and the `v0.minor.patch` mapping in both directions is gone.

## Design

```
kubernetes-version-dependencies/
  latest/.pre-commit-config.yaml
  release-1-31/.pre-commit-config.yaml
  ...
  release-1-36/.pre-commit-config.yaml
```

```yaml
repos:
  - repo: https://github.com/containerd/containerd
    rev: v2.3.2
    hooks: []
```

`images/capi/hack/kubernetes-version-matrix.py` renders the manifests from the matrix and parses them back through the `yq` helper it already uses to load the matrix. `verify` compares the parsed repo and rev pairs, so a reordered or reformatted manifest still passes. `render-tracking --write` regenerates the manifests from the matrix without network access. `sync-tracking --write` folds Dependabot's revs back into the matrix YAML, as before.

`.github/dependabot.yml` carries one `pre-commit` entry per tracking directory, each naming its directory exactly, with its own group and the schedule, labels and commit-message settings the previous entries used. Every entry ignores each of the five repositories above a single explicit bound: the next minor of the pinned version for a release pin, the next major for `latest`.

```yaml
  ignore:
    # Release pins take patch updates only.
    - dependency-name: "https://github.com/containerd/containerd"
      versions: [ ">= 2.4.0" ]
```

The entries cannot be collapsed behind a `directories` glob, because the bounds differ per directory. `verify` checks that every tracking directory has an entry and that its bounds are exactly what the matrix implies, so a pin that crosses a minor fails until the bounds are updated; the error names the value to write. `sync-tracking` applies the same policy to the revs themselves, as a second guard for anything the bounds miss.

`update-types` conditions are not used and `verify` rejects them, for the reason in the next section.

The workflow changes only its `paths` filter and the two `tracking_pattern` regexes.

Three generator fixes are shared with #2148: `update --write` preserves `containerd_version`, `runc_version` and `kubernetes_cni_semver`, which Dependabot owns; the rolling `latest` entry is refreshed from its own values rather than from the release pin for the same minor, which would silently roll it back; and `validate_entry` checks the CNI semver's shape plus DEB and RPM agreement, instead of requiring the tarball version to equal the distro package version.

## Two defects in dependabot-core's pre-commit ecosystem

`ignore` conditions using `update-types` do not work on a pre-commit dependency, in two independent ways. Both are reproduced with the dependabot CLI against this branch.

`Dependabot::Config::IgnoreCondition#correct_version_for` parses the current version with the ecosystem's version class. `Dependabot::PreCommit::Version` is an empty subclass of `Dependabot::Version` and inherits RubyGems' anchored version pattern, which rejects a leading `v`; the `go_modules` and `github_actions` version classes both override `correct?` to strip it. With the release-pin ignore block written as `update-types`, Dependabot still proposed `kubernetes/kubernetes v1.32.13 -> v1.37.0` in the `release-1-32` directory. The condition produced no range at all.

Dropping the `v` from the revs makes it worse rather than better. The condition then produces `">= 2.4.a, < 3"`, and `Dependabot::PreCommit::Requirement.requirements_array` hands the whole string to `Gem::Requirement` without splitting on commas, raising `Gem::Requirement::BadRequirementError: Illformed requirement` and failing the entire update job.

Single-clause `versions` bounds are unaffected by both, which is why the config uses them. They are generated from the matrix and checked by `verify` so they cannot drift from the pins they describe. `update-types` conditions are rejected by `verify` rather than merely unused: if only the first defect were fixed upstream, an `update-types` condition would begin emitting the two-clause range and hit the second.

## Relationship to #2148

Same bug, same three generator fixes, different mechanism.

|  | #2148 (gomod) | this PR (pre-commit) |
| --- | --- | --- |
| files changed | 116 | 18 |
| files in the tracking tree | 105 (35 modules with `go.mod`, `go.sum`, `tools.go`) | 7 |
| Go toolchain | required by the generator's write path (`go mod tidy`) | not used |
| ecosystem age | long established | GA 2026-03-10 |
| pin policy | `update-types` in `dependabot.yml`, works as written | explicit per-directory `versions` bounds in `dependabot.yml`, generated from the matrix and checked by `verify` |
| policy backstop | none needed | `sync-tracking` rejects an out-of-policy rev |
| tracking file content | a module path, a `go` directive, a requirement, a checksum file and a build-tagged blank import per dependency | one `repo` and one `rev` per dependency |

Both express the same policy declaratively in `dependabot.yml` and both were validated end to end with the dependabot CLI. #2148 relies on a mature ecosystem and a mechanism the project already uses elsewhere, at the cost of 105 generated files in the tracking tree and a Go toolchain in the generator. This PR relies on an ecosystem that is about six months old and on two of its rough edges being worked around, in exchange for a tracking format whose entire content is the pins and a generator with no toolchain dependency. The workarounds are pinned down by `verify`, so they degrade loudly rather than silently, but they are workarounds and the ecosystem may still change under them.

Whichever is merged, the other should be closed; they touch the same files and cannot coexist.

## Live demo

Both approaches were deployed as `main` of a fresh repository on my account, so the real Dependabot service ran against them with the exact `dependabot.yml` from each branch, and the `update-kubernetes-version-matrix` workflow ran on the resulting pull requests. Nothing was triggered by hand: Dependabot's first pass after the config landed opened the PRs, and every workflow run finished successfully and pushed the `dependabot: update Kubernetes version matrix` commit onto the Dependabot branch.

Go modules (#2148, https://github.com/MaxRink/image-builder-dependabot-gomod):

| Dependabot PR | What it bumped | Regenerated matrix |
| --- | --- | --- |
| [gomod PR 2](https://github.com/MaxRink/image-builder-dependabot-gomod/pull/2) | `latest`: containerd 2.3.2 to 2.3.4, runc 1.4.3 to 1.5.1, Kubernetes 1.36.3 to 1.37.0, cri-tools 1.36.0 to 1.37.0 (four modules, one grouped PR) | [run](https://github.com/MaxRink/image-builder-dependabot-gomod/actions/runs/33920467529) wrote `kubernetes-version-latest.yaml` |
| [gomod PR 3](https://github.com/MaxRink/image-builder-dependabot-gomod/pull/3) | release pins across eleven module directories: containerd 2.3.4 everywhere, CNI plugins 1.6.2 for 1.32 and 1.33, Kubernetes patch bumps for 1.34, 1.35 and 1.36, no minor bumps | [run](https://github.com/MaxRink/image-builder-dependabot-gomod/actions/runs/33920492382) wrote `kubernetes-version-matrix.yaml` |

pre-commit configs (#2163, https://github.com/MaxRink/image-builder-dependabot-precommit):

| Dependabot PR | What it bumped | Regenerated matrix |
| --- | --- | --- |
| [pre-commit PR 7](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/7) | `latest`: the same four revs as gomod PR 2, each a one-line `rev:` change | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920652950) |
| [pre-commit PR 4](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/4) | release-1-32: containerd v2.3.4 and CNI plugins v1.6.2, no Kubernetes, cri-tools or runc update | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920634111) |
| [pre-commit PR 5](https://github.com/MaxRink/image-builder-dependabot-precommit/pull/5) | release-1-31: containerd only | [run](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/runs/33920640647) |
| [pre-commit PRs 2, 3, 6, 8](https://github.com/MaxRink/image-builder-dependabot-precommit/pulls) | release-1-34, 1-35, 1-36, 1-33: containerd plus the Kubernetes or CNI patch level, one PR per directory | all [runs](https://github.com/MaxRink/image-builder-dependabot-precommit/actions/workflows/update-kubernetes-version-matrix.yaml) green |

The matrix diffs are identical between the two demos for the same selector. The visible operational difference is PR volume: the gomod config groups across directories and produced two matrix PRs, the pre-commit config needs one entry per directory because the version bounds differ, and produced seven.

## Validation

- `verify` passes offline; `render latest` and `render 1.33` are byte-identical to `main`.
- Simulating a Dependabot update by rewriting only the `rev` lines (`release-1-32` runc to `v1.4.4`, `latest` containerd to `v2.3.4`) and running `sync-tracking --write` then `verify` produces exactly the two expected matrix edits and nothing else.
- dependabot CLI, `release-1-32`, with this branch's ignore bounds: two pull requests, containerd `v2.3.2 -> v2.3.4` and cni-plugins `v1.6.0 -> v1.6.2`, no kubernetes, cri-tools or runc update. Every diff is a single `rev` line.
- dependabot CLI, `latest`: four pull requests, containerd `-> v2.3.4`, runc `-> v1.5.1`, kubernetes `-> v1.37.0`, cri-tools `-> v1.37.0`, no major bumps. As a control, lowering only containerd's bound to `>= 2.3.3` removes exactly the containerd update and leaves the other three, confirming the bounds bind in that directory.
- `update --write` against the network preserves the three Dependabot-owned values across all seven entries and leaves a tree that verifies.
- A reordered, requoted manifest with comments still verifies. An out-of-policy rev, a missing entry, a stale bound after a hand pin bump, an `update-types` condition, an untracked `dependency-name`, a globbed directory, a stray file in the tracking tree and an entry outside it all fail.
- `ruff check` reports the same findings as `main`, `yq` parses every changed YAML file, and both workflow patterns match exactly the seven manifests.


MaxRink:
In general dependabot is sadly much more limited than the renovate we use downstream, where we can just use packageRules combined with customManagers of type regex

images/capi/hack/kubernetes-version-matrix.py
```diff
@@ -437,26 +627,17 @@ def entry_from_tracking(selector: str, current: dict[str, Any]) -> dict[str, Any
             f"expected {selector}"
         )
 
-    cni_semver = versions["github.com/containernetworking/plugins"]
-    cni_version = cni_semver.removeprefix("v")
-    cni_rpm_version = resolve_cni_rpm_version(kubernetes_minor, cni_version)
+    deb_version = current.get("kubernetes_deb_version")
+    if current.get("kubernetes_rpm_version") != kubernetes_version or not deb_version:
+        deb_version = resolve_kubernetes_deb_version(kubernetes_version)
```
drew-viles, line 636: Could we also re-resolve `kubernetes_cni_deb_version` and `kubernetes_cni_rpm_version` here when `kubernetes_minor` changes? On `latest` a 1.36 to 1.37 bump keeps the 1.36 repo's `kubernetes-cni` pin, and `validate_entry` only checks deb and rpm agree with each other, so the build fails on apt install rather than here.

images/capi/hack/kubernetes-version-matrix.py
```diff
@@ -336,99 +362,263 @@ def tracking_selector_name(selector: str) -> str:
...
+    the ranges are only as good as their last regeneration, and a rev can also
+    reach the tree by hand, so a release pin takes patch updates only and the
+    rolling latest entry takes everything below a major bump.
+    """
+    segments = 1 if selector == "latest" else 2
+    errors = []
+    for tracked in TRACKED_REPOS:
+        before = str(current.get(tracked.entry_key, "")).removeprefix("v")
+        after = values[tracked.entry_key].removeprefix("v")
+        if not before or before.split(".")[:segments] == after.split(".")[:segments]:
```
drew-viles: This only compares the leading segments, so a downgrade (v1.31.14 to v1.31.13) passes and gets written. Since the docstring says this guards hand edits too, could we compare with `version_sort_key` and reject anything not strictly newer?

.github/dependabot.yml
```diff
@@ -23,45 +23,203 @@ updates:
     - "kind/cleanup"
     - "release-note-none"
 
-# Kubernetes version matrix release pins
+# Kubernetes version matrix tracking manifests.
+# One entry per directory: the ignore ranges below hold each pinned
+# repository to the pin policy of its directory, and those ranges differ
+# per directory, so a single globbed entry cannot express them. Regenerate
```
drew-viles: Nothing regenerates these, `verify` only checks them. Could we either add a `render-dependabot` step or reword this to say the ranges are hand-maintained and `verify` names the value to write?

drew-viles (commented):
Tested offline: `verify`, `render`, `render-tracking` all pass and the 35 bounds match the matrix. I also checked dependabot-core and the `PreCommit::Version` / `requirements_array` analysis holds.

Three things I would like before LGTM, inline: the CNI package pins no longer re-resolve when `latest` crosses a minor, the policy check accepts downgrades, and the dependabot.yml header says "regenerate" but nothing generates it. The rest are nits or follow-ups.

images/capi/hack/kubernetes-version-matrix.py
```diff
@@ -563,6 +741,19 @@ def update(write: bool) -> int:
...
+    if changed and not write:
+        print(
+            "Dependabot tracking manifests do not match the matrix; rerun with --write",
+            file=sys.stderr,
+        )
+        return 1
+    return verify()
+
+
 def sync_tracking(write: bool) -> int:
```
drew-viles, line 778: Policy failures from `entry_from_tracking` surface as a raw traceback rather than an `ERROR:` line like `verify`. Could we catch the `ValueError` here and print it the same way?

.github/workflows/update-kubernetes-version-matrix.yaml
```diff
@@ -7,8 +7,7 @@ on:
       - reopened
       - synchronize
     paths:
-      - "images/capi/packer/config/kubernetes-version-dependencies/**/go.mod"
-      - "images/capi/packer/config/kubernetes-version-dependencies/**/go.sum"
+      - "images/capi/packer/config/kubernetes-version-dependencies/**/.pre-commit-config.yaml"
```
drew-viles, line 18: `verify` now checks manifests and dependabot.yml bounds but this workflow only runs for Dependabot PRs. A human PR adding release-1-37 without the dependabot block merges green and breaks the next unrelated Dependabot PR. Could we wire `verify` into a presubmit, or is that a follow-up?

images/capi/hack/kubernetes-version-matrix.py
```diff
@@ -336,99 +362,263 @@ def tracking_selector_name(selector: str) -> str:
...
+def ignore_bound(selector: str, tracked: TrackedRepo, entry: dict[str, Any]) -> str:
+    """Return the ignore range that holds one repository to the selector's pin policy.
+
+    A release pin is held below the next minor of its pinned version, the
+    rolling latest entry below the next major. The range is a single clause on
+    purpose: Dependabot::PreCommit::Requirement passes the whole string to
+    Gem::Requirement without splitting on commas, so a two-clause range raises
+    Gem::Requirement::BadRequirementError and fails the update job.
+    """
+    major, minor = str(entry[tracked.entry_key]).removeprefix("v").split(".")[:2]
```
drew-viles, line 490: A single-segment `containerd_version` crashes here with a bare `ValueError`. Could `validate_entry` shape-check the three unprefixed keys the way it does for the two semvers?

images/capi/hack/kubernetes-version-matrix.py
MaxRink: Addressed in c51f3ea6adc5aefb481e88850272af29152abb41: sync_tracking refreshes both CNI package versions when the Kubernetes minor changes, while patch updates preserve them. The focused tests cover both cases.

images/capi/hack/kubernetes-version-matrix.py
MaxRink: Addressed in c51f3ea6: pin_policy_errors compares version_sort_key and rejects backward movement. An unchanged pin remains valid.

.github/dependabot.yml
MaxRink: Addressed in c51f3ea6: the Dependabot configuration header now explicitly says these bounds are hand-maintained; verify reports the required value when they drift.

images/capi/hack/kubernetes-version-matrix.py
MaxRink: Addressed in c51f3ea6: malformed tracking input raises ValueError, which is caught and reported as ERROR with exit status 1 before writes. A dedicated regression test covers this.

.github/workflows/update-kubernetes-version-matrix.yaml
MaxRink: Addressed in c51f3ea6: the existing update workflow now has a pull_request trigger covering the matrix, tracking manifests, generator, tests, workflow and Dependabot config, plus a read-only verification job. Current-head verify-kubernetes-version-matrix passed.

images/capi/hack/kubernetes-version-matrix.py
MaxRink: Addressed in c51f3ea6: validate_entry checks the shape of all three unprefixed version fields, and verify returns on malformed input before computing bounds. A dedicated regression test covers this.

